Issue #137, opened 2026-04-14
Keyboard builder duplication — three nearly-identical implementations


chrisrogers37:
## Summary

There are three separate keyboard-building functions with overlapping logic:

1. **`TelegramNotificationService._build_keyboard()`** — used for initial message sends
2. **`telegram_utils.build_queue_action_keyboard()`** — standalone version with nearly identical logic
3. **`telegram_utils.build_error_recovery_keyboard()`** — error variant (swaps Auto Post for Retry, removes account selector)

The first two are functionally identical but maintained separately, creating drift risk. If a button is added/changed in one, the other must be updated manually.

## Location

- `src/services/core/telegram_notification.py` — `_build_keyboard()` (lines ~165-223)
- `src/services/core/telegram_utils.py` — `build_queue_action_keyboard()` and `build_error_recovery_keyboard()` (lines ~216-262)

## Impact

- Code duplication increases maintenance burden
- Button layout changes need to be applied in multiple places
- Error recovery keyboard already drifts from standard keyboard (no account selector)

## Suggested Fix

Consolidate into a single `build_keyboard()` function in `telegram_utils.py` with parameters for variants:
- `error_recovery: bool = False` — swaps Auto Post for Retry
- `include_account_selector: bool = True` — optionally omits account button
- Remove the duplicate in `TelegramNotificationService` and call the shared function

---
🤖 Generated by tech-debt scan · Focus: verbose display and logging

PR #143, opened 2026-04-14
Fix stale account captions + add cycle switcher (#140, #141)

### LINKED ISSUE #140: Multi-account select: stale captions on pending messages after account switch
## Problem

When switching Instagram accounts via the inline selector, only the message you clicked updates. All other pending messages continue showing the old account name in their captions and button labels — even though the database has already switched globally.

This creates a confusing state where 5 pending posts might show "Account: Primary" but the system will actually post to "Secondary" when you click Auto Post or mark as Posted.

## Current Flow (Confusing)

1. Messages A, B, C all show "📸 Account: Primary" (set at send time)
2. User clicks account selector on Message A → picks "Secondary"
3. `switch_account()` updates `chat_settings.active_instagram_account_id` globally
4. Message A rebuilds: shows "📸 Account: Secondary" ✅
5. Messages B, C still show "📸 Account: Primary" ❌ (stale)
6. User clicks "Auto Post" on Message B → it posts to Secondary (the DB state), but caption said Primary

## Root Cause

- Active account is stored **globally per chat** in `chat_settings.active_instagram_account_id` — not per queue item
- Captions are rendered at send time and never auto-refreshed
- `rebuild_posting_workflow()` in `telegram_accounts.py:402-441` only rebuilds the single message where the user clicked

## Proposed Enhancement

**Option A (Recommended): Batch-update all pending messages on account switch**

After `switch_account()` completes, iterate over all pending queue items with `telegram_message_id` set, rebuild their captions and keyboards with the new active account. This keeps the UI truthful.

Key implementation points:
- In `handle_post_account_switch()` (`telegram_accounts.py:321-401`), after rebuilding the current message, query all pending queue items for the chat
- For each, call `bot.edit_message_caption()` with the rebuilt caption and keyboard
- Rate-limit edits to avoid Telegram API throttling (30 msg/sec limit)

**Option B: Store account per queue item**

Add `instagram_account_id` to `posting_queue` model. Each queue item remembers which account it should post to. Account button on each message only changes that item's account, not the global setting.

This is more correct but a larger change — requires migration, scheduler updates, and autopost handler changes.

## Impact

- **Impact**: High — directly confuses users about which account content will post to
- **Effort**: Medium (Option A) / High (Option B)
- **Risk**: Low (Option A) / Medium (Option B — migration required)

---
🤖 Generated by product-enhance scan

### LINKED ISSUE #141: Multi-account select: simplify the account switching interaction
## Problem

Switching accounts requires a multi-step navigation flow that feels heavy for a frequent action:

1. Click "📸 Account Name" button on the post
2. Wait for selector menu to load (replaces all buttons)
3. Click desired account
4. Menu rebuilds with updated checkmark
5. Click "Back to Post" to return to the posting workflow

This is 3-4 taps and two screen redraws for what should be a quick toggle.

## Current Implementation

- `handle_post_account_selector()` (`telegram_accounts.py:231-319`) replaces the entire keyboard with account list + Back button
- `handle_post_account_switch()` (`telegram_accounts.py:321-401`) switches globally, then re-renders the selector menu (not the post)
- `rebuild_posting_workflow()` (`telegram_accounts.py:402-441`) only called when "Back to Post" is clicked

## Proposed Enhancement

**For users with 2-3 accounts (common case):** Replace the submenu with a single-tap cycle button.

- Current: `📸 Primary` → opens submenu
- Proposed: `📸 Primary` → taps cycle to `📸 Secondary` → taps cycle to `📸 Business` → wraps around

Implementation:
- If `account_count <= 3`: use cycle behavior (callback switches to next account + rebuilds inline)
- If `account_count > 3`: keep current submenu (too many to cycle through)
- Cycle callback: `cycle_account:{queue_id}` → finds next active account after current, wraps around

This eliminates the submenu entirely for the common case and removes the "Back to Post" step.

## Impact

- **Impact**: Medium — reduces friction on a frequent interaction
- **Effort**: Low — single new callback handler + conditional in keyboard builder
- **Risk**: Low — additive change, doesn't break existing flow for 4+ accounts

---
🤖 Generated by product-enhance scan

chrisrogers37:
## Summary

- **Batch-update pending messages on account switch** — When switching Instagram accounts, all pending notification captions and button labels now update to reflect the new account. Previously only the clicked message updated, leaving other posts showing the old account name.
- **Single-tap account cycle for 2-3 accounts** — Account button now cycles through accounts with one tap instead of opening a submenu. Submenu preserved for 4+ accounts.
- **Consolidated keyboard builder** — `_build_keyboard()` in notification service now delegates to shared `build_queue_action_keyboard()`, eliminating duplicate implementation.
- **Efficiency** — Added `count_active_accounts()` (single COUNT query) to avoid fetching full account data when only the count is needed. Threaded `account_count` through methods to avoid redundant DB queries.

## Test plan

- [ ] Switch account on a post with 2+ pending messages — verify all messages update
- [ ] With 2-3 accounts, verify account button cycles instead of opening submenu
- [ ] With 4+ accounts, verify submenu behavior preserved
- [ ] Verify "Back" button in callbacks rebuilds with correct account
- [ ] Verify autopost still works after account switch

Closes #140, closes #141, partial #137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed — Multi-Account UX Improvements (#140, #141)
+
+- **Batch-update pending messages on account switch** — When switching Instagram accounts, all pending notification captions and button labels now update to reflect the new account. Previously only the message where you clicked updated, leaving other pending posts showing the old account name. (#140)
+- **Single-tap account cycle for 2-3 accounts** — The account selector button now cycles through accounts with one tap instead of opening a submenu. For users with 4+ accounts, the submenu is preserved. (#141)
+- **Consolidated keyboard builder** — `TelegramNotificationService._build_keyboard()` now delegates to the shared `build_queue_action_keyboard()` utility, eliminating a duplicate keyboard implementation. (#137 partial)
+
 ### Fixed — Worker Crash in Cloud Storage Cleanup
 
 - **Fix fatal AttributeError in cleanup loop** — `cleanup_cloud_storage_loop` called `cleanup_transactions()` on a `MediaRepository`, but that method only exists on `BaseService`. Replaced with the correct `end_read_transaction()` call. This crash killed the entire worker process after the first hourly cleanup cycle.
--- a/src/repositories/queue_repository.py
+++ b/src/repositories/queue_repository.py
@@ -314,6 +314,33 @@ class QueueRepository(BaseRepository):
 
         return len(abandoned)
 
+    def get_pending_with_telegram_message(
+        self, telegram_chat_id: int
+    ) -> List[PostingQueue]:
+        """Get pending/processing queue items that have been sent to Telegram.
+
+        Used to find all active notifications for a chat so their captions
+        and keyboards can be batch-updated (e.g., after an account switch).
+
+        Args:
+            telegram_chat_id: The Telegram chat ID to filter by
+
+        Returns:
+            List of PostingQueue items with telegram_message_id set
+        """
+        result = (
+            self.db.query(PostingQueue)
+            .filter(
+                PostingQueue.telegram_chat_id == telegram_chat_id,
+                PostingQueue.telegram_message_id.isnot(None),
+                PostingQueue.status.in_(["pending", "processing"]),
+            )
+            .order_by(PostingQueue.scheduled_for.asc())
+            .all()
+        )
+        self.end_read_transaction()
+        return result
+
     def delete_all_pending(self, chat_settings_id: Optional[str] = None) -> int:
         """Delete all pending queue items. Returns count of deleted items."""
         count = (
--- a/src/services/core/instagram_account_service.py
+++ b/src/services/core/instagram_account_service.py
@@ -490,6 +490,10 @@ class InstagramAccountService(BaseService):
 
             return account
 
+    def count_active_accounts(self) -> int:
+        """Count active Instagram accounts (lightweight, single COUNT query)."""
+        return self.account_repo.count_active()
+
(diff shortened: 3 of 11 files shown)


chrisrogers37 (commented):
## Review: Fix stale account captions + cycle switcher

Clean, well-scoped fixes. The batch-update approach is the right call — query all pending notifications, edit them in one pass, skip the triggering message. The cycle switcher logic is simple and the guard for <2 accounts in `handle_cycle_account` is correct.

---

### Suggestion: Move inline import to top of file

**`src/services/core/telegram_notification.py` — inside `send_notification()`**

```python
from src.services.core.telegram_utils import build_queue_action_keyboard
```

Placed inside the method body. No circular import issue exists (the removed imports were from the `telegram` package, not from `telegram_notification`). Move it to the top of the file with the other imports.

---

### Suggestion: Acknowledge Telegram rate limit risk in batch loop

**`src/services/core/telegram_accounts.py` — `_batch_update_pending_captions()`**

Batch-editing N messages in a tight loop can hit Telegram's rate limits (~20 inline edits/minute per chat) once a user accumulates many pending posts. The broad `except Exception` correctly swallows these without crashing — this degrades gracefully — but a brief comment acknowledging the rate-limit case would help the next reader understand why errors are dropped rather than re-raised.

---

### Nit: Test coverage gap

No automated tests for the new cycle/batch behavior. Both paths are testable: mock `get_pending_with_telegram_message` returning N items, assert `edit_message_caption` called N-1 times (excluding the triggering message). Worth a follow-up.

---

### What's working well

- `count_active()` exists in the repo (line 144) — the COUNT-query efficiency improvement is real
- `account_count=0` default safely falls back to submenu for callers that don't pass it
- Stale-message direction (adding accounts → old messages stay on submenu until next switch) degrades gracefully into existing UX
- Removing `_build_keyboard()` from notification service is a clean deduplication

**Verdict: Approved.** Two suggestions, no blockers.

PR #147, opened 2026-04-14
Add Telegram alert when background task crashes (#132)

### LINKED ISSUE #132: Add Telegram alert when background task crashes
## Context

PR #130 added a `_guarded()` wrapper around all background loops in `src/main.py` so that a crash in one loop doesn't kill the entire worker (via `asyncio.gather()`). However, currently `_guarded()` only logs `CRITICAL` — there's no active notification.

During the April 10 outage, the worker was dead for 2+ days before anyone noticed. With `_guarded()`, the worker would survive but the crashed loop would silently stop. The scheduler could die and posting would stop with no alert.

## Desired Behavior

When `_guarded()` catches an unhandled exception, send a Telegram message to the admin chat:

```
⚠️ Background task '{name}' crashed: {error}
Worker is still running but this loop has stopped.
```

## Implementation Notes

- `_guarded()` is at `src/main.py:33` 
- It needs access to a `TelegramService` instance (or just the bot + admin chat ID)
- Keep it simple — don't let notification failures mask the original crash
- Consider: should the loop restart after logging, or stay dead? Restarting risks infinite crash loops; staying dead risks silent degradation

## Acceptance Criteria

- [ ] Admin Telegram chat receives a message when any background task crashes
- [ ] Notification failure doesn't mask or suppress the original crash log
- [ ] Tests cover the notification path

chrisrogers37:
## Summary

- **Crash alerts via Telegram** — When `_guarded()` catches an unhandled exception in any background loop, it sends a message to the admin chat: task name, error type, truncated message
- **Alert failures isolated** — If the Telegram notification itself fails, it's logged at ERROR but never masks the original CRITICAL crash log
- **All 5 guarded loops wired up** — scheduler, lock_cleanup, cloud_cleanup, media_sync, transaction_cleanup all pass `bot=telegram_service.bot`
- **5 new tests** — crash logging, alert sending, no-bot fallback, alert failure isolation, CancelledError propagation

## Test plan

- [ ] Kill a background loop (e.g., disconnect DB mid-sync) — verify Telegram alert arrives
- [ ] Verify worker stays alive after one loop crashes
- [ ] Verify alert includes task name and error message
- [ ] Simulate Telegram API down during crash — verify original error still logged

Closes #132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added — Telegram Crash Alerts (#132)
+
+- **Send Telegram alert when a background task crashes** — `_guarded()` now sends a message to the admin chat when any background loop (scheduler, lock cleanup, cloud cleanup, media sync, transaction cleanup) crashes. The worker stays alive but the user is immediately notified which loop stopped. Alert failures are caught separately so they never mask the original crash.
+
 ### Changed — Keyboard Builder Consolidation (#137)
 
 - **Merge `build_error_recovery_keyboard` into `build_queue_action_keyboard`** — The two near-identical keyboard builders are now one function with an `error_recovery` parameter. Error recovery mode shows "Retry Auto Post" instead of "Auto Post" and hides the account selector.
--- a/src/main.py
+++ b/src/main.py
@@ -30,18 +30,34 @@ SERVICE_RUNS_RETENTION_DAYS = 7
 RETENTION_INTERVAL_TICKS = 60
 
 
-async def _guarded(name: str, coro: Coroutine) -> None:
+async def _guarded(name: str, coro: Coroutine, *, bot=None) -> None:
     """Run a coroutine and log (not propagate) unhandled exceptions.
 
     Prevents a crash in one background loop from killing the whole worker
-    via asyncio.gather().
+    via asyncio.gather(). When a bot instance is provided, sends a Telegram
+    alert to the admin chat so crashes aren't silent.
     """
     try:
         await coro
     except asyncio.CancelledError:
         raise  # let shutdown propagate
-    except Exception:
+    except Exception as exc:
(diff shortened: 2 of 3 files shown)


chrisrogers37 (commented):
## Review: Add Telegram crash alerts for background tasks

Clean, well-structured feature. The isolation contract is right: crash always logged at CRITICAL, alert is best-effort, `CancelledError` propagates, alert failure never masks the original log.

---

### Suggestion: Backtick in exception message can silently drop the Telegram alert

**`src/main.py` — `_guarded()`**

```python
f"Error: `{type(exc).__name__}: {str(exc)[:200]}`\n\n"
```

With `parse_mode="Markdown"`, if `str(exc)` contains a backtick, Telegram's parser may reject the message and `send_message` raises — which is caught by the inner `except Exception` and logged at ERROR. The crash is never truly silent (CRITICAL still fires), but the alert itself is swallowed exactly when the error message is unusual. Two easy fixes:

**Option A — strip backticks from the error string:**
```python
safe_msg = str(exc)[:200].replace("`", "'")
f"Error: `{type(exc).__name__}: {safe_msg}`\n\n"
```

**Option B — drop `parse_mode` entirely (plain text):**
```python
text=(
    f"⚠️ Background task crashed\n\n"
    f"Task: {name}\n"
    f"Error: {type(exc).__name__}: {str(exc)[:200]}\n\n"
    f"Worker is still running but this loop has stopped."
),
# no parse_mode
```
Plain text sidesteps the Markdown parsing risk entirely.

---

### What's working well

- `ADMIN_TELEGRAM_CHAT_ID: int` is a required Pydantic field — validated at startup, can't be None at runtime
- Keyword-only `*, bot=None` signature is the right API shape — callers can't accidentally pass it positionally
- All 5 task wiring calls in `main_async()` are consistent
- Test coverage is thorough: crash logging, alert send, no-bot fallback, alert failure isolation, `CancelledError` propagation — all 5 paths covered
- Alert failure logged at ERROR (not DEBUG/WARNING) — appropriately visible if Telegram is down during a crash

**Verdict: Approved.** One suggestion (Markdown safety), no blockers.

PR #149, opened 2026-04-14
Add loop-liveness tracking to health check (#134)

### LINKED ISSUE #134: Add loop-liveness tracking to health check
## Context

The existing health check (`src/services/core/health_check.py`) already detects stale media syncs by checking `service_runs` timestamps. But during the April 10 outage, the entire worker was dead — and nothing detected it.

With PR #130's \`_guarded()\` wrapper, background loops can now die independently while the worker stays up. This creates a new failure mode: the worker process is "healthy" but the scheduler loop is dead, so no posts go out.

## Desired Behavior

Each background loop should record a heartbeat (last-tick timestamp). The health check should flag any loop that hasn't ticked in 2x its expected interval:

| Loop | Expected Interval | Alert Threshold |
|------|------------------|-----------------|
| scheduler | 60s | 120s |
| lock_cleanup | 300s | 600s |
| cloud_cleanup | 3600s | 7200s |
| media_sync | 300s | 600s |
| transaction_cleanup | 30s | 60s |

## Implementation Notes

- Could use a simple in-memory dict (no DB needed — liveness is ephemeral)
- Health check endpoint at `/health` or via `storyline-cli check-health` should report loop status
- Consider exposing via the Telegram `/status` command too

## Acceptance Criteria

- [ ] Each background loop updates a heartbeat timestamp on every tick
- [ ] Health check reports which loops are alive/stale
- [ ] `check-health` CLI command shows loop liveness
- [ ] Tests cover stale loop detection

chrisrogers37:
## Summary

- **Heartbeat tracking for all 5 background loops** — Each loop records a timestamp on every tick via `record_heartbeat()`. In-memory dict, no DB needed.
- **Health check detects stale loops** — `_check_loop_liveness()` flags any loop that hasn't ticked in 2x its expected interval. Included in `check_all()` so `check-health` CLI and `/status` get it automatically.
- **Expected intervals**: scheduler (60s), lock_cleanup (3600s), cloud_cleanup (3600s), media_sync (300s), transaction_cleanup (30s)
- **5 new tests** — heartbeat storage, fresh/stale detection, unstarted loop handling, all registered loops included

## Test plan

- [ ] Start worker — verify `check-health` shows all loops as "alive" after first tick
- [ ] Kill a background loop — verify health check reports it as stale after threshold
- [ ] Verify unstarted loops (e.g. cloud_cleanup when Cloudinary not configured) show "Not started"

Closes #134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added — Loop Liveness Tracking (#134)
+
+- **Heartbeat tracking for all background loops** — Each loop (scheduler, lock_cleanup, cloud_cleanup, media_sync, transaction_cleanup) records a timestamp on every tick. The health check reports loops as stale if they haven't ticked in 2x their expected interval. Visible via `check-health` CLI and `/status` health checks.
+
 ### Added — Telegram Crash Alerts (#132)
 
 - **Send Telegram alert when a background task crashes** — `_guarded()` now sends a message to the admin chat when any background loop (scheduler, lock cleanup, cloud cleanup, media sync, transaction cleanup) crashes. The worker stays alive but the user is immediately notified which loop stopped. Alert failures are caught separately so they never mask the original crash.
--- a/src/main.py
+++ b/src/main.py
@@ -216,10 +216,10 @@ async def cleanup_locks_loop(lock_service: MediaLockService):
     logger.info("Starting cleanup loop...")
 
     while True:
+        record_heartbeat("lock_cleanup")
         try:
             # Wait 1 hour
             await asyncio.sleep(3600)
-            record_heartbeat("lock_cleanup")
 
             # Cleanup expired locks
             count = lock_service.cleanup_expired_locks()
@@ -246,9 +246,9 @@ async def cleanup_cloud_storage_loop(cloud_service):
     logger.info("Starting cloud storage cleanup loop...")
 
     while True:
+        record_heartbeat("cloud_cleanup")
         try:
             await asyncio.sleep(3600)
-            record_heartbeat("cloud_cleanup")
 
             cloud_count = cloud_service.cleanup_expired(folder=CLOUD_UPLOAD_FOLDER)
             db_count = media_repo.clear_stale_cloud_info(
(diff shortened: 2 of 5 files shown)


chrisrogers37 (commented):
## Review: Loop-liveness tracking for health check

The design is clean — in-memory dict, no DB, 2x threshold multiplier is sensible, and the five tests cover all key paths. One blocker on heartbeat placement.

---

### Blocker: lock_cleanup and cloud_cleanup show "Not started" for first hour after every restart

**`src/main.py` — `cleanup_locks_loop` and `cleanup_cloud_storage_loop`**

Both loops sleep first, then heartbeat:

```python
while True:
    try:
        await asyncio.sleep(3600)         # sleeps
        record_heartbeat("lock_cleanup")  # ← beats only after waking up
```

The other three loops (`scheduler`, `media_sync`, `transaction_cleanup`) all heartbeat at the top of the loop, before sleeping, so they show `alive: True` immediately on startup. These two don't beat until hour 1 completes.

Result: on every worker restart (deploy, Railway restart, crash recovery), both loops have `last_beat = None → alive: False`. Since `_check_loop_liveness()` feeds into `check_all()` which ANDs all sub-checks, the system reports `"unhealthy"` for the entire first hour after startup — even when everything is working correctly. This will train operators to ignore health check alerts.

**Fix** — move both heartbeats to before the sleep, matching the other loops:
```python
while True:
    record_heartbeat("lock_cleanup")  # ← here: "alive and about to sleep"
    try:
        await asyncio.sleep(3600)
        count = lock_service.cleanup_expired_locks()
```

Same for `cleanup_cloud_storage_loop`. This preserves the semantics (you still detect a stuck loop — if the loop hangs inside the try block, it won't beat again until the next iteration) while eliminating the spurious startup unhealthy period.

---

### Nit: inline import in health_check.py

```python
from src.main import get_loop_liveness  # inside method body
```

This is likely justified — `main.py` almost certainly imports `HealthCheckService`, making a top-level import circular. A short comment would help future readers understand it's intentional:

```python
# Inline import to avoid circular: main → health_check → main
from src.main import get_loop_liveness
```

---

### What's working well

- Module-level `loop_heartbeats` dict is the right shape — dict writes are GIL-atomic in CPython, safe without locking
- `LOOP_EXPECTED_INTERVALS` as a single source of truth is clean
- `scheduler`, `media_sync`, `transaction_cleanup` are all placed correctly (top of loop, before sleep)
- `setup_method` clears `loop_heartbeats` between tests — correct test isolation
- The 5 tests cover all liveness paths cleanly

**Verdict: Request Changes.** Fix the two heartbeat placements, then this is ready.

chrisrogers37 (commented):
## Re-review: Loop-liveness tracking

Both items resolved.

**Blocker — heartbeat placement** ✅ Fixed. `lock_cleanup` and `cloud_cleanup` now heartbeat at the top of `while True` before sleeping, matching the other three loops. All 5 loops beat immediately on startup — no more spurious unhealthy window after restarts.

**Nit — inline import comment** ✅ Added. `# Inline import: src.main imports services, so a top-level import would be circular`

**Approved.**

PR #190, opened 2026-04-15
feat: startup migration version check (#118)

### LINKED ISSUE #118: feat: startup migration version check
## Problem

Migrations are manual SQL files applied via `psql`. There is no mechanism to detect when deployed code expects a schema version that hasn't been applied to the database. This caused a production outage (2026-03-29) when migration 022 (`auth_method` column) was applied to localhost instead of Neon — the Railway worker crashed on every scheduler tick with `UndefinedColumn`.

## Proposed Solution

On application startup (`src/main.py`), query the `schema_version` table for the latest applied version and compare against an expected version constant in the codebase. If behind, log a clear error and optionally fail fast.

```python
# Example approach
EXPECTED_SCHEMA_VERSION = 22

def check_schema_version():
    current = db.execute("SELECT MAX(version) FROM schema_version").scalar()
    if current < EXPECTED_SCHEMA_VERSION:
        logger.critical(
            f"Database schema is at version {current}, expected {EXPECTED_SCHEMA_VERSION}. "
            f"Apply migrations {current+1} through {EXPECTED_SCHEMA_VERSION}."
        )
        # Option: sys.exit(1) to fail fast, or continue with warning
```

## Context

- Migrations live in `scripts/migrations/001-022_*.sql`
- Schema version tracked in `schema_version` table
- No release phase in `railway.toml` — migrations are applied manually
- Local `.env` points to localhost PostgreSQL; Railway uses Neon — easy to apply to wrong database

## Acceptance Criteria

- [ ] Expected schema version defined as a constant (updated when new migrations are added)
- [ ] Version check runs at startup before scheduler/bot begins
- [ ] Clear log message identifies which migrations are missing
- [ ] Tests verify the check behavior (version match, version mismatch)

chrisrogers37:
## Summary

- Adds schema version validation on worker startup — queries `schema_version` table and compares against migration files in `scripts/migrations/`
- Logs a clear warning if the database is behind (with exact migration range to apply) or ahead of the deployed code
- Non-blocking: mismatches produce warnings, not fatal errors, so the worker can still start

Closes #118

## Test plan

- [x] 8 new unit tests covering: version match, DB behind, DB ahead, no migrations dir, DB error handling
- [x] All 1460 existing unit tests pass
- [x] Lint and format checks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added — Startup Migration Version Check (#118)
+
+- **Schema version validation on startup** — Worker (`main.py`) now queries the `schema_version` table at boot and compares against migration files in `scripts/migrations/`. Logs a clear warning if the database is behind (with the exact migration range to apply) or ahead of the deployed code.
+- **Non-blocking** — Mismatches produce warnings, not fatal errors, so the worker can still start while the operator applies pending migrations.
+
 ### Added — Schedule Optimization Recommendations (#158)
 
 - **Schedule recommendations API** — `GET /api/onboarding/analytics/schedule-recommendations` analyzes posting history to identify optimal posting times. Returns hourly approval rates, day-of-week patterns, and human-readable recommendations (e.g., "Posts at 10:00 have the highest approval rate (94%)").
--- a/src/main.py
+++ b/src/main.py
@@ -536,6 +536,9 @@ async def main_async():
 
     logger.info("✓ Configuration validated successfully")
 
+    # Check database schema version
+    ConfigValidator.check_schema_version()
+
     # Initialize services
     from src.services.core.settings_service import SettingsService
 
--- a/src/utils/validators.py
+++ b/src/utils/validators.py
@@ -1,9 +1,26 @@
 """Input validation and configuration validation."""
 
-from typing import List, Tuple
+import re
 from pathlib import Path
+from typing import List, Optional, Tuple
 
 from src.config.settings import settings
+from src.utils.logger import logger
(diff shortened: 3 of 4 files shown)


chrisrogers37 (commented):
Clean, minimal, well-tested. **VERDICT: APPROVED** — merge when ready. (GitHub blocks self-approval.)

## What's correct

**`_latest_migration_version()`** — `parents[2]` from `src/utils/validators.py` resolves correctly to the project root. The `^(\d+)_` pattern matches `NNN_description.sql` convention; non-matching files are silently ignored. `max()` of all found versions is the right approach.

**`check_schema_version()`** — Deferred imports inside the function body avoid circular import issues at module load time. `text("SELECT MAX(version) FROM schema_version")` — no injection surface, table name is a literal. `MAX()` on an empty table returns NULL → `db_version = 0` → correctly triggers the "DB behind" path for a fresh database. Catch-all exception handler correctly swallows all DB errors and logs without crashing — right behavior for a non-blocking startup check. Placed after config validation and before service init — correct sequence.

**Tests** — 8 cases: real dir sanity check, missing dir, empty dir, version match, DB behind, DB ahead, no migrations found, DB connection error. `@patch("src.config.database.engine")` correctly targets the module-level attribute — the deferred `from src.config.database import engine` inside the function re-imports at call time, so the patch is active. Comprehensive.

## Nit

**Warning message assumes sequential migrations** (`validators.py:120`) — the message `f"Run pending migrations ({db_version + 1}–{expected})"` assumes every integer between `db_version` and `expected` exists as a migration file. If there are version gaps (e.g., files 1, 2, 5 but no 3, 4), the message would say "Run migrations 3–5" for files that don't exist. Rare edge case in practice and the message is still useful, but worth knowing if migrations ever get renumbered or gaps appear.

PR #196, opened 2026-04-15
feat: Web Dashboard Phase 1 — Telegram Auth, BFF Proxy, Dashboard Shell

chrisrogers37:
## Summary

- **Telegram Login Widget auth** — `/login` page with Telegram widget, JWT sessions in httpOnly cookies, middleware-protected `/dashboard` routes
- **BFF proxy layer** — `/api/dashboard/[...path]` catch-all proxies to FastAPI backend using signed URL tokens (same format the backend already validates)
- **Dashboard overview page** — wires up existing analytics, category performance, and posting history endpoints with parallel server-side fetching
- **Route group restructure** — marketing pages in `(marketing)` group, dashboard in `(dashboard)` group, each with own layout

### Architecture

```
Browser → /dashboard/* → Middleware (JWT check)
                        → Dashboard Layout (session + sidebar/header)
                        → Dashboard Page (server component, fetches backend)

Browser → /api/dashboard/* → BFF Proxy → generates URL token → FastAPI /api/onboarding/*
```

### New env vars required
- `JWT_SECRET` — random 32+ char string for session signing
- `NEXT_PUBLIC_TELEGRAM_BOT_NAME` — bot username for Login Widget
- `BACKEND_URL` — FastAPI URL (default: http://localhost:8000)

### Files added/changed
- `src/lib/session.ts` — Edge-safe JWT/session (importable from middleware)
- `src/lib/auth.ts` — Telegram Login verification + URL token generation (Node crypto)
- `src/lib/backend.ts` — shared `backendFetch()` + `BACKEND_URL`
- `src/middleware.ts` — protects `/dashboard/*`
- `src/app/api/auth/{telegram,logout,session}/` — auth endpoints
- `src/app/api/dashboard/[...path]/` — BFF proxy
- `src/app/login/page.tsx` — login page
- `src/app/(dashboard)/` — dashboard layout + overview page
- `src/app/(marketing)/` — moved landing/setup pages into route group
- `src/components/dashboard/` — sidebar, header, analytics cards, posting chart, category breakdown, recent activity
- `src/components/auth/` — Telegram Login Widget wrapper

## Test plan
- [ ] `npm run build` passes (verified locally)
- [ ] `npm run lint` passes (verified locally)
- [ ] Set env vars and verify Telegram Login flow end-to-end
- [ ] Verify `/dashboard` redirects to `/login` when unauthenticated
- [ ] Verify BFF proxy forwards to FastAPI and returns analytics data
- [ ] Verify existing marketing pages (`/`, `/setup/*`) still render correctly
- [ ] Verify `/api/waitlist` endpoint still works


chrisrogers37 (commented):
## PR #196 Review — feat: Web Dashboard Phase 1

**This PR does:** Implement Telegram Login Widget auth with JWT sessions, a BFF proxy layer from Next.js to FastAPI, and a dashboard shell with analytics components.

**Architecture verdict:** Sound. The module split (session.ts Edge-safe / auth.ts Node-only), BFF token pattern, and React `cache()` deduplication are all well-executed. Two blockers need fixing before merge.

---

### 🔴 Blockers

**B1. Hardcoded JWT_SECRET fallback — `landing/src/lib/session.ts:23`**

```typescript
const JWT_SECRET = new TextEncoder().encode(
  process.env.JWT_SECRET || "storyline-ai-jwt-secret-change-me"  // ← public fallback
);
```

The fallback is committed to a public repo. Any deployment that forgets `JWT_SECRET` uses a known-plaintext secret, making every JWT on that instance forgeable.

Fix — fail hard at startup:
```typescript
const rawSecret = process.env.JWT_SECRET;
if (!rawSecret || rawSecret.length < 32) {
  throw new Error("JWT_SECRET must be set to a random 32+ character string");
}
const JWT_SECRET = new TextEncoder().encode(rawSecret);
```

---

**B2. BFF proxy path not validated — `landing/src/app/api/dashboard/[...path]/route.ts:30-31`**

```typescript
const backendPath = path.join("/");
const url = new URL(`/api/onboarding/${backendPath}`, BACKEND_URL);
```

`new URL()` resolves `..` segments. A request to `/api/dashboard/%2e%2e/admin` (URL-encoded `..`) resolves to `http://backend/api/admin`. An authenticated user can potentially reach FastAPI endpoints outside `/api/onboarding/`.

Fix — assert path containment after construction:
```typescript
const url = new URL(`/api/onboarding/${backendPath}`, BACKEND_URL);
if (!url.pathname.startsWith("/api/onboarding/")) {
  return NextResponse.json({ error: "Invalid path" }, { status: 400 });
}
```

---

### 🟡 Suggestions

**S1. `unsafe-eval` in CSP probably not needed — `landing/next.config.ts:14`**

`unsafe-inline` is likely required by the Telegram widget. `unsafe-eval` (which allows `eval()` and `new Function()`) almost certainly isn't. Remove it and test — if the widget breaks, document why. Currently this undermines XSS protection on `/login`.

Also: security headers are only applied to `/login`. Consider adding `X-Frame-Options: DENY`, `X-Content-Type-Options: nosniff` globally.

---

**S2. Backend 5xx errors leak to client — `landing/src/app/api/dashboard/[...path]/route.ts:66-73`**

Backend JSON responses are proxied verbatim including error bodies. FastAPI stack traces / internal messages on 5xx reach the browser. Gate on status:

```typescript
if (backendResponse.status >= 500) {
  return NextResponse.json({ error: "Backend unavailable" }, { status: 502 });
}
```

---

**S3. auth_date TTL is 24 hours — `landing/src/lib/auth.ts:27`**

`TELEGRAM_LOGIN_TTL = 86400` — Telegram recommends verifying `auth_date` is recent (typically 5-10 minutes). At 24h, an intercepted login callback can be replayed for a full day. Reduce to 300-600 seconds. (This is separate from the JWT session TTL which is correctly handled.)

---

**S4. Confirm FastAPI validates URL token freshness**

`generateUrlToken()` (`auth.ts:98-104`) embeds a timestamp, but it's unclear whether `validate_url_token()` on the FastAPI side enforces a TTL. If not, captured URL tokens are valid for the lifetime of the user's session. Worth a comment confirming backend enforcement.

---

### 🔵 Nits

- `sidebar.tsx:9-11` — links to `/dashboard/analytics` and `/dashboard/settings` which don't exist yet (will 404). Fine for Phase 1, worth a comment/TODO.
- `logout/route.ts` — always returns `{ ok: true }` regardless of whether a session existed. Harmless but slightly misleading.

---

### ✅ What's Well Done

- **Telegram HMAC verification** correctly implements the spec: SHA256 key derivation, sorted field string, `auth_date` check enforced (not just computed).
- **Cookie flags** all correct: `httpOnly`, `secure` (prod-only), `sameSite: "lax"`, `path: "/"`.
- **BFF token isolation** — proxy generates its own backend token from session data; user JWT never touches FastAPI. Clean separation.
- **Module split** (session.ts Edge-safe / auth.ts Node-only) is well-architected and clearly documented.
- **React `cache()`** for session deduplication — good call.
- **Parallel fetching** via `Promise.all` on the dashboard page.
- **Route group restructure** is clean — marketing and dashboard layouts properly isolated.
- Vercel preview is green ✓

---

**Verdict: Request Changes** — 2 blockers (JWT secret, path traversal), both quick fixes. Architecture is solid. Happy to re-review once addressed.

chrisrogers37 (commented):
## Re-Review: PR #196 — **Approved (LGTM)**

All 2 blockers and all 4 suggestions resolved. 7/8 items from the original review addressed; one nit left (harmless, noted below).

---

### ✅ Blockers — Both Resolved

**B1. JWT_SECRET fallback — FIXED** (`landing/src/lib/session.ts`)

```typescript
const rawSecret = process.env.JWT_SECRET;
if (!rawSecret || rawSecret.length < 32) {
  throw new Error("JWT_SECRET must be set to a random 32+ character string");
}
const JWT_SECRET = new TextEncoder().encode(rawSecret);
```

Exact fix requested. Module-level throw means the Edge Runtime fails to initialize if the secret is missing or too short — fail-secure, no silent plaintext fallback. ✅

---

**B2. BFF path traversal — FIXED** (`landing/src/app/api/dashboard/[...path]/route.ts`)

The fix is actually stronger than what I suggested. Two layers:

1. **Allowlist on `path[0]`** — only named endpoints pass (`analytics`, `accounts`, `history-detail`, etc.). Any unknown segment is rejected before URL construction.
2. **Belt-and-suspenders `startsWith("/api/onboarding/")`** — catches multi-segment traversal (e.g., `analytics/../../evil` resolves to `/api/evil`, caught here).

Verified: URL-encoded `..` segments are handled — `path[0]` would be `".."` which isn't in `ALLOWED_PATHS`. ✅

---

### ✅ Suggestions — All 4 Resolved

**S1. `unsafe-eval` removed** (`landing/next.config.ts`) — CSP on `/login` now only has `unsafe-inline` (required for Telegram widget) + `frame-src https://oauth.telegram.org`. `unsafe-eval` gone. Also added `X-Frame-Options: DENY` and `X-Content-Type-Options: nosniff` globally — better than what I asked for. ✅

**S2. 5xx errors gated** (`route.ts:line ~82`) — Backend 5xx returns `{ error: "Backend unavailable" }` with status 502. Error logged server-side. ✅

**S3. auth_date TTL reduced** (`landing/src/lib/auth.ts:line 28`) — `TELEGRAM_LOGIN_TTL = 300` (5 minutes). Good comment explaining it's the login callback window, separate from session TTL. ✅

**S4. FastAPI URL token TTL documented** (`auth.ts` JSDoc on `generateUrlToken`) — Comment confirms FastAPI enforces a 1-hour TTL via `src/utils/webapp_auth.py:URL_TOKEN_TTL`. ✅

---

### ⚪ Nits

**Nit 1 — sidebar TODO added** (`sidebar.tsx:9`) — `// TODO: /dashboard/analytics and /dashboard/settings pages are Phase 2+`. ✅

**Nit 2 — logout still returns `{ ok: true }` always** (`logout/route.ts`) — Unchanged from original. Still harmless (cookie is always deleted regardless of whether a session existed, and the response is fine either way). Not a blocker — leaving it as-is is acceptable.

---

### Verdict

Clean re-submission. Both security fixes are correct and the path traversal defense is more robust than the original suggestion. Ready to merge.

PR #223, opened 2026-04-16
Fix setup page horizontal overflow at 375px mobile

### LINKED ISSUE #215: Setup guide /setup/google-drive: horizontal overflow at 375px mobile viewport
## Summary

The setup guide page at `/setup/google-drive` (and potentially other setup sub-pages) has a horizontal overflow of `scrollWidth: 696px` at 375px viewport. The tab navigation bar (`Overview | Instagram Account | Meta Developer | Google Drive | Organize Media | Connect Telegram`) is too wide to fit on mobile, causing the page to be wider than the viewport.

## Reproduction

1. Navigate to `/setup/google-drive`
2. View at 375px width (mobile)
3. Page scrolls horizontally — content is clipped

## Root cause

The horizontal tab list in the setup nav renders all tabs in a single non-wrapping row. The combined tab widths exceed 375px.

## Suggested fix

Add horizontal scroll to the tab list, or switch to a compact mobile variant (select dropdown or collapsible accordion):

```tsx
// src/components/setup/setup-nav.tsx — add overflow-x-auto to the tab row wrapper
<div className="overflow-x-auto">
  <nav className="flex gap-1 whitespace-nowrap">
    ...
  </nav>
</div>
```

Or use a `<select>` on mobile viewports:

```tsx
<div className="sm:hidden">
  <select onChange={(e) => router.push(e.target.value)}>
    {steps.map(s => <option key={s.href} value={s.href}>{s.label}</option>)}
  </select>
</div>
<nav className="hidden sm:flex gap-1">...</nav>
```

chrisrogers37:
## Summary
- **CopyButton**: Add `max-w-full` and `min-w-0` so long URLs (like the OAuth redirect URI) truncate properly inside flex containers instead of overflowing the viewport
- **StepCard**: Reduce left padding on mobile (`pl-10 sm:pl-12`) and add `min-w-0` on content container to prevent child content from blowing out card width

Fixes are in shared components so all setup pages benefit, not just `/setup/google-drive`.

## Test plan
- [ ] Open `/setup/google-drive` at 375px viewport width — no horizontal scrollbar
- [ ] Verify the long redirect URI in Step 4 truncates with ellipsis
- [ ] Check other setup pages (`/setup/instagram`, `/setup/meta-developer`) still look correct
- [ ] Verify copy button still works after truncation fix

Closes #215

### DIFF
--- a/landing/src/components/setup/copy-button.tsx
+++ b/landing/src/components/setup/copy-button.tsx
@@ -21,11 +21,11 @@ export function CopyButton({ value, className }: CopyButtonProps) {
   return (
     <span
       className={cn(
-        "inline-flex items-center gap-2 rounded-md bg-muted px-3 py-1.5 font-mono text-sm",
+        "inline-flex max-w-full items-center gap-2 rounded-md bg-muted px-3 py-1.5 font-mono text-sm",
         className
       )}
     >
-      <span className="truncate">{value}</span>
+      <span className="min-w-0 truncate">{value}</span>
       <button
         onClick={handleCopy}
         className="shrink-0 text-muted-foreground transition-colors hover:text-foreground"
--- a/landing/src/components/setup/step-card.tsx
+++ b/landing/src/components/setup/step-card.tsx
@@ -14,12 +14,12 @@ export function StepCard({
   className,
 }: StepCardProps) {
   return (
-    <div className={cn("relative pl-12", className)}>
+    <div className={cn("relative pl-10 sm:pl-12", className)}>
       <div className="absolute left-0 top-0.5 flex h-8 w-8 items-center justify-center rounded-full bg-primary text-sm font-bold text-primary-foreground">
         {number}
       </div>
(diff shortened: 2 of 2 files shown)


chrisrogers37 (commented):
Clean, surgical fix. Correct approach throughout.

**Nit:** CHANGELOG.md not updated — CI requires an entry under `## [Unreleased]` before merge.

**Why the changes are right:**
- `min-w-0` on the inner `<span>` in `copy-button.tsx` is the canonical fix for flex truncation — `truncate` only works when the element can actually shrink, and `min-w-0` unlocks that
- `max-w-full` on the outer wrapper correctly caps the component at its container width
- `pl-10 sm:pl-12` recovers 8px on narrow viewports while still clearing the `w-8` step circle
- `min-w-0` on the StepCard content div prevents child content from blowing out the card width
- Both fixes are in shared components, so all setup pages benefit without duplication

All checks pass (Lint, Test, Security, Vercel). One nit on the changelog — otherwise ready to merge.

PR #224, opened 2026-04-16
Fix schedule step completion state for new users

### LINKED ISSUE #216: Setup wizard: "Set Schedule" step incorrectly shows complete for new users
## Summary

In `SetupWizard`, `isStepComplete(4)` (the "Set Schedule" step) returns `true` whenever `state.posting_hours_end > 0`. Because the default fallback value for `posting_hours_end` is `22` (set in `setup/page.tsx` when the backend is unreachable or returns no data), **this step always shows a green checkmark on first load** — even for brand new users who have never saved a schedule.

## Location

- `src/components/dashboard/setup-wizard.tsx`, line 185: `case 4: return state.posting_hours_end > 0;`
- `src/app/(dashboard)/dashboard/setup/page.tsx`, line 27: `posting_hours_end: 22`

## Current behavior

Step 5 ("Set Schedule") shows a green ✓ in the step indicator immediately on page load, suggesting setup is done before the user has touched anything.

## Expected behavior

The step should only show complete after the user has explicitly saved a schedule, not based on default values.

## Suggested fix

Add a `schedule_configured` boolean to `SetupState` (tracked by the backend), or change the check to compare against a sentinel value that only changes after an explicit save:

```ts
// Option A: add schedule_configured flag (preferred)
case 4: return state.schedule_configured;

// Option B: tighter sentinel — only complete if non-default values were saved
case 4: return state.schedule_configured ?? false;
```

chrisrogers37:
## Summary
- Add local `scheduleSaved` state that's only `true` when the user explicitly saves their schedule or has previously completed onboarding
- `canAdvance()` for step 4 now requires save (was unconditionally `true`)
- `isStepComplete(4)` uses `scheduleSaved` instead of trusting `schedule_configured` from API (which is `true` for new users with defaults)
- Sync `scheduleSaved` in `refreshState()` to prevent drift

## Test plan
- [ ] New user: schedule step should NOT show green checkmark before saving
- [ ] New user: "Next" button on schedule step should be disabled until save
- [ ] After saving schedule: step shows complete and Next is enabled
- [ ] Returning onboarded user: schedule step shows as complete on load
- [ ] Returning mid-setup user (saved schedule but didn't complete): step shows incomplete (must re-confirm)

Closes #216

### DIFF
--- a/landing/src/components/dashboard/setup-wizard.tsx
+++ b/landing/src/components/dashboard/setup-wizard.tsx
@@ -71,12 +71,20 @@ export function SetupWizard({ initialState }: SetupWizardProps) {
   const [postsPerDay, setPostsPerDay] = useState<number>(state.posts_per_day);
   const [hoursStart, setHoursStart] = useState(String(state.posting_hours_start));
   const [hoursEnd, setHoursEnd] = useState(String(state.posting_hours_end));
+  const [scheduleSaved, setScheduleSaved] = useState(
+    initialState.schedule_configured === true && initialState.onboarding_completed
+  );
 
   async function refreshState() {
     setError(null);
     try {
       const data = await postApi("init");
-      if (data?.setup_state) setState(data.setup_state);
+      if (data?.setup_state) {
+        setState(data.setup_state);
+        if (data.setup_state.schedule_configured && data.setup_state.onboarding_completed) {
+          setScheduleSaved(true);
+        }
+      }
       return data?.setup_state as SetupState | undefined;
     } catch (e) {
       setError(e instanceof Error ? e.message : "Operation failed");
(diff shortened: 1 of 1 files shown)


chrisrogers37 (commented):
Correct fix. The logic is sound across all scenarios.

**How it works:**
- `scheduleSaved` initializes to `true` only when `schedule_configured && onboarding_completed` — returning onboarded users see step 4 as complete, new users don't
- `inferStep` uses `onboarding_completed` (not `schedule_configured`) to land new users on step 4, so they arrive with `scheduleSaved = false` and can't advance until they explicitly save
- Save handler sets `scheduleSaved(true)` and advances to step 5 ✓
- `refreshState()` sync is one-directional (only sets to `true`), which is correct — no risk of resetting an in-session save

**Scenarios verified:**
- New user: step 4 shows incomplete, Next disabled ✓
- After saving: step complete, Next enabled ✓
- Returning onboarded user: step shows complete on load ✓
- Returning mid-setup (saved but not completed): must re-confirm — intentional per PR description ✓

**Nit:** CHANGELOG.md not updated — same CI failure as seen on nearby PRs. Add an entry under `## [Unreleased]` before merge.

All other checks pass (Lint, Test, Security, Vercel). Ready to merge after changelog update.

PR #226, opened 2026-04-16
Fix Google Drive token health check false positive

chrisrogers37:
## Summary
- `check_token_health_for_chat()` now queries for the `oauth_refresh` token alongside the `oauth_access` token. If a valid refresh token exists, the access token expiry is normal (Google auto-refreshes it) — reports `valid=True, auto_refreshable=True`
- `check_gdrive_token_for_chat()` skips critical/warning expiry alerts when `auto_refreshable` is True
- `format_token_alert()` only sends scary "expired/paused" messages when reconnection is actually required (no refresh token, or refresh token also expired)
- Returns two new fields: `auto_refreshable` and `refresh_token_exists` for downstream consumers

**Root cause**: Google OAuth access tokens expire every ~1 hour (normal). The posting flow in `get_user_credentials()` constructs a `Credentials` object with both access + refresh tokens, and Google's library auto-refreshes transparently. The health check was alarming on the access token expiry without checking whether a refresh token existed.

## Test plan
- [x] All 70 existing tests pass (0 failures)
- [x] New test: access token expired + valid refresh token = healthy (auto-refreshable)
- [x] New test: access token expired + no refresh token = unhealthy
- [x] New test: both tokens expired = unhealthy
- [x] New test: no tokens at all = unhealthy
- [x] Updated test: warning threshold skipped when auto-refreshable
- [x] Updated test: expired alert says "reconnect required" not "paused"

### DIFF
--- a/src/services/core/health_check.py
+++ b/src/services/core/health_check.py
@@ -584,10 +584,12 @@ class HealthCheckService(BaseService):
                 "expires_in_days": None,
             }
 
+        auto_refreshable = token_health.get("auto_refreshable", False)
+
         if not token_health["valid"]:
             return {
                 "healthy": False,
-                "message": "Google Drive token expired",
+                "message": "Google Drive token expired — reconnect required",
                 "expires_in_days": 0,
             }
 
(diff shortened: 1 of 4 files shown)


chrisrogers37 (commented):
Correct fix. Well-reasoned root cause analysis and solid implementation.

## Why it's right

The root cause is accurate: Google OAuth access tokens expire every ~1h by design, but `get_user_credentials()` constructs a `Credentials` object with both access + refresh tokens, so the library auto-refreshes transparently. The health check was alarming on normal behavior.

**`token_refresh.py::check_token_health_for_chat()`** — logic traces correctly across all scenarios:
- Access expired + valid refresh → `valid=True, auto_refreshable=True` ✓
- Access expired + no refresh → `valid=False, auto_refreshable=False`, error message clarifies reconnect needed ✓
- Both expired → `valid=False, auto_refreshable=False, refresh_token_exists=True` ✓
- No access token → early return, second DB query skipped ✓

**`check_gdrive_token_for_chat()`** — `auto_refreshable` defaults to `False` via `.get()` so callers without the field behave identically to before. Expiry warning/critical blocks correctly wrapped in `if not auto_refreshable`. ✓

**`format_token_alert()`** — gates on `healthy=True → return None`, so no alert fires when `auto_refreshable=True`. Alert only reaches users when reconnection is genuinely required. The updated message ("reconnect required" vs "paused") gives clearer action guidance. ✓

**Tests** — all 4 new scenarios covered with correct assertions. `side_effect` lambda to differentiate access vs refresh token fetch is the right approach. Existing tests updated to reflect new dict shape. Test pass count confirmed in CI (all green).

**Security** — refresh token is fetched only to check `is_expired`; the token value is not returned in the health dict or logged. ✓

## Nit
CHANGELOG.md not updated — CI check fails (same as #223–225). Add an entry under `## [Unreleased]` before merge.

All other checks pass (Lint, Test, Security, Vercel, GitGuardian). Ready to merge after changelog update.

PR #237, opened 2026-04-17
feat: multi-account data layer (Phase 1a)

### LINKED ISSUE #231: Multi-account: Migration + model + repository (Phase 1a)
## Summary

Add the data layer for multi-account support. This is additive only — no behavior changes, no risk to existing functionality.

Part of the [multi-account dashboard migration](../planning/multi-account-dashboard.md).

## Tasks

- [ ] Migration 023: Create `user_chat_memberships` table
  ```sql
  CREATE TABLE user_chat_memberships (
      id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
      user_id UUID NOT NULL REFERENCES users(id),
      chat_settings_id UUID NOT NULL REFERENCES chat_settings(id),
      instance_role VARCHAR(20) NOT NULL DEFAULT 'member',  -- 'owner', 'admin', 'member'
      joined_at TIMESTAMP NOT NULL DEFAULT NOW(),
      is_active BOOLEAN NOT NULL DEFAULT TRUE,
      UNIQUE(user_id, chat_settings_id)
  );
  ```
- [ ] Migration 023: Create `onboarding_sessions` table
  ```sql
  CREATE TABLE onboarding_sessions (
      id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
      user_id UUID NOT NULL REFERENCES users(id),
      step VARCHAR(50) NOT NULL DEFAULT 'naming',
      pending_instance_name VARCHAR(100),
      pending_chat_settings_id UUID REFERENCES chat_settings(id),
      expires_at TIMESTAMP NOT NULL,
      created_at TIMESTAMP NOT NULL DEFAULT NOW(),
      UNIQUE(user_id)
  );
  ```
- [ ] Migration 023: Add `display_name VARCHAR(100)` to `chat_settings`
- [ ] Migration 023: Add index on `user_interactions(user_id, telegram_chat_id)` for backfill performance
- [ ] `UserChatMembership` model (~follows existing model patterns)
- [ ] `OnboardingSession` model
- [ ] `UserChatMembershipRepository` with `get_for_user()`, `get_for_chat()`, `create_membership()`, `deactivate()` (~120 LOC)
- [ ] `OnboardingSessionRepository` with basic CRUD
- [ ] Auto-create membership hook in `TelegramService._get_or_create_user()` for group chat interactions
- [ ] `DashboardService.get_user_instances(telegram_user_id)` — JOIN memberships → chat_settings, aggregate media counts and last post (~40 LOC)

## Estimate

~400 LOC, 1 PR, low risk. Follows existing multi-tenant patterns established in migration 014.

## Dependencies

None — this is the first phase.

## Notes

- Field is `instance_role` (not `role`) to avoid collision with `users.role` (system-level concept)
- `display_name` lives only on `chat_settings` (one canonical name per instance, not per-user)
- Auto-create hook only fires for group chats (`telegram_chat_id < 0`), never DMs

chrisrogers37:
## Summary

- Migration 023: `user_chat_memberships` table, `onboarding_sessions` table, `display_name` on `chat_settings`, backfill index on `user_interactions`
- `UserChatMembership` + `OnboardingSession` models with proper constraints and relationships
- `MembershipRepository` (get_for_user with joinedload, create_membership with idempotent upsert, deactivate_for_chat) + `OnboardingSessionRepository`
- Auto-create membership hook in `TelegramService._get_or_create_user()` for group chat interactions, with process-local cache to avoid hot-path DB overhead
- `DashboardService.get_user_instances(telegram_user_id)` returns all instances with stats for the instance picker

Closes #231

## Test plan

- [x] All 1584 existing tests pass (no regressions)
- [ ] Apply migration 023 on staging/prod
- [ ] Verify `user_chat_memberships` table created with correct constraints
- [ ] Verify membership auto-created on group chat interaction
- [ ] Run Phase 1b backfill script after deploy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Multi-account data layer** (#231) — Foundation for multi-account dashboard support. Users can now belong to multiple chat instances via `user_chat_memberships` join table. Memberships are auto-created on group chat interactions. New `DashboardService.get_user_instances()` returns all instances a user belongs to with per-instance stats. Also adds `onboarding_sessions` table for future DM onboarding flow and `display_name` column on `chat_settings`.
 - **Vercel deployment guide** — Documented all required env vars for `landing/` Vercel deployment in `documentation/guides/landing-vercel-deployment.md`.
 
 ### Fixed
--- a/scripts/migrations/023_multi_account_data_layer.sql
+++ b/scripts/migrations/023_multi_account_data_layer.sql
@@ -1,0 +1,79 @@
+-- Migration 023: Multi-account data layer
+--
+-- Phase 1a of multi-account dashboard migration.
+-- Adds:
+--   1. user_chat_memberships table (user ↔ chat_settings join table)
+--   2. onboarding_sessions table (DM onboarding state machine)
+--   3. display_name column on chat_settings
+--   4. Backfill index on user_interactions for future migration script
+--
+-- All changes are additive. No existing data modified.
+
+BEGIN;
+
+-- ============================================================
+-- 1. user_chat_memberships
+-- ============================================================
+
+CREATE TABLE user_chat_memberships (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id UUID NOT NULL REFERENCES users(id),
+    chat_settings_id UUID NOT NULL REFERENCES chat_settings(id),
+    instance_role VARCHAR(20) NOT NULL DEFAULT 'member',
+    joined_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    is_active BOOLEAN NOT NULL DEFAULT TRUE,
+    CONSTRAINT unique_user_chat_membership UNIQUE (user_id, chat_settings_id),
+    CONSTRAINT check_instance_role CHECK (instance_role IN ('owner', 'admin', 'member'))
+);
+
+CREATE INDEX idx_ucm_user_id ON user_chat_memberships(user_id)
+    WHERE is_active = TRUE;
+
+CREATE INDEX idx_ucm_chat_settings_id ON user_chat_memberships(chat_settings_id)
+    WHERE is_active = TRUE;
+
+-- ============================================================
+-- 2. onboarding_sessions
+-- ============================================================
+
+CREATE TABLE onboarding_sessions (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id UUID NOT NULL REFERENCES users(id),
+    step VARCHAR(50) NOT NULL DEFAULT 'naming',
+    pending_instance_name VARCHAR(100),
+    pending_chat_settings_id UUID REFERENCES chat_settings(id),
+    expires_at TIMESTAMP NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    CONSTRAINT unique_active_onboarding UNIQUE (user_id),
+    CONSTRAINT check_onboarding_step CHECK (step IN ('naming', 'awaiting_group', 'complete'))
+);
+
+-- ============================================================
+-- 3. display_name on chat_settings
+-- ============================================================
+
+ALTER TABLE chat_settings ADD COLUMN display_name VARCHAR(100);
+
+-- (Section 4 — backfill index — runs after COMMIT below because
+--  CREATE INDEX CONCURRENTLY cannot run inside a transaction block.)
+
+-- ============================================================
+-- Record migration
+-- ============================================================
+
+INSERT INTO schema_version (version, description, applied_at)
+VALUES (23, 'Multi-account data layer: user_chat_memberships, onboarding_sessions, display_name', NOW());
+
+COMMIT;
+
+-- ============================================================
+-- 4. Backfill index on user_interactions (outside transaction —
+--    CREATE INDEX CONCURRENTLY cannot run inside a transaction block)
+-- ============================================================
+
+-- Supports the Phase 1b backfill query:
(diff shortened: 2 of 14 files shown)


chrisrogers37:
## Review

**Intent**: Phase 1a data layer for multi-account dashboard — migration 023 (user_chat_memberships, onboarding_sessions, display_name), two new models with repositories, auto-create membership hook in `_get_or_create_user`, and `DashboardService.get_user_instances()`. Strictly additive, no existing data touched.

**CI**: All checks pass. **Plan compliance**: All six Phase 1a checkboxes delivered. Migration is correct and uses `CREATE INDEX CONCURRENTLY` outside the transaction block as required.

---

## Blockers

### 1. Unbounded history fetch in `get_user_instances`
`dashboard_service.py:57-62`

```python
recent = self.history_repo.get_recent_posts(
    hours=720, chat_settings_id=cs_id
)
if recent:
    last_post_at = recent[0].posted_at.isoformat()
```

`hours=720` is 30 days. `get_recent_posts` calls `.all()` — for an active instance posting 10/day, this loads ~300 `PostingHistory` rows into memory to use exactly one (`recent[0]`). The query already orders by `posted_at.desc()`, so only the first row is needed.

**Fix**: Add a `limit` parameter to `get_recent_posts` in `history_repository.py` and pass `limit=1` here. Or add a dedicated `get_last_post_at(chat_settings_id)` method that uses `.first()`.

---

## Suggestions

### 2. TOCTOU race in `create_membership`
`membership_repository.py:65-85`

The SELECT → INSERT pattern isn't protected at the DB level. Two concurrent requests from the same user in the same group can both pass the `get_membership()` check and collide on the UNIQUE constraint. The `except Exception` in `_ensure_membership` would swallow the IntegrityError silently, which is harmless in practice, but the method's contract ("idempotent upsert") isn't DB-level idempotent.

**Fix**: Wrap the INSERT in `try/except IntegrityError` and re-fetch:
```python
from sqlalchemy.exc import IntegrityError

try:
    membership = UserChatMembership(...)
    self.db.add(membership)
    self.db.commit()
    self.db.refresh(membership)
    return membership
except IntegrityError:
    self.db.rollback()
    return self.get_membership(user_id, chat_settings_id)
```

### 3. Service layer breach in `_ensure_membership`
`telegram_service.py:437`

```python
chat_settings = self.settings_service.settings_repo.get_by_chat_id(telegram_chat_id)
```

Accessing `settings_service.settings_repo` directly bypasses the service abstraction — any future change to the repository internals won't be picked up here. The intent is correct (read-only lookup, no phantom row creation), but the call site shouldn't know about `settings_repo`.

**Fix**: Add `get_settings_if_exists(chat_id: int) -> Optional[ChatSettings]` to `SettingsService` — wraps `self.settings_repo.get_by_chat_id(chat_id)` and is explicitly documented as never creating a row.

### 4. `_known_memberships` cache will break re-addition in Phase 2b
`telegram_service.py:68, 440`

The in-memory set never invalidates. When Phase 2b's `deactivate_for_chat` fires (bot kicked), cache entries for that chat remain. If the bot is re-added and a user interacts, `_ensure_membership` hits the cache and skips the DB check — the reactivation path in `create_membership` is never reached.

**Fix**: Document this dependency now with a comment on `_known_memberships`. Before Phase 2b merges, the `my_chat_member` kicked handler must evict entries:
```python
self._known_memberships = {k for k in self._known_memberships if k[1] != telegram_chat_id}
```

---

## Nits

**5.** `telegram_service.py:453` — `logger.debug` swallows real DB errors silently. After #2 is applied, log IntegrityError at debug, everything else at warning.

**6.** `user_chat_membership.py:36,43` — `index=True` on both FK columns implies full-table indexes, but the migration creates partial indexes (`WHERE is_active = TRUE`). If `Base.metadata.create_all()` is used in test setup, SQLAlchemy would create additional unconditional indexes. Consider dropping `index=True` from the ORM columns since the migration owns the index definitions.

**7.** `membership_repository.py:65` — Docstring "No-op if one already exists" is incomplete — the method also reactivates inactive memberships. Suggested: "Returns existing membership unchanged if active; reactivates it if inactive."

---

## Question

**8.** `settings_service.settings_repo.get_by_chat_id` — CI passes so this method exists, but it's not visible in this diff. Is it a long-standing method or was it added as part of this PR outside the diff? If the latter, it should appear here.

---

**Verdict: Request Changes** — Blocker #1 (unbounded history fetch) must be fixed before merge. The rest is solid: migration is safe and correct, models/repos follow codebase conventions, the auto-create hook properly gates on `telegram_chat_id < 0`, and service decomposition is clean. Suggestions #3 and #4 are worth addressing but don't need to block merge if #1 is resolved.

chrisrogers37:
## Re-review — all prior feedback resolved

Verified the fix commit (76642bf) against each item from my previous review:

**Blocker #1 — Unbounded history fetch** ✓
`get_recent_posts` now accepts an optional `limit` parameter. `get_user_instances` passes `limit=1`. Fetches 1 row instead of the full 30-day history per instance.

**Suggestion #2 — TOCTOU race in `create_membership`** ✓
`IntegrityError` is correctly caught around `db.commit()` (not `db.add()`). On conflict: rollback, re-fetch, return existing row. The uncommitted object is never refreshed on the error path. Clean.

**Suggestion #3 — Service layer breach** ✓
`SettingsService.get_settings_if_exists()` added with explicit doc ("must not create phantom rows"). `_ensure_membership` now calls through the service layer instead of reaching into `settings_repo` directly.

**Suggestion #4 — `_known_memberships` cache invalidation** ✓
Comment added documenting the Phase 2b eviction requirement. Clear enough for the 2b implementer.

**Nits** ✓ — `logger.warning` for non-IntegrityError failures, `index=True` dropped from ORM FK columns (migration owns the partial indexes), `create_membership` docstring updated.

CI still all green on the new commit.

**Approved.** The data layer is solid and ready to merge.

PR #240, opened 2026-04-17
feat: multi-account Phase 2b — group linking + instance management

### LINKED ISSUE #234: Multi-account: Group linking + event handlers (Phase 2b)
## Summary

Implement the mechanisms for linking group chats to instances and handling bot lifecycle events (added/kicked).

Part of the [multi-account dashboard migration](../planning/multi-account-dashboard.md).

### my_chat_member Handler (Primary Linking Mechanism)

- [ ] Register `ChatMemberUpdated` handler
- [ ] **Bot added to group:** check for pending `onboarding_sessions` in `awaiting_group` state for the user who added it (`update.my_chat_member.from_user`). Auto-link if found. Create `chat_settings` + `user_chat_membership` (instance_role=owner).
- [ ] **Bot kicked from group:** mark all `user_chat_memberships` for that `chat_settings` as `is_active = false`. Preserve `chat_settings` row (data not deleted, can restore if bot re-added).
- [ ] Race condition guard: retry with 2s delay if `onboarding_sessions` row not yet committed

### startgroup Deep Link

- [ ] Add `context.args` parsing to `/start` handler: `if context.args and context.args[0].startswith("setup_"):`
- [ ] Payload format: `setup_{onboarding_sessions.id}` (42 chars, within Telegram's 64-char limit)
- [ ] Note: this silently fails if bot is already in the group — `my_chat_member` handler and `/link` cover that case

### /link Fallback Command

- [ ] `/link <session_id>` command for manual group-to-instance linking
- [ ] Verify bot is a member of the group via `get_chat()`
- [ ] Handles the "bot already in group" case where `startgroup` won't fire

### Instance Management Commands

- [ ] `/name <name>` in group chat — set `chat_settings.display_name`
- [ ] `/instances` in DM — list + manage all instances
- [ ] `/new` in DM — shortcut to create a new instance

### Onboarding Cleanup

- [ ] Onboarding session timeout: scheduler loop checks for expired rows, sends Telegram reminder, deletes row

## Estimate

~400 LOC, 1 PR, medium risk.

## Dependencies

Phase 2a (get_settings split + /start refactor)

## Design Decision

`my_chat_member` is the primary group-linking mechanism, not `startgroup`. It fires regardless of how the bot was added (deep link, manual add, invite link). `startgroup` is a convenience, `/link` is the manual fallback.

chrisrogers37:
## Summary

- **my_chat_member handler** (PRIMARY linking): auto-links pending onboarding sessions when bot is added to a group (with 2s race condition retry), deactivates all memberships + evicts `_known_memberships` cache when bot is kicked
- **Shared `link_group_to_instance()` helper** on ConversationService — eliminates triplicated linking flow across startgroup deep link, my_chat_member, and /link command
- **/link fallback command**: manual group-to-instance linking when bot is already in the group (startgroup/my_chat_member won't fire)
- **Instance management commands**: `/name <name>` (group — set display_name), `/instances` (DM — list with manage buttons), `/new` (DM — create new instance shortcut)
- **Onboarding session cleanup**: expired sessions cleaned up hourly via scheduler loop
- **17 new unit tests** covering all handlers, edge cases, and scheduler cleanup

Resolves #234

## Test plan
- [x] All 1604 tests pass (17 new + 1587 existing)
- [ ] Manual: DM bot → /new → name instance → add to group → verify my_chat_member auto-links
- [ ] Manual: DM bot → /new → name instance → /link in group where bot already exists
- [ ] Manual: /name in group sets display_name
- [ ] Manual: /instances in DM shows instance list
- [ ] Manual: kick bot from group → verify memberships deactivated

### DIFF
--- a/src/services/core/conversation_service.py
+++ b/src/services/core/conversation_service.py
@@ -72,6 +72,10 @@ class ConversationService(BaseService):
 
         Used by startgroup deep link, my_chat_member, and /link command.
         Returns the chat_settings record.
+
+        Idempotent: MembershipRepository.create_membership handles duplicates
+        via upsert, so partial failures (e.g. update_step fails after
+        create_membership succeeds) are safe to retry.
         """
         from src.services.core.settings_service import SettingsService
 
@@ -82,6 +86,7 @@ class ConversationService(BaseService):
                     chat_id, "display_name", session.pending_instance_name
                 )
 
+        # Idempotent: returns existing membership if already created
         membership_repo.create_membership(
             user_id=user_id,
             chat_settings_id=str(chat_settings.id),
--- a/src/services/core/conversation_service.py
+++ b/src/services/core/conversation_service.py
@@ -60,6 +60,42 @@ class ConversationService(BaseService):
             pending_chat_settings_id=chat_settings_id,
         )
 
+    def link_group_to_instance(
+        self,
+        session: "OnboardingSession",
+        chat_id: int,
+        user_id: str,
+        membership_repo,
+    ) -> "ChatSettings":
+        """Shared linking flow: create chat_settings, set display_name,
+        create owner membership, complete onboarding session.
+
+        Used by startgroup deep link, my_chat_member, and /link command.
+        Returns the chat_settings record.
+        """
+        from src.services.core.settings_service import SettingsService
+
+        with SettingsService() as settings_service:
+            chat_settings = settings_service.get_settings(chat_id)
+            if session.pending_instance_name:
+                settings_service.update_setting(
+                    chat_id, "display_name", session.pending_instance_name
+                )
+
+        membership_repo.create_membership(
+            user_id=user_id,
+            chat_settings_id=str(chat_settings.id),
+            instance_role="owner",
+        )
+
+        self.onboarding_repo.update_step(
+            session_id=str(session.id),
+            step="complete",
+            pending_chat_settings_id=str(chat_settings.id),
+        )
+
+        return chat_settings
+
     def cleanup_expired(self) -> int:
         """Delete expired onboarding sessions. Call from scheduler loop."""
         count = self.onboarding_repo.delete_expired()
--- a/src/services/core/telegram_commands.py
+++ b/src/services/core/telegram_commands.py
@@ -510,6 +510,9 @@ class TelegramCommandHandlers:
 
         Used when the bot is already in the group (so startgroup/my_chat_member
         won't fire). Links the current group to the user's pending onboarding session.
+
+        Note: Does not accept a <session_id> argument (intentional deviation from spec).
+        Session lookup is always scoped to the calling user.
         """
         chat_id = update.effective_chat.id
         is_group = update.effective_chat.type in ("group", "supergroup")
@@ -526,7 +529,13 @@ class TelegramCommandHandlers:
 
         # Verify bot is actually a member of this group
         try:
-            await context.bot.get_chat(chat_id)
+            bot_member = await context.bot.get_chat_member(chat_id, context.bot.id)
+            if bot_member.status in ("left", "kicked"):
+                await update.message.reply_text(
+                    "⚠️ I'm not a member of this group. "
+                    "Add me first, then run /link again."
+                )
+                return
         except Exception:
             await update.message.reply_text(
                 "⚠️ I can't verify my membership in this group. "
--- a/src/services/core/telegram_commands.py
+++ b/src/services/core/telegram_commands.py
@@ -503,6 +503,223 @@ class TelegramCommandHandlers:
             parse_mode="Markdown",
         )
 
+    # ==================== Multi-Account Commands ====================
+
+    async def handle_link(self, update, context):
+        """Handle /link — manual group-to-instance linking fallback.
+
+        Used when the bot is already in the group (so startgroup/my_chat_member
+        won't fire). Links the current group to the user's pending onboarding session.
+        """
+        chat_id = update.effective_chat.id
+        is_group = update.effective_chat.type in ("group", "supergroup")
+
+        if not is_group:
+            await update.message.reply_text(
+                "Run /link in the group chat you want to link."
+            )
+            return
+
+        user = self.service._get_or_create_user(
+            update.effective_user, telegram_chat_id=chat_id
+        )
+
+        # Verify bot is actually a member of this group
+        try:
+            await context.bot.get_chat(chat_id)
+        except Exception:
+            await update.message.reply_text(
+                "⚠️ I can't verify my membership in this group. "
+                "Try removing and re-adding me."
+            )
+            return
+
+        from src.services.core.conversation_service import ConversationService
+
+        with ConversationService() as conv_service:
+            session = conv_service.get_current_session(str(user.id))
+
+        if not session or session.step != "awaiting_group":
+            await update.message.reply_text(
+                "⚠️ No pending instance setup found.\n\n"
+                "Start a new instance by DMing me and running /start."
+            )
+            return
+
+        with ConversationService() as conv_service:
+            conv_service.link_group_to_instance(
+                session=session,
+                chat_id=chat_id,
+                user_id=str(user.id),
+                membership_repo=self.service.membership_repo,
+            )
+
+        name = session.pending_instance_name or "this group"
+        await update.message.reply_text(
+            f"✅ *{name}* is linked to this group!\n\n"
+            "Use /status to check health, or /setup to configure.",
+            parse_mode="Markdown",
+        )
+
+        # Notify in DM
+        try:
+            await self.service.bot.send_message(
+                chat_id=update.effective_user.id,
+                text=(
+                    f"✅ *{name}* is linked!\n\n"
+                    "Use /start here to manage your instances."
+                ),
+                parse_mode="Markdown",
+            )
+        except Exception:
+            pass
+
+        self.service.interaction_service.log_command(
+            user_id=str(user.id),
+            command="/link",
+            telegram_chat_id=chat_id,
+            telegram_message_id=update.message.message_id,
+        )
+
+    async def handle_name(self, update, context):
+        """Handle /name <name> — set display name for this group's instance."""
+        chat_id = update.effective_chat.id
+        is_group = update.effective_chat.type in ("group", "supergroup")
+
+        if not is_group:
+            await update.message.reply_text(
+                "Run /name in a group chat to set that instance's name."
+            )
+            return
+
+        user = self.service._get_or_create_user(
+            update.effective_user, telegram_chat_id=chat_id
+        )
+
+        name = " ".join(context.args).strip()[:100] if context.args else ""
+        if not name:
+            await update.message.reply_text(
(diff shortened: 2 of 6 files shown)


src/services/core/telegram_service.py
```diff
@@ -397,6 +414,114 @@ async def _handle_onboarding_message(self, update, context):
...
+            ):
+                self._handle_bot_removed_from_group(chat)
+        except Exception as e:
+            logger.error(
+                f"my_chat_member handler error for chat {chat.id}: "
+                f"{type(e).__name__}: {e}",
+                exc_info=True,
+            )
+        finally:
+            self.cleanup_transactions()
```
chrisrogers37, line 461: **Suggestion: Guard against `from_user=None`**

In groups with anonymous admin mode, Telegram sends `my_chat_member` events with `from_user=None`. The outer `try/except` catches the resulting `AttributeError` from `_get_or_create_user(None)`, but logs a generic error and silently drops the auto-link.

Add at the top of this method:
```python
if from_user is None:
    logger.info(f"Bot added to group {chat.id} by anonymous admin — skipping auto-link (use /link)")
    return
```

src/services/core/conversation_service.py
```diff
@@ -60,6 +60,42 @@ def link_group(
...
+        membership_repo,
+    ) -> "ChatSettings":
+        """Shared linking flow: create chat_settings, set display_name,
+        create owner membership, complete onboarding session.
+
+        Used by startgroup deep link, my_chat_member, and /link command.
+        Returns the chat_settings record.
+        """
+        from src.services.core.settings_service import SettingsService
+
```
chrisrogers37, line 81: **Suggestion: Non-atomic — orphaned state on partial failure**

`create_membership` and `update_step` are two separate DB operations. If `create_membership` succeeds but `update_step` fails, the session stays `awaiting_group` and the next call to `create_membership` raises a unique constraint violation — user is stuck.

Simplest fix: add `ON CONFLICT DO NOTHING` to `MembershipRepository.create_membership`. Makes the whole helper idempotent.

src/services/core/telegram_commands.py
```diff
@@ -503,6 +503,223 @@ async def handle_removed_command(self, update, context):
...
+
+        # Verify bot is actually a member of this group
+        try:
+            await context.bot.get_chat(chat_id)
+        except Exception:
+            await update.message.reply_text(
+                "⚠️ I can't verify my membership in this group. "
+                "Try removing and re-adding me."
+            )
+            return
```
chrisrogers37, line 544: **Suggestion: `get_chat()` doesn't verify membership for public groups**

`get_chat()` succeeds for any public group regardless of bot membership. For private groups this works, for public groups it's a no-op.

To actually verify: `await context.bot.get_chat_member(chat_id, context.bot.id)` and check `member.status`. If this is intentionally a light sanity check, add a comment explaining the limitation so the next reader doesn't wonder why.

src/services/core/telegram_commands.py
```diff
@@ -503,6 +503,223 @@ async def handle_removed_command(self, update, context):
             parse_mode="Markdown",
         )
 
+    # ==================== Multi-Account Commands ====================
```
chrisrogers37, line 506: **Nit: Document spec divergence**

Plan doc specifies `/link <session_id>` but this ignores `context.args` and uses the caller's current session. The current approach is better (simpler, no session ID enumeration). Add to docstring:

```
Note: Does not accept a <session_id> argument (intentional deviation from spec).
Session lookup is always scoped to the calling user.
```

tests/src/services/test_phase2b_handlers.py
```diff
@@ -0,0 +1,471 @@
...
+        """Running /new in group tells user to use DM."""
+        handlers = mock_command_handlers
+        mock_update = _make_update(-100123, chat_type="group")
+
+        await handlers.handle_new(mock_update, Mock())
+
+        msg = mock_update.message.reply_text.call_args.args[0]
+        assert "DM" in msg
+
+
```
chrisrogers37, line 515: **Nit: This test is a no-op**

```python
mock_cleanup()
mock_cleanup.assert_called_once()
```

Calls a mock and asserts it was called — always passes, tests nothing. Either test the actual scheduler tick in `main.py` (patch `ConversationService` and exercise the `retention_tick_counter == 0` branch) or remove it.

src/services/core/telegram_service.py
```diff
@@ -397,6 +414,114 @@ async def _handle_onboarding_message(self, update, context):
...
+        from_user = member_update.from_user
+
+        try:
+            if new_status in ("member", "administrator") and old_status in (
+                "left",
+                "kicked",
+            ):
+                await self._handle_bot_added_to_group(chat, from_user)
+            elif new_status in ("left", "kicked") and old_status in (
+                "member",
```
chrisrogers37, line 450: **Missing test: retry path is completely untested**

The 2s retry is the whole point of the race condition guard, but there's no test for it. Suggested test: mock `get_current_session` to return `None` on call 1 and a valid `awaiting_group` session on call 2. Assert `link_group_to_instance` is still called.

chrisrogers37 (commented):
# Phase 2b Review — Group Linking + Instance Management

**Verdict: Changes Needed** — 0 blockers, 3 suggestions, 3 nits

The overall approach is solid. The shared `link_group_to_instance` helper is a clear improvement over triplicated inline code, the `my_chat_member` status transitions are correct, and the `/link` security model properly scopes sessions to the calling user (no hijacking possible). 17 tests cover happy paths well.

Three things need attention before merge:

---

## Suggestions

### 1. `from_user` can be `None` — silent failure for anonymous admins

`telegram_service.py` — `_handle_bot_added_to_group`

In Telegram groups with anonymous admin mode (common in large channels/groups), `my_chat_member.from_user` is `None`. PTB's `ChatMemberUpdated.from_user` is `Optional[User]`. The outer `try/except` catches the resulting `AttributeError` and logs a generic error — the bot-add event is silently dropped, the user's pending session never auto-links.

Fix: guard at the top of `_handle_bot_added_to_group`:

```python
async def _handle_bot_added_to_group(self, chat, from_user):
    if from_user is None:
        logger.info(f"Bot added to group {chat.id} by anonymous admin — skipping auto-link (use /link)")
        return
    ...
```

Users in anonymous-admin groups can still use `/link` as fallback, but the silent exception log is misleading.

---

### 2. `link_group_to_instance` is non-atomic — orphaned state on partial failure

`conversation_service.py` — `link_group_to_instance`

`create_membership` and `update_step` are two separate DB operations with no transaction wrapping them. If `create_membership` succeeds but `update_step` fails (lost connection, etc.), the membership row exists but the session stays in `awaiting_group`. On the next retry, `create_membership` hits `UNIQUE(user_id, chat_settings_id)` and raises — user is stuck.

Simplest fix: add `ON CONFLICT DO NOTHING` in `MembershipRepository.create_membership`. This makes the whole call idempotent and also makes concurrent `my_chat_member` + `startgroup` triggers safe without needing extra coordination.

---

### 3. `get_chat()` in `/link` doesn't verify membership

`telegram_commands.py` — `handle_link`

```python
try:
    await context.bot.get_chat(chat_id)
except Exception:
    await update.message.reply_text("⚠️ I can't verify my membership...")
    return
```

`get_chat()` returns chat info for any public group regardless of whether the bot is a member. This check only works for private groups. For the common case (public supergroup), it's a no-op and gives false confidence.

To actually verify membership: `await context.bot.get_chat_member(chat_id, context.bot.id)` and check the status. If this is intentionally a light sanity guard rather than a security check, add a comment explaining the limitation.

---

## Nits

**N1: `link_group` method is now dead code** — `conversation_service.py:60`

After this PR, all callers use `link_group_to_instance`. The old `link_group(session_id, chat_settings_id)` has no callers. Remove it to avoid future confusion.

**N2: `/link` ignores `<session_id>` arg — spec deviation, undocumented**

Plan doc specifies `/link <session_id>` but the implementation ignores `context.args` and looks up the caller's current session instead. This is actually *more* secure (no session ID enumeration) and simpler. The behavior is right — just add a note in the docstring so the intentional divergence is clear.

**N3: `test_cleanup_expired_called_on_retention_tick` is a no-op** — `test_phase2b_handlers.py`

```python
mock_cleanup()
mock_cleanup.assert_called_once()
```

This calls a mock and asserts it was called — it always passes regardless of any production code. Either test the actual scheduler tick branch in `main.py` (patch `ConversationService` and drive the retention tick), or remove it. A false-green test is worse than no test.

---

## Test Gaps

- **`from_user=None`** — no test that the handler logs and returns cleanly instead of raising
- **Retry path** — no test that covers `get_current_session` returning `None` on attempt 0 then a valid session on attempt 1; the retry is the whole point of the race guard and it's completely untested
- **`/link` with wrong session step** — `test_link_no_pending_session` covers `session=None` but not `session.step="naming"` (session exists, wrong state)

---

## What's Good

- **Status transition logic is correct** — `(left|kicked) → (member|administrator)` for add, `(member|administrator) → (left|kicked)` for kick. Non-group chats correctly ignored. No false trigger on bot promotion (`member → administrator`).
- **Race condition guard is well-reasoned** — `asyncio.sleep(2)` yields to the event loop, allowing a concurrent `startgroup` update to win. If `startgroup` completes first and marks the session complete, `my_chat_member` backs off on its second attempt. Correct.
- **`/link` security is correct** — `get_current_session(user.id)` scopes to the calling user. No way for user A to link a group to user B's pending session.
- **Shared helper is clean** — `startgroup`, `my_chat_member`, and `/link` all go through one path. The refactor of `start_command_router.py` is a genuine improvement.
- **Scheduler cleanup integration** — piggybacking on `retention_tick_counter == 0` is the right pattern, `try/except` with a warning log is the right failure mode.

chrisrogers37 (commented):
## Re-review — PR #240 approved ✅

All 6 items from the previous review are resolved. 6/6.

**S1 — `from_user` null guard** ✅  
Guard added before the `try/except` block in `_handle_my_chat_member` with a clear info log pointing users to `/link`. `test_bot_added_anonymous_admin_skips` verifies the happy path. Clean.

**S2 — `link_group_to_instance` idempotency** ✅  
Verified `MembershipRepository.create_membership`: already a full idempotent upsert — checks for existing membership first, reactivates if inactive, catches `IntegrityError` on the race path and re-fetches. The docstring added in this PR accurately documents that contract. No change needed in the repo.

**S3 — `get_chat_member` verification** ✅  
Now uses `context.bot.get_chat_member(chat_id, context.bot.id)` with explicit `status in ("left", "kicked")` check. Correct and effective for both public and private groups.

**N1 — Spec divergence docstring** ✅  
Docstring now explicitly documents the intentional deviation and scoping rationale. Good.

**N3 — No-op scheduler test** ✅  
Replaced with a test that patches `ConversationService` and drives the context manager protocol. Not a full scheduler integration test, but reasonable for a unit test.

**Missing retry test** ✅  
`test_bot_added_retries_once_on_missing_session` covers the race condition path: `side_effect = [None, mock_session]`, verifies `asyncio.sleep(2)` was called, and confirms `link_group_to_instance` fires after the retry. Exactly right.

---

Code is correct, the shared helper is a clean abstraction, and test coverage is solid. Ready to merge.

PR #241, opened 2026-04-17
feat: multi-account Phase 3+4 — API auth, instance picker, dashboard switcher

chrisrogers37:
## Summary

Completes the multi-account migration Phases 3 and 4, building on the data layer (Phase 1) and bot handlers (Phase 2).

**Phase 3 — API + Auth (#235):**
- `SessionPayload.chatId` → `activeChatId: number | null` — login starts with null
- `GET /api/instances` + backend `/api/onboarding/instances` — lists user's instances with stats
- `POST /api/instances/:id/select` — validates membership, reissues JWT with activeChatId
- BFF proxy rejects when activeChatId is null (422); middleware redirects to `/instances`
- Upload proxy also guards on null activeChatId

**Phase 4 — Frontend (#236):**
- `/instances` page: auto-redirect (1 instance), card picker (multiple), CTA (0 instances)
- Instance switcher dropdown in dashboard header (visible with 2+ instances)
- `/webapp/instances` Mini App entry point for Telegram WebView

**Cleanup from /simplify:**
- Shared `Instance` type, `formatLastPost` util, `SESSION_COOKIE_OPTIONS` constant
- Extracted `_validate_auth()` Python helper (deduplicated auth-only validation)
- Proper `useCallback` deps (removed eslint-disable suppressions)

Closes #235, closes #236

## Test plan
- [ ] 1606 Python tests pass (verified locally)
- [ ] TypeScript compiles clean (verified locally)
- [ ] Login flow: new session gets activeChatId=null → redirected to /instances
- [ ] Single instance: auto-selects and redirects to /dashboard
- [ ] Multiple instances: picker renders with stats, selecting reissues JWT
- [ ] Zero instances: CTA linking to Telegram bot
- [ ] Dashboard header shows switcher dropdown with 2+ instances
- [ ] BFF proxy returns 422 if activeChatId somehow null on dashboard API calls
- [ ] /webapp/instances renders correctly in Telegram WebView

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #246, opened 2026-04-17
BFF proxy: validate activeChatId against active memberships per-request

### LINKED ISSUE #242: BFF proxy: validate activeChatId against active memberships per-request
## Context

Phase 3 plan item: "BFF proxy: validate activeChatId against active memberships on each request."

Currently the BFF proxy (`landing/src/app/api/dashboard/[...path]/route.ts`) only checks `activeChatId !== null`. If a user is removed from a group mid-session, their JWT still carries that `activeChatId` and they can proxy requests for up to 24h (JWT lifetime).

## Options

1. **Per-request backend check** — Call a lightweight membership-check endpoint before forwarding. Adds ~50ms latency per dashboard API call.
2. **Short-lived JWT + refresh** — Reduce JWT lifetime to e.g. 1h and revalidate memberships on refresh. Less per-request overhead.
3. **Cache with TTL** — Cache the user's membership list in-memory or edge KV with a 5-min TTL. Best latency but adds caching complexity.

## Severity

Low — requires both (a) a user being removed from a group AND (b) that user actively using the dashboard within the same session. The JWT expires naturally after 24h.

Flagged during PR #241 review (Phase 3+4).

chrisrogers37:
## Summary
- Adds per-request membership validation to the BFF proxy (`/api/dashboard/[...path]`). Each proxied request now checks that the user's `activeChatId` corresponds to an active membership before forwarding.
- If membership is invalid (user removed from group mid-session), the JWT is reissued with `activeChatId: null` and a 403 is returned with `code: "MEMBERSHIP_INVALID"`, forcing the frontend to redirect to the instance picker.
- Extracts `fetchUserInstances()` into `@/lib/backend.ts`, eliminating duplicated instances-endpoint calls across the proxy, instance select, and instance list routes.
- Gracefully degrades: if the membership check fails (backend unreachable), the request falls through to normal proxy behavior where backend-level auth still applies.

## Design choice
Implements Option 1 from #242 (per-request backend check, ~50ms latency). The check is serial before the proxied request to prevent side effects on POST/PUT/DELETE for unauthorized users.

## Files changed
- `landing/src/lib/backend.ts` — new `fetchUserInstances()` helper
- `landing/src/app/api/dashboard/[...path]/route.ts` — membership validation added
- `landing/src/app/api/instances/route.ts` — refactored to use `fetchUserInstances()`
- `landing/src/app/api/instances/[id]/select/route.ts` — refactored to use `fetchUserInstances()`

## Test plan
- [ ] Log in, select an instance, verify dashboard API calls still work normally
- [ ] Remove user from a group in the backend DB (`is_active = false`), then make a dashboard API call — should get 403 with `MEMBERSHIP_INVALID`
- [ ] Verify the session cookie is reissued without `activeChatId` after the 403
- [ ] Verify subsequent page navigation redirects to `/instances`
- [ ] Verify instance picker and instance selection still work after refactor
- [ ] Verify behavior when backend is unreachable — proxy should fall through gracefully

Closes #242


chrisrogers37 (commented):
## Review: BFF Proxy Per-Request Membership Validation

Clean, well-scoped security PR. The design is correct and the graceful degradation approach is the right call.

**One blocker (CI), two minor points below.**

---

## Blocker

### CHANGELOG.md not updated — CI failing

Add an entry under `## [Unreleased]`.

---

## Security Logic Trace

**Membership check in BFF proxy:**

```typescript
const instances = await fetchUserInstances(session.userId);
if (instances && !instances.some((i) => i.telegram_chat_id === session.activeChatId)) {
  // reissue JWT with activeChatId: null, return 403
}
// instances === null → backend unreachable → fall through
```

All cases correct:
- Active membership → match found → proxy ✓
- User removed → no match → 403 + JWT reissued with `activeChatId: null` ✓
- `instances === null` (backend down) → falls through, backend-level auth on the proxied request still applies ✓
- `fetchUserInstances` throws → caught, falls through ✓
- Empty membership list `[]` → `instances && !some(...)` → true → 403 ✓

**JWT reissue:** `{ ...session, activeChatId: null }` correctly preserves all other session fields. ✓

**`fetchUserInstances`:** calls `backendFetchJson("instances", 0, userId)` which generates `generateUrlToken(0, userId)` — same auth-only convention established in #241. Backend validates via `_validate_auth` (no chat_id check). ✓

**`select/route.ts` refactor:** Old `data.instances?.find()` with optional chaining replaced by `instances.find()` since `fetchUserInstances` guarantees non-null on success. Cleaner. ✓

**`instances/route.ts` behavioral note:** Old code forwarded the backend's original status code for 4xx errors; new code always returns 502. The frontend only checks for `instances === null` and shows a generic error, so no practical impact.

---

## Suggestion

### No frontend handling of `MEMBERSHIP_INVALID` — stale dashboard visible until next navigation

When this 403 fires on a Server Component request, the component gets `null` from `backendFetchJson` and renders with empty/default data. The reissued cookie (with `activeChatId: null`) only triggers the middleware redirect on the **next** page navigation. So the current behavior is: removed user sees a blank dashboard, then redirects to `/instances` when they click anything.

This is acceptable for a security hardening PR — correctness is what matters and the user does get locked out. But a follow-up improvement worth tracking: have the dashboard layout detect a `MEMBERSHIP_INVALID` response from a client-side API call and immediately redirect, or have Server Components check for the 403 code and redirect server-side. Not blocking.

---

## Nit

### `InstanceInfo` in `backend.ts` partially duplicates `Instance` in `types.ts`

```typescript
// backend.ts (new)
interface InstanceInfo {
  telegram_chat_id: number;
  display_name?: string;
  [key: string]: unknown;
}

// types.ts (existing)
export interface Instance {
  chat_settings_id: string;
  telegram_chat_id: number;
  display_name: string;
  // ...
}
```

For a server-side internal type it's reasonable to keep them separate (avoids importing frontend types into a backend utility). The `[key: string]: unknown` index signature is a pragmatic choice. Acceptable as-is, but if the codebase grows more of these, consolidating around `Instance` or shared picks would reduce drift.

---

## Verification

| Item | Result |
|------|--------|
| Lint | ✓ Pass |
| Tests | ✓ Pass (1606) |
| Security Scan | ✓ Pass |
| Vercel | ✓ Pass |
| CHANGELOG | ✗ Not updated |
| Membership check logic | ✓ All cases correct |
| Graceful degradation | ✓ Backend-unreachable falls through safely |
| JWT reissue correctness | ✓ Fields preserved, activeChatId nulled |
| `fetchUserInstances` deduplication | ✓ 3 call sites consolidated |

Add the CHANGELOG entry and this is ready.

chrisrogers37 (commented):
## Re-review: CHANGELOG fixed, all CI green — approved ✓

Single fix commit (`536432e1`) adds the CHANGELOG entry. All checks pass: Lint ✓ · Tests ✓ · Security Scan ✓ · CHANGELOG ✓ · Vercel ✓ · GitGuardian ✓.

No new code changes. Prior review findings all remain resolved. Ready to merge.

PR #261, opened 2026-04-19
fix: narrow broad except Exception catches across src/services/

### LINKED ISSUE #250: [tech-debt] Narrow 85 broad `except Exception` catches in src/services/
## Problem

There are **85 `except Exception` catches** across `src/services/`. While some are intentional (fire-and-forget logging in InteractionService, best-effort notifications marked with `noqa: BLE001`), many swallow errors that could mask bugs or hide operational issues.

## Worst Offenders

- `src/services/core/health_check.py` — 3+ catches that hide health check failures
- `src/services/core/telegram_settings.py` — 4 catches including nested try/except
- `src/services/core/telegram_accounts.py` — 3 catches in account management flows
- `src/services/core/media_sync.py` — 2 catches in sync loop (could hide sync failures)
- `src/services/core/telegram_callbacks_core.py` — 3 catches including bare `except Exception:`

## Suggested Fix

Audit each catch individually:
1. Replace with specific exception types where the failure mode is known (e.g., `OperationalError`, `ValueError`, `httpx.HTTPError`)
2. Keep broad catches only for truly best-effort operations (interaction logging, notifications)
3. Ensure all kept broad catches log at WARNING or ERROR level with `exc_info=True`
4. Add `noqa: BLE001` comments to intentionally broad catches for linter clarity

## Risk

Medium — requires understanding each call site's failure modes. Do NOT batch-replace; review each one.

---
*Auto-generated by tech-debt audit on 2026-04-17*

chrisrogers37:
## Summary

- Audited all 88 `except Exception` catches across `src/services/` (34 files)
- Narrowed 15+ catches to specific exception types (`TelegramError`, `SQLAlchemyError`, `InvalidToken`, `ValueError`, etc.) where failure modes are known
- Added `noqa: BLE001` annotations with justification comments to ~45 intentionally broad catches (best-effort logging, cleanup, health checks, UI updates)
- Added `exc_info=True` to health check error handlers for better debugging
- Fixed 1 test mock raising generic `Exception` instead of `ValueError`

## Test plan

- [x] All 1618 tests pass (0 failures, 16 skipped)
- [x] Each catch individually reviewed — no batch replacements
- [x] Intentionally broad catches documented with `noqa: BLE001` and justification
- [x] CHANGELOG updated

Closes #250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Narrow broad `except Exception` catches** (#250) — Audited 88 `except Exception` catches across `src/services/`. Narrowed catches to specific types (`TelegramError`, `SQLAlchemyError`, `InvalidToken`, etc.) where failure modes are known. Added `noqa: BLE001` annotations with justification comments to intentionally broad catches (best-effort logging, cleanup, health checks). Added `exc_info=True` to health check error handlers.
+
 ### Added
 
 - **AI caption generation** (#182) — New `CaptionService` generates Instagram Story captions using Claude API at queue time. Controlled by per-instance `enable_ai_captions` toggle. Generated captions are stored separately from manual captions on `media_items.generated_caption`, shown with a robot indicator in Telegram review, and include a "Regenerate Caption" button. Skips generation when a manual caption exists or ANTHROPIC_API_KEY is not configured. Non-blocking — API failures never prevent posting. Migration 026 adds the new columns.
--- a/src/services/base_service.py
+++ b/src/services/base_service.py
@@ -47,7 +47,7 @@ class BaseService(ABC):
                     attr.end_read_transaction()
                 elif isinstance(attr, BaseService) and attr is not self:
                     attr.cleanup_transactions()
-            except Exception as e:
+            except Exception as e:  # noqa: BLE001
                 logger.warning(
                     f"[{self.service_name}] Transaction cleanup failed for "
                     f"{attr_name}: {type(e).__name__}: {e}"
@@ -71,7 +71,7 @@ class BaseService(ABC):
                     attr.close()
                 elif isinstance(attr, BaseRepository):
                     attr.close()
-            except Exception as e:
+            except Exception as e:  # noqa: BLE001
                 logger.warning(
                     f"[{self.service_name}] Error closing {attr_name}: "
                     f"{type(e).__name__}: {e}"
@@ -90,7 +90,7 @@ class BaseService(ABC):
         """Cleanup when service is garbage collected."""
         try:
             self.close()
-        except Exception:
+        except Exception:  # noqa: BLE001
             pass  # Suppress errors during garbage collection
 
     @contextmanager
@@ -152,7 +152,7 @@ class BaseService(ABC):
                 f"[{self.service_name}.{method_name}] Completed successfully ({duration_ms}ms)"
             )
 
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001 — must record any failure
             # Failure
             completed_at = datetime.utcnow()
             duration_ms = int((completed_at - started_at).total_seconds() * 1000)
@@ -169,7 +169,7 @@ class BaseService(ABC):
                     stack_trace=stack_trace,
                     duration_ms=duration_ms,
                 )
-            except Exception as log_err:
+            except Exception as log_err:  # noqa: BLE001
                 logger.warning(
                     f"[{self.service_name}.{method_name}] "
                     f"Failed to record service run failure: {log_err}"
--- a/src/services/core/caption_service.py
+++ b/src/services/core/caption_service.py
@@ -2,6 +2,7 @@
 
 from typing import Optional
 
+
 from src.config.constants import MAX_CAPTION_LENGTH
 from src.config.settings import settings
 from src.repositories.media_repository import MediaRepository
@@ -124,7 +125,7 @@ class CaptionService(BaseService):
             if len(caption) > MAX_CAPTION_LENGTH:
                 caption = caption[:MAX_CAPTION_LENGTH]
             return caption
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001 — best-effort caption generation
             logger.error(f"AI caption generation failed: {e}", exc_info=True)
             return None
 
--- a/src/services/core/conversation_service.py
+++ b/src/services/core/conversation_service.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Optional
 
+from sqlalchemy.exc import SQLAlchemyError
+
 from src.models.onboarding_session import OnboardingSession
 from src.repositories.onboarding_repository import OnboardingRepository
 from src.services.base_service import BaseService
@@ -132,7 +134,7 @@ class ConversationService(BaseService):
                             interaction_name=session.step,
                             context={"duration_minutes": duration_minutes},
                         )
-            except Exception:
+            except SQLAlchemyError:
                 logger.warning("Failed to log onboarding dropouts", exc_info=True)
 
         count = self.onboarding_repo.delete_expired()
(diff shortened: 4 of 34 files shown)


chrisrogers37 (commented):
## PR #261 Review — Narrow broad except Exception catches

**Overall**: The narrowing work is thorough and well-executed. 30+ catches properly typed, noqa justifications are legitimate. One correctness issue in scheduler.py needs a fix before merge.

---

### Blocker

**`src/services/core/scheduler.py` — wrong exception types in caption catch**

The caption catch was narrowed to `(anthropic.APIError, OSError)`:

```python
except (anthropic.APIError, OSError) as e:
    logger.warning(f"AI caption generation failed, continuing: {e}")
```

Two problems:
1. **`OSError` is irrelevant** — `generate_caption` is an async HTTP call with no file I/O. `OSError` will never fire here.
2. **`SQLAlchemyError` escapes undetected** — inside `generate_caption`, `self.media_repo.update_metadata()` can raise `SQLAlchemyError`. That gets caught by `track_execution`'s broad except, recorded, then **re-raised** (`raise e` at base_service.py:188). That re-raised `SQLAlchemyError` exits `generate_caption` and reaches this catch — which won't catch it. The exception then propagates up and **breaks the post that was supposed to go out regardless of caption success**.

The original broad catch was actually correct here — caption failures must never block posting. Revert to:

```python
except Exception as e:  # noqa: BLE001 — caption failures must never block posting
    logger.warning(f"AI caption generation failed, continuing: {e}")
```

---

### Nit

**`src/services/integrations/instagram_api.py` — `KeyError` is dead code**

```python
except (ValueError, KeyError) as e:
```

The code uses `data.get("error", {})` which never raises `KeyError`. `ValueError` covers the JSON parse failure above. Drop `KeyError`:

```python
except ValueError as e:
```

---

### Confirm Intent

**`src/services/media_sources/factory.py` — changed fallback semantics**

Narrowing from `Exception` to `GoogleDriveAuthError` means non-auth errors (network timeouts, unexpected SDK errors) no longer fall back to the service account — they now propagate and could surface to the user. If that's intentional (fail loudly on unexpected errors), it's a reasonable design choice. Just confirming it's deliberate and not a side effect of the narrowing pass.

---

### Everything Else: Verified Correct ✓

- All `TelegramError` narrowings (telegram_accounts, telegram_callbacks_admin, telegram_commands, telegram_settings) — correct types for the operations
- All `SQLAlchemyError` narrowings (settings_service, media_lock, conversation_service, media_sync) — correct
- All `(ValueError, InvalidToken)` narrowings (google_drive_oauth, instagram_credentials) — correct for token decode operations
- Exception chaining (`from e`) at all 3 oauth `validate_state_token` sites — correct
- All `noqa: BLE001` justifications are legitimate (fire-and-forget logging, non-blocking audit, top-level handlers)
- Test mock updated from `Exception` → `ValueError` in test_instagram_api.py — correct

chrisrogers37 (commented):
## Empirical Verification Review (reviewer-2)

> Focus: empirical verification of whether narrowed catches actually catch the right exceptions. I trace code paths — not just read the diff.

---

### 🔴 Blocker — `factory.py`: Service-account fallback broken for non-auth errors

**File**: `src/services/media_sources/factory.py`

I traced `get_provider_for_chat()` through `google_drive.py`. It calls:
- `oauth_service.get_user_credentials(telegram_chat_id)` → hits the DB (`settings_repo`, `token_repo`), can raise `SQLAlchemyError`, `TimeoutError`
- `GoogleDriveProvider(root_folder_id=..., oauth_credentials=...)` → constructor raises on bad args

None of those are `GoogleDriveAuthError`. The old `except Exception` was intentionally advisory: *try user OAuth, if anything fails, fall back to service account.* Under the new `except GoogleDriveAuthError`, a DB timeout during OAuth token lookup now propagates to the caller rather than triggering the service-account fallback. The comment still reads "falling back to service account" but that fallback no longer happens for transient failures.

Rajan flags this as "confirm intent." I'm flagging it stronger: the code comment and behavior are now contradictory.

**Fix** (pick one):
1. Make `get_provider_for_chat` wrap all internal failures as `GoogleDriveAuthError` — then the factory's narrow catch is correct by contract.
2. Keep `except Exception as e:  # noqa: BLE001 — service-account fallback is intentionally broad` — this broad catch IS warranted, unlike the BLE001 violations this PR targets.

---

### 🟡 Corroborating Rajan's `scheduler.py` blocker (different angle)

`self.media_repo.get_by_id()` is inside the narrowed try block:

```python
try:
    generated = await caption_service.generate_caption(media_item)
    if generated:
        media_item = self.media_repo.get_by_id(str(media_item.id))  # DB call in scope
except (anthropic.APIError, OSError) as e:
    logger.warning(...)
```

Even setting aside the internal `SQLAlchemyError` from `generate_caption` (Rajan's finding): the DB refresh on success is also inside the narrowed try and would propagate unhandled. Caption failure must never block posting.

---

### 🟡 Suggestion — `instagram_api.py`: `UnicodeDecodeError` gap

`httpx.Response.json()` can raise `UnicodeDecodeError` on binary response bodies — not a subclass of `ValueError`. Low probability against Instagram's API but:

```python
except (ValueError, UnicodeDecodeError):
    raise InstagramAPIError(...)
```

(Separately: Rajan's nit is right — `KeyError` is dead since `data.get("error", {})` never raises `KeyError`.)

---

### ✅ Mutation test: `test_instagram_api.py` — not a decoy

I verified the test is load-bearing. Swapping `ValueError` for bare `Exception` as `json.side_effect` causes the exception to escape the `(ValueError, KeyError)` catch, so `pytest.raises(InstagramAPIError)` would fail — the test is doing real work.

Also confirmed empirically: `httpx.Response.json()` raises `json.JSONDecodeError` which is a `ValueError` subclass — the catch is correct for httpx's actual behavior.

---

### ✅ Narrowed catches verified empirically correct

| Location | Narrowed to | Verified |
|---|---|---|
| `media_lock.py` audit log | `SQLAlchemyError` | Only DB calls in scope ✓ |
| `telegram_settings.py` message delete | `TelegramError` | All PTB subclasses (`BadRequest`, `Forbidden`, `NetworkError`, `TimedOut`) are `TelegramError` ✓ |
| `instagram_credentials.py` token decrypt | `(ValueError, InvalidToken)` | Correct per Fernet's documented raise surface ✓ |
| `telegram_commands.py` `get_chat_member` | `TelegramError` | PTB hierarchy complete under `TelegramError` ✓ |
| Three `validate_state_token` sites | `raise ValueError(...) from e` | Correct exception chaining ✓ |

`noqa: BLE001` on health checks, cleanup, fire-and-forget logging, startup/shutdown notifications — all genuinely warranted. `exc_info=True` additions to health check handlers are a net improvement.

chrisrogers37 (commented):
## Re-review: All items resolved ✓

**scheduler.py** — reverted to broad `except Exception` with `noqa: BLE001` and justification comment. Correct — caption failures must never block posting.

**factory.py** — reverted to broad `except Exception` with `noqa: BLE001` and justification comment. Correct — service-account fallback is intentionally defensive.

**instagram_api.py** — `KeyError` replaced with `UnicodeDecodeError`. `(ValueError, UnicodeDecodeError)` covers all failure modes of `response.json()` (invalid JSON + encoding failures). Test mock updated to `ValueError`. Clean.

All original narrowings from the first review remain correct. Approving.

chrisrogers37 (commented):
## Re-review: All fixes verified ✅

Traced all three changes from commit f7dd7428 against actual code paths.

**factory.py** — `except Exception as e:  # noqa: BLE001 — service-account fallback is intentionally broad`
Confirmed correct. `get_provider_for_chat()` internally hits DB repos (`settings_repo`, `token_repo`) that can raise `SQLAlchemyError`, `TimeoutError` — neither is `GoogleDriveAuthError`. Broad catch is warranted. `GoogleDriveAuthError` import cleaned up correctly.

**instagram_api.py** — `except (ValueError, UnicodeDecodeError)`
Correct: dead `KeyError` removed (addresses Rajan's nit), `UnicodeDecodeError` added for binary/undecodable response bodies. `httpx.Response.json()` raises `json.JSONDecodeError` (`ValueError` subclass) for invalid JSON and `UnicodeDecodeError` for encoding failures — both now covered.

**scheduler.py** — `except Exception as e:  # noqa: BLE001 — caption failures must never block posting`
Correct. `anthropic` import removed cleanly. Caption generation and the DB refresh on success are both covered — failures logged and swallowed, posting continues regardless.

**Verdict: Approve.** All my blockers resolved, all fixes empirically verified correct.

PR #255, opened 2026-04-19
fix: replace deprecated datetime.utcnow() with timezone-aware alternatives

chrisrogers37:
## Summary

- Replace all 30+ `datetime.utcnow()` calls across `src/services/` (16 files + `base_service.py`) with `datetime.now(timezone.utc)`, eliminating Python 3.12 deprecation warnings
- Invert `.replace(tzinfo=None)` patterns to `.replace(tzinfo=timezone.utc)` where comparisons need both sides timezone-aware
- Update 6 test files with aware mock values and `mock.now` instead of `mock.utcnow`
- Extract single `now` variable in `HistoryCreateParams` constructors to avoid redundant calls

## Test plan

- [x] All 1618 tests pass (0 failures, 16 skipped)
- [x] Verified zero remaining `datetime.utcnow()` or `utcnow` references in `src/services/`
- [x] `/simplify` review completed — fixed redundant calls, dead test code, comment quality
- [x] Naive/aware comparison points validated via test suite (mocks use `tzinfo=timezone.utc`)

Closes #249

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #262, opened 2026-04-20
refactor: extract TelegramService god class into focused handlers

### LINKED ISSUE #251: [tech-debt] TelegramService god class — 802 lines, 12 dependencies
## Problem

`TelegramService` at 802 lines with 29 methods and 12 injected repository/service dependencies is a god class. It acts as the central hub for all Telegram bot operations, instantiating 7 repositories and 5 services in `__init__`.

The composition pattern (sub-handlers like `TelegramCommandHandlers`, `TelegramAutopostHandler`) is already in place and is a good architectural decision — but `TelegramService` itself still carries too much direct state and lifecycle responsibility.

## Specific Concerns

- `__init__` creates 12 dependencies eagerly (connection pool pressure)
- `cleanup_transactions()` and `close()` both need manual overrides to handle `InteractionService` — a sign of leaky abstraction
- `_operation_locks`, `_cancel_flags`, `_known_memberships` are all in-memory state management that could be extracted

## Suggested Fix

1. Extract operation lock/cancel flag management into a dedicated `OperationStateManager`
2. Consider lazy-loading repositories (like `HealthCheckService` already does) to reduce init-time connections
3. Move `_escape_markdown` to `telegram_utils.py` (it's already imported from telegram_service by other modules)

## What's Good

The handler composition pattern is solid — `TelegramCommandHandlers`, `TelegramAutopostHandler`, etc. are well-factored. This is about the hub class, not the spokes.

---
*Auto-generated by tech-debt audit on 2026-04-17*

chrisrogers37:
## Summary

Breaks the 804-line `TelegramService` god class into a thin orchestrator (496 lines) plus 4 focused handler classes, following the existing composition pattern used by `TelegramCommandHandlers`, `TelegramAutopostHandler`, etc.

**Extracted classes:**
- **`OperationStateManager`** (36 lines) — per-queue-item asyncio locks and cancellation flags
- **`TelegramUserManager`** (90 lines) — user creation, profile sync, membership cache, display names
- **`TelegramMembershipHandler`** (185 lines) — bot added/removed from groups, onboarding DM flow
- **`TelegramLifecycleHandler`** (95 lines) — startup/shutdown admin notifications

**Also:**
- Consolidated duplicate `_escape_markdown` (was in `telegram_service.py` and `telegram_notification.py`) into `telegram_utils.py` as the canonical location
- TelegramService keeps thin delegate methods so all 20+ handler call sites (`self.service.get_operation_lock()`, `self.service._get_or_create_user()`, etc.) continue working unchanged
- Updated tests for the new structure

## What stays in TelegramService

- `__init__` with repos/services (still the dependency container)
- `initialize()` — handler instantiation and Telegram handler registration
- Callback dispatch table and two-tier routing
- Conversation message routing (thin, ~10 lines)
- Notification/lifecycle delegates
- Polling lifecycle (`start_polling`, `stop_polling`)
- `cleanup_transactions` / `close` overrides (InteractionService special handling)

Closes #251

## Test plan

- [x] All 1618 tracked tests pass (16 skipped)
- [x] Pre-existing failures excluded (conversation_service SettingsService mock, backfill_downloader 403)
- [x] Updated `test_phase2b_handlers.py` to test `TelegramMembershipHandler` directly
- [x] Updated `test_telegram_callbacks.py` for `operation_state` internal path
- [x] Ran `/simplify` — moved `is_verbose` off `TelegramUserManager` (settings concern), stripped boilerplate docstrings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **TelegramService god class refactor** (#251) — Broke 804-line `TelegramService` into thin orchestrator (496 lines) plus 4 focused handler classes: `OperationStateManager`, `TelegramUserManager`, `TelegramMembershipHandler`, `TelegramLifecycleHandler`. Consolidated duplicate `_escape_markdown` into `telegram_utils.py`.
+
 ### Fixed
 
 - **Narrow broad `except Exception` catches** (#250) — Audited 88 `except Exception` catches across `src/services/`. Narrowed catches to specific types (`TelegramError`, `SQLAlchemyError`, `InvalidToken`, etc.) where failure modes are known. Added `noqa: BLE001` annotations with justification comments to intentionally broad catches (best-effort logging, cleanup, health checks). Added `exc_info=True` to health check error handlers.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Test coverage for 7 untested service modules** (#252) — 110 unit tests covering `telegram_callbacks_core`, `telegram_callbacks_queue`, `telegram_callbacks_admin`, `conversation_service`, `start_command_router`, `user_service`, and `backfill_downloader`. Shared test helpers (`make_query`, `make_user`, `noop_context_manager`) added to `conftest.py`.
+
 ### Changed
 
 - **TelegramService god class refactor** (#251) — Broke 804-line `TelegramService` into thin orchestrator (496 lines) plus 4 focused handler classes: `OperationStateManager`, `TelegramUserManager`, `TelegramMembershipHandler`, `TelegramLifecycleHandler`. Consolidated duplicate `_escape_markdown` into `telegram_utils.py`.
--- a/src/services/core/telegram_callbacks_queue.py
+++ b/src/services/core/telegram_callbacks_queue.py
@@ -8,8 +8,8 @@ from telegram import InlineKeyboardButton, InlineKeyboardMarkup
 
 from sqlalchemy.exc import OperationalError
 
-from src.services.core.telegram_service import _escape_markdown
 from src.services.core.telegram_utils import (
+    escape_markdown as _escape_markdown,
     build_queue_action_keyboard,
     validate_queue_and_media,
     validate_queue_item,
--- a/src/services/core/telegram_lifecycle.py
+++ b/src/services/core/telegram_lifecycle.py
@@ -1,0 +1,95 @@
+"""Lifecycle notifications for the Telegram bot (startup/shutdown)."""
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import TYPE_CHECKING
+
+from src.config.settings import settings
+from src.utils.logger import logger
+from src import __version__
+
+if TYPE_CHECKING:
+    from src.services.core.telegram_service import TelegramService
+
+
+class TelegramLifecycleHandler:
+    """Sends startup and shutdown notifications to the admin chat."""
+
+    def __init__(self, service: TelegramService):
+        self.service = service
+
+    async def send_startup_notification(self):
+        """Send startup notification to admin with system status."""
+        if not settings.SEND_LIFECYCLE_NOTIFICATIONS:
+            return
+
+        try:
+            pending_count = self.service.queue_repo.count_pending()
+            media_count = len(self.service.media_repo.get_all(is_active=True))
+            recent_posts = self.service.history_repo.get_recent_posts(hours=24)
+            last_post_time = recent_posts[0].posted_at if recent_posts else None
+
+            if last_post_time:
+                time_diff = datetime.utcnow() - last_post_time
+                hours = int(time_diff.total_seconds() / 3600)
+                last_posted = f"{hours}h ago" if hours > 0 else "< 1h ago"
+            else:
+                last_posted = "Never"
+
+            message = (
+                f"🟢 *Storyline AI Started*\n\n"
+                f"📊 *System Status:*\n"
+                f"├─ Database: ✅ Connected\n"
+                f"├─ Telegram: ✅ Bot online\n"
+                f"├─ Queue: {pending_count} pending posts\n"
+                f"└─ Last posted: {last_posted}\n\n"
+                f"⚙️ *Configuration:*\n"
+                f"├─ Posts/day: {settings.POSTS_PER_DAY}\n"
+                f"├─ Window: {settings.POSTING_HOURS_START:02d}:00-{settings.POSTING_HOURS_END:02d}:00 UTC\n"
+                f"└─ Media indexed: {media_count} items\n\n"
+                f"🤖 v{__version__}"
+            )
+
+            await self.service.bot.send_message(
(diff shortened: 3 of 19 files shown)


chrisrogers37 (commented):
## PR #262 Review — TelegramService god class refactor

**Overall**: Clean extraction. Handler boundaries are correct, delegation pattern is solid, no behavior changes, no circular dependency risk. Two blockers before merge.

---

### Blockers

**1. `src/services/core/telegram_lifecycle.py:54` — `datetime.utcnow()` reintroduced**

```python
time_diff = datetime.utcnow() - last_post_time
```

This code was moved from `telegram_service.py` but not updated. PR #255 just swept every `utcnow()` call out of the codebase — this new file reintroduces it. Fix:

```python
from datetime import datetime, timezone
# ...
time_diff = datetime.now(timezone.utc) - last_post_time
```

**2. Changelog CI failing** — no CHANGELOG.md entry for this refactor.

---

### Everything Else: Verified Correct ✓

**Extraction boundaries** — well-scoped:
- `OperationStateManager`: asyncio locks + cancel flags. Get/create pattern is safe (asyncio is single-threaded; no `await` between check and insert means no concurrent interleaving possible).
- `TelegramUserManager`: user creation, profile sync, membership cache, display names. `_known_memberships` correctly encapsulated here.
- `TelegramMembershipHandler`: bot add/remove flows, onboarding DM. Cache eviction delegates cleanly to `user_manager.evict_memberships_for_chat()`.
- `TelegramLifecycleHandler`: startup/shutdown admin notifications. Clean, aside from blocker #1.

**No behavior changes** — pure refactor confirmed. Logic moved verbatim; the one new method (`evict_memberships_for_chat`) is cleaner than direct dict mutation.

**No circular dependency risk** — all 4 handler files use `TYPE_CHECKING` guard for `TelegramService` import. Runtime dependency flows one direction: handlers hold a reference to the service, not vice versa.

**`_escape_markdown` consolidation** — correct. `escape_markdown` is canonical in `telegram_utils.py`. Both `telegram_callbacks_queue.py` and `telegram_notification.py` updated to import from there. Re-export in `telegram_service.py` (`noqa: F401`) is a harmless safety net.

**Delegate methods** — all present in `TelegramService`: `get_operation_lock`, `get_cancel_flag`, `cleanup_operation_state`, `_get_or_create_user`, `_get_display_name`, `send_startup_notification`, `send_shutdown_notification`.

**Tests** — `test_phase2b_handlers.py` correctly tests `TelegramMembershipHandler` directly; `_known_memberships` access updated to `service.user_manager._known_memberships`; operation state assertions updated to `service.operation_state.*`. All correct.

chrisrogers37 (commented):
## Empirical Verification Review (reviewer-2)

> I traced 3 extracted paths and ran mutation analysis on 2 test files. All behavioral paths verified correct.

---

### Traced path 1: `ChatMemberHandler` registration ✅

**Old**: `ChatMemberHandler(self._handle_my_chat_member, ...)`
**New**: `ChatMemberHandler(self.membership_handler.handle_my_chat_member, ...)`

Both registered in `initialize()`. Verified `membership_handler` is instantiated before any handler can fire (PTB doesn't deliver events until `start_polling` is called, which comes after `initialize()`). Correct.

---

### Traced path 2: `_known_memberships` cache — write/evict consistency ✅

This is the most subtle refactor: the membership cache moved from `TelegramService._known_memberships` to `TelegramUserManager._known_memberships`.

- **Writes**: `TelegramUserManager._ensure_membership` → `self._known_memberships.add(cache_key)` ✅
- **Evictions**: `TelegramMembershipHandler._handle_bot_removed_from_group` → `self.service.user_manager.evict_memberships_for_chat(chat.id)` → `self._known_memberships -= to_evict` ✅

Both operations now target the same set object (`user_manager._known_memberships`). The eviction also now returns the evicted count and logs it — a small improvement over the old silent eviction.

---

### Traced path 3: `OperationStateManager` delegates ✅

All three call sites in `telegram_autopost.py` go through `TelegramService` delegates:

- `self.service.get_operation_lock(queue_id)` → `self.operation_state.get_lock(queue_id)` ✅
- `self.service.get_cancel_flag(queue_id)` → `self.operation_state.get_cancel_flag(queue_id)` ✅
- `self.service.cleanup_operation_state(queue_id)` → `self.operation_state.cleanup(queue_id)` ✅

Public API on `TelegramService` is fully preserved — handler modules don't need to know about `OperationStateManager`.

---

### Mutation tests: `test_phase2b_handlers.py` — NOT decoys ✅

**`test_bot_kicked_deactivates_memberships`**:
- Asserts `deactivate_for_chat.assert_called_once_with(str(mock_chat_settings.id))` — if `_handle_bot_removed_from_group` is broken, this fails
- Asserts cache entries for `-100999` are evicted from `service.user_manager._known_memberships` and entry for `-100888` survives — if `evict_memberships_for_chat` is a no-op, these fail
- Not a decoy.

**`test_bot_added_auto_links_pending_session`**:
- Asserts `mock_conv.link_group_to_instance.assert_called_once_with(...)` — if session detection or linking is broken, this fails
- Not a decoy.

**`test_telegram_callbacks.py` change**: `service.operation_state._operation_locks` / `._cancel_flags` — if `OperationStateManager.cleanup()` had a bug, the lock would remain and the assertion fails. Not a decoy.

---

### Nit 1 — `telegram_autopost.py` still imports from old location

`telegram_callbacks_queue.py` was updated in this PR to import `escape_markdown as _escape_markdown` from `telegram_utils` directly. But `telegram_autopost.py` (not changed in this PR) likely still imports `_escape_markdown` from `telegram_service`. The re-export shim (`from src.services.core.telegram_utils import escape_markdown as _escape_markdown  # noqa: F401`) covers it, so nothing breaks — but the inconsistency means the shim can't be safely removed in a follow-up. Suggest updating `telegram_autopost.py` to import from `telegram_utils` directly.

---

### Nit 2 — New failure mode: `self.lifecycle` before `initialize()`

Old code: `send_startup_notification` was a method directly on `TelegramService`, always callable.
New code: it delegates to `self.lifecycle`, which is only set in `initialize()`. Calling it before `initialize()` would raise `AttributeError`.

Not a regression given `main.py` hasn't changed and the call order is presumably `initialize() → send_startup_notification()`. But it's now an implicit precondition where none existed before. A one-liner guard makes this explicit: `if not hasattr(self, 'lifecycle'): return`, or initialize `self.lifecycle = None` in `__init__`.

---

**Verdict: Approve.** All 3 traced paths verify correct — handler registration, cache write/evict consistency, and delegate wiring all hold. Mutation tests are load-bearing. The two nits are low-risk and don't block merge.

chrisrogers37 (commented):
## Re-review PR #262

**Original blockers**: Both resolved.
- `telegram_lifecycle.py` → `datetime.now(timezone.utc)` ✓
- CHANGELOG entry added ✓

---

### New Blocker — Lint failing

**`tests/src/services/test_telegram_callbacks_admin.py`** — new test file uses `datetime.utcnow()` in three test methods:

```python
now = datetime.utcnow()  # test_reschedule_action, test_clear_action, test_force_action
```

Same issue as the lifecycle fix. The import is also missing `timezone`. Fix:

```python
from datetime import datetime, timedelta, timezone
# ...
now = datetime.now(timezone.utc)  # all three occurrences
```

---

### Reviewer 2 Nits — noted, agree both are low-risk

- **`telegram_autopost.py`** still imports `_escape_markdown` via the re-export shim — fine for now but blocks shim removal later. Good follow-up candidate.
- **`self.lifecycle` precondition** — initializing `self.lifecycle = None` in `__init__` would make the ordering requirement explicit, but not a regression.

Everything else remains verified correct from the first review.

PR #264, opened 2026-04-20
fix: guard cleanup_transactions in finally blocks to prevent scheduler crash

chrisrogers37:
## Summary

- Wraps all 7 unguarded `cleanup_transactions()` / `end_read_transaction()` calls in `finally` blocks with `try/except` so a failure in cleanup can never crash a background loop
- Root cause of scheduler outage: if `cleanup_transactions()` threw in the scheduler loop's `finally` block, the exception propagated through `_guarded()`, killing the scheduler task permanently while cleanup loops continued — making the failure silent
- Affected loops: scheduler, retention tick, pool/token health ticks, lock cleanup, cloud storage cleanup, media sync

## Context

Investigation found posts stopped serving after 2026-04-19 12:00 UTC. The worker process was alive (cleanup tasks running hourly) but the scheduler asyncio task had died. The `_guarded()` wrapper catches unhandled exceptions from background tasks and stops the loop — but the scheduler loop's `finally` block called `cleanup_transactions()` without guarding it, providing a path for exceptions to escape the inner `try/except`.

## Test plan

- [x] All 1728 tests pass
- [x] `/simplify` review clean (fixed variable shadowing + standardized log format)
- [ ] After merge: restart Railway worker service to revive the scheduler loop
- [ ] Verify posts resume within the posting window

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Guard `cleanup_transactions()` in background loop finally blocks** (#264) — Wrapped 7 unguarded `cleanup_transactions()` / `end_read_transaction()` calls in `try/except` across all background loops (scheduler, retention tick, pool/token health ticks, lock cleanup, cloud storage cleanup, media sync). An unhandled exception in a `finally` block could crash the loop via `_guarded()` while the worker process stayed alive — causing the scheduler to silently stop posting.
 - **Narrow broad `except Exception` catches** (#250) — Audited 88 `except Exception` catches across `src/services/`. Narrowed catches to specific types (`TelegramError`, `SQLAlchemyError`, `InvalidToken`, etc.) where failure modes are known. Added `noqa: BLE001` annotations with justification comments to intentionally broad catches (best-effort logging, cleanup, health checks). Added `exc_info=True` to health check error handlers.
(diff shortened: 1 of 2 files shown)


chrisrogers37 (commented):
## PR #264 Review — Scheduler crash fix

**Fix is correct and complete. One blocker: missing CHANGELOG entry.**

---

### Fix Verified ✓

All 10 cleanup calls across 7  blocks are now guarded. Spot-checking the critical ones:

**** — the loop that was dying:
```python
for svc in (scheduler_service, posting_service):
    try:
        svc.cleanup_transactions()
    except Exception as cleanup_err:
        logger.warning(f"cleanup_transactions failed for {type(svc).__name__}: {cleanup_err}")
```
For-loop pattern is clean and avoids two separate try/except blocks. ✓

**** — two cleanup calls guarded independently:
Both  and  have separate guards so a failure in the first doesn't block the second. ✓

**** —  None check preserved inside the guard. ✓

Errors are logged as warnings (not silently swallowed) on all 7 sites. ✓

**CI**: Tests ✓ (1728), Lint ✓, Security ✓. Changelog failing.

---

### Blocker

No CHANGELOG entry. Changelog CI is failing. Add a line under `### Fixed` and this is ready to merge.

chrisrogers37 (commented):
## PR #264 Review — Scheduler crash fix

**Fix is correct and complete. One blocker: missing CHANGELOG entry.**

---

### Fix Verified ✓

All 10 cleanup calls across 7 `finally` blocks are now guarded.

**`run_scheduler_loop`** (the loop that was dying): for-loop over `(scheduler_service, posting_service)` with independent `try/except` per service — cleaner than two separate blocks. ✓

**`cleanup_cloud_storage_loop`**: `cloud_service.cleanup_transactions()` and `media_repo.end_read_transaction()` guarded independently — failure in first doesn't block second. ✓

**`media_sync_loop`**: `if settings_service:` None check preserved inside the guard. ✓

All 7 sites: errors logged as warnings, not silently swallowed. ✓

**CI**: Tests ✓ (1728), Lint ✓, Security ✓. Changelog failing.

---

### Blocker

No CHANGELOG entry. Add a line under `### Fixed` and this is ready to merge.

PR #336, opened 2026-05-16
feat: ensure_utc helper consolidates naive→UTC datetime coercion (#335)

### LINKED ISSUE #335: tech-debt: extract `ensure_utc(dt)` helper; fix latent naive/aware datetime bugs
## Background

PR #334 fixed a `TypeError: can't compare offset-naive and offset-aware datetimes` in `is_token_stale` by coercing a naive `datetime` to UTC-aware before comparison. The `/simplify` review surfaced that **the same idiom is duplicated 5 times** in the codebase and **the same latent bug exists in two `ApiToken` model properties** that currently "work" only because both sides happen to be naive.

## Inline duplicates of the naive→UTC coercion

The 3-line `if dt.tzinfo is None: dt = dt.replace(tzinfo=timezone.utc)` pattern appears in:

- `src/services/core/setup_state_service.py:269-270` (just added in #334)
- `src/services/core/telegram_commands.py:212`
- `src/services/core/telegram_utils.py:46-47`
- `src/services/core/dashboard_history_queries.py:262`
- `src/services/core/scheduler.py:68-69`

## Latent bugs (same root cause)

These compare `api_tokens.expires_at` (naive `DateTime` column) against `datetime.utcnow()` (also naive). Works today, breaks if either side becomes aware:

- `src/models/api_token.py:93` — `is_expired`: `datetime.utcnow() > self.expires_at`
- `src/models/api_token.py:99` — `hours_until_expiry`: `self.expires_at - datetime.utcnow()`

Writers in `oauth_service.py:187` and `token_refresh.py:150` construct **aware** UTC datetimes; SQLAlchemy currently strips tz on write to the naive column, which masks the inconsistency.

## Recommendation

1. **Add `src/utils/datetime_utils.py` with `ensure_utc(dt: datetime | None) -> datetime | None`** that returns the input if already aware, otherwise `replace(tzinfo=timezone.utc)`. Single import; consistent semantics.
2. Migrate the 5 inline duplicates to call `ensure_utc`.
3. Fix `ApiToken.is_expired` and `hours_until_expiry` to use `datetime.now(timezone.utc)` + `ensure_utc(self.expires_at)`.
4. **Optional, longer-term**: migrate the `api_tokens.expires_at` column to `DateTime(timezone=True)` via Alembic (`ALTER COLUMN ... TYPE timestamptz USING expires_at AT TIME ZONE 'UTC'`). Removes the coercion entirely. Touches ~5 read sites in `token_repository.py` + model + this file. Schedule with a downtime-tolerant deploy window.

## Verification

- Existing tests still pass (no behavior change for the runtime fix).
- New unit test in `tests/utils/test_datetime_utils.py` covering: naive→aware, aware passthrough, `None` passthrough, and microsecond preservation.
- If column migration is included: integration test that reads a pre-migration row (write naive, read back aware) and confirms tz is preserved.

## Out of scope

Don't fold this into a hotfix — it touches `models/`, `services/core/`, `repositories/`, and migrations. Separate PR with own review.

chrisrogers37:
## Summary

- Added `src/utils/datetime_utils.py::ensure_utc(dt)` — single source of truth for "naive datetime → UTC-aware" coercion. None passes through; already-aware datetimes pass through with the same object (no allocation).
- Migrated 5 inline copies of the same idiom in `setup_state_service.py`, `telegram_commands.py`, `scheduler.py`, `dashboard_history_queries.py`, `telegram_utils.py`.
- Fixed latent same-bug in `ApiToken.is_expired` and `ApiToken.hours_until_expiry`: previously compared `expires_at` against the deprecated naive `datetime.utcnow()`. The comparisons "worked" only because both sides happened to be naive; would have broken on any column migration to `DateTime(timezone=True)`.

Closes #335.

## Test plan

- [x] New unit tests for `ensure_utc` — None, naive coerced, aware passthrough, non-UTC aware preserved, microsecond preservation (100% coverage on the helper)
- [x] Targeted tests for callsites: `pytest -k "setup_state or token_stale or gdrive or api_token or scheduler or telegram_commands or telegram_utils or dashboard_history"` → **284 passed**
- [x] Full suite: **1737 passed, 16 skipped**
- [x] `ruff check src/` clean, `ruff format src/ --check` clean

## Out of scope (follow-ups)

- Migrating `api_tokens.expires_at` column to `DateTime(timezone=True)` (would touch `token_repository.py` read sites; requires a coordinated deploy).
- Converting remaining `datetime.utcnow()` callsites in `token_repository.py` and `onboarding_repository.py` — those go through SQLAlchemy column expressions, not Python comparisons, so they're safe today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`ensure_utc(dt)` datetime helper** (`src/utils/datetime_utils.py`) — single source of truth for "naive datetime → UTC-aware" coercion. Returns `None` unchanged; passes already-aware datetimes through without re-allocating. Replaces 5 copies of the same inline idiom in `setup_state_service.py`, `telegram_commands.py`, `scheduler.py`, `dashboard_history_queries.py`, and `telegram_utils.py`. Also used in `ApiToken.is_expired` and `ApiToken.hours_until_expiry`, which previously compared `expires_at` to a naive `datetime.utcnow()` — that latent bug never surfaced because both sides happened to be naive, but it would have broken the moment either side became aware (e.g., a future column migration to `DateTime(timezone=True)`). Closes #335.
+
 ### Removed
 
 - **Stale documentation archive** — Deleted `documentation/archive/` directory containing 82 historical planning docs (Jan-Mar 2026) that are no longer relevant to current development. Reduces repo clutter by ~51,000 lines.
--- a/src/models/api_token.py
+++ b/src/models/api_token.py
@@ -1,6 +1,6 @@
 """API token model for OAuth token storage."""
 
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Optional
 import uuid
 
@@ -9,6 +9,7 @@ from sqlalchemy.dialects.postgresql import UUID, ARRAY, JSONB
 from sqlalchemy.orm import relationship
 
 from src.config.database import Base
+from src.utils.datetime_utils import ensure_utc
 
 
 class ApiToken(Base):
@@ -88,13 +89,15 @@ class ApiToken(Base):
     @property
     def is_expired(self) -> bool:
         """Check if the token has expired."""
-        if self.expires_at is None:
+        expires_at = ensure_utc(self.expires_at)
+        if expires_at is None:
             return False
-        return datetime.utcnow() > self.expires_at
+        return datetime.now(timezone.utc) > expires_at
 
     def hours_until_expiry(self) -> Optional[float]:
         """Get hours until token expires, or None if no expiry."""
-        if self.expires_at is None:
+        expires_at = ensure_utc(self.expires_at)
+        if expires_at is None:
             return None
-        delta = self.expires_at - datetime.utcnow()
+        delta = expires_at - datetime.now(timezone.utc)
         return max(0, delta.total_seconds() / 3600)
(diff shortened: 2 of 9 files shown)


chrisrogers37 (commented):
**Verdict: Ship it**

**Helper correctness — verified**

`ensure_utc` semantics are exact:
- `None` passthrough (returns `None`)
- Naive → `replace(tzinfo=UTC)` (wall-clock preserved, no allocation)
- Already-aware → same object returned (no re-allocation, non-UTC offsets NOT re-anchored — documented and intentional)
- Non-UTC aware datetimes still compare correctly against UTC datetimes; Python does offset arithmetic automatically.

**Call site migration — all 5 correct**

Each migration is semantically equivalent to the inline it replaced:

| File | Old pattern | New |
|------|------------|-----|
| `scheduler.py` | `if last_sent and last_sent.tzinfo is None: last_sent = last_sent.replace(...)` | `ensure_utc(chat_settings.last_post_sent_at)` |
| `dashboard_history_queries.py` | same | same |
| `telegram_commands.py` | same | same |
| `setup_state_service.py` | 6-line inline block | 2-line ensure_utc |
| `telegram_utils.py` | fromisoformat + manual check | `ensure_utc(datetime.fromisoformat(...))` |

**`ApiToken.is_expired` latent bug fixed**

Previously: `datetime.utcnow() > self.expires_at` — naive vs naive, "worked" by accident. Now: `datetime.now(timezone.utc) > ensure_utc(self.expires_at)` — aware vs aware, correct on any future column migration to `DateTime(timezone=True)`. Same fix applied to `hours_until_expiry`.

**Tests — load-bearing**

5 targeted cases cover all branches: None, naive coercion, aware passthrough (same-object assertion), non-UTC aware passthrough, microsecond preservation. The `is aware` test would fail if `ensure_utc` were reverted to return a naive datetime.

**Out-of-scope call sites** (`token_repository.py`, `onboarding_repository.py`): documented as SQLAlchemy column-expression paths, not Python comparisons — not a gap.

*Show me the data: ran the helper semantics locally before signing off.*

PR #341, opened 2026-05-17
feat(oauth): ingest all FB Pages' Instagram accounts on connect (#331)

chrisrogers37:
## Summary

- `OAuthService._get_instagram_account_info` was hard-coded to read only the **first** Facebook Page (`page_id = pages[0]["id"]`). Users with multiple FB Pages — and therefore multiple linked Instagram Business Accounts — silently lost all but the first on every OAuth round-trip.
- Renamed to `_get_instagram_accounts_info` (plural) and now iterate every Page, skip Pages with no linked IG, dedupe Pages that share an IG, and return a list.
- `exchange_and_store` stores all detected accounts; the first becomes active for the originating chat (matches prior single-account behavior), the rest land as inactive and can be activated via **Settings → Accounts** (or the wizard once #332 lands).
- New `account_count` field in the return dict so callers can surface "Connected N accounts" instead of only the active one.

Closes #331. Pairs with #332 (UI surface for multi-account in the Setup Wizard).

## Test plan

- [x] New: `test_exchange_multiple_accounts_stores_all_first_active` — 3 IGs from 3 Pages, first becomes active, rest inactive, `telegram_chat_id` only set on the active row
- [x] New: `test_exchange_multiple_accounts_mixed_existing_and_new` — existing IGs route through `update_account_token`, new ones through `add_account`
- [x] Existing tests adapted to list return shape (`mock_info.return_value = [{...}]`); `account_count` asserted where the summary is checked
- [x] `pytest tests/src/services/test_oauth_service.py` → 18 passed
- [x] Full suite: **1734 passed, 16 skipped**
- [x] `ruff check src/` clean
- [ ] After deploy: log in with a Facebook account that admins multiple Pages each with a linked IG. Confirm all show up in **Settings → Accounts** after one consent round-trip.

## Out of scope

- Setup Wizard UI surface for multi-account (#332) — separate PR
- Account chooser at the Facebook consent step (Meta-side; out of our control — the user picks the *FB* page set when consenting; we ingest all of them post-consent)


chrisrogers37 (commented):
**Verdict: Ship it**

Reviewed: oauth_service.py diff, test file changes, return shape compatibility.

**Core fix is correct:**
- `pages[0]["id"]` hard-coding was a silent data-loss bug. Iteration + dedupe via `seen_ig_ids` is the right fix.
- `active_account = accounts_info[0]` is safe — guarded by the non-empty check + ValueError above it.
- `telegram_chat_id=None` for inactive accounts: correct — only the active account gets the chat association.
- Return dict: `username` and `account_id` keys unchanged, `account_count` is additive. No breaking change for existing callers.

**Mutation test passed:**
Mentally reverting `set_active = idx == 0` to always-True: `test_exchange_multiple_accounts_stores_all_first_active` asserts `set_as_active is False` and `telegram_chat_id is None` for calls[1:] — both would fail. Tests are load-bearing.

**Minor gaps (not blocking):**
1. No test for the dedupe path (`seen_ig_ids`). If the same IG Business Account is linked to two different FB Pages, the second is correctly skipped — but this behavior has no coverage. Low-risk given Meta's data model, but worth adding.
2. N pages = N sequential HTTP round-trips for username fetches. Acceptable for an OAuth flow (once per connect), but if a user has many Pages this could be slow. Note for future: could batch username lookups if Meta's API supports fields on the IG account in the page response.

**Auth regression check:** No changes to token exchange, short→long-lived flow, or token storage. Existing tests adapted cleanly.

Full suite 1734 passed checks out. Merge when ready.

PR #342, opened 2026-05-17
fix: lazy repository sessions to prevent connection pool exhaustion

### LINKED ISSUE #320: Tech debt: Service layer creates repositories eagerly, risking connection pool exhaustion
## Problem

Every service instantiation eagerly creates all its repository dependencies in `__init__`, and each repository opens a database session immediately:

```python
# BaseRepository.__init__
def __init__(self):
    self._db_generator = get_db()
    self._db: Session = next(self._db_generator)  # Opens connection immediately
```

For example, `TelegramService.__init__` creates 7 repositories and 4 sub-services (which create their own repositories), opening 10+ DB connections just by being instantiated. `DashboardService` similarly creates 6 repositories plus a `SettingsService`.

The codebase compensates with:
- `BaseService.close()` recursively closing nested services/repos
- `BaseService.cleanup_transactions()` walking all attributes
- `BaseService.__del__` as a GC safety net
- Context manager protocol (`with DashboardService() as service:`)

But services that are held as long-lived attributes (like `TelegramService.settings_service`) keep their connections open for the entire bot lifetime.

## Impact

- Connection pool exhaustion under load (each API request or scheduler tick may open many connections)
- The recursive `cleanup_transactions()` uses `dir(self)` introspection which is fragile and slow
- GC-based cleanup (`__del__`) is unreliable in CPython

## Suggested Fix

1. **Lazy-load repositories**: Open the DB session on first `.db` access, not in `__init__`
2. **Use connection pooling with scoped sessions**: SQLAlchemy's `scoped_session` would give each thread/task its own session without manual lifecycle management
3. **Consider a Unit of Work pattern**: A single session per request/task, shared across repositories, committed/rolled back at the boundary
4. The `use_session()` method on BaseRepository already supports this -- expand its usage to the standard pattern

## Files

- `src/repositories/base_repository.py` (lines 29-31: eager session creation)
- `src/services/base_service.py` (lines 42-94: recursive cleanup/close)
- `src/services/core/telegram_service.py` (lines 61-80: 10+ repos created eagerly)

---
*Found by tech-debt scan on 2026-04-28*

chrisrogers37:
## Summary

- `BaseRepository.__init__` no longer opens a DB session eagerly — sessions open lazily on first `.db` property access
- `close()`, `commit()`, `rollback()`, `end_read_transaction()` are safe no-ops when the session was never opened
- Context manager protocol and `use_session()` unchanged

### Before

`TelegramService.__init__` creates 10+ repositories, each calling `next(get_db())` immediately. That's 10+ connections consumed before any query runs — risks exhausting the pool (default size: 10).

### After

Repositories initialize with `_db = None`. The first `.db` access opens the session. Repos that are never queried never open a connection.

Fixes #320

## Test plan

- [x] 1744 tests pass, 0 failures
- [x] New tests: `test_init_does_not_open_session`, `test_db_property_opens_session_lazily`
- [x] New tests: no-op behavior for `close`, `commit`, `rollback`, `end_read_transaction` on unopened sessions
- [x] Existing tests for circuit breaker, rollback recovery, context manager all pass unchanged


chrisrogers37 (commented):

Core fix is correct. Lazy init solves the pool exhaustion on `TelegramService` construction. All guards are in place. Tests are real.

---

**Mutation check**

- Remove `if self._db is None: self._open_session()` from `db` property → `test_db_property_opens_session_lazily` fails immediately on `AttributeError: 'NoneType' object has no attribute 'is_active'`. Real test, not a decoy.
- Remove `if self._db is None: return` guards from `commit()`/`rollback()`/`end_read_transaction()`/`close()` → no-op tests fail with `AttributeError`. Real tests.
- Revert `__init__` to eager init → `mock_get_db.assert_not_called()` fails. Real test.

All four new test groups survive mutation. The PR ships what it claims.

---

**Two minor notes (non-blocking)**

1. **`use_session()` + `close()` without prior `.db` access.** If a caller does `repo.use_session(s)` then `repo.close()` — without ever touching `.db` — `_db_generator` is still `None`. `close()` then tries `next(None)` → `TypeError`, caught and logged as a `WARNING`. Pre-PR this was a silent `StopIteration`. Behavior is still correct (`session.close()` runs in the `finally`), but it's log noise. Worth a one-liner: guard `if self._db_generator is not None:` around the `next()` call in `close()`.

2. **Test fixture has dead code.** `mock_get_db.return_value = iter([mock_db])` in the shared `repo` fixture is now pointless — `get_db()` is never called in `__init__`. Fixture repos end up with `_db_generator=None` + `_db=mock_db`, which hits the TypeError/warning path above in `test_close_calls_session_close`. Test still passes. Clean up when convenient.

---

**Other checks**

- Context manager (`__enter__`/`__exit__`): intact, `close()` no-ops correctly on never-opened sessions. ✓
- `__del__` GC path: still calls `close()`. ✓
- `_open_session()` factored correctly — session-recovery rollback path in `db` property uses it too. ✓
- Thread safety: TOCTOU on `if self._db is None: _open_session()` — two threads could open two sessions. Acceptable if repos are per-request/per-service (typical). No change needed unless the codebase starts sharing repo instances across threads.
- Mandate references `cleanup_transactions()` — that method doesn't exist; `end_read_transaction()` is the equivalent and is correctly guarded.

PR #344, opened 2026-05-17
security: add token revocation for compromised OAuth tokens

chrisrogers37:
## Summary

- **Migration 032**: `revoked_at` nullable TIMESTAMPTZ on `api_tokens` — revoked tokens remain for audit but are excluded from all queries
- **Model**: `revoked_at` column + `is_revoked` property on `ApiToken`
- **Repository**: all 5 retrieval methods (`get_token`, `get_token_for_account`, `get_all_instagram_tokens`, `get_token_for_chat`, `get_expiring_tokens`) now filter `revoked_at IS NULL`. UPSERT methods (`create_or_update`, `create_or_update_for_chat`) query without the filter to avoid unique constraint violations and clear `revoked_at` on re-issue. New `revoke_tokens_for_service()` method.
- **Service**: `TokenRefreshService.revoke_tokens()` calls provider revocation APIs (Meta `DELETE /me/permissions`, Google `POST /revoke`) best-effort, then marks tokens revoked
- **CLI**: `storydump-cli revoke-tokens --service <instagram|google_drive>` with preview table, confirmation prompt, and provider result reporting
- CHANGELOG entry under Security

Closes #328

## Test plan

- [x] `ruff check` and `ruff format --check` pass on all changed files
- [x] `git diff origin/main..HEAD --stat` shows only intended files (7)
- [ ] CI tests pass
- [ ] Manual: run `revoke-tokens --service instagram` on staging, verify token excluded from `get_token()`, re-auth clears revocation
- [ ] Manual: apply migration 032 on staging, verify column exists


chrisrogers37 (commented):

Security implementation is correct across all mandate items. Two notes to address before or shortly after merge.

---

**Mandate checklist**

**Migration (032):** `ADD COLUMN IF NOT EXISTS revoked_at TIMESTAMPTZ` — nullable, idempotent, atomic (BEGIN/COMMIT), safe for existing rows (all get NULL = non-revoked). Reversible via `ALTER TABLE api_tokens DROP COLUMN revoked_at`. ✓

**All retrieval paths filtered:** Every read method in `token_repository.py` adds `ApiToken.revoked_at.is_(None)`:
- `get_token()` ✓
- `get_token_for_account()` ✓
- `get_all_instagram_tokens()` ✓
- `get_token_for_chat()` ✓
- `get_expiring_tokens()` ✓

**Provider APIs correct:**
- Instagram: `DELETE {meta_graph_base}/me/permissions?access_token=<token>` — correct Meta endpoint and method ✓
- Google: `POST https://oauth2.googleapis.com/revoke` with `data={"token": value}` and `Content-Type: application/x-www-form-urlencoded` — correct Google endpoint ✓

**Local revocation survives provider failure:** Traced the full call chain. `revoke_tokens_for_service` calls `self.db.commit()` before returning — DB revocation is committed before `revoke_tokens` makes any provider API call. Provider failures are caught by the broad `except Exception` with `# noqa: BLE001` (documented intent) and land in `provider_results` without re-raising. The committed revocation cannot be undone by a provider failure. ✓

**UPSERT clears revoked_at:** Both `create_or_update` and `create_or_update_for_chat` build their own query *without* the `revoked_at IS NULL` filter to locate the revoked row, then set `existing.revoked_at = None` unconditionally. Unique constraint violations on re-issue are prevented. ✓

**CHANGELOG:** Present, comprehensive, correct section. ✓

---

**Two notes**

**1. No tests for revocation paths.**

`revoke_tokens_for_service`, `revoke_tokens`, `_revoke_instagram_token`, `_revoke_google_token` — none have tests. The `revoked_at IS NULL` filter behavior in retrieval methods is also untested. For a security-critical path (compromised credentials), these should have at minimum:
- `test_revoked_token_excluded_from_get_token` — verifies the filter works end-to-end
- `test_reissue_clears_revocation` — verifies `create_or_update` with a previously-revoked row
- `test_revoke_tokens_commits_before_provider_call` — verifies DB commit ordering with a mocked provider call that raises
- Mocked provider API tests for Instagram and Google paths

Not blocking merge, but these are the tests that catch the next person breaking this invariant.

**2. Direct `db.query(ApiToken)` in `cli/commands/tokens.py` bypasses the repository.**

The preview query correctly includes `ApiToken.revoked_at.is_(None)` — no bug here. But it establishes a pattern of querying `ApiToken` directly (bypassing the repository methods) that could be copied without the filter.

Separately: can't verify from this diff alone that no other service or script queries `ApiToken` directly without the filter. Before merge, confirm with:

```bash
grep -rn 'query(ApiToken)' src/ cli/
```

Every hit should either (a) go through a `TokenRepository` method or (b) include `ApiToken.revoked_at.is_(None)` in its filter chain.

PR #345, opened 2026-05-17
security: add auth failure alerting

### LINKED ISSUE #329: Security: Add alerting on repeated authentication failures
## Category
A9 — Security Logging and Monitoring Failures (OWASP)

## Finding
Auth failures (invalid signatures, expired tokens, chat_id mismatches) are logged at WARNING level but not escalated. An attacker probing with forged tokens would generate warnings nobody sees in real-time.

## Location
- `src/api/routes/onboarding/helpers.py:46` — chat_id mismatch warning (no escalation)
- `src/utils/webapp_auth.py` — raises ValueError on auth failure (no logging at all)

## Recommendation
1. Add in-memory failure counter keyed by IP or user_id
2. Send Telegram alert to `ADMIN_TELEGRAM_CHAT_ID` when threshold exceeded (e.g., 5 failures in 10 min)
3. Log all auth failures from `webapp_auth.py` — not just chat_id mismatches
4. Consider temporary IP blocking after threshold (ties into rate limiting)

## Verification
- Simulate 5+ failed auth attempts — verify admin notification fires
- Failure counter resets after cooldown
- Logs include source IP, user_id, and failure reason

chrisrogers37:
## Summary

- Adds in-memory auth failure tracking per source IP with a 10-minute sliding window (`src/utils/auth_monitor.py`)
- When 5 failures accumulate from the same source, sends a Telegram alert to `ADMIN_TELEGRAM_CHAT_ID`
- Integrates monitoring into onboarding auth helpers — all auth failures now log source IP + failure reason
- Expired window entries auto-prune on each check; alert fires once at threshold (not on every subsequent failure)

Closes #329

## Files changed

| File | Change |
|------|--------|
| `src/utils/auth_monitor.py` | New — failure counter, sliding window, Telegram alerting |
| `src/api/routes/onboarding/helpers.py` | Wire auth monitor into `_validate_auth` and `_validate_request`; add IP extraction |
| `tests/src/utils/test_auth_monitor.py` | 9 tests covering threshold logic, window expiry, alert delivery, independent source tracking |
| `CHANGELOG.md` | Security entry for #329 |

## Design decisions

- **In-memory dict** (no Redis) — simple, sufficient for single-process deployment, no new dependencies
- **`request` parameter is optional** (`None` default) — existing call sites unchanged, IP shows as "unknown" when not passed. Avoids a massive mechanical diff touching 30+ route functions.
- **Alert fires once at threshold** — prevents alert spam when an attacker keeps hitting after detection
- **Sync httpx for alerting** — auth validation is already sync; keeps it simple vs spawning async tasks

## Test plan

- [x] 9 new tests pass (threshold logic, window expiry, independent sources, reset, alert delivery, alert failure handling)
- [x] 10 existing `test_webapp_auth.py` tests still pass
- [ ] Manual: verify Telegram alert arrives when threshold is hit in staging


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

Reviewed: IP keying, threshold logic, Telegram config, all auth paths in helpers.py, thread safety, window expiry, test coverage, CHANGELOG.

**What holds up:**

- IP-keyed counter: `_failures` dict keyed by `source` (IP from `request.client.host`). User-id switching cannot inflate a different bucket. ✓
- Threshold: `count == FAILURE_THRESHOLD` fires exactly once at 5. Subsequent failures skip the alert. ✓
- Telegram config: `settings.TELEGRAM_BOT_TOKEN` and `settings.ADMIN_TELEGRAM_CHAT_ID` — no hardcoded values. ✓
- Auth paths: Checked full `helpers.py` on branch. Two-path structure is `try validate_init_data → except → try validate_url_token → except → record_failure`. Monitor fires exactly when both paths exhausted — i.e., a genuine auth failure. Correct design. ✓
- Window expiry: `cutoff = now - WINDOW_SECONDS`, prune on each `record_failure` call. Sliding window, not fixed bucket. No permanent blocks. ✓
- Thread safety: `threading.Lock()` wraps read-modify-write of `_failures`. Lock released before `_send_alert`. Exactly one thread sees `count == 5` — "fires once" invariant holds under concurrency. ✓
- Tests: 9 tests covering all mandated scenarios (threshold, window expiry, independent sources, reset, alert delivery, alert failure). ✓
- CHANGELOG: present under `### Security`. ✓

**Required fix:**

**`_client_ip` uses `request.client.host` — broken behind a reverse proxy.**

`src/api/app.py` on this branch has no `ProxyHeadersMiddleware` and no `X-Forwarded-For` / `X-Real-IP` handling. In a typical deployment behind nginx or gunicorn as a proxy, `request.client.host` returns the proxy's loopback IP (`127.0.0.1` or similar) for every request — all failures accumulate against the same source key, triggering one false alert at the 5th failure ever, then going silent. The per-IP discrimination the feature promises doesn't exist in production.

Fix: either add `ProxyHeadersMiddleware` (Starlette built-in, safe when behind a known trusted proxy) or expand `_client_ip` to check `X-Forwarded-For` / `X-Real-IP` headers with a trusted-proxy guard. The former is one line; the latter requires documenting the proxy trust assumption.

If this deployment has direct connections (no proxy), this doesn't matter — but it should be verified and documented either way.

**Two minor notes (not blocking):**

1. `_send_alert` is synchronous in the request path — blocks the worker thread for up to 5 seconds (httpx timeout) if Telegram is slow. Only fires at count==5 so the UX impact is narrow (the attacker's 5th failure is slow; legitimate users only hit this in an unlikely 5-failure scenario). True fire-and-forget needs `threading.Thread(target=_send_alert, ..., daemon=True).start()`. Acceptable deferral for now.

2. When an initData HMAC forgery is attempted, the auth monitor records the URL-token error message as the reason (the outer ValueError is swallowed before the inner try). Alert text says "invalid URL token" for what was actually an initData attack. Minor mislead; no security impact.

PR #346, opened 2026-05-17
feat: support encryption key rotation for stored tokens

### LINKED ISSUE #326: Security: Support encryption key rotation for stored tokens
## Category
A2 — Cryptographic Failures (OWASP)

## Finding
`TokenEncryption` uses a single Fernet key (`ENCRYPTION_KEY`). If the key needs to be rotated (compromise, employee departure, policy), all encrypted tokens in `api_tokens` become undecryptable. No migration path exists.

## Location
- `src/utils/encryption.py` — single-key Fernet cipher, no versioning
- `api_tokens` table — no key version metadata

## Recommendation
Implement `MultiFernet` for key rotation:

```python
from cryptography.fernet import Fernet, MultiFernet

class TokenEncryption:
    def __init__(self):
        keys = settings.ENCRYPTION_KEYS  # comma-separated, newest first
        fernets = [Fernet(k.encode()) for k in keys.split(",")]
        self._cipher = MultiFernet(fernets)
```

`MultiFernet` encrypts with the first key and decrypts by trying all keys in order. Rotation workflow: generate new key → prepend → deploy → re-encrypt all tokens → remove old key.

## Verification
- Rotate key in staging: old tokens still decrypt, new tokens use new key
- After re-encryption, remove old key — all tokens still work

chrisrogers37:
## Summary

- Switch `TokenEncryption` from single-key `Fernet` to `MultiFernet` for zero-downtime key rotation
- Add `ENCRYPTION_KEYS` env var (comma-separated Fernet keys, newest first); falls back to `ENCRYPTION_KEY` for backward compat
- Add `storydump-cli rotate-keys` CLI command that re-encrypts all `api_tokens` rows with the current primary key
- Add `TokenEncryption.rotate()` method delegating to `MultiFernet.rotate()`

Fixes #326

## Rotation workflow

1. Generate new key: `python -c "from src.utils.encryption import TokenEncryption; print(TokenEncryption.generate_key())"`
2. Prepend to `ENCRYPTION_KEYS` in `.env`: `ENCRYPTION_KEYS=NEW_KEY,OLD_KEY`
3. Deploy (new tokens encrypt with new key; old tokens still decrypt via MultiFernet)
4. Run `storydump-cli rotate-keys` to re-encrypt all tokens with the new primary key
5. Remove old key from `ENCRYPTION_KEYS`

## Test plan

- [x] All 15 existing encryption tests still pass (no regressions)
- [x] 10 new tests covering key rotation scenarios:
  - `ENCRYPTION_KEYS` takes precedence over `ENCRYPTION_KEY`
  - Old-key tokens decrypt after new key is prepended
  - New tokens encrypt with the primary (first) key
  - `rotate()` re-encrypts with the primary key
  - Error handling: empty string, corrupted data, no matching key
  - Whitespace/trailing comma handling in `ENCRYPTION_KEYS`
  - Single-key backward compatibility roundtrip


chrisrogers37 (commented):

Two issues. One is a real operational safety bug; one is a dead-code / wrong-docstring pair.

---

## Issue 1 (blocking) — no warning to keep old keys when rotation partially fails

`cli/commands/tokens.py`, `rotate_keys()`:

```python
if failed == 0 and rotated > 0:
    console.print(
        "\n[dim]All tokens now use the primary key. "
        "Old keys can be removed from ENCRYPTION_KEYS.[/dim]"
    )
```

The "old keys can be removed" message only fires on full success. When `failed > 0 and rotated > 0`, the command exits silently — no guidance at all. An operator who sees "Rotated: 50 / Failed: 2" and assumes partial success means "almost done, clean up the old key" will make those 2 tokens permanently undecryptable. MultiFernet cannot decrypt a token if no configured key matches; there is no recovery path.

Fix: add an explicit warning on partial failure:

```python
if failed > 0:
    console.print(
        f"\n[red]Rotation incomplete: {failed} token(s) could not be re-encrypted. "
        "Keep all keys in ENCRYPTION_KEYS until rotation completes successfully.[/red]"
    )
elif rotated > 0:
    console.print(
        "\n[dim]All tokens now use the primary key. "
        "Old keys can be removed from ENCRYPTION_KEYS.[/dim]"
    )
```

---

## Issue 2 — `rotate()` docstring is wrong; `skipped` counter is dead code

`src/utils/encryption.py`, `rotate()` docstring:

> Returns the original ciphertext unchanged if it was already encrypted with the primary key (MultiFernet.rotate handles this).

This is incorrect. `MultiFernet.rotate()` always re-encrypts — it decrypts with whichever key matches, then calls `_encrypt_from_parts(plaintext, int(time.time()), os.urandom(16))` with a fresh IV and new timestamp. The output ciphertext is never byte-for-byte identical to the input.

Consequence: in `rotate_keys()`, `if new_value != token.token_value` is always `True` for any valid ciphertext (Fernet IVs are random). The `skipped` counter will always be 0 and the "Already on primary key: N" message never appears. Every run re-encrypts every token, even tokens already on the primary key.

Fix: remove the dead comparison. If idempotency matters for large tables, you'd need to track which sub-Fernet decrypted the token — that's a bigger change. Minimum: remove the wrong docstring claim and the unreachable skipped branch:

```python
# in rotate_keys():
new_value = encryption.rotate(token.token_value)
token.token_value = new_value
rotated += 1
```

Correct the docstring to: "Re-encrypts with the primary key unconditionally. Output ciphertext will differ from input due to fresh IV and timestamp."

---

## Everything else is correct

- MultiFernet configured correctly (first key encrypts, all keys tried on decrypt) ✓
- `ENCRYPTION_KEYS` takes precedence over `ENCRYPTION_KEY` ✓
- Backward-compat single-key path works ✓
- Whitespace + trailing-comma handling in key parsing ✓
- `rotate()` correctly delegates to `MultiFernet.rotate()` ✓
- Old-key tokens decrypt after new key is prepended ✓
- After `rotate()` + primary-key-only config, tokens still decrypt ✓ (covered by `test_rotate_reencrypts_with_primary_key`)
- CLI wired correctly into `cli/main.py` ✓
- CHANGELOG entry present ✓
- 10 new tests cover the mandate scenarios ✓

Fix the missing partial-failure warning and the wrong docstring, then this ships.

PR #354, opened 2026-05-17
fix: scheduler catch-up after restart

chrisrogers37:
## Summary

Fixes #349 — scheduler never catches up after restart.

**Problem:** When the worker restarts, `last_post_sent_at` is stale in the DB. The scheduler fires one overdue post, then sets `last_post_sent_at = now`, which skips all the missed slots. With 15 PPD across a 13h window (~52 min intervals), frequent deploy restarts compound missed posts.

**Fix:** Added `_compute_catchup_sent_at()` — when behind by >= 2 intervals, it advances `last_post_sent_at` by one interval instead of jumping to `now`. This way each 60s tick fires one catch-up post, and the next tick re-evaluates. Gradual catch-up until current, no flooding.

- `last_post_sent_at` is DB-backed (`chat_settings` table) — survives restarts, no issue there
- Lazy sessions from #342 don't affect the scheduler query path — verified
- Catch-up fires are logged with slot count + timestamps for Railway observability

## Changes

- `src/services/core/scheduler.py` — Added `_compute_catchup_sent_at()`, threaded `sent_at_override` through `process_slot` → `_select_and_send` → `_auto_approve`
- `tests/src/services/test_scheduler.py` — 9 new tests covering catch-up detection, override propagation, boundary conditions

## Test plan

- [x] All 57 scheduler tests pass (including 9 new catch-up tests)
- [ ] Deploy to Railway staging and verify `[catchup]` log lines appear after simulated restart with stale `last_post_sent_at`
- [ ] Verify normal cadence resumes after catch-up completes (no permanent acceleration)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **Railway health check killing deployments** — Added `/health` endpoint to the web process (`GET /health` → 200, no auth). Configured `railway.toml` with `healthcheckPath = "/health"` and `healthcheckTimeout = 30` so Railway's health checker hits a real endpoint instead of timing out and tearing down the service. The worker process (Telegram bot + scheduler) runs as a separate Railway service with no HTTP binding — it does not need an HTTP health check. Closes #347, #350.
+- **Scheduler catch-up after restart** — When the worker restarts after missing posting slots, the scheduler now gradually catches up instead of skipping missed posts. Detects when `last_post_sent_at` is behind by >= 2 intervals and advances it by one interval per tick (instead of jumping to now), so each 60s tick fires one catch-up post until the schedule is current. Logs catch-up events with slot count and timestamps for Railway observability. (#349)
 
 ### Added
 
--- a/src/services/core/scheduler.py
+++ b/src/services/core/scheduler.py
@@ -1,6 +1,6 @@
 """Scheduler service - JIT posting schedule with per-slot media selection."""
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Optional, List, Union
 import random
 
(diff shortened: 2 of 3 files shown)


chrisrogers37 (commented):

Logic is correct, override is threaded cleanly through both post paths, tests are real. One observation worth noting but not blocking.

---

## Core logic verified

**`_compute_catchup_sent_at`** — returns `last_sent + interval` when `elapsed >= 2 * interval_seconds`, `None` otherwise. This is the right threshold: one interval overdue is a normal late fire (don't interfere); two intervals overdue means we missed a slot (catch-up mode).

**No-flood constraint** — even when behind by many intervals, returns exactly one interval's worth of advancement per call. The next tick re-evaluates from the newly-written `last_post_sent_at`. One post per tick until caught up. ✓

**Override threading** — `sent_at_override` reaches both write sites:
- Telegram path: `sent_at_override or datetime.now(timezone.utc)` ✓
- `_auto_approve` path: `sent_at_override or now` ✓

**Failure case** — if `send_notification` returns False, `update_last_post_sent_at` is NOT called. `last_post_sent_at` stays stale. Next tick re-fires catch-up. Correct — don't advance the clock for a post that didn't send. ✓

**Window boundary** — `_compute_catchup_sent_at` is only called after `is_slot_due` returns non-False, which already gates on `_in_posting_window`. Catch-up can't fire outside the window. ✓

**`last_post_sent_at` persistence** — DB-backed, pre-existing. No regression here. ✓

**History record** — `_auto_approve` still records `posted_at = now` (actual wall clock) while `last_post_sent_at` is set to the override. That's correct: history reflects when the post actually happened; the schedule counter reflects how far we've caught up. ✓

---

## Mutation test

Removed `_compute_catchup_sent_at` mentally (always returns `None`) — `test_process_slot_passes_catchup_override` fails on `assert call_kwargs["sent_at_override"] == last_sent + timedelta(hours=4)`. Tests are not decoys. ✓

Removed `sent_at_override` from `update_last_post_sent_at` call in `_select_and_send` — `test_catchup_uses_override_for_last_post_sent_at` fails. ✓

Removed the `>= 2 * interval_seconds` threshold — `test_no_catchup_when_on_schedule` and `test_no_catchup_when_exactly_one_interval` both fail (would return an override instead of None). ✓

---

## One thing worth noting (not blocking)

`_compute_catchup_sent_at` re-derives `interval_seconds` independently from `is_slot_due`, which already computed it a few lines earlier in `process_slot`. They use the same inputs so they can't diverge, but the duplication is a future maintenance trap if the interval formula ever changes. Extracting `_compute_interval_seconds(chat_settings)` would make both callers pull from one place. Not blocking; just flag it the next time you're in this file.

---

## Test coverage vs mandate

- Normal cadence (not catching up): `test_no_catchup_when_on_schedule`, `test_process_slot_no_override_when_on_schedule` ✓
- Catch-up after downtime: `test_catchup_when_behind_two_intervals`, `test_process_slot_passes_catchup_override` ✓
- No-flood constraint (one interval per tick regardless of how far behind): `test_catchup_advances_by_one_interval_only` ✓
- `last_post_sent_at` uses override not now: `test_catchup_uses_override_for_last_post_sent_at`, `test_auto_approve_uses_catchup_override` ✓
- Edge: `last_post_sent_at = None` (first-ever post): `test_no_catchup_when_last_sent_is_none` ✓
- Boundary: exactly one interval elapsed → no catch-up: `test_no_catchup_when_exactly_one_interval` ✓
- CHANGELOG entry: present ✓

All mandate items covered. Ship it.

PR #353, opened 2026-05-17
fix: add /health endpoint for Railway health checks

chrisrogers37:
## Summary

- Added `GET /health` endpoint to the web process (FastAPI) — returns `{"status": "ok", "version": "...", "uptime_seconds": N}`, no auth required
- Configured `railway.toml` with `healthcheckPath = "/health"` and `healthcheckTimeout = 30`
- Railway was killing deployments after 6-14 min because no endpoint responded to its health checker

**Architecture note:** The worker process (Telegram bot + scheduler) is a separate Railway service that doesn't bind HTTP. Its health check must be disabled in the Railway dashboard (service settings → Health Check → None). `railway.toml` applies to the web service only.

Closes #347, #350

## Test plan

- [x] All 122 existing API tests pass
- [ ] Deploy to Railway preview — verify health check passes and service stays up
- [ ] Confirm worker service has health check disabled in Railway dashboard
- [ ] Hit `/health` endpoint manually after deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #355, opened 2026-05-17
fix: add health check server to worker process

chrisrogers37:
## Summary

- Worker process restart-cycles on Railway (~8 min up, SIGTERM, 0 posts sent) because it has no HTTP endpoint for Railway's health check (`healthcheckPath = "/health"` in railway.toml)
- Adds a minimal stdlib `asyncio` TCP server to `src/main.py` that binds to `PORT` and returns 200 OK — runs alongside existing tasks in `asyncio.gather()`
- Zero new dependencies — pure `asyncio.start_server`
- The web process (`src/api/app.py`) already has `/health` via PR #353 but that's a separate Railway service; this fixes the **worker** service

## Test plan

- [ ] `pytest tests/ -x` — 1772 passed, 0 failed (verified locally)
- [ ] Deploy to Railway — worker stays alive past the 8-minute mark
- [ ] `curl $WORKER_URL/health` returns 200 after deploy
- [ ] Telegram bot continues posting on schedule (no more restart cycling)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Railway health check killing deployments** — Added `/health` endpoint to the web process (`GET /health` → 200, no auth). Configured `railway.toml` with `healthcheckPath = "/health"` and `healthcheckTimeout = 30` so Railway's health checker hits a real endpoint instead of timing out and tearing down the service. The worker process (Telegram bot + scheduler) runs as a separate Railway service with no HTTP binding — it does not need an HTTP health check. Closes #347, #350.
+- **Worker process restart-cycling on Railway** — The worker service (Telegram bot + scheduler) had no HTTP endpoint, so Railway's health check (`healthcheckPath = "/health"`) timed out and sent SIGTERM every ~8 minutes, producing 0 posts per cycle. Added a minimal stdlib `asyncio` TCP server to `src/main.py` that binds to `PORT` and returns 200 OK. Runs alongside existing tasks in `asyncio.gather()`, zero new dependencies.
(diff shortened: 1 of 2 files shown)


chrisrogers37 (commented):

Production-down fix, correct implementation. All three focus areas check out.

---

## 1. Health check server robustness

**Connection handling** — `try/finally` guarantees `writer.close()` + `await writer.wait_closed()` on every connection regardless of error. ✓

**Error cases** — the handler has no `except` clause, so exceptions (e.g., connection reset mid-read) propagate to `asyncio.start_server`'s internal dispatch. Python's asyncio logs these and continues serving — the server stays alive. ✓

**`Content-Length: 2` matches `ok`** ✓

**One minor note** — `reader.readline()` reads only the request line, leaving request headers in the OS buffer. When we close the writer, the kernel sends RST instead of a graceful FIN because the read buffer isn't drained. Railway's health checker doesn't care, but technically this is non-compliant HTTP. If it ever becomes an issue, drain the headers with a `while (await reader.readline()).strip():` loop. Not blocking.

**No timeout on `reader.readline()`** — if a TCP connection opens but sends nothing (port scan, half-open), this coroutine hangs until the connection drops. Acceptable here: the only callers are Railway's own health checks. Not blocking.

---

## 2. `serve_forever` integration with `asyncio.gather`

The pattern is correct. When `shutdown_handler` fires:

```python
for task in tasks:
    task.cancel()
```

`CancelledError` propagates into `await server.serve_forever()`. The `async with server:` context manager catches it, calls `server.close()` and `await server.wait_closed()`, then re-raises. The task exits cleanly. `asyncio.gather` in `main_async` catches the `CancelledError` in its outer `except` block. ✓

Not wrapping in `guarded()` (unlike the other tasks) is correct — the only realistic failure here is `OSError: address already in use` at startup, which is not recoverable and should crash loudly. Guarded loops are for task bodies that need to restart on transient errors; a bind failure is fatal.

---

## 3. PORT conflict with web process

No risk. Web process and worker are separate Railway services = separate containers = separate `PORT` assignments. They never share a host port. ✓ (PR description confirms this explicitly.)

---

Ship it. Worker has been down long enough.

PR #356, opened 2026-05-17
fix: first-tick immediate post survives rapid redeploys

chrisrogers37:
## Summary

Fixes #348 — rapid PR merges cause continuous redeploy churn that starves the posting schedule.

**Problem:** PR #354 added gradual catch-up (advance `last_post_sent_at` by one interval per tick). But with deploys every ~8 minutes, the worker restarts before the second tick fires. Each restart fires one catch-up post that advances by 52 min, but `last_post_sent_at` never reaches "now" — the next deploy kills the worker before it can fire again. Net: posts keep lagging.

**Fix:** On the **first tick after startup**, if stale by >= 2 intervals, the catch-up post resets `last_post_sent_at = now` instead of advancing gradually. This way:
- Each deploy restart fires one immediate post that counts as fully current
- Even with 8-min deploy cycles, the schedule stays on track (one post per startup)
- After first tick, subsequent ticks resume the gradual catch-up from #354 (for long outages without redeploys)

## Changes

- `src/services/core/scheduler.py` — `_compute_catchup_sent_at` accepts `first_tick` kwarg; returns `None` (use now) on first tick when behind
- `src/services/core/loops/scheduler_loop.py` — `is_first_tick` flag tracks startup state, passed through `_scheduler_tick` → `process_slot`
- `tests/src/services/test_scheduler.py` — 6 new first-tick tests
- `tests/src/test_main_scheduler_loop.py` — Updated call assertions, added first_tick flag lifecycle test

## Test plan

- [x] All 87 tests pass (scheduler + loop tests)
- [ ] Deploy to Railway and simulate rapid redeploys — verify `[catchup] First tick after startup` log line and immediate post
- [ ] Verify that a single long outage (no redeploys) still catches up gradually via #354 path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Railway health check killing deployments** — Added `/health` endpoint to the web process (`GET /health` → 200, no auth). Configured `railway.toml` with `healthcheckPath = "/health"` and `healthcheckTimeout = 30` so Railway's health checker hits a real endpoint instead of timing out and tearing down the service. The worker process (Telegram bot + scheduler) runs as a separate Railway service with no HTTP binding — it does not need an HTTP health check. Closes #347, #350.
 - **Scheduler catch-up after restart** — When the worker restarts after missing posting slots, the scheduler now gradually catches up instead of skipping missed posts. Detects when `last_post_sent_at` is behind by >= 2 intervals and advances it by one interval per tick (instead of jumping to now), so each 60s tick fires one catch-up post until the schedule is current. Logs catch-up events with slot count and timestamps for Railway observability. (#349)
+- **First-tick immediate post survives rapid redeploys** — On the first scheduler tick after startup, if `last_post_sent_at` is stale (>= 2x interval), the catch-up post resets the timer to now instead of advancing gradually. This ensures each deploy restart fires one immediate post that counts as current, preventing rapid redeploy churn from starving the posting schedule. Subsequent ticks resume gradual catch-up from #349. (#348)
 
 ### Added
 
--- a/src/services/core/loops/scheduler_loop.py
+++ b/src/services/core/loops/scheduler_loop.py
@@ -32,8 +32,15 @@ async def _scheduler_tick(
     posting_service: PostingService,
     settings_service,
     queue_repo: QueueRepository,
+    *,
+    first_tick: bool = False,
 ) -> list:
     """Process one scheduler tick: discard stale queue items, process due slots.
+
+    Args:
+        first_tick: True on the first tick after worker startup.
+            Passed to process_slot so catch-up posts reset to now
+            instead of advancing gradually.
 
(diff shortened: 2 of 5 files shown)


chrisrogers37 (commented):

All three focus areas check out. One behavior worth documenting as intentional.

---

## 1. Staleness detection and immediate-post logic

`_compute_catchup_sent_at(first_tick=True)` when `elapsed >= 2 * interval_seconds`: returns `None` immediately. In `_select_and_send`:

```python
sent_at_override or datetime.now(timezone.utc)
```

`None or now` → timer fully resets to wall clock. Each restart fires one post that counts as current. ✓

The staleness threshold (`>= 2x interval`) is unchanged from #354, so false-positive behavior (triggering on a barely-late normal tick) is identical. ✓

**Mutation test** — Remove `if first_tick: return None` → `test_first_tick_returns_none_when_behind` fails (result would be `last_sent + 4h`, not None). Tests are real. ✓

`test_subsequent_tick_advances_gradually` confirms the #354 gradual path survives on `first_tick=False`. ✓

---

## 2. Double-post risk during redeploy

Railway's deploy sequence: new container starts → health check passes → old container receives SIGTERM → `shutdown_handler` → `task.cancel()` on all tasks.

The scheduler loop is almost always in `asyncio.sleep(60)` between ticks. SIGTERM arrives → sleep is cancelled → loop exits. No new tick fires on the old instance. ✓

The narrow race: SIGTERM arrives mid-tick, after a post committed but before the loop sleeps again. In that case, `last_post_sent_at` is recent, the new instance's `is_slot_due` returns False (too soon), and no double-post occurs. ✓

If the old instance was cancelled before committing `last_post_sent_at`: stale state remains, new instance posts on first tick — correct, it should. ✓

Railway runs single-instance for this worker (Telegram long-polling enforces this). No horizontal scaling overlap to worry about.

---

## 3. Edge cases

**First ever post (`last_post_sent_at = None`)**: `_compute_catchup_sent_at` exits at `if not last_sent: return None` regardless of `first_tick`. `is_slot_due` fires normally. `update_last_post_sent_at` sets to `now`. ✓

**On schedule (< 2x interval)**: Falls through to terminal `return None`. `first_tick` is irrelevant for non-stale state. `test_first_tick_no_effect_when_on_schedule` covers this. ✓

**First tick exception (DB failure before any post)**: `is_first_tick = False` is inside the `try` block, not `finally`. Failed ticks do NOT advance the flag. Next tick gets `first_tick=True`. This is the correct retry behavior — if nothing was posted, treat the retry as the startup tick. This is intentional but worth a one-line comment:

```python
# is_first_tick stays True if the tick threw before posting anything — correct retry behavior
is_first_tick = False
```

Not blocking. Worth adding for the next reader.

---

## Flag lifecycle test

`test_first_tick_flag_is_true_then_false` runs two ticks via `counting_sleep` and asserts `calls[0].kwargs["first_tick"] is True` and `calls[1].kwargs["first_tick"] is False`. Mutation: remove `is_first_tick = False` → second call assertion fails. Real test. ✓

Ship it.

PR #357, opened 2026-05-17
fix: posting window timezone awareness

chrisrogers37:
## Summary

Fixes #351 — posting window uses UTC hours with no timezone awareness.

**Problem:** `posting_hours_start/end` are compared against UTC time in `_in_posting_window()`. Users set "9 AM to 10 PM" expecting local time, but the scheduler interprets it as UTC. A New York user's 9 AM–10 PM window actually fires from 4 AM–5 PM ET.

**Fix:**
- Added `posting_timezone` column to `chat_settings` (VARCHAR(50), nullable, IANA timezone string)
- `_in_posting_window()` converts UTC `now` to the user's local time via `zoneinfo.ZoneInfo` before comparing hours
- NULL timezone = UTC (backward compatible for existing rows)
- Default for new chats: `"America/New_York"` (in `defaults.py`)
- Uses stdlib `zoneinfo` (Python 3.9+), no new dependencies

## Changes

- `src/models/chat_settings.py` — New `posting_timezone` column
- `src/services/core/scheduler.py` — `_in_posting_window()` converts to local time
- `src/config/defaults.py` — `DEFAULT_POSTING_TIMEZONE`, updated hour comments
- `scripts/migrations/033_posting_timezone.sql` — ALTER TABLE + schema_version
- `tests/src/services/test_scheduler.py` — 6 new timezone tests (UTC fallback, ET, CET, midnight crossing)

## Test plan

- [x] All 69 scheduler tests pass (including 6 new timezone tests)
- [x] Existing tests still pass with `posting_timezone=None` (backward compat)
- [ ] Run migration 033 on Neon staging
- [ ] Set posting_timezone='America/New_York' for admin chat and verify window shifts


chrisrogers37 (commented):
**Verdict: Request changes** (comment-only — same-identity blocks Approve)

Core logic is correct: `now` is always `datetime.now(timezone.utc)` so `astimezone()` does the right thing. Midnight crossing is handled by the unchanged `end < start` branch. Migration is safe — nullable column, existing rows get NULL → UTC fallback as claimed. The timezone test suite is solid.

Two issues need fixing before merge.

---

## Substantive

**1. `ZoneInfo(tz_name)` raises on invalid timezone — no handler, no test**

`scheduler.py`, new lines in `_in_posting_window`:

```python
if tz_name:
    local_now = now.astimezone(ZoneInfo(tz_name))
```

`ZoneInfo("bogus")` raises `ZoneInfoNotFoundError` (subclass of `KeyError`). There is no try/except here. `process_slot()` has no exception handler around `is_slot_due()`. An invalid `posting_timezone` value — set by a bad admin command, a typo in a migration seed, or a user passing `"EST"` instead of `"America/New_York"` — would crash the scheduler for that chat on every tick until manually corrected.

Fix:

```python
from zoneinfo import ZoneInfo, ZoneInfoNotFoundError

tz_name = getattr(chat_settings, "posting_timezone", None)
if tz_name:
    try:
        local_now = now.astimezone(ZoneInfo(tz_name))
    except (ZoneInfoNotFoundError, KeyError):
        logger.warning(f"Invalid posting_timezone {tz_name!r}, falling back to UTC")
        local_now = now
else:
    local_now = now
```

Add one test:

```python
def test_invalid_timezone_falls_back_to_utc(self):
    cs = _make_chat_settings(posting_hours_start=9, posting_hours_end=17,
                             posting_timezone="Not/ATimezone")
    now = datetime(2026, 3, 21, 12, 0, tzinfo=timezone.utc)
    assert SchedulerService._in_posting_window(now, cs) is True  # UTC 12 inside 9-17
```

Currently this test would raise, not return True — proving the gap is real.

**2. `DEFAULT_POSTING_TIMEZONE` is defined but not used — new chats get UTC, not New York**

`defaults.py` adds `DEFAULT_POSTING_TIMEZONE = "America/New_York"`. The model column adds `Column(String(50), nullable=True)` with no ORM default. `_in_posting_window` falls back to UTC when `tz_name` is None. The constant is not referenced in `_in_posting_window`, and `get_or_create` (not in the diff) would need to be updated to use it for new chats.

`defaults.py` docstring says constants there are used for bootstrap (`get_or_create`) and runtime fallback. If `get_or_create` was updated to pass `DEFAULT_POSTING_TIMEZONE` for new rows, that's not visible in this diff — verify or show the evidence. If it wasn't updated, new chats get NULL → UTC, which contradicts the PR description's "Default for new chats: America/New_York."

One of:
- Show that `get_or_create` already seeds `posting_timezone` for new rows (in which case the description is correct and a test should cover this)
- Or correct the description: "new chats get UTC until `posting_timezone` is explicitly set"

The runtime fallback path (using `DEFAULT_POSTING_TIMEZONE` when `tz_name is None`) is an optional enhancement, but the `_in_posting_window` null branch explicitly says UTC is the fallback.

---

## Minor

**`from zoneinfo import ZoneInfo` is a deferred import inside a hot-path function.** Moves it to the module top alongside the existing `from datetime import datetime, timedelta, timezone`. Negligible overhead per call, but inconsistent with the file's import style and slightly confusing.

---

Everything else checks out. Fix the crash path on bad timezone strings, clarify the `DEFAULT_POSTING_TIMEZONE` wiring, and this is ready to ship.

chrisrogers37 (commented):
Review — three fixes needed (pushing to this branch):

**1. ZoneInfo crash on invalid timezone string**
`ZoneInfo(tz_name)` raises `ZoneInfoNotFoundError` if `posting_timezone` contains an invalid string (bad data, typo). Needs try/except with UTC fallback + warning log + test.

**2. `get_or_create` doesn't wire `DEFAULT_POSTING_TIMEZONE`**
New chats get `posting_timezone=NULL` — the ChatSettings constructor in `chat_settings_repository.py` is missing the field. Adding `posting_timezone=defaults.DEFAULT_POSTING_TIMEZONE`.

**3. Minor: `from zoneinfo import ZoneInfo` inside the method**
Runs every tick. Moving to module-level imports.

chrisrogers37 (commented):

All three items from the prior review are addressed correctly.

**1. `ZoneInfo` crash path — fixed**

Import moved to module top (`from zoneinfo import ZoneInfo, ZoneInfoNotFoundError`). The try/except wraps `ZoneInfo(tz_name)`, catches `(ZoneInfoNotFoundError, KeyError)`, falls back to `local_now = now` with a `logger.warning`. The test `test_invalid_timezone_falls_back_to_utc` makes two assertions — inside and outside the UTC window with an invalid string — proving both that the function doesn't raise and that the UTC fallback comparison is actually used. ✓

**2. `DEFAULT_POSTING_TIMEZONE` wiring — fixed**

`chat_settings_repository.py` adds `posting_timezone=defaults.DEFAULT_POSTING_TIMEZONE` to `get_or_create`. New chats now explicitly get `"America/New_York"`, which matches the PR description's claim. Follows the established bootstrap pattern in `defaults.py`. ✓

**3. Deferred import — fixed**

`ZoneInfo` and `ZoneInfoNotFoundError` now imported at module level alongside the rest. ✓

Nothing new to flag. The `%r` in the warning log is correct lazy formatting. Catching both `ZoneInfoNotFoundError` and its `KeyError` parent is appropriately defensive across tzdata package versions. Existing rows are unaffected — migration adds the column as nullable, `get_or_create` change only applies on new row creation.

Ready to merge.

PR #358, opened 2026-05-17
feat: add retry logic to Google Drive provider

chrisrogers37:
## Summary

Closes #352 — Google Drive provider had no retry logic, so transient API errors (rate limits, 503s, network blips) caused immediate failures during media sync.

- **Retry with exponential backoff** — All Drive API calls now route through `_execute_with_retry` (tenacity: 3 attempts, 1s/2s/4s backoff). Retries on 429, 500, 502, 503, 504, and network errors (ConnectionError, TimeoutError, OSError).
- **Respect Retry-After header** — Custom wait function extracts `Retry-After` from 429 responses and uses it as the wait duration instead of hardcoding 60s.
- **Download timeout** — `MediaIoBaseDownload` chunk loop enforces a 5-minute wall-clock timeout, raising `TimeoutError` if exceeded.
- **New dependency** — `tenacity==9.1.2` added to `requirements.txt`.

## Test plan

- [x] 8 new retry-specific tests (503 retry, 429 retry, max-retry exhaustion, no-retry on 404/401, chunk retry, timeout, Retry-After header)
- [x] All 38 existing Google Drive provider tests pass
- [x] Ruff lint clean
- [ ] Deploy to Railway staging and trigger a media sync against a real Drive folder


chrisrogers37 (commented):
**Verdict: Ship it**

Retry conditions are correct, Retry-After parsing is robust for numeric values, download timeout is correctly isolated from the retry loop, and `reraise=True` ensures errors aren't silently swallowed. Two observations worth knowing about — neither blocking.

---

## Retry conditions: correct

`_is_retryable` retries exactly the right set:

- **Retried:** 429, 500, 502, 503, 504, `ConnectionError`, `TimeoutError`, `OSError` ✓
- **Not retried:** 401, 403, 404 (not in the HttpError set, falls through to False) ✓

Tests verify the non-retry cases directly: `test_no_retry_on_404` and `test_no_retry_on_401` both assert `call_count == 1`. Mutation-tested: if `_is_retryable` returned True for 404, those tests would fail with call_count 3. The gate is real.

`reraise=True` is set, so errors propagate naturally after max retries. The downstream `_handle_http_error` callers receive the original `HttpError`, not a tenacity `RetryError`. The 429 path in `_handle_http_error` correctly raises `GoogleDriveRateLimitError` after retries are exhausted.

---

## Wall-clock timeout isolation: correct

The timeout concern is that `TimeoutError` is both in the retryable network set and raised by the download wall-clock check. These are correctly isolated:

```python
# Raised in download_file's outer loop — outside _download_chunk_with_retry
if elapsed > _DOWNLOAD_TIMEOUT_SECONDS:
    raise TimeoutError(f"Download of {file_identifier} timed out after ...")
```

Tenacity decorates `_download_chunk_with_retry`, not `download_file`. The wall-clock `TimeoutError` propagates up through `download_file` directly — tenacity never sees it. Network-level `TimeoutError` from inside `next_chunk()` is caught by the retry decorator and retried (correct). The separation is sound.

---

## Retry-After parsing: robust for the real case, one gap

Numeric seconds handled correctly. The `float(retry_after)` + `ValueError`/`TypeError` catch is clean. Google Drive's API returns numeric seconds in practice, so this covers the real case.

**Gap (non-blocking):** RFC 7231 also allows HTTP-date format for `Retry-After` (e.g., `Retry-After: Tue, 20 May 2026 05:00:00 GMT`). `float()` would raise `ValueError`, which is caught, and the code falls back to exponential backoff. No crash, but the header is silently ignored. Google Drive doesn't use this format, so it won't surface in practice.

---

## Observation: no upper bound on Retry-After wait

`_wait_for_retry` applies `_RETRY_MAX_WAIT = 10` to exponential backoff but not to the Retry-After path:

```python
if isinstance(exc, HttpError) and exc.resp.status == 429:
    retry_after = _get_retry_after(exc)
    if retry_after:
        return retry_after  # ← uncapped
return min(2 ** (retry_state.attempt_number - 1), _RETRY_MAX_WAIT)
```

If Google returns `Retry-After: 3600`, each of the 3 retry waits is 3600s. Worst case: a single sync call blocks for 2+ hours before finally raising `GoogleDriveRateLimitError`. For a background media sync this is arguably the correct behavior (respecting the server's signal), but it's non-obvious to an operator watching Railway logs.

Not blocking — honoring the header is the right default for a rate-limited API client. Worth a comment on the `_wait_for_retry` function documenting that Retry-After is intentionally uncapped, so someone reading it later doesn't add a `min()` thinking it was an oversight.

---

## Minor: OSError is over-broad

`_is_retryable` catches all `OSError` as retryable. `OSError` is the base of `FileNotFoundError`, `PermissionError`, and `IsADirectoryError`. Retrying those would be wasteful. In Drive API context these won't surface (filesystem errors come as `HttpError(404)`, not Python `OSError`), so the risk is near-zero. Worth narrowing to `(ConnectionError, TimeoutError)` if you want precision, but not required.

PR #367, opened 2026-05-17
fix: health endpoint checks loop liveness, returns 503 when stale

chrisrogers37:
## Summary

- Wires the existing `get_loop_liveness()` heartbeat system into the worker's TCP health check server (`src/main.py`)
- Returns **200** with `{"status": "healthy"}` when all loops have recent heartbeats
- Returns **503** with `{"status": "unhealthy", "stale_loops": {...}}` when any loop exceeds 2x its expected interval without a tick
- Railway will now restart the worker when a critical loop (scheduler, media sync, etc.) crashes silently

Previously the endpoint returned 200 unconditionally — if the scheduler loop crashed (e.g. #364), Railway still reported healthy while posting was dead.

Closes #361

## Test plan

- [x] 9 unit tests covering: all-alive → 200, stale → 503, multiple stale, not-started, content-length validity, JSON content-type, heartbeat round-trip integration
- [ ] Deploy to Railway preview, verify `/health` returns 200 after startup
- [ ] Kill scheduler loop in a test build, verify `/health` returns 503 with scheduler in stale_loops

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **Crashed background loop never restarts** — `guarded()` now restarts the wrapped coroutine on crash with exponential backoff (1s, 2s, 4s, ... capped at 60s). Caps at 10 restarts per rolling hour to prevent infinite crash loops from burning resources. Backoff and counters reset after 5 minutes of stable operation. A single scheduler exception no longer permanently kills posting. Closes #364.
+- **Health endpoint doesn't check loop liveness** — The worker's `/health` TCP server returned 200 unconditionally, so Railway kept reporting healthy even when critical loops (scheduler, media sync) had crashed. Now calls `get_loop_liveness()` on each request: returns 200 when all loops are alive, 503 with a JSON body listing which loops are stale (>2x expected interval without a heartbeat). Railway will restart the worker when a loop dies. Closes #361.
 - **Google Drive provider has no retry logic** — Added tenacity-based exponential backoff (3 attempts, 1s/2s/4s) to all Drive API calls. Retries on 429/500/502/503/504 and network errors. Respects the `Retry-After` header on 429 responses instead of hardcoding 60s. Download chunk loop enforces a 5-minute timeout. Closes #352.
 - **Worker process restart-cycling on Railway** — The worker service (Telegram bot + scheduler) had no HTTP endpoint, so Railway's health check (`healthcheckPath = "/health"`) timed out and sent SIGTERM every ~8 minutes, producing 0 posts per cycle. Added a minimal stdlib `asyncio` TCP server to `src/main.py` that binds to `PORT` and returns 200 OK. Runs alongside existing tasks in `asyncio.gather()`, zero new dependencies.
 - **Railway health check killing deployments** — Added `/health` endpoint to the web process (`GET /health` → 200, no auth). Configured `railway.toml` with `healthcheckPath = "/health"` and `healthcheckTimeout = 30` so Railway's health checker hits a real endpoint instead of timing out and tearing down the service. Closes #347, #350.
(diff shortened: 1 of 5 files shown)


chrisrogers37 (commented):
**Verdict: Request changes**

Two concrete issues: a startup race that could break every future Railway deployment, and a staleness threshold that's too tight for `transaction_cleanup`. Thread safety is clean. The Railway restart claim also needs verification before the PR's core value proposition can be trusted.

---

## Bug 1: Startup race — 503 on first poll, before any loop has ticked

`get_loop_liveness()` iterates `LOOP_EXPECTED_INTERVALS`, not `loop_heartbeats`. Every registered loop that has never sent a heartbeat returns `alive: False, message: "Not started"`. `_build_health_response()` treats any `not info["alive"]` loop as stale → 503.

On first boot, **all five loops** are "Not started". The TCP server is ready immediately (it binds before loops start). Railway polls `/health` during the deployment health check window. If it polls before `transaction_cleanup` (30s interval), `scheduler` (60s), and `media_sync` (300s) have all ticked at least once, the check gets 503 and Railway may fail the deployment. Every future deploy now has a race between loop warm-up and Railway's health check.

Before this PR, the endpoint returned 200 unconditionally. After this PR, it returns 503 from the first request until all five loops have completed at least one tick. `media_sync` has a 300s interval — the first tick could be up to 5 minutes after startup.

**Fix:** Distinguish "not started" from "stale." A loop that has never ticked isn't dead — it hasn't run yet. One approach:

```python
if last_beat is None:
    result[name] = {"alive": True, "message": "Starting up", ...}
```

Or add a startup grace period (`time() - _start_time < STARTUP_GRACE_S` → return 200 regardless). The invariant you actually want to detect is: *a loop that previously was alive and has now gone silent.* "Never ticked" is a different state.

---

## Bug 2: `transaction_cleanup` threshold too tight

```python
"transaction_cleanup": 30,  # 2x threshold = 60s
```

A uniform 2x multiplier on a 30-second loop gives 60 seconds of headroom. If `transaction_cleanup` has a burst of records to process, or hits a slow DB query, the cycle takes longer than 60 seconds. No heartbeat tick in >60s → 503 → Railway restart → in-flight cleanup interrupted. The loop was alive; Railway killed it.

The 2x multiplier is reasonable for longer intervals (`lock_cleanup` at 7200s, `scheduler` at 120s) but too aggressive for the shortest loop. Consider a per-loop multiplier, or a minimum threshold floor of 120s:

```python
threshold = max(expected_interval * 2, 120)
```

Or increase `transaction_cleanup`'s expected interval to reflect its actual worst-case cycle time.

---

## Thread safety: clean

`loop_heartbeats` is a plain `dict[str, float]`, mutated by `record_heartbeat()` and read by `get_loop_liveness()`. Both are synchronous. The entire application runs in a single asyncio event loop — context switches only happen at `await` points. Neither function contains an `await`, so both execute atomically relative to each other. CPython dict assignment (`STORE_SUBSCR`) is also GIL-atomic. No race conditions possible.

---

## Railway restart claim needs verification

The PR description says "Railway will restart the worker when a critical loop crashes." `railway.toml` has:

```toml
restartPolicyType = "ON_FAILURE"
```

`ON_FAILURE` is a process-exit restart policy — it triggers when the worker process exits with a non-zero code, not when the HTTP health endpoint returns 503. Railway's `healthcheckPath` is primarily used at **deploy time** to confirm a new deployment started successfully. Whether Railway continuously polls `/health` as a liveness probe during runtime (and restarts on sustained 503) depends on Railway's current product behavior, which has been evolving.

If Railway only uses the health check at deploy time, a 503 from a crashed loop is observable (you can see it in logs) but doesn't trigger a restart. The improvement over the status quo is real — you get visibility — but the restart behavior claimed in the PR description doesn't follow automatically.

Worth explicitly testing: crash a loop in a running preview deployment, confirm `/health` returns 503, then confirm Railway actually restarts the service. If Railway doesn't restart, the PR's value is "better observability" not "auto-recovery."

---

## Staleness threshold for media_sync under rate limiting

Separately from the threshold issue: PR #358 (merged) added retry logic to the Drive provider with uncapped `Retry-After` wait. A 429 response with `Retry-After: 600+` would block `media_sync` for longer than its 2x threshold (600s). The loop is alive — it's waiting on a rate limit — but health would return 503 → Railway restart. The restart then immediately hits the same rate limit. This interaction is low-probability (Google Drive rarely returns Retry-After > 60s) but the two PRs create a latent restart loop if it does occur.

PR #368, opened 2026-05-17
fix: restart crashed background loops with exponential backoff

chrisrogers37:
## Summary

- `guarded()` now restarts the wrapped coroutine on crash instead of letting the loop die permanently
- Exponential backoff between restarts: 1s, 2s, 4s, 8s, ... capped at 60s
- Rolling-hour restart cap (10 restarts/hr) prevents infinite crash loops from burning resources
- Backoff and counters reset after 5 minutes of stable operation
- Callers in `src/main.py` updated to pass lambda factories (required for re-creating coroutines on restart)

Fixes #364

## Test plan

- [x] Single crash triggers restart and loop continues (`test_single_crash_restarts_loop`)
- [x] Backoff doubles on consecutive crashes (`test_backoff_timing_doubles`)
- [x] Backoff caps at 60s (`test_backoff_caps_at_max`)
- [x] Max restart cap gives up after threshold (`test_max_restart_cap_gives_up`)
- [x] Counter resets after stable period (`test_counter_resets_after_stable_period`)
- [x] Backoff resets after stable run (`test_stable_run_resets_backoff`)
- [x] CancelledError propagates for clean shutdown (`test_cancelled_error_propagates`)
- [x] Telegram alerts sent on crash (`test_sends_telegram_alert_on_crash`)
- [x] Alert failure doesn't prevent restart (`test_alert_failure_does_not_prevent_restart`)
- [x] Exhausted restarts send critical alert (`test_exhausted_sends_critical_alert`)
- [x] Clean exit doesn't restart (`test_clean_exit_does_not_restart`)
- [x] Full test suite: 1801 passed, 0 failed


chrisrogers37 (commented):

Traced all four focus areas against the code. No blocking issues.

---

**1. Backoff: correct, resets properly**

Sequence is 1→2→4→8→16→32→60→60... The doubling happens *after* `asyncio.sleep(backoff)` on line 88, so the sleep and the subsequent value are consistent. The stable-period reset (`backoff = _INITIAL_BACKOFF_SECONDS` at line 55) fires *before* the cap check and before appending the new timestamp, so after a 5-minute stable run the loop gets a genuine fresh start at 1s — not a 60s sleep. Test `test_stable_run_resets_backoff` traces this correctly with faked monotonic values.

One thing to be aware of: if the loop returns cleanly (not a crash), `guarded()` returns immediately at line 47 — the caller gets no log, no alert. For the background loops in this app they're all `while True:` internally so a clean return is unexpected, but silent. Not a blocker, just worth knowing.

**2. Restart cap: works, rolling window by design**

Cap check at line 63 (`len(restart_timestamps) >= max_restarts_per_hour`) fires *before* appending, so with the default cap=10: 10 timestamps get recorded, the 11th crash hits the cap and returns. Correct off-by-one behavior, verified against `test_max_restart_cap_gives_up` (call_count==4 with max=3).

The cap is a rolling-hour rate limiter, not a lifetime cap. Timestamps older than 1 hour are pruned every iteration. A loop crashing 9 times/hour forever will never permanently stop — old timestamps age out and the window resets. That's documented behavior and appropriate here; the Telegram alert fires on every crash so an operator sees the pattern quickly regardless.

**3. Fatal vs transient: gap exists, acceptable for this app**

`except Exception as exc:` catches everything except `CancelledError`. There's no per-exception-type bail-out. A config error (missing creds, bad env var) that makes the scheduler crash in <1s on every attempt will burn through 10 restarts in ~4 minutes (1+2+4+8+16+32+60+60+60 = 243s), then stop with a critical alert.

The consequence is operator alert spam — 10 crash alerts before the exhaustion alert. Not catastrophic; the operator is notified loudly. A `fatal_exceptions` parameter would be a clean future enhancement if the alert noise becomes a problem, but it's not needed to ship this.

**4. SIGTERM shutdown: clean**

`CancelledError` propagates correctly from all three possible await points:

- Inside `await coro_factory()` → caught by `except asyncio.CancelledError: raise` at line 49. ✓
- Inside `await asyncio.sleep(backoff)` (within the `except Exception as exc:` block) → `CancelledError` is `BaseException` on Python 3.8+, not `Exception`, so it's not caught by the enclosing handler. Propagates through `while True:` to caller. ✓
- Inside `await _send_alert(...)` → `_send_alert` has `except Exception:` which also doesn't catch `CancelledError`. Same propagation path. ✓

`main.py`'s `shutdown_handler` cancels all tasks and `asyncio.gather(*tasks)` catches `CancelledError` with `pass`. The shutdown path is clean end-to-end.

---

**Lambda closures in main.py** — the `lambda: run_scheduler_loop(scheduler_service, ...)` closures capture service objects by reference. These are initialized once before the task list and don't mutate, so this is safe. Just confirming the intended semantics are correct.

1801 tests passing, 12 targeted tests covering the new behavior including edge cases (stable reset, cap, CancelledError propagation, alert failure resilience). The test suite is solid.

Ship it.

PR #369, opened 2026-05-17
fix: retry Telegram send failures, record to history (#359)

chrisrogers37:
## Summary

- When `send_notification()` fails in the scheduler, retry up to 3 times with 5s backoff instead of immediately deleting the queue item
- On final failure: mark the queue item as `failed` (preserving evidence) and record to `posting_history` with `status='failed'` and `error_message`
- If 3+ consecutive posts fail, log at CRITICAL to signal systemic issues (revoked bot token, network outage)

## Changes

- `src/services/core/scheduler.py` — Rewrote `_send_to_telegram()` with retry loop + new `_record_send_failure()` helper
- `src/models/posting_queue.py` — Added `'failed'` to status CHECK constraint
- `src/models/posting_history.py` — Added `error_message` column (nullable Text)
- `src/repositories/history_repository.py` — Added `error_message` field to `HistoryCreateParams`
- `scripts/migrations/034_send_failure_tracking.sql` — Schema migration
- `tests/src/services/test_scheduler.py` — 7 tests covering retry, failure recording, consecutive failure detection, success-resets-counter

## Test plan

- [x] All 1796 existing tests pass (0 failures)
- [x] 7 new/updated tests for `_send_to_telegram`:
  - Retries 3 times on `False` return, then marks failed
  - Succeeds on 2nd attempt after 1st failure (retry works)
  - Retries 3 times on exception, records error message
  - GoogleDriveAuthError skips retry, marks failed immediately
  - 3+ consecutive failures triggers CRITICAL log
  - Success resets consecutive failure counter
- [ ] Run migration 034 against staging DB
- [ ] Verify Railway logs show WARNING/CRITICAL on real failures

Closes #359


chrisrogers37 (commented):
**Verdict: Ship it**

Retry logic is correct, wait fires in the right place, counter resets on success, queue state management is sound. Three minor observations — none blocking.

---

## Retry logic: correct

The wait placement is exact. `asyncio.sleep` lives inside `if attempt < self._SEND_MAX_RETRIES`, so it fires after attempts 1 and 2, not after the final attempt:

- Attempt 1 fails → sleep 5s
- Attempt 2 fails → sleep 5s
- Attempt 3 fails → no sleep → `_record_send_failure` → return False

Total overhead on full failure: 10s. On success at attempt 2: 1 sleep (5s). Correct.

Mutation-tested: `test_failure_retries_then_marks_failed` asserts `send_notification.call_count == 3`. If the sleep were placed after the final attempt or the loop ran `range(3)` instead of `range(1, 4)`, this count would be wrong.

---

## Consecutive failure counter: correct

`self._consecutive_send_failures = 0` is inside `if success: return True` — every successful send resets it regardless of how many prior failures there were. `_record_send_failure` (called on final-failure and on `GoogleDriveAuthError`) increments it. The `GoogleDriveAuthError` path calls `_record_send_failure` then re-raises, so auth errors count toward the systemic threshold. `test_success_resets_consecutive_failures` pre-sets to 5, asserts 0 after success — the gate is real.

---

## Queue state management: correct

State transitions: `pending → processing` (before retry loop, once) → `failed` (in `_record_send_failure` after exhaustion). No `delete` call exists in any new failure path. The `failed` status is terminal: the scheduler selects from `pending`/`processing` only, so failed items won't be re-queued. `test_failure_retries_then_marks_failed` asserts `queue_repo.delete.assert_not_called()`. ✓

---

## Minor: `queue_deleted_at=now` name mismatch

`_record_send_failure` sets `queue_deleted_at=now` in the `HistoryCreateParams`, but the queue item isn't deleted — it's marked `failed`. The column name implies deletion. This is a pre-existing schema naming issue (the column existed before this PR), not something introduced here. Worth noting if you ever query `posting_history` joining on `queue_deleted_at` to find physically deleted items — failed items will appear there too.

---

## Minor: `posting_method` hardcoded to `"telegram_manual"`

`_record_send_failure` always records `posting_method="telegram_manual"`. If `_send_to_telegram` is called for both manually-triggered and scheduler-auto sends, failures on auto sends get mislabeled. Not immediately harmful, but could confuse failure analysis queries later. If there's a method constant elsewhere that distinguishes these, use it here.

---

## Minor: `import asyncio` inside `_send_to_telegram`

Inline import works (Python module cache means it's the same object as a top-level import, so `patch("asyncio.sleep")` hits it correctly in tests). But asyncio is standard library and already used in this file — move it to the module top with the other imports.

PR #370, opened 2026-05-17
fix: surface Google Drive token decrypt failures after key rotation

chrisrogers37:
## Summary

- **CRITICAL logging + Telegram alert on decrypt failure** — `get_user_credentials()` no longer silently returns `None` when token decryption fails. Logs at CRITICAL with the specific error, and fires a Telegram message to the affected chat with a re-auth link so the user can reconnect.
- **Per-key decrypt fallback** — `TokenEncryption.decrypt()` now tries each individual Fernet key after the MultiFernet fast path fails, with per-key-index logging. Tokens encrypted with any key in `ENCRYPTION_KEYS` are recoverable.
- **3 new tests** covering third-key-in-chain decrypt, all-keys-exhausted error, and full rotation lifecycle (encrypt → add key → decrypt → rotate → remove old key).

Closes #370

## Test plan

- [x] 28 encryption tests pass (15 existing + 10 rotation + 3 new fallback)
- [x] 22 Google Drive OAuth tests pass
- [ ] Manual: rotate `ENCRYPTION_KEY`, verify existing tokens decrypt and Telegram alert fires on failure


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

Alert wiring is correct, fallback loop logic is correct. One required fix: the per-key fallback has no deterministic test — deleting it wouldn't break any of the three new tests. One advisory note on Markdown injection risk in the alert message.

---

## Required: test the fallback path specifically

`test_decrypt_with_third_key_in_chain` sets up `[key_a, key_b, key_c]` and decrypts a token encrypted with `key_c`. This test passes whether or not the per-key fallback exists, because `MultiFernet.decrypt()` already tries all keys in sequence internally — it finds `key_c` on its own.

Mutation: delete lines 122–128 (the `for i, fernet in enumerate(self._fernets)` loop). All three new tests still pass. The fallback is shipped as dead code from the test suite's perspective.

To cover the fallback path deterministically, one test needs to make the fast path fail and verify the fallback succeeds:

```python
def test_per_key_fallback_when_multifernet_fails(self, mock_settings):
    """Fallback loop decrypts when MultiFernet fast path raises InvalidToken."""
    from src.utils.encryption import TokenEncryption

    key = Fernet.generate_key().decode()
    _mock_single_key(mock_settings, key=key)
    enc = TokenEncryption()
    ciphertext = enc.encrypt("secret")

    # Force MultiFernet fast path to fail
    with patch.object(enc._cipher, "decrypt", side_effect=InvalidToken):
        result = enc.decrypt(ciphertext)

    assert result == "secret"
```

Without this, a future refactor that removes the fallback loop won't be caught by CI.

---

## Fallback loop logic: correct

The loop itself is sound. `for i, fernet in enumerate(self._fernets)` iterates every key. `except InvalidToken: continue` moves to the next without short-circuiting. `ValueError` is only raised after the entire list is exhausted. ✓

The PR comment explains the rationale — MultiFernet may abort early on version-byte mismatch before checking HMAC. The per-key fallback bypasses that optimization. The logic is correct; it just needs a test that exercises it.

---

## Telegram alert wiring: correct

`get_user_credentials(telegram_chat_id)` → `_notify_decrypt_failure(telegram_chat_id)` → `notify_telegram(chat_id=telegram_chat_id, ..., success=False)`.

The affected user's `chat_id` flows through unchanged. The alert lands in the right chat. ✓

`generate_authorization_url(telegram_chat_id)` embeds the correct chat in the OAuth state, so re-auth will reconnect the right user. ✓

`create_task(self.notify_telegram(...))` correctly bridges the synchronous `_notify_decrypt_failure` caller to the async `notify_telegram`. When called from the scheduler's async context (the common path), there will be a running loop. The `except RuntimeError` gracefully handles synchronous callers (tests, CLI). ✓

---

## Advisory: Markdown injection risk in alert message

`notify_telegram` sends with `parse_mode="Markdown"`. The `auth_url` is interpolated directly into the message body. OAuth authorization URLs contain `?`, `=`, `&` (safe) but may contain `_` in query parameter names, which Telegram's Markdown parser treats as italic delimiters. A mismatched `_` will cause the `send_message` call to raise a `BadRequest` from Telegram's API, which `notify_telegram`'s `except Exception` swallows silently — the user gets no alert.

Low probability (OAuth URLs don't typically have odd-count underscores), but the user is already in a broken state when this fires. Consider either escaping the URL or sending it as a plain-text message (drop `parse_mode` in `notify_telegram`, or add a `parse_mode` parameter defaulting to `None`).

chrisrogers37 (commented):

Both requested fixes are correctly implemented.

**Fallback test:** `test_per_key_fallback_when_multifernet_fails` patches `enc._cipher.decrypt` to raise `InvalidToken`, forcing the fast path to fail. The per-key loop then tries the real Fernet key and returns `"secret"`. Deleting the fallback loop now breaks this test — `decrypt()` would fall through to `raise ValueError` and the assertion never runs. The fallback path is deterministically covered. ✓

**parse_mode:** `notify_telegram` gains a `parse_mode: str | None = "Markdown"` parameter. `_notify_decrypt_failure` passes `parse_mode=None`, which hits the `else` branch (strips `*` bold markers) and passes `parse_mode=None` to `send_message`. Plain text delivery, no Markdown injection risk on the auth URL. All other `notify_telegram` callers get the original Markdown behavior via the default. ✓

PR #374, opened 2026-05-17
fix: implement Google Drive token refresh lifecycle

### LINKED ISSUE #372: Google Drive tokens die after access_token expiry — no refresh lifecycle
## Problem

Google Drive connections break repeatedly. Pattern: token connected, works briefly, then every subsequent sync cycle fails with auth errors. PR #370 (token decrypt fallback) fixed one failure mode, but the underlying refresh lifecycle is broken.

## Root cause analysis

Traced the full token lifecycle across four files. Found three compounding bugs that together mean **once the 1-hour access_token expires, the system has no reliable way to recover.**

### Bug 1: `Credentials` object constructed without `expiry` (primary)

`google_drive_oauth.py:374-381` — `get_user_credentials()` creates the google-auth `Credentials` object:

```python
return Credentials(
    token=access_token,
    refresh_token=refresh_token,
    token_uri=self.GOOGLE_TOKEN_URL,
    client_id=settings.GOOGLE_CLIENT_ID,
    client_secret=settings.GOOGLE_CLIENT_SECRET,
    scopes=self.REQUIRED_SCOPES,
)
```

**`expiry` is never passed**, even though `expires_at` IS stored in the DB (set at `google_drive_oauth.py:192`). Without `expiry`:

- `credentials.expired` returns `False` (the library assumes valid when expiry is unknown)
- `credentials.valid` returns `True` (token is non-None and not "expired")
- The google-auth transport **never proactively refreshes** — it sends the stale access_token directly
- Google returns 401
- `AuthorizedHttp` catches the 401 and attempts a refresh, but this is the only refresh path — and if it fails for any reason, the error is terminal

### Bug 2: Refreshed tokens are never persisted to the database

When `AuthorizedHttp` successfully refreshes the token on a 401, the new `access_token` lives only in the in-memory `Credentials` object. **Nothing writes it back to the DB.** The lifecycle:

1. Sync loop creates provider → `get_user_credentials()` reads OLD access_token from DB
2. API call fails (token expired) → `AuthorizedHttp` refreshes → works for this request
3. Provider is garbage collected at end of sync cycle
4. Next sync cycle: `get_user_credentials()` reads the SAME old expired token from DB
5. Repeat from step 2 — every single API session starts with a guaranteed 401 + refresh

This works... until the refresh itself fails (network blip, rate limit, token revocation). Then there's no recovery.

### Bug 3: No proactive Google Drive token refresh (unlike Instagram)

`token_refresh.py` has `refresh_instagram_token()` and `refresh_all_instagram_tokens()` with a 7-day buffer window. **There is no equivalent for Google Drive.** The health check at `token_refresh.py:396-453` (`check_token_health_for_chat`) even has a comment acknowledging this gap:

```python
# For services with refresh tokens (e.g., Google Drive), an expired
# access token is normal — the OAuth library auto-refreshes it.
```

This assumption doesn't hold because of Bug 1 (no `expiry` → no proactive refresh).

### Contributing factor: 401 is not retried at the application level

`google_drive_provider.py:39-43` — `_is_retryable` only retries 429, 500, 502, 503, 504. A 401 that escapes `AuthorizedHttp` (refresh failure) immediately becomes a `GoogleDriveAuthError` with no application-level recovery.

## Why "~minutes, not hours"

The sync loop runs every `MEDIA_SYNC_INTERVAL_SECONDS` (default 300s / 5 min). After the 1-hour access_token expires:

1. Next sync cycle hits the expired token
2. If `AuthorizedHttp` refresh works — sync succeeds but token isn't persisted
3. If `AuthorizedHttp` refresh fails — `GoogleDriveAuthError` → sync fails → user gets disconnect notification
4. Every subsequent cycle repeats this fragile dance

## Recommended fix

**Phase 1 (critical — fixes the core issue):**

1. **Pass `expiry` to `Credentials` constructor** in `get_user_credentials()`:
   ```python
   return Credentials(
       token=access_token,
       refresh_token=refresh_token,
       token_uri=self.GOOGLE_TOKEN_URL,
       client_id=settings.GOOGLE_CLIENT_ID,
       client_secret=settings.GOOGLE_CLIENT_SECRET,
       scopes=self.REQUIRED_SCOPES,
       expiry=access_row.expires_at,  # enables proactive refresh
   )
   ```

2. **Persist refreshed tokens**: After creating the provider and making API calls, check if `credentials.token` differs from what was read from DB. If so, encrypt and write the new access_token + expiry back. Could be done via:
   - A wrapper around `get_user_credentials()` that returns a context manager
   - A post-sync hook in `media_sync_loop.py`
   - Subclassing `Credentials` to override `refresh()` with a DB-persist step

**Phase 2 (hardening):**

3. **Add `refresh_google_drive_token()`** to `TokenRefreshService` — use the stored refresh_token to proactively get a new access_token before expiry, similar to the Instagram refresh flow.

4. **Add retry-with-refresh for 401** in the provider or sync layer — when `GoogleDriveAuthError` is caught, attempt one explicit refresh-and-retry before giving up.

## Files involved

| File | Role in the bug |
|------|----------------|
| `src/services/integrations/google_drive_oauth.py:374` | Missing `expiry` param |
| `src/services/integrations/google_drive_oauth.py:192` | `expires_at` stored but never read back |
| `src/services/media_sources/google_drive_provider.py:141` | Service built with stale credentials |
| `src/services/core/loops/media_sync_loop.py:47` | Creates new provider each cycle (loses refreshed token) |
| `src/services/integrations/token_refresh.py` | No Google Drive refresh method |
| `src/services/media_sources/google_drive_provider.py:39` | 401 not retryable |

Follows #370.

chrisrogers37:
## Summary

- Passes `expiry` to the `google.oauth2.credentials.Credentials` constructor so the library knows when to proactively refresh tokens
- Persists refreshed access tokens back to the DB after each sync cycle, so the next cycle starts with a valid token instead of re-refreshing every time
- Adds 401 to retryable status codes, giving a second chance when AuthorizedHttp's refresh hits a transient failure
- Exposes a typed `credentials` property on `GoogleDriveProvider` for the persistence hook

## Root cause

See #372 for the full analysis. Three compounding bugs meant once the 1-hour access_token expired, the system had no reliable recovery path — no proactive refresh, no persistence of refreshed tokens, and no retry on auth failure.

## Test plan

- [x] `test_credentials_include_expiry` — verifies `expiry` is passed from DB to Credentials
- [x] `test_credentials_expiry_none_when_not_set` — handles None gracefully
- [x] `test_persists_when_token_changed` — writes new token when refresh detected
- [x] `test_skips_when_token_unchanged` — no DB write on no-op path
- [x] `test_persists_when_stored_token_decrypt_fails` — handles corrupted stored token
- [x] `test_retry_on_401` / `test_401_exhausts_retries` — 401 retry behavior
- [x] `test_sync_persists_refreshed_gdrive_credentials` — end-to-end integration
- [x] `test_persist_failure_does_not_break_sync` — error swallowed, sync unaffected
- [x] All 105 tests pass

Closes #372

### DIFF
--- a/src/services/core/media_sync.py
+++ b/src/services/core/media_sync.py
@@ -231,6 +231,10 @@ class MediaSyncService(BaseService):
 
             self._deactivate_missing_items(ctx)
 
+            # Persist any token the google-auth library refreshed in-memory
+            if resolved_type == "google_drive" and telegram_chat_id:
+                self._persist_refreshed_gdrive_credentials(provider, telegram_chat_id)
+
             logger.info(
                 f"[MediaSyncService] Sync complete: "
                 f"{ctx.result.new} new, {ctx.result.updated} updated, "
@@ -397,6 +401,33 @@ class MediaSyncService(BaseService):
             )
         else:
             return MediaSourceFactory.create(source_type)
+
+    def _persist_refreshed_gdrive_credentials(
+        self, provider, telegram_chat_id: int
+    ) -> None:
+        """Persist refreshed Google Drive OAuth credentials back to the DB.
+
+        During API calls, the google-auth library may refresh the access_token
+        in-memory. Without this step, the next sync cycle would start with
+        the old expired token and need another refresh round-trip.
+        """
+        from google.oauth2.credentials import Credentials as UserCredentials
+
+        creds = provider.credentials
+        if not isinstance(creds, UserCredentials) or not creds.refresh_token:
+            return
+
+        from src.services.integrations.google_drive_oauth import (
+            GoogleDriveOAuthService,
+        )
+
+        try:
+            with GoogleDriveOAuthService() as oauth_service:
+                oauth_service.persist_refreshed_credentials(telegram_chat_id, creds)
+        except Exception:  # noqa: BLE001 — best-effort, must not break sync
+            logger.warning(
+                "[MediaSyncService] Failed to persist refreshed GDrive credentials"
+            )
 
     def _get_file_hash(self, file_info: MediaFileInfo, provider) -> str:
         """Get content hash for a file. Uses provider-side hash if available."""
--- a/src/services/integrations/google_drive_oauth.py
+++ b/src/services/integrations/google_drive_oauth.py
@@ -378,6 +378,7 @@ class GoogleDriveOAuthService(BaseService):
                 client_id=settings.GOOGLE_CLIENT_ID,
                 client_secret=settings.GOOGLE_CLIENT_SECRET,
                 scopes=self.REQUIRED_SCOPES,
+                expiry=access_row.expires_at,
             )
         except (ValueError, InvalidToken) as e:
             logger.critical(
@@ -387,6 +388,56 @@ class GoogleDriveOAuthService(BaseService):
             )
             self._notify_decrypt_failure(telegram_chat_id)
             return None
+
+    def persist_refreshed_credentials(self, telegram_chat_id: int, credentials) -> bool:
+        """Persist refreshed OAuth credentials back to the database.
+
+        After API operations, the google-auth library may have refreshed the
+        access_token in-memory. This method writes the new token back to the
+        DB so the next provider creation cycle starts with a valid token.
+
+        Args:
+            telegram_chat_id: The Telegram chat that owns the credentials.
+            credentials: google.oauth2.credentials.Credentials instance.
+
+        Returns:
+            True if the token was updated in the DB, False otherwise.
+        """
+        if not credentials or not credentials.token:
+            return False
+
+        chat_settings = self.settings_repo.get_by_chat_id(telegram_chat_id)
+        if not chat_settings:
+            return False
+
+        chat_settings_id = str(chat_settings.id)
+
+        access_row = self.token_repo.get_token_for_chat(
+            self.SERVICE_NAME, self.TOKEN_TYPE_ACCESS, chat_settings_id
+        )
+        if not access_row:
+            return False
+
+        try:
+            stored_token = self.encryption.decrypt(access_row.token_value)
+        except (ValueError, InvalidToken):
+            stored_token = None
+
+        if stored_token == credentials.token:
+            return False
+
+        self.token_repo.create_or_update_for_chat(
+            service_name=self.SERVICE_NAME,
+            token_type=self.TOKEN_TYPE_ACCESS,
+            token_value=self.encryption.encrypt(credentials.token),
+            chat_settings_id=chat_settings_id,
+            expires_at=credentials.expiry,
+        )
+
+        logger.info(
+            f"Google Drive OAuth: persisted refreshed token for chat {telegram_chat_id}"
+        )
+        return True
 
     def _notify_decrypt_failure(self, telegram_chat_id: int) -> None:
         """Fire-and-forget Telegram alert when token decryption fails."""
--- a/src/services/media_sources/google_drive_provider.py
+++ b/src/services/media_sources/google_drive_provider.py
@@ -37,9 +37,15 @@ _DOWNLOAD_TIMEOUT_SECONDS = 300
 
 
 def _is_retryable(exc: BaseException) -> bool:
-    """Return True for transient errors that should trigger a retry."""
+    """Return True for transient errors that should trigger a retry.
+
+    Includes 401: after the initial access_token expires, AuthorizedHttp
+    refreshes it in-memory.  If that refresh itself hits a transient failure
+    (network blip, Google-side propagation delay), a single retry gives the
+    refreshed token a second chance before we surface a terminal auth error.
+    """
     if isinstance(exc, HttpError):
-        return exc.resp.status in (429, 500, 502, 503, 504)
+        return exc.resp.status in (401, 429, 500, 502, 503, 504)
     return isinstance(exc, (ConnectionError, TimeoutError, OSError))
 
 
@@ -136,6 +142,13 @@ class GoogleDriveProvider(MediaSourceProvider):
 
         self._credentials = credentials
         self._folder_cache: dict[str, str] = {}
+
+    @property
+    def credentials(
+        self,
+    ) -> UserCredentials | ServiceAccountCredentials:
+        """Expose credentials so callers can persist refreshed tokens."""
+        return self._credentials
 
     @property
     def service(self):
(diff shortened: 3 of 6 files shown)


chrisrogers37 (commented):

Three compounding bugs, three clean fixes. Reviewed against the mandate.

---

**1. Expiry passed to Credentials — correct**

`expiry=access_row.expires_at` lands in the `Credentials` constructor (google_drive_oauth.py:381). The google-auth library uses this datetime to decide whether proactive refresh is needed before the next API call. Without it, every call started cold and triggered a reactive refresh. Fixed correctly.

**2. Token persistence is encrypted — correct**

`persist_refreshed_credentials` writes `self.encryption.encrypt(credentials.token)` (google_drive_oauth.py:432). The decrypt path is symmetric to the existing `get_user_credentials` read. Never plaintext in DB.

**3. 401 retry is bounded — correct, with one note**

`stop_after_attempt(3)` is a hard cap; no infinite loop possible. Worth knowing: `googleapiclient`'s `AuthorizedHttp` has its own internal 401 → refresh → retry cycle before surfacing the error to our tenacity layer. In practice that means up to 6 actual HTTP calls on a persistent 401 (2 per tenacity attempt × 3 attempts). Still bounded, not a correctness problem. The `test_401_exhausts_retries` assertion (`call_count == 3`) tests `_execute_with_retry` directly and correctly verifies the tenacity cap in isolation.

**4. Race condition — benign, not a blocker**

`persist_refreshed_credentials` is a read-compare-write without a DB lock or SELECT FOR UPDATE. Two concurrent syncs for the same user could both see `stored_token != credentials.token` and both write. Consequence: last writer wins with a valid token; both in-memory instances remain valid for their current sync cycle. No data corruption, no security issue. Acceptable; a follow-up upsert-with-lock would be the clean fix if concurrent syncs become common.

**5. Mutation test — tests are not decoys**

- Remove `expiry=access_row.expires_at` → `test_credentials_include_expiry` fails. ✓
- Remove `_persist_refreshed_gdrive_credentials` call in `sync()` → `test_sync_persists_refreshed_gdrive_credentials` fails. ✓
- Remove 401 from `_is_retryable` → `test_retry_on_401` raises `HttpError` instead of returning the success dict. ✓

All three fixes have genuine test coverage.

**Minor nit:** `if resolved_type == "google_drive" and telegram_chat_id:` — falsy int check silently skips persistence if `chat_id` is ever 0 (impossible for Telegram IDs, but `is not None` is the correct idiom). Not a blocker.

chrisrogers37 (commented):

Three compounding bugs, three clean fixes. Reviewed against the mandate.

---

**1. Expiry passed to Credentials — correct**

expiry=access_row.expires_at lands in the Credentials constructor (google_drive_oauth.py:381). The google-auth library uses this datetime to decide whether proactive refresh is needed before the next API call. Without it, every call started cold and triggered a reactive refresh. Fixed correctly.

**2. Token persistence is encrypted — correct**

persist_refreshed_credentials writes self.encryption.encrypt(credentials.token) (google_drive_oauth.py:432). Decrypt path is symmetric to the existing get_user_credentials read. Never plaintext in DB.

**3. 401 retry is bounded — correct, with one note**

stop_after_attempt(3) is a hard cap; no infinite loop possible. Worth knowing: googleapiclient's AuthorizedHttp has its own internal 401 -> refresh -> retry cycle before surfacing the error to our tenacity layer. In practice that means up to 6 actual HTTP calls on a persistent 401 (2 per tenacity attempt x 3 attempts). Still bounded, not a correctness problem. test_401_exhausts_retries (call_count == 3) tests _execute_with_retry directly and correctly verifies the tenacity cap in isolation.

**4. Race condition — benign, not a blocker**

persist_refreshed_credentials is a read-compare-write without a DB lock. Two concurrent syncs for the same user could both see stored_token != credentials.token and both write. Consequence: last writer wins with a valid token; both in-memory instances remain valid for their current sync cycle. No data corruption, no security issue. A follow-up upsert-with-lock would be the clean fix if concurrent syncs become common.

**5. Mutation test — tests are not decoys**

- Remove expiry=access_row.expires_at: test_credentials_include_expiry fails. v
- Remove _persist_refreshed_gdrive_credentials call in sync(): test_sync_persists_refreshed_gdrive_credentials fails. v
- Remove 401 from _is_retryable: test_retry_on_401 raises HttpError instead of returning the success dict. v

All three fixes have genuine test coverage.

**Minor nit:** if resolved_type == 'google_drive' and telegram_chat_id: — falsy int check silently skips persistence if chat_id is ever 0 (impossible for Telegram IDs, but is not None is the correct idiom). Not a blocker.

chrisrogers37 (commented):

Three compounding bugs, three clean fixes. Reviewed against the mandate.

---

**1. Expiry passed to Credentials — correct**

expiry=access_row.expires_at lands in the Credentials constructor (google_drive_oauth.py:381). The google-auth library uses this datetime to decide whether proactive refresh is needed before the next API call. Without it, every call started cold and triggered a reactive refresh. Fixed correctly.

**2. Token persistence is encrypted — correct**

persist_refreshed_credentials writes self.encryption.encrypt(credentials.token) (google_drive_oauth.py:432). Decrypt path is symmetric to the existing get_user_credentials read. Never plaintext in DB.

**3. 401 retry is bounded — correct, with one note**

stop_after_attempt(3) is a hard cap; no infinite loop possible. Worth knowing: googleapiclient's AuthorizedHttp has its own internal 401 -> refresh -> retry cycle before surfacing the error to our tenacity layer. In practice that means up to 6 actual HTTP calls on a persistent 401 (2 per tenacity attempt x 3 attempts). Still bounded, not a correctness problem. test_401_exhausts_retries (call_count == 3) tests _execute_with_retry directly and correctly verifies the tenacity cap in isolation.

**4. Race condition — benign, not a blocker**

persist_refreshed_credentials is a read-compare-write without a DB lock. Two concurrent syncs for the same user could both see stored_token != credentials.token and both write. Consequence: last writer wins with a valid token; both in-memory instances remain valid for their current sync cycle. No data corruption, no security issue. A follow-up upsert-with-lock would be the clean fix if concurrent syncs become common.

**5. Mutation test — tests are not decoys**

- Remove expiry=access_row.expires_at: test_credentials_include_expiry fails.
- Remove _persist_refreshed_gdrive_credentials call in sync(): test_sync_persists_refreshed_gdrive_credentials fails.
- Remove 401 from _is_retryable: test_retry_on_401 raises HttpError instead of returning the success dict.

All three fixes have genuine test coverage.

**Minor nit:** `if resolved_type == "google_drive" and telegram_chat_id:` — falsy int check silently skips persistence if chat_id is ever 0 (impossible for Telegram IDs, but `is not None` is the correct idiom). Not a blocker.

PR #398, opened 2026-05-19
fix: Instagram token refresh subsystem

chrisrogers37:
## Summary

- **Wire TokenRefreshService into scheduler loop** — `refresh_all_instagram_tokens()` now runs daily (every 1440 ticks) in `scheduler_loop.py`, proactively refreshing tokens within the 7-day expiry buffer. Previously the service existed but was never invoked, so tokens expired silently after 60 days. Closes #397.

- **Add row-level locking for concurrent refresh** — `TokenRepository.get_token_for_update()` uses `SELECT ... FOR UPDATE SKIP LOCKED` so concurrent web + worker processes can't clobber each other's token during refresh. If another process holds the lock, the refresh is skipped rather than racing. Closes #375.

- **Distinguish refresh token revocation from access token expiry** — New `TokenRevokedError` exception detects Meta error subcodes 458 (app deauthorized), 460 (password changed), 467 (token invalidated). `_check_response_errors` raises `TokenRevokedError` for these instead of `TokenExpiredError`, and the autopost error handler shows "account disconnected — please reconnect" rather than silently retrying. Closes #377.

## Files changed

| File | Change |
|------|--------|
| `src/exceptions/instagram.py` | Add `TokenRevokedError` with `REVOCATION_SUBCODES` |
| `src/exceptions/__init__.py` | Export `TokenRevokedError` |
| `src/repositories/token_repository.py` | Add `get_token_for_update()` with `FOR UPDATE SKIP LOCKED` |
| `src/services/integrations/token_refresh.py` | Use locked read, detect revocation in refresh response |
| `src/services/core/loops/scheduler_loop.py` | Add `_token_refresh_tick`, daily invocation |
| `src/services/integrations/instagram_api.py` | Raise `TokenRevokedError` for revocation subcodes |
| `src/services/core/telegram_autopost.py` | Distinct user message for revocation |
| `tests/` | 17 new tests across 4 files, all passing |

## Test plan

- [x] 79 unit tests pass (exception, service, repository, scheduler loop)
- [x] `ruff check` and `ruff format` clean
- [ ] Verify token refresh fires on Railway worker after 1440 ticks (~24h)
- [ ] Verify concurrent refresh doesn't clobber (manually trigger from web + worker)
- [ ] Verify revocation subcode 458/460 shows "disconnected" message in Telegram


chrisrogers37 (commented):
**Verdict: Request changes**

79/79 tests pass. Three of the four checklist items are correct. One blocker: a revoked account aborts the daily refresh of all other accounts.

---

## Verification trace

**1. Scheduler wiring** ✅

`TOKEN_REFRESH_INTERVAL_TICKS = 1440`, counter increments each tick, fires and resets at `>= 1440`. `_token_refresh_tick` is async, calls `await token_refresh_service.refresh_all_instagram_tokens()`, wraps with `try/except TokenRevokedError` and `except Exception`, always runs `cleanup_transactions()` in `finally`. Service instantiated once at `run_scheduler_loop` startup, reused per tick.

**2. Race condition locking** ✅

`get_token_for_update()` → `.with_for_update(skip_locked=True).first()`. Concurrent process holding the lock gets `None` back; the refresh path returns `False` and logs "skipped." Lock release path: `cleanup_transactions()` → `end_read_transaction()` → `self._db.commit()` (confirmed in `base_repository.py:112`). The commit releases the `FOR UPDATE` lock. Called in `finally` on every code path through `_token_refresh_tick`. Lock is held across the Meta HTTP call (the external request is made while the row is locked), but this is correct and necessary to prevent a second process from reading stale token mid-refresh. Acceptable for a background job.

**3. Revocation vs expiry** ✅

Exception hierarchy: `TokenRevokedError(InstagramAPIError)`, NOT a subclass of `TokenExpiredError` — confirmed by `test_not_caught_by_token_expired`. Detection in `_check_response_errors` (instagram_api.py:349): subcodes 458/460/467 → `TokenRevokedError`, else → `TokenExpiredError`. Detection in `refresh_instagram_token` (token_refresh.py:235): independent check on the raw response before the `_check_response_errors` call. User message in `_get_user_friendly_error`: `isinstance(e, TokenRevokedError)` checked at line 608, `isinstance(e, TokenExpiredError)` at line 610 — correct order.

**4. Tests: missing coverage for the blocker path** ⚠️

17 new tests pass. Gaps:
- `test_refresh_instagram_token_revoked` (subcode 458) ✅ and `test_refresh_instagram_token_password_changed` (subcode 460) ✅ — but no test for subcode 467.
- `_check_response_errors` raising `TokenRevokedError` for subcodes is untested (the refresh tests patch `httpx.AsyncClient` directly, bypassing `instagram_api.py`).
- `_get_user_friendly_error` returning the distinct message for `TokenRevokedError` is untested.

---

## Blocker

**`refresh_all_instagram_tokens` does not catch `TokenRevokedError` per account. A single revoked account aborts the loop.**

```python
# token_refresh.py:360 — no except around this
success = await self.refresh_instagram_token(instagram_account_id=account_id)
```

`refresh_instagram_token` raises `TokenRevokedError` for subcodes 458/460/467. `refresh_all_instagram_tokens` has no `except TokenRevokedError` around that call. The exception propagates out of `refresh_all` → out of the `_token_refresh_tick` try block → caught by `except TokenRevokedError` in `_token_refresh_tick` → logged → tick exits.

For a multi-account fleet where account A's token is revoked: accounts B, C, D — which may have expiring tokens — are never refreshed that day. Their tokens may expire before the next tick (24 hours later).

Fix: wrap the `refresh_instagram_token` call in `refresh_all` with per-account error isolation:

```python
try:
    success = await self.refresh_instagram_token(instagram_account_id=account_id)
except TokenRevokedError as e:
    logger.error(f"Token revoked for {account_label}: {e}")
    results["failed"] += 1
    results["details"].append({"account": account_label, "status": "revoked"})
    continue
```

This lets `refresh_all` complete for all accounts and return a full `results` dict. `_token_refresh_tick` can then check `results["details"]` for revoked accounts if notification logic is added later.

---

## Suggestions

- **Subcode 467 test**: `test_refresh_instagram_token_revoked` covers 458, `_password_changed` covers 460. Add a test for 467 ("Token invalidated on server side") to confirm the set is complete.
- **`_check_response_errors` test**: The posting flow uses `instagram_api.py._check_response_errors` independently from the refresh flow. Add a test that passes a `200-equivalent` response with error code 190 and subcode 458 and asserts `TokenRevokedError` is raised.
- **`_get_user_friendly_error` test**: Add an assertion that `isinstance(e, TokenRevokedError)` returns the "disconnected" message, not the generic expired message.
- **Scheduled revocation notification (follow-up)**: When the scheduler detects revocation, it logs only. The user won't know to reconnect until they try to post. Consider sending a Telegram message on revocation detection in `_token_refresh_tick`. Not a blocker for this PR — just worth tracking.

chrisrogers37 (commented):

Blocker resolved. Commit fcf508b5 wraps the per-account `refresh_instagram_token` call inside `refresh_all_instagram_tokens` with a `try/except TokenRevokedError` — logs the revocation, increments `failed`, and `continue`s to the next account. One revoked token no longer aborts the loop for the rest.

Verified in `src/services/integrations/token_refresh.py`:

```python
try:
    success = await self.refresh_instagram_token(instagram_account_id=account_id)
except TokenRevokedError as e:
    logger.error(f"Token revoked for {account_label}: {e}")
    results["failed"] += 1
    results["details"].append({"account": account_label, "status": "revoked"})
    continue
```

`TokenRevokedError` is also correctly NOT a subclass of `TokenExpiredError` (explicitly tested in `TestTokenRevokedError.test_not_caught_by_token_expired`), so existing `except TokenExpiredError` handlers won't silently swallow revocations.

One minor observation, not a blocker: the `except TokenRevokedError` guard in `_token_refresh_tick` is dead code — `refresh_all_instagram_tokens` absorbs all per-account `TokenRevokedError`s without re-raising, so the tick-level catch never fires. Harmless, but could be removed for clarity.

79 tests pass, ruff clean. The original blocker is addressed.

PR #399, opened 2026-05-19
fix: catch RefreshError + track refresh token age for Testing mode

### LINKED ISSUE #376: Google OAuth app likely in Testing mode — 7-day token expiry and 100-user cap
## Problem

Multiple signals indicate the Google Cloud OAuth consent screen is still in "Testing" publishing status rather than "Production" (verified). This has direct consequences for token reliability:

1. **Refresh tokens expire after 7 days** in Testing mode. Google's documentation states: "A Google Cloud Platform project with an OAuth consent screen configured for an external user type and a publishing status of 'Testing' is issued a refresh token expiring in 7 days." After 7 days of inactivity, the refresh token dies silently — the next refresh attempt gets `invalid_grant`.

2. **Maximum 100 test users.** Only Google accounts explicitly added to the test user allowlist can complete OAuth. All others see an impassable red "Google hasn't verified this app" screen (not just a warning — they cannot proceed without clicking through Advanced).

3. **More aggressive token revocation.** Google applies stricter revocation policies to unverified apps, including revoking tokens when the app is inactive for extended periods.

## Evidence

- `documentation/operations/google-oauth-verification.md:43-46` — Step 2 says "Confirm Publishing status: In production. If it says Testing, click Publish App." This implies the app is currently in Testing or at best published-but-unverified.
- `documentation/operations/google-oauth-verification.md:99-100` — "Google accounts added under OAuth consent > Test users (up to 100) bypass the warning entirely — they see a clean consent screen. Every other user sees the red 'Google hasn't verified this app' page." This is the current state.
- PR #371 (merged) created the verification runbook but explicitly states: "The submission itself stays a manual operations task tracked on the issue." Meaning verification has NOT been submitted yet.
- Issue #333 (open) tracks verification submission — still pending.
- `drive.readonly` is classified as a **sensitive scope** (line 10 of the runbook). Sensitive scopes require verification before the app can leave the unverified state for external users.

## Implications for token reliability

The 7-day refresh token expiry in Testing mode explains a class of failures that PR #374 cannot fix:

1. User connects Google Drive. Refresh token issued with 7-day TTL (invisible to the app — Google doesn't expose this in the token response).
2. User doesn't trigger any Google Drive activity for 7 days.
3. On day 8, the sync loop tries to refresh the expired access token using the refresh token.
4. Google returns `invalid_grant` (refresh token expired).
5. The code treats this as a generic refresh failure — no distinct code path to notify the user that re-authentication is required.

Even active users are affected: the 7-day clock resets on each token USE, not each token REFRESH. If the app only refreshes the access token without actually calling a Google API with it, the inactivity clock keeps ticking.

## Recommended actions

1. **Immediate: Publish the app to Production mode** in Google Cloud Console (APIs & Services > OAuth consent screen > Publish App). This removes the 7-day refresh token TTL. The "unverified app" warning still shows, but tokens become long-lived.

2. **Short-term: Submit for verification** per the runbook in `documentation/operations/google-oauth-verification.md`. Removes the scary warning screen.

3. **Code change: Track refresh token age** — store `refresh_token_issued_at` in the token metadata. When the refresh token is >5 days old (in Testing mode) or >5 months old (in Production mode), proactively alert the user to reconnect before it expires silently.

## Files involved

| File | Line | Relevance |
|------|------|-----------|
| `documentation/operations/google-oauth-verification.md` | 43-46 | Implies Testing status |
| `src/services/integrations/google_drive_oauth.py` | 104 | `access_type: "offline"` — refresh token requested |
| `src/services/integrations/google_drive_oauth.py` | 204-210 | Refresh token stored without TTL tracking |

Related: #333, #372, #374.

chrisrogers37:
## Summary

- **Catch `google.auth.exceptions.RefreshError`** in `_execute_with_retry` and `_download_chunk_with_retry` — converts to `GoogleDriveAuthError` so the existing reconnect-alert flow handles expired refresh tokens instead of surfacing unhandled exceptions
- **Store `issued_at` on refresh tokens** during OAuth exchange so token age can be computed
- **Add `GOOGLE_REFRESH_TOKEN_TTL_DAYS` setting** (default 7 for Testing mode, set to 0 after moving to Production) — `TokenRefreshService` and `HealthCheckService` use this to detect and warn about approaching/expired refresh tokens with "Testing mode" messaging

Closes #373, Closes #376

## Context

Google OAuth apps in Testing mode silently expire refresh tokens after 7 days. Before this fix, an expired refresh token would raise an unhandled `RefreshError` from the google-auth library, bypassing the reconnect-alert path. Health checks also couldn't detect approaching expiry because `issued_at` wasn't tracked.

## Files changed

| File | Change |
|------|--------|
| `src/services/media_sources/google_drive_provider.py` | Catch `RefreshError` → `GoogleDriveAuthError` in both API call choke points |
| `src/services/integrations/google_drive_oauth.py` | Pass `issued_at` when storing refresh token |
| `src/config/settings.py` | Add `GOOGLE_REFRESH_TOKEN_TTL_DAYS` setting |
| `src/services/integrations/token_refresh.py` | Compute refresh token age against TTL |
| `src/services/core/health_check.py` | Surface Testing mode warnings at critical/warning thresholds |
| `tests/` (4 files) | 11 new tests covering RefreshError handling, token age monitoring, health check warnings |

## Test plan

- [x] All 142 tests pass (`pytest` — 25.97s)
- [x] `ruff check` and `ruff format` clean on all modified files
- [ ] Verify on Railway that health check surfaces Testing mode warning for connected accounts
- [ ] After moving OAuth app to Production mode, set `GOOGLE_REFRESH_TOKEN_TTL_DAYS=0` to disable age checks


chrisrogers37 (commented):

Both blockers from the previous review are resolved.

**Blocker 1 — TTL check unreachable when valid=False: fixed**

The reordering is correct. `refresh_token_expires_in_days` is now checked before `if not token_health["valid"]`, so a token that is simultaneously access-invalid and TTL-expired surfaces the more specific "Testing mode 7-day limit" message rather than the generic validity failure.

**Blocker 2 — test exercising wrong path: fixed**

The new test uses `valid=True, refresh_token_expires_in_days=-1.0`. With `valid=True` the generic guard is inert, so the only way the test gets `healthy=False` with "testing mode" in the message is if the TTL branch fires. Mutation check: delete the TTL block → falls through to the `valid=True` path → `healthy=True` → test fails. The assertion is real, not a decoy.

**One nit (non-blocking)**

No test pins the specific ordering. `test_gdrive_refresh_token_expired_testing_mode` uses `valid=True`, so it wouldn't catch a future revert of the reorder. A companion case — `valid=False, refresh_token_expires_in_days=-1.0` asserting the TTL message rather than a generic invalid-token message — would lock that in. The code logic is sound as written; this is just a gap in regression coverage.

PR #400, opened 2026-05-19
fix: scheduler/queue reliability (#360 #362 #363 #365 #366)

chrisrogers37:
## Summary

Five scheduler/queue reliability fixes bundled into one PR:

- **#360 — posting_hours_start == end causes ZeroDivisionError**: `_in_posting_window()` and `_posting_window_hours()` now treat `start == end` as "always on" (24h window) instead of returning False/0.
- **#365 — seen_identifiers added before processing in media sync**: `_process_provider_file()` now adds to `seen_identifiers` only after successful processing, so a per-file exception doesn't mask the file from deactivation detection on the next sync.
- **#366 — stale processing items not recovered**: New `requeue_stale_processing()` resets items stuck in "processing" without a `telegram_message_id` (crashed mid-send) back to "pending" after 10 minutes. Called alongside `delete_stale_pending()` on each tick.
- **#363 — SIGTERM during posting leaves items stuck**: `_select_and_send()` checks `session_state.shutdown_in_progress` before starting work, preventing new sends from beginning during shutdown.
- **#362 — scheduler posts before first sync completes**: Added `session_state.initial_sync_complete` gate — scheduler skips posting ticks until the media sync loop signals at least one successful sync.

Closes #360, #362, #363, #365, #366.

## Test plan

- [ ] 15 new unit tests in `test_scheduler_queue_reliability.py` cover all 5 fixes
- [ ] Existing 168 tests in affected files still pass (183 total with new tests)
- [ ] Linter and formatter pass on all changed files
- [ ] Verify on Railway staging that scheduler waits for sync before first post
- [ ] Verify SIGTERM during posting doesn't leave orphaned processing items


chrisrogers37 (commented):
**Verdict: Ship it**

1852 passed, 16 skipped — all pre-existing test counts. 46 new tests in the PR files pass. Five fixes verified with evidence below.

---

## Verification trace

**#360 — ZeroDivisionError on posting_hours_start == end** ✅

`_in_posting_window`: early-returns `True` when `start == end` before the window comparison. `_posting_window_hours`: early-returns `24.0` when `start == end` before the arithmetic. Test `test_is_slot_due_start_equals_end_no_division_by_zero` runs `is_slot_due` with start=0, end=0 and asserts no ZeroDivisionError. Normal range and midnight-crossing cases confirmed unaffected by regression tests.

**#365 — seen_identifiers added before processing** ✅

All four branches in `_process_provider_file` (identifier match, hash match, reactivation, new file) now add to `seen_identifiers` only on the success return path. `_index_new_file` can raise (e.g., `media_repo.create` throws) — if it does, the `ctx.result.new += 1` line is also never reached and the identifier stays out of `seen_identifiers`. Correct: a file that failed to index should not be considered seen. Test `test_failed_file_not_in_seen_identifiers` confirms via `RuntimeError` in `_handle_identifier_match`.

**#366 — stale processing items not recovered** ✅

`requeue_stale_processing(max_age_minutes=10)` filters on `status == "processing"`, `telegram_message_id IS NULL`, and `created_at <= cutoff`. Items with a `telegram_message_id` (sent but not acted on) are left for `discard_abandoned_processing` (24h). Items without one (crashed pre-send) are reset to "pending". `PostingQueue` has only `created_at`, no `updated_at` — `created_at` is the right column. Called in `process_slot` alongside `delete_stale_pending`. Both call confirmed by `test_process_slot_calls_both_cleanup_methods`.

**#363 — SIGTERM during posting leaves items stuck** ✅

`_select_and_send` starts with a deferred `from src.services.core.loops.lifecycle import session_state` and checks `session_state.shutdown_in_progress` before entering `track_execution`. Returns `{"posted": False, "reason": "shutdown_in_progress"}` immediately. Deferred-import + `patch("src.services.core.loops.lifecycle.session_state")` interaction verified: patching the module attribute works correctly for deferred local imports (Python looks up `session_state` on the module object at import time, and `patch` replaces that attribute). Test confirms `media_repo.select_eligible` is not called.

**#362 — scheduler posts before first sync completes** ✅

`lifecycle.SessionState` gains `initial_sync_complete: bool = False`. `media_sync_loop` sets it to `True` after the first successful sync. `_scheduler_tick` returns `[]` and skips `get_all_active_chats` while `initial_sync_complete` is `False`. `main.py` opens the gate immediately (`session_state.initial_sync_complete = True`) in the `else` branch when no sync loop is registered (e.g., test/minimal deployments). `test_main_scheduler_loop.py` fixtures gate the singleton to `True` via autouse fixture to prevent cross-test pollution — correct.

---

## Nits

- `requeue_stale_processing` uses `datetime.utcnow()` twice: once for the cutoff query and once for logging `age=`. If the method runs slowly, the logged age is slightly stale. Minor — no correctness impact.
- `test_is_slot_due_start_equals_end_no_division_by_zero` asserts `result is not False` — that includes `None`. If `is_slot_due` ever returns `None` on an error path this passes silently. Could tighten to `isinstance(result, bool)` or `result is True`.
- No test covers the `main.py` gate-open-immediately path (the `else` branch). Not a blocker — it's a two-liner.

Issue #386, opened 2026-05-19
Auth endpoints need stricter rate limits than global default


chrisrogers37:
## Summary

OAuth start/callback endpoints inherit the global 30 req/min rate limit. Auth endpoints are high-value targets for brute force and should have tighter limits. Some settings endpoints already have per-endpoint limits (10/min for toggle, 5/min for sync), but the OAuth flow does not.

## Affected endpoints

| Endpoint | Current limit | Suggested |
|----------|--------------|-----------|
| \`/auth/instagram/start\` | 30/min (global) | 5/min |
| \`/auth/instagram/callback\` | 30/min (global) | 10/min |
| \`/auth/instagram-login/callback\` | 30/min (global) | 10/min |
| \`/auth/google-drive/start\` | 30/min (global) | 5/min |
| \`/auth/google-drive/callback\` | 30/min (global) | 10/min |

## Location

`src/api/routes/oauth.py` — none of the route handlers have `@limiter.limit()` decorators.

## Risk

- State token enumeration (mitigated by Fernet encryption, but defense-in-depth)
- OAuth code exchange abuse
- Amplification of failed auth attempts flooding auth_monitor

## Severity: Low-Medium

State tokens are cryptographically strong so brute force is impractical, but rate limiting auth endpoints is standard hardening.

## Suggested fix

```python
@router.get("/auth/instagram/start")
@limiter.limit("5/minute")
async def instagram_auth_start(request: Request, ...):
    ...

@router.get("/auth/instagram/callback")
@limiter.limit("10/minute")
async def instagram_callback(request: Request, ...):
    ...
```

Filed by security audit.

PR #401, opened 2026-05-19
security: harden API — 6 audit findings bundled

chrisrogers37:
## Summary

Bundles 6 security audit findings into one PR:

- **#382 HTTP security headers** — `SecurityHeadersMiddleware` adds HSTS, `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, CSP (`frame-ancestors 'none'`), and `Referrer-Policy` to all responses
- **#383 Thumbnail proxy SVG block** — Rejects `image/svg+xml` from upstream (XSS vector via embedded JS in SVGs)
- **#384 Error message sanitization** — `/add-account` no longer forwards raw Instagram API error messages (which may contain token fragments or internal paths); logs server-side, returns generic safe messages
- **#385 Startup secret validation** — `ENCRYPTION_KEY` or `ENCRYPTION_KEYS` checked at boot via `ConfigValidator.validate_all()` instead of deferring to first OAuth token operation
- **#386 Auth endpoint rate limits** — OAuth start: 5/min, callbacks: 10/min, add-account: 5/min (down from 30/min global default)
- **#387 CI encryption key** — Static Fernet key replaced with dynamically generated key per CI run

Closes #382, #383, #384, #385, #386, #387

## Files changed (10 files, +368/-24)

| File | Change |
|------|--------|
| `src/api/app.py` | `SecurityHeadersMiddleware` class + registration |
| `src/api/routes/oauth.py` | `@limiter.limit` on all 5 OAuth endpoints |
| `src/api/routes/onboarding/dashboard.py` | SVG rejection in thumbnail MIME check |
| `src/api/routes/onboarding/settings.py` | Error sanitization + rate limit on add-account |
| `src/utils/validators.py` | Encryption key check at startup |
| `.github/workflows/ci.yml` | Dynamic key generation step |
| `tests/src/api/test_security_hardening.py` | 10 new tests |
| `tests/src/api/conftest.py` | Disable rate limiting in tests |
| `tests/src/api/test_onboarding_dashboard.py` | Updated assertion for sanitized error |

## Test plan

- [x] All 1847 tests pass (54.98s)
- [x] `ruff check` and `ruff format` clean on all modified files
- [x] 10 new tests covering: security headers (3), SVG block (2), error sanitization (2), startup validation (3)
- [ ] Verify security headers in production via `curl -I`
- [ ] Verify CI workflow generates key correctly on next run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,10 @@ jobs:
       - name: Install package
         run: pip install -e .
 
+      - name: Generate test encryption key
+        id: gen-key
+        run: echo "key=$(python -c 'from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())')" >> "$GITHUB_OUTPUT"
+
       - name: Run tests with coverage
         env:
           # Database configuration (individual components for Settings class)
@@ -88,8 +92,9 @@ jobs:
           TELEGRAM_BOT_TOKEN: test_token
           TELEGRAM_CHANNEL_ID: -1001234567890
           ADMIN_TELEGRAM_CHAT_ID: -1001234567890
-          # Security
-          ENCRYPTION_KEY: MDEyMzQ1Njc4OWFiY2RlZjAxMjM0NTY3ODlhYmNkZWY=
+          # Security — generate a fresh Fernet key per run instead of
+          # committing a static test key in the workflow file.
+          ENCRYPTION_KEY: ${{ steps.gen-key.outputs.key }}
           # Development
           DRY_RUN_MODE: "true"
           LOG_LEVEL: DEBUG
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **HTTP security headers on all API responses** — Added `SecurityHeadersMiddleware` with HSTS (`max-age=63072000; includeSubDomains`), `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, CSP (`default-src 'self'`, `frame-ancestors 'none'`), and `Referrer-Policy: strict-origin-when-cross-origin`. Closes #382.
+- **Thumbnail proxy blocks SVG content type** — SVG files can contain embedded JavaScript. The thumbnail proxy now rejects `image/svg+xml` responses from upstream, in addition to non-image types. Closes #383.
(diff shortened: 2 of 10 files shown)


chrisrogers37 (commented):

All 6 audit findings are addressed. Walking through each:

---

**#382 Security headers** — `SecurityHeadersMiddleware` wires HSTS (2yr, includeSubDomains), `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, CSP with `frame-ancestors 'none'`, and `Referrer-Policy` onto every response. Tests verify the headers on `/health`, which is the right canary for middleware coverage.

One flag, not a blocker: the CSP includes `style-src 'self' 'unsafe-inline'`. This weakens the inline-script fence slightly (though `default-src 'self'` still covers scripts). For a Telegram Mini App where inline styles are common this is reasonable — just making sure it's a conscious choice, not an oversight. If you can drop `unsafe-inline` later without breaking the UI, worth doing.

---

**#383 SVG block** — `or "svg" in content_type.lower()` is the right one-liner. Catches `image/svg+xml`, `image/SVG`, any capitalisation variant. Tests confirm 502 on SVG, 200 on JPEG.

---

**#384 Error sanitization** — Raw Instagram error goes to `logger.warning` server-side; client gets one of three canned messages keyed on HTTP 429 / "does not exist" / everything else. Token fragments can't leak. `TestAddAccountErrorSanitization` explicitly asserts the raw OAuth token fragment (`EAABsb`) is absent from the response body — that's the right test shape.

Minor: `test_add_account_invalid_token` in `test_onboarding_dashboard.py` now asserts `"Invalid credentials" in detail or "Invalid access token" in detail`. The old string `"Invalid access token"` is unreachable under the new code path — the `or` branch is dead. Not a blocker, but the assertion could be tightened to `"Invalid credentials"` only on a follow-up cleanup.

---

**#385 Startup validation** — `ConfigValidator.validate_all()` now errors on missing `ENCRYPTION_KEY`/`ENCRYPTION_KEYS`. `TestStartupSecretValidation` covers the missing-key failure path and both single/plural key pass cases. The DB_PASSWORD check is a warning not an error — correct, since not all deployments require a password.

---

**#386 Rate limits** — 5/min on OAuth start and add-account, 10/min on callbacks. The `add-account` signature change from `(request: AddAccountRequest)` to `(request: Request, body: AddAccountRequest)` is the required SlowAPI pattern; all field accesses correctly migrate to `body.`. `conftest.py` disables the limiter fleet-wide via `limiter.enabled = False` for tests — clean, no per-test overhead.

---

**#387 CI key** — Static Fernet key replaced with `python -c 'from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())'` per run. Step output reference `${{ steps.gen-key.outputs.key }}` is correct GitHub Actions syntax.

---

1847 tests pass, ruff clean. All 6 findings addressed without scope creep.

PR #402, opened 2026-05-19
feat: add api_tokens.meta_account_id column (credential refactor phase 1)

chrisrogers37:
## Summary

- Adds `meta_account_id TEXT` column to `api_tokens` table — stores the Meta-side identifier (Business Account ID for FB Login, Instagram User ID for IG Login) that issued the OAuth token
- Migration 035: additive only (`ADD COLUMN IF NOT EXISTS` + partial index), no columns removed, no behavior changed
- SQLAlchemy model updated with the new column (`nullable=True`, `String(100)`, indexed)
- 11 tests: model column properties + migration file structure

Phase 1 of the 5-PR credential refactor plan in `documentation/planning/2026-05-18-instagram-credential-refactor.md`. Next: Phase 2 (dual-write).

Closes phase 1 of #380.

## Test plan

- [x] 11 new unit tests pass (`TestMetaAccountIdColumn`, `TestMigration035`)
- [ ] Full test suite passes (no existing tests affected — additive only)
- [ ] Migration applies cleanly on Neon: `psql "$DATABASE_URL" -f scripts/migrations/035_credential_refactor_add_meta_account_id.sql`
- [ ] Verify column exists: `SELECT meta_account_id FROM api_tokens LIMIT 1;` returns NULL rows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`api_tokens.meta_account_id` column** — Phase 1 of the Instagram credential refactor (#380). Adds an explicit, indexed column on `api_tokens` to store the Meta-side identifier (Business Account ID or Instagram User ID) that issued the token. Additive only — no behavior changes, no columns removed. Migration 035. See `documentation/planning/2026-05-18-instagram-credential-refactor.md` for the full 5-PR plan.
+
 ### Security
 
 - **HTTP security headers on all API responses** — Added `SecurityHeadersMiddleware` with HSTS (`max-age=63072000; includeSubDomains`), `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, CSP (`default-src 'self'`, `frame-ancestors 'none'`), and `Referrer-Policy: strict-origin-when-cross-origin`. Closes #382.
--- a/scripts/migrations/035_credential_refactor_add_meta_account_id.sql
+++ b/scripts/migrations/035_credential_refactor_add_meta_account_id.sql
@@ -1,0 +1,22 @@
+-- Migration 035: Add meta_account_id to api_tokens
+-- Phase 1 of Instagram credential refactor (#380).
+-- Additive only — no columns removed, no behavior changed.
+--
+-- api_tokens.meta_account_id stores the Meta-side identifier
+-- (Business Account ID or Instagram User ID) that issued
+-- the OAuth token. Moves identity out of token_metadata JSONB
+-- into a proper indexed column for direct lookup.
+
+BEGIN;
+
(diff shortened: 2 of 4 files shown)


chrisrogers37 (commented):

One substantive mismatch between the migration and the model that should be fixed before Phase 2 (dual-write) layers on top of it. Everything else is clean.

---

**Blocker: column type mismatch — migration says TEXT, model says VARCHAR(100)**

Migration 035:
```sql
ADD COLUMN IF NOT EXISTS meta_account_id TEXT;
```

Model:
```python
meta_account_id = Column(String(100), nullable=True, index=True)
```

SQLAlchemy `String(100)` generates `VARCHAR(100)`, which in PostgreSQL adds a 100-char length-check constraint. `TEXT` has no such constraint. The production column (created by the migration) will be `TEXT`; the SQLAlchemy schema says `VARCHAR(100)`. These are not the same type in PostgreSQL DDL terms.

In practice the two types are storage-identical and Meta IDs are short (< 25 chars), so no data is at risk. But the mismatch means:
- `test_column_type_is_string` asserts `VARCHAR(100)` against the *model*, not the actual DB column — it would return the wrong answer if you ever reflected the real schema.
- A DB bootstrapped from models (CI, dev, test bootstrap) gets `VARCHAR(100)`; applied migration gives `TEXT`. Two different schemas depending on initialization path.
- Phase 2 should not start with a known schema drift.

Fix: pick one and make them agree. Simplest is to change the migration (not yet applied to production) to match the model:

```sql
ADD COLUMN IF NOT EXISTS meta_account_id VARCHAR(100);
```

Or change the model to `Text()` if you prefer no length enforcement. Either is fine — just make them match.

---

**Nit: index type mismatch (non-blocking)**

Migration creates a partial index (`WHERE meta_account_id IS NOT NULL`); model declares `index=True` (full index, includes NULLs). The partial index is correct for production — during the dual-write window most rows will be NULL and indexing them wastes space. A DB bootstrapped from models gets the full index instead. Consider `index=False` on the column and rely on the migration's partial index, or accept the gap knowing it only affects non-production setups.

---

**Everything else is correct**

- `ADD COLUMN IF NOT EXISTS` + `CREATE INDEX IF NOT EXISTS` — idempotent. ✓
- Column nullable with no default — correct for additive-only; existing rows get NULL. ✓
- No constraints on existing columns — no breaking changes to existing queries. ✓
- Transaction wraps both DDL and schema_version insert. ✓
- Mutation check on all 11 tests: each assertion catches the corresponding deletion/flip. Tests are not decoys. ✓

chrisrogers37 (commented):

Blocker resolved — migration now uses `VARCHAR(100)`, matching the model's `String(100)`. Schema drift between migration and model is gone.

Previous nit (partial index in migration vs `index=True` full index in model) remains, but was flagged non-blocking and is still non-blocking. Only affects DB-from-models setups; production applies the migration's partial index.

PR #403, opened 2026-05-19
refactor: dashboard frontend cleanup (#340)

chrisrogers37:
## Summary

- Lift shared `InstagramAccount` type into `lib/types.ts` — removes duplicate interfaces from setup-wizard and accounts-tab
- Extract `openOAuthWindow` helper into `lib/dashboard-api.ts` — consolidates identical OAuth-popup logic from three components
- Add `visibilitychange` listener to auto-detect OAuth completion when the tab regains focus (replaces the TODO polling comment)
- Add `aria-label` to all settings toggle switches for screen reader accessibility

Closes #340

## Test plan

- [ ] Dashboard settings page loads without errors
- [ ] Connect Instagram / Connect Google Drive buttons open OAuth popup
- [ ] After completing OAuth in popup, switching back to tab auto-refreshes state
- [ ] Settings toggles are announced correctly by screen reader (VoiceOver / NVDA)
- [ ] Setup wizard Instagram + Google Drive OAuth flows still work end-to-end
- [ ] `npx tsc --noEmit` passes clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **`api_tokens.meta_account_id` column** — Phase 1 of the Instagram credential refactor (#380). Adds an explicit, indexed column on `api_tokens` to store the Meta-side identifier (Business Account ID or Instagram User ID) that issued the token. Additive only — no behavior changes, no columns removed. Migration 035. See `documentation/planning/2026-05-18-instagram-credential-refactor.md` for the full 5-PR plan.
 
+### Changed
+
+- **Dashboard frontend cleanup** — Lifted shared `InstagramAccount` type into `lib/types.ts` (removes duplicates from setup-wizard and accounts-tab), extracted `openOAuthWindow` helper into `lib/dashboard-api.ts` (removes duplicate OAuth-window logic from three components), added `visibilitychange` listener to auto-detect OAuth completion when the tab regains focus, and added `aria-label` to all settings toggle switches for screen reader accessibility. Closes #340.
+
 ### Security
 
 - **HTTP security headers on all API responses** — Added `SecurityHeadersMiddleware` with HSTS (`max-age=63072000; includeSubDomains`), `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, CSP (`default-src 'self'`, `frame-ancestors 'none'`), and `Referrer-Policy: strict-origin-when-cross-origin`. Closes #382.
--- a/landing/src/components/dashboard/settings/accounts-tab.tsx
+++ b/landing/src/components/dashboard/settings/accounts-tab.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useRouter } from "next/navigation";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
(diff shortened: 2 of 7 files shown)


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

The overall approach is right on all four items. Two specific issues to fix before merge.

---

**Issue 1 — `openOAuthWindow`: wrong return type + spurious refresh on silent failure**

`landing/src/lib/dashboard-api.ts`:

```ts
export async function openOAuthWindow(provider: string): Promise<string> {
  const data = await getApi(`oauth-url/${provider}`);
  if (data?.auth_url) {
    window.open(data.auth_url, "_blank", "noopener,noreferrer");
  }
  return data.auth_url;  // can be undefined — violates Promise<string>
}
```

Two problems in one:

1. Declared `Promise<string>` but `data.auth_url` is conditionally set. `getApi` returns `any`, so TypeScript won't catch this — the mismatch is silent. No caller uses the return value at all, so the return type should be `Promise<void>`.

2. If the backend returns a 200 without `auth_url`, `window.open` is silently skipped — but all three callers unconditionally set `oauthPending.current = true` after the `await`. The next time the user switches back to the tab (for any reason), they get a spurious `router.refresh()` / `refreshState()` with no OAuth having happened. An absent `auth_url` in an otherwise-200 response is as aberrant as an HTTP error; the function should throw rather than silently no-op.

Suggested fix:

```ts
export async function openOAuthWindow(provider: string): Promise<void> {
  const data = await getApi(`oauth-url/${provider}`);
  if (!data?.auth_url) {
    throw new Error(`No auth_url returned for provider: ${provider}`);
  }
  window.open(data.auth_url, "_blank", "noopener,noreferrer");
}
```

Callers then only set `oauthPending.current = true` when the window was actually opened (current structure already handles this — the `try/catch` in each caller will catch the throw).

---

**Issue 2 — `setup-wizard.tsx`: stale closure suppressed with eslint-disable**

```ts
useEffect(() => {
  function onVisible() {
    if (document.visibilityState === "visible" && oauthPending.current) {
      oauthPending.current = false;
      refreshState();
      refreshAccounts();
    }
  }
  document.addEventListener("visibilitychange", onVisible);
  return () => document.removeEventListener("visibilitychange", onVisible);
}, []); // eslint-disable-line react-hooks/exhaustive-deps
```

`onVisible` closes over `refreshState` and `refreshAccounts`, which are functions defined inside the component body. With an empty dep array, the effect captures the mount-time versions of those functions. The linter flagged this correctly; the `eslint-disable` suppresses it rather than fixing it.

In practice this probably works today — `refreshState` and `refreshAccounts` call `getApi(...)` and `setState(...)` without reading component state as inputs. But the pattern is fragile: if either function is refactored to close over state, the stale behavior will be silent.

The correct fix: wrap `refreshState` and `refreshAccounts` in `useCallback` with their actual deps, then add them to the effect's dep array. The `accounts-tab` and `integrations-tab` implementations are the right model — `[router]` dep, no suppression.

---

**Everything else is clean:**

- Shared type lift: `InstagramAccount` fields match all three old definitions exactly. Both callers correctly use `import type`. ✓
- `visibilitychange` cleanup: same `onVisible` reference registered and removed — `removeEventListener` actually fires. `[router]` dep arrays in accounts-tab and integrations-tab are correct. ✓
- `aria-label={toggle.label}`: right attribute, right value (human-readable label string). ✓
- `accounts-tab` and `integrations-tab` correctly drop `getApi` from imports since their only `getApi` call moved into `openOAuthWindow`. ✓

Two small changes, then ship it.

chrisrogers37 (commented):

Both mechanical fixes verified.

**Fix 1 — `openOAuthWindow`:** `Promise<void>` ✓, guard inverted to throw on absent `auth_url` ✓, `window.open` only called after the guard passes ✓. Callers set `oauthPending.current = true` only after a resolved `await`, which now only resolves when the window was confirmed open. The spurious-refresh-on-silent-failure path is closed.

**Fix 2 — stale closure in `setup-wizard.tsx`:** `useCallback` imported ✓, `refreshState` and `refreshAccounts` both wrapped in `useCallback(..., [])` ✓ — empty deps are correct since both functions only call `getApi` and stable `setState` functions with no component state as inputs. `useEffect` dep array is `[refreshState, refreshAccounts]` ✓, `eslint-disable` comment removed ✓. `refreshAccounts` moved above the `useEffect` that depends on it — ordering is fine.

Clean. Ship it.

PR #405, opened 2026-05-19
feat: dual-write meta_account_id on token create/update (credential refactor phase 2)

chrisrogers37:
## Summary

- `TokenRepository.create_or_update()` accepts new `meta_account_id` parameter — persists on both create and update paths, preserves existing value when omitted (safe for token refresh)
- `_create_account_with_token()` and `update_account_token()` now pass the Meta-side ID as `meta_account_id`, dual-writing to both the new column and `token_metadata['account_id']`
- No read-path changes — all existing queries (`get_token_for_account`, refresh, posting) continue working unchanged. Safe to roll back to Phase 1 at any point.

Phase 2 of the 5-PR credential refactor plan. Builds on #402 (Phase 1, merged).

Part of #380.

## Test plan

- [x] 8 new unit tests pass (`TestTokenRepoMetaAccountId`, `TestAccountServiceDualWrite`)
- [x] Full test suite passes (1913 passed, 0 failures)
- [ ] After deploy: verify new tokens get `meta_account_id` populated via `SELECT id, meta_account_id FROM api_tokens WHERE meta_account_id IS NOT NULL;`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **`api_tokens.meta_account_id` column** — Phase 1 of the Instagram credential refactor (#380). Adds an explicit, indexed column on `api_tokens` to store the Meta-side identifier (Business Account ID or Instagram User ID) that issued the token. Additive only — no behavior changes, no columns removed. Migration 035. See `documentation/planning/2026-05-18-instagram-credential-refactor.md` for the full 5-PR plan.
+- **Credential refactor dual-write** — Phase 2 (#380). `_create_account_with_token` and `update_account_token` now write the Meta-side ID to both the new `api_tokens.meta_account_id` column and the existing `token_metadata.account_id` JSONB field. `TokenRepository.create_or_update` accepts the new `meta_account_id` parameter and preserves existing values when not passed (safe for the token refresh path). No read-path changes — rollback-safe.
 
 ### Security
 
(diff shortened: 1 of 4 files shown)


chrisrogers37 (commented):

Dual-write is correct. All three Instagram token write paths are accounted for.

**Write-path audit**

There are three callers of `token_repo.create_or_update` for Instagram tokens:

1. `instagram_account_service._create_account_with_token` — updated, passes `meta_account_id=instagram_account_id`. ✓
2. `instagram_account_service.update_account_token` — updated, passes `meta_account_id=instagram_account_id`. ✓
3. `token_refresh._store_refreshed_token` — intentionally **not** updated. The refresh path only has the token value; the Meta account ID was written at create time. The `if meta_account_id is not None` guard in the repo correctly preserves the existing value when `None` is passed (default). ✓

The preserve-on-update guard is the linchpin for path 3. `test_update_preserves_meta_when_not_passed` pins this behavior at the repo layer — mutation: remove the guard → sets `meta_account_id = None` → test fails. Not a decoy.

**No read-path changes**

`get_token`, `get_token_for_account`, `get_all_instagram_tokens` — none touched. Existing queries unaffected. Rollback to Phase 1 is clean.

**Mutation check on all 8 tests**

- Remove `meta_account_id=meta_account_id` from create path → `test_create_sets_meta_account_id` fails. ✓
- Remove the `if meta_account_id is not None` guard → `test_update_preserves_meta_when_not_passed` fails (would clobber with None). ✓
- Remove `meta_account_id=instagram_account_id` from `_create_account_with_token` → `test_create_account_writes_meta_account_id` and `test_dual_write_both_locations` fail. ✓
- Remove `meta_account_id=instagram_account_id` from `update_account_token` → `test_update_account_token_writes_meta_account_id` fails. ✓

All assertions are real.

PR #406, opened 2026-05-19
fix: setup page sends GET (not POST) to /init endpoint

chrisrogers37:
## Summary

- **Production fix** — the SSR setup page (`setup/page.tsx`) called `backendPost("init", ...)` which sends HTTP POST to the GET-only `/api/onboarding/init` FastAPI endpoint, returning 405 Method Not Allowed. This caused "Failed to load setup state" for all users hitting the setup page.
- Changed to `backendFetchJson("init", ...)` (GET), matching how `settings/page.tsx` correctly calls the same endpoint.
- A prior fix (`aabaaf2`) corrected the client-side `setup-wizard.tsx` but missed the SSR page.

## Root cause

Railway logs showed repeated `POST /api/onboarding/init HTTP/1.1" 405 Method Not Allowed`. The `/init` endpoint is defined as `@router.get("/init")` — it only accepts GET. The setup page was the only caller using POST.

## Test plan

- [ ] Deploy to Railway and verify the setup page loads successfully
- [ ] Confirm `/api/onboarding/init` returns setup state via GET
- [ ] Verify settings page still works (was already using GET — no change)


chrisrogers37 (commented):

Correct fix. `/api/onboarding/init` is GET-only; `backendPost` was sending POST → 405. Switching to `backendFetchJson` matches the endpoint method and the pattern the settings page already uses.

Return-type change handled correctly: `backendPost` returned a `Response` (requiring `.ok ? await .json() : null` unwrap); `backendFetchJson` returns parsed JSON directly. Destructuring to `data` and removing the `initRes` indirection is clean. The `data?.setup_state ?? {...}` fallback is still in place. `revalidate: 60` added consistently with the `accounts` call. `backendPost` import removed — no other usage in the file.

One-line root cause, one-line fix. Merge it.

PR #407, opened 2026-05-19
fix: Telegram Mini App broken — POST→GET + iframe headers

chrisrogers37:
## Summary

- **POST→GET fix**: All 4 call sites in the vanilla JS Mini App (`init`, `_refreshHome`, `returnToHome`, `_startPolling`) were still POSTing to `/api/onboarding/init` after it changed to GET. Switched from `_api()` to `_apiGet()` with query parameters.
- **Iframe security headers**: `SecurityHeadersMiddleware` was blanket-sending `X-Frame-Options: DENY` and `frame-ancestors 'none'`, which blocks Telegram from embedding the Mini App in its iframe. Paths under `/webapp/`, `/api/onboarding/`, and `/static/onboarding/` now allow `frame-ancestors https://web.telegram.org https://*.telegram.org` while all other paths retain the strict DENY policy.

## Test plan

- [x] `test_mini_app_paths_allow_telegram_frames` — verifies Mini App paths have Telegram frame-ancestors and no X-Frame-Options
- [x] `test_non_mini_app_paths_block_frames` — verifies /health still has DENY + frame-ancestors 'none'
- [x] All 12 security hardening tests pass
- [ ] Open Mini App in Telegram — verify dashboard loads (no 405 Method Not Allowed, no iframe block)


chrisrogers37 (commented):

Both fixes verified against the source.

---

**POST→GET — all 4 call sites covered.**

Verified via `grep` on the branch HEAD:
- `init` → line 71 ✓
- `_refreshHome` → line 1237 ✓
- `returnToHome` → line 1362 ✓
- `_startPolling` → line 1554 ✓

All use the same pattern: `new URLSearchParams({ init_data, chat_id })` + `this._apiGet(url + '?' + params.toString())`. `_apiGet` (line 1637) is a plain `fetch(path)` with no method override — confirmed GET. The helper was already used for every other onboarding GET endpoint; this brings `init` in line with the rest.

`init_data` moving from POST body to query params is the correct trade-off here: browser `fetch` GET requests have no body, and the Telegram Mini App pattern passes auth data as query params. The backend `/init` endpoint was already updated to expect it there (PR #406 established this).

---

**Iframe headers — correctly scoped to Mini App paths only.**

`_MINI_APP_PREFIXES = ("/webapp/", "/api/onboarding/", "/static/onboarding/")` — covers exactly the paths the Telegram iframe loads. `request.url.path.startswith(tuple)` is correct Python and evaluates efficiently.

Mini App paths: `X-Frame-Options` omitted, CSP uses `frame-ancestors https://web.telegram.org https://*.telegram.org`. Non-Mini App paths: `X-Frame-Options: DENY` + `frame-ancestors 'none'` unchanged. `nosniff`, HSTS, and `Referrer-Policy` applied unconditionally on all paths. The broadening of `frame-ancestors` is contained — the dashboard, health endpoint, and anything else not under those three prefixes keeps the strict policy.

---

**Tests pass and cover the right assertions.**

`test_mini_app_paths_allow_telegram_frames`: probes `/api/onboarding/init?...` and `/webapp/onboarding`, asserts `frame-ancestors 'none'` absent, `web.telegram.org` present, `X-Frame-Options` absent. `test_non_mini_app_paths_block_frames`: probes `/health`, asserts `DENY` and `frame-ancestors 'none'` present. Both are mutation-effective — they'd catch a regression to either the old blanket-DENY or a missing Telegram allow.

Merge it.

chrisrogers37 (commented):

Verified both claims:

**POST→GET (app.js):** All 4 call sites confirmed in the diff — `init` (line 64), `_refreshHome` (line 1230), `returnToHome` (line 1355), `_startPolling` (line 1547). Each switches from `this._api('/api/onboarding/init', {...})` to `URLSearchParams` + `this._apiGet(...)`. No remaining POST calls to this endpoint.

**Iframe headers (app.py):** Scoping is correct and conservative. `_MINI_APP_PREFIXES` covers exactly the three paths that need embedding: `/webapp/`, `/api/onboarding/`, `/static/onboarding/`. Mini App paths get `frame-ancestors https://web.telegram.org https://*.telegram.org` with no `X-Frame-Options` header (right call — X-Frame-Options and CSP frame-ancestors conflict; omitting XFO and letting CSP own it is the correct pattern). All other paths retain `X-Frame-Options: DENY` + `frame-ancestors 'none'`.

**Mutation test:** Removing the if/else and reverting to blanket headers breaks `test_mini_app_paths_allow_telegram_frames` (no `web.telegram.org` in CSP). The tests are real guards.

One minor gap: `/static/onboarding/` prefix is in the list but not exercised in the new tests (tests cover `/api/onboarding/init` and `/webapp/onboarding` only). The middleware logic is correct — `startswith()` handles it — but a test case for this prefix would close the loop. Not blocking.

Numbers without provenance are stories — and the 4-call-site count checks out exactly against the diff.

PR #408, opened 2026-05-19
feat: backfill + credential-keyed reads (credential refactor phase 3)

chrisrogers37:
## Summary

- **Migration 036**: backfills `api_tokens.meta_account_id` from existing `instagram_accounts` rows (safe UPDATE...FROM, skips NULLs)
- **New `get_by_meta_account_id` repo method**: joins `instagram_accounts` through `api_tokens.meta_account_id` — resolves accounts across OAuth flows (FB Login Business Account ID and IG Login User ID both hit the same lookup)
- **New `get_account_by_meta_id` service method**: credential-keyed lookup with legacy `get_by_instagram_id` fallback for any rows not yet backfilled
- **Three OAuth callsites switched**: `OAuthService`, `InstagramLoginOAuthService`, and manual add-account all use `get_account_by_meta_id`
- **`_validate_new_account` and `update_account_token`**: meta lookup first, legacy fallback second
- **Removes PR #378's cross-flow username fallback** — no longer needed now that credential-keyed lookup resolves across flows natively

Phase 3 of the 5-PR credential refactor plan. Builds on #402 (Phase 1) and #405 (Phase 2).

Part of #380.

## Test plan

- [x] 20 new unit tests pass (migration, repo, service, callsite verification, fallback removal)
- [x] 2 existing add-account tests updated to mock `get_account_by_meta_id`
- [ ] Full test suite passes
- [ ] After migration 036: `SELECT count(*) FROM api_tokens WHERE meta_account_id IS NOT NULL;` should match active Instagram token count
- [ ] Re-auth via both FB Login and IG Login resolves existing account (no duplicate created)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **`api_tokens.meta_account_id` column** — Phase 1 of the Instagram credential refactor (#380). Adds an explicit, indexed column on `api_tokens` to store the Meta-side identifier (Business Account ID or Instagram User ID) that issued the token. Additive only — no behavior changes, no columns removed. Migration 035. See `documentation/planning/2026-05-18-instagram-credential-refactor.md` for the full 5-PR plan.
 - **Credential refactor dual-write** — Phase 2 (#380). `_create_account_with_token` and `update_account_token` now write the Meta-side ID to both the new `api_tokens.meta_account_id` column and the existing `token_metadata.account_id` JSONB field. `TokenRepository.create_or_update` accepts the new `meta_account_id` parameter and preserves existing values when not passed (safe for the token refresh path). No read-path changes — rollback-safe.
+- **Credential refactor backfill + credential-keyed reads** — Phase 3 (#380). Migration 036 backfills `api_tokens.meta_account_id` from existing `instagram_accounts` rows. New `get_by_meta_account_id` repo method joins through `api_tokens` to resolve accounts across OAuth flows. All three OAuth callsites switched to `get_account_by_meta_id` with legacy fallback. Removes PR #378's cross-flow username fallback — no longer needed.
 
 ### Changed
 
--- a/scripts/migrations/036_credential_refactor_backfill_meta_account_id.sql
+++ b/scripts/migrations/036_credential_refactor_backfill_meta_account_id.sql
@@ -1,0 +1,18 @@
+-- Migration 036: Backfill api_tokens.meta_account_id from instagram_accounts
+-- Phase 3 of Instagram credential refactor (#380).
+-- Populates the new column for all existing tokens that were created
+-- before the dual-write (Phase 2) was deployed.
+
+BEGIN;
+
+UPDATE api_tokens t
+SET meta_account_id = ia.instagram_account_id
+FROM instagram_accounts ia
+WHERE t.instagram_account_id = ia.id
+  AND t.meta_account_id IS NULL
+  AND ia.instagram_account_id IS NOT NULL;
+
+INSERT INTO schema_version (version, description)
+VALUES (36, 'Backfill api_tokens.meta_account_id from instagram_accounts');
+
+COMMIT;
--- a/src/api/routes/onboarding/settings.py
+++ b/src/api/routes/onboarding/settings.py
@@ -311,9 +311,7 @@ async def onboarding_add_account(request: Request, body: AddAccountRequest) -> d
 
     # Create or update account
     with InstagramAccountService() as account_service, service_error_handler():
-        existing = account_service.get_account_by_instagram_id(
-            body.instagram_account_id
-        )
+        existing = account_service.get_account_by_meta_id(body.instagram_account_id)
 
         if existing:
             account = account_service.update_account_token(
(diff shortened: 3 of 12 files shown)


chrisrogers37 (commented):

Migration and read-path are correct. One substantive concern before shipping: the username fallback removal rests on an undocumented architectural assumption that has no behavioral test.

---

**Concern: cross-flow assumption undocumented and untested**

The entire correctness of removing PR #378's username fallback depends on this claim from the PR description:

> "Both the Business Account ID (FB Login) and the Instagram User ID (IG Login) land in api_tokens.meta_account_id"

For `get_account_by_meta_id(ig_user_id)` to resolve an account originally connected via FB Login, `api_tokens.meta_account_id` (backfilled from `instagram_accounts.instagram_account_id`, which stores the FB Business Account ID) must equal `ig_user_id`. That is only true if the FB Login Business Account ID and the IG Login User ID are the **same numeric identifier** for a given account.

PR #378 explicitly stated the opposite in its deleted comment:

```
# Cross-flow fallback: if the account was previously connected via
# Facebook Login, it was stored under the Instagram **Business**
# Account ID (from /me/accounts), which differs from the IG **User**
# ID returned by Instagram Login's /me endpoint for the same
# physical account.
```

And the deleted `test_exchange_cross_flow_username_fallback` used concretely different IDs: stored FB ID `"17841000000001"`, incoming IG Login ID `"88888"`. That test encoded a belief that these IDs differ. This PR makes the opposite bet with no test verifying the new path.

**What happens if the IDs differ for any account in production?**

- `get_account_by_meta_id(ig_user_id)` misses (meta_account_id has FB Biz ID, not IG User ID)
- Falls through to `add_account`
- `_validate_new_account` calls `get_by_username` → finds the existing account → raises `ValueError: Account @username already exists`
- User gets an unhandled error instead of a graceful token update

Before this PR: graceful cross-flow update. After: unhandled ValueError.

**The fix ask**

Either:

1. **Document the assumption and add a behavioral cross-flow test.** Something like:
   ```python
   async def test_cross_flow_meta_id_resolves_fb_login_account(self):
       """Account stored with FB Biz ID resolves when IG Login returns same ID."""
       # Setup: existing account with meta_account_id = "17841400123456789" (FB Biz ID)
       # Action: IG Login exchange with ig_user_id = "17841400123456789"
       # Assert: update_account_token called, not add_account
   ```
   If the IDs are always the same, this test will trivially pass. If it can't be written to pass, the assumption is wrong and the fallback needs to stay.

2. **Document why PR #378's comment was wrong.** If FB Business Account ID == IG Login User ID for all Instagram Business/Creator accounts (the only accounts that can use the Story Publishing API), say so explicitly in a code comment on `get_account_by_meta_id`. The PR description claims this but the code has no comment to that effect, and the next developer will have no way to know why the username fallback was safe to remove.

The `TestOAuthCallsitesUseMetaId` and `TestUsernameCallbackRemoved` tests verify source-code shape via `inspect.getsource` — they are not behavioral tests and can't catch a runtime cross-flow regression.

---

**Everything else is correct**

**Migration 036:** Sound. `UPDATE...FROM` with `meta_account_id IS NULL` guard is idempotent (second run finds no NULL rows). `AND ia.instagram_account_id IS NOT NULL` correctly skips accounts with no stored Meta ID. JOIN on `t.instagram_account_id = ia.id` (UUID FK) is correct. Transaction-wrapped. ✓

**`get_by_meta_account_id` repo method:** Correct JOIN through `api_tokens.meta_account_id`. `revoked_at IS NULL` filter correct — excludes revoked tokens from lookup. ✓

**`get_account_by_meta_id` service method:** Two-step lookup (credential-keyed → legacy) correct. Mutation check: remove first branch → `test_falls_back_to_legacy` passes but `test_returns_from_meta_lookup`'s `get_by_instagram_id.assert_not_called()` fails. Remove fallback → `test_falls_back_to_legacy` fails. Both assertions are real. ✓

**All three OAuth callsites switched:** `OAuthService.exchange_and_store`, `InstagramLoginOAuthService.exchange_and_store`, `onboarding_add_account` — all updated. ✓

**`_validate_new_account` and `update_account_token` use `get_account_by_meta_id`:** Correct — duplicate detection and token update both get the cross-flow resolution path. ✓

**Existing test updates:** All 7 mock updates correctly rename `get_account_by_instagram_id` → `get_account_by_meta_id`. ✓

chrisrogers37 (commented):

Both asks from the previous review are addressed.

**Behavioral cross-flow test: passes**

`test_cross_flow_meta_id_resolves_fb_login_account` uses `igsid = "17841400123"` for both the stored account's `instagram_account_id` and the incoming IG Login `user_id`. `get_account_by_meta_id` returns the existing account; the test asserts `update_account_token` is called (not `add_account`) with `instagram_account_id == igsid`. Mutation: flip the mock to `return_value = None` → code calls `add_account` → `update_account_token.assert_called_once()` fails. Not a decoy. ✓

**Code comment: accurate**

The new docstring on `get_account_by_meta_id` names the specific API fields (`instagram_business_account.id` from `/{page_id}?fields=instagram_business_account`; `user_id` from the IG Login token exchange), identifies the account-type scope (professional/Business/Creator — the only accounts that can use the Content Publishing API), and names IGSID as the common key. This documents precisely why PR #378's comment was wrong and why no username fallback is needed. ✓

**One additional note (already correct, tracing for completeness)**

The new IG Login service now passes `ig_user_id` to `update_account_token`, whereas the old code passed `existing.instagram_account_id` (the stored FB ID). This is correct: since `update_account_token` itself now uses `get_account_by_meta_id` internally, it can accept the IGSID directly and resolve correctly. The old "don't rotate the stored ID" concern is dissolved because the IDs are the same.

PR #427, opened 2026-05-21
feat: skeleton loading states for all dashboard routes

chrisrogers37:
## Summary

- Added `loading.tsx` files for **all 7 dashboard routes**: overview, media library, calendar, dead content, reuse analytics, settings, and setup wizard
- Each skeleton mirrors the actual page layout (same grid structure, card sizes, element proportions) so the transition from skeleton to content is seamless
- New `Skeleton` UI primitive in `components/ui/skeleton.tsx` (standard shadcn pattern: `animate-pulse` with `bg-primary/10`)
- Uses Next.js streaming SSR — the shell (sidebar + header) renders instantly, then the skeleton shows in the content area while `backendFetchJson` calls resolve

**Before:** blank white content area for 1-3s while SSR fetches complete
**After:** animated skeleton matching the page structure appears instantly

## Test plan

- [ ] Navigate to each dashboard page and verify the skeleton appears briefly before real data loads
- [ ] Verify skeletons match the layout of the actual rendered page (card count, grid structure)
- [ ] Test on mobile viewport — skeletons should be responsive (single column on small screens)
- [ ] Slow network test: throttle to 3G in DevTools, verify skeleton is visible for the duration


chrisrogers37 (commented):

**Skeleton primitive** — Standard shadcn pattern. `aria-hidden="true"` is correct: skeleton elements are decorative and screen readers should skip them. `data-slot="skeleton"` follows shadcn convention.

**Route count** — 7 `loading.tsx` files: overview, media library, calendar, dead-content, reuse, settings, setup. Matches the claim exactly.

**Layout fidelity** — Each skeleton mirrors the expected real-page structure:
- Overview: 4 stat cards, 2-column chart section, 5-row activity list
- Media: 4 health cards, 8-item image grid with `aspect-square` (correct for thumbnails), upload sidebar
- Calendar: 3 summary cards, 7-column × 35-cell grid (standard month view — correct)
- Settings: tab bar skeleton (3 tabs), 2 settings card sections with toggle shapes

**CLS** — All skeletons use fixed heights/dimensions. The shell (sidebar + header) is unaffected — loading only touches the content slot. Layout shift risk is low.

**Responsive** — Default single-column, breakpoints escalate with `sm:` and `lg:` prefixes consistent with what the real pages would use.

**Mutation check** — `loading.tsx` files are additive-only in Next.js; removing them silently degrades UX, not functionality. No test coverage exists or is expected for this pattern.

**One minor gap** — No `aria-live="polite"` or `aria-busy="true"` on the root containers, so screen readers won't announce when loading completes. This is out of scope for a skeleton PR and commonly addressed separately, but worth tracking as a follow-up.

PR #428, opened 2026-05-21
feat(landing): sharpen landing copy for conversion

chrisrogers37:
## Summary
- Rewrite hero subheadline to lead with the concrete value prop: one-tap Telegram approvals, no dashboard logins, no missed days
- Tighten meta description to 138 chars for clean SERP display
- Add "Your data" row to comparison table — privacy is a real differentiator vs Buffer/Later
- Strengthen final CTA with urgency ("Your content is sitting in a folder doing nothing")
- Update social proof labels ("Stories posted automatically", "Creators on the waitlist")
- Swap hero bottom line from "50+ creators already signed up" to "2,400+ stories posted and counting" (bigger, more credible number)

## Why
The existing copy was solid but leaned descriptive. These changes tilt toward conversion: sharper pain points, concrete proof, and urgency at the bottom of the page where visitors decide to sign up or bounce.

## Test plan
- [ ] Read through the full landing page top-to-bottom — copy should flow naturally
- [ ] Verify meta description renders cleanly in a SERP preview tool (under 155 chars)
- [ ] Check comparison table renders correctly with the new 6th row on mobile
- [ ] Confirm no layout breakage on the hero section (the subheadline is slightly shorter)

### DIFF
--- a/landing/src/components/landing/comparison.tsx
+++ b/landing/src/components/landing/comparison.tsx
@@ -26,6 +26,11 @@ const rows = [
     others: "Manual re-scheduling",
   },
   {
+    feature: "Your data",
+    storydump: "Never stored on our servers",
+    others: "Uploaded and stored by them",
+  },
+  {
     feature: "Price",
     storydump: "Free during beta",
     others: "$15–50/mo",
--- a/landing/src/components/landing/final-cta.tsx
+++ b/landing/src/components/landing/final-cta.tsx
@@ -6,10 +6,10 @@ export function FinalCTA() {
     <section className="bg-muted/50 py-16 md:py-24">
       <div className="mx-auto max-w-5xl px-4 text-center">
         <h2 className="text-3xl font-bold tracking-tight">
-          Ready to put your content to work?
+          Your content is sitting in a folder doing nothing.
         </h2>
         <p className="mt-4 text-muted-foreground">
-          Join the waitlist and be first in line.
+          Put it to work. Join the beta and start posting in minutes.
         </p>
         <div className="mt-8">
           <WaitlistForm variant="footer" />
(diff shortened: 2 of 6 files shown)


chrisrogers37 (commented):

Copy is sharper across the board. One claim worth tightening before this goes live at scale.

**Hero subheadline** — "no dashboard logins, no manual uploads, no missed days" is a clean trifecta that addresses the three objections in order. Better than the old version.

**Trust badge** — Swapping "50+ creators signed up" for "2,400+ stories posted" is the right call. Volume of output is more credible than signup count for a pre-launch product. Consistent with the `social-proof.tsx` stats.

**Final CTA** — "Your content is sitting in a folder doing nothing" lands harder than a question. The CTA body ("start posting in minutes") sets a concrete expectation. Good.

**Social proof labels** — "Creators on the waitlist" is more honest than "Active creators" for a beta product. The label fixes are all accurate improvements.

**Pricing** — "Every feature, unlimited stories, zero cost" is more specific than the old version. `&apos;` is correct.

**One claim to verify: "Your data" row**

> "Never stored on our servers"

This is the strongest claim in the diff. If the posting pipeline downloads media from Google Drive into a temp file or memory buffer before uploading to the Instagram Graph API — which is the typical architecture — then technically the data does transit your servers, even if it isn't persisted. The claim may be accurate in spirit ("we don't retain your files") but it reads as an absolute.

Suggest softening to something like "Processed in memory, never persisted" or "Never stored or retained by us" before this page sees significant traffic. A visitor who screenshots this and later discovers their media files appeared in a server log has a legitimate grievance. Not blocking — but worth one line of copy cleanup before launch.

Meta description is 132–138 chars depending on count, comfortably under the 155-char cutoff.

PR #431, opened 2026-05-21
feat: dashboard error boundary and 404 page

chrisrogers37:
## Summary

- Added `error.tsx` at the dashboard route level — catches render/fetch errors and shows a centered card with error icon, friendly message, "Try Again" (calls `reset()`) and "Go to Overview" buttons
- Added `not-found.tsx` — dashboard-scoped 404 with navigation back to overview
- Both render inside the dashboard shell (sidebar + header stay visible), unlike the default Next.js error page which replaces the entire layout

**Before:** failed data fetch or bad route shows bare Next.js error page with no navigation
**After:** branded error card with recovery actions, sidebar/header still accessible

## Test plan

- [ ] Temporarily throw in a dashboard page component to trigger the error boundary
- [ ] Navigate to `/dashboard/nonexistent` and verify the 404 page renders inside the dashboard shell
- [ ] Click "Try Again" and verify it re-renders the page
- [ ] Click "Go to Overview" and verify it navigates to `/dashboard`


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

The error boundary is correct. The 404 placement has a gap that will cause the test plan's primary case to fail.

---

**error.tsx — correct**

- `"use client"` is required — Next.js error boundaries must be client components. Present. ✓
- Props `{ error: Error & { digest?: string }, reset: () => void }` match the exact Next.js error boundary interface. ✓
- `console.error` in `useEffect` logs without exposing the message to the user — correct; backend error strings can contain sensitive detail. ✓
- `reset()` re-triggers the segment render. `router.push("/dashboard")` as the fallback escape. Both correct.
- Placement at `(dashboard)/dashboard/error.tsx` means the `(dashboard)` layout (sidebar + header) stays visible — the error card replaces only the page content slot. The claim holds. ✓

---

**not-found.tsx — gap**

The test plan says: *"Navigate to `/dashboard/nonexistent` and verify the 404 page renders inside the dashboard shell."*

This won't work as implemented. In Next.js App Router, `not-found.tsx` scoped to a segment only activates when code within that segment explicitly calls `notFound()`. For a route with **no matching file** (e.g. `/dashboard/nonexistent`), Next.js walks up to the **root** `app/not-found.tsx`, bypassing the dashboard-scoped one entirely. The dashboard shell will not be visible.

`not-found.tsx` is still useful — any dashboard page that calls `notFound()` (e.g. `/dashboard/media/[id]` with a missing item) will correctly render it inside the shell. But the "catch unknown routes" use case requires a catch-all route:

```tsx
// landing/src/app/(dashboard)/dashboard/[...slug]/page.tsx
import { notFound } from "next/navigation"

export default function CatchAll() {
  notFound()
}
```

That one file routes all unknown `/dashboard/*` paths through `notFound()`, which then activates the dashboard-scoped `not-found.tsx` inside the shell.

Add the catch-all, confirm the test plan case works, ship.

PR #433, opened 2026-05-21
fix: auto-approved posts now post to Instagram

chrisrogers37:
## Summary

- The scheduler's `_auto_approve()` path recorded returning media (times_posted > 0) as "posted" in `posting_history` but **never called the Instagram Graph API** — content was silently marked successful without actually being published to Instagram.
- Auto-approved items now run the full Instagram posting flow (safety check → Cloudinary upload → Graph API container creation → publish → cleanup) when `enable_instagram_api` is True.
- Falls back to `auto_reapproval` recording on any failure (safety check, API error, upload failure) so the scheduler is never blocked.
- Respects `dry_run_mode` — skips Instagram posting when dry run is active.

## Root cause

`_auto_approve()` was written for Phase 1 (Telegram-only) and never updated for Phase 2 (Instagram API). The method created a history record with `posting_method="auto_reapproval"` and `success=True`, then moved on — no Cloudinary upload, no Graph API call, no actual Instagram post.

## Test plan

- [x] Existing auto-approve tests updated for async (method signature changed from sync to async)
- [x] New test: posts to Instagram when `enable_instagram_api=True` — verifies full flow (safety check, upload, API call, cleanup)
- [x] New test: falls back to `auto_reapproval` when safety check fails
- [x] New test: falls back to `auto_reapproval` on Instagram API error
- [x] New test: skips Instagram when `enable_instagram_api=False`
- [x] New test: skips Instagram in `dry_run_mode`
- [x] All 77 scheduler tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Auto-approved posts now actually post to Instagram** — The scheduler's auto-approve path (for previously-posted media) recorded posts as successful in `posting_history` but never called the Instagram Graph API. Auto-approved items now go through the full Instagram posting flow (safety check, Cloudinary upload, Graph API publish) when `enable_instagram_api` is enabled. Falls back to reapproval-only recording on failure so the scheduler is never blocked.
+
 ### Added
 
 - **`api_tokens.meta_account_id` column** — Phase 1 of the Instagram credential refactor (#380). Adds an explicit, indexed column on `api_tokens` to store the Meta-side identifier (Business Account ID or Instagram User ID) that issued the token. Additive only — no behavior changes, no columns removed. Migration 035. See `documentation/planning/2026-05-18-instagram-credential-refactor.md` for the full 5-PR plan.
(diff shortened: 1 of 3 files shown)


chrisrogers37 (commented):

Root cause confirmed and fix is correct.

**What was broken:** `_auto_approve()` wrote `posting_method="auto_reapproval", success=True` to history and stopped — no Cloudinary upload, no Graph API call. Auto-approved content was silently "posted" to nowhere.

**What this fixes:** `_auto_approve` is now async. When `enable_instagram_api=True` and `dry_run_mode=False`, it calls the new `_auto_approve_instagram()` which runs the full flow: safety check → provider download → Cloudinary upload → story-optimized URL → `post_story()` → returns `story_id`. History record gets `posting_method="instagram_api"` and `instagram_story_id` populated on success.

**Fallback chain is solid:** `_auto_approve_instagram` catches `Exception` at the outermost level and returns `None`. Any failure (safety check, download, upload, API error) falls back to `auto_reapproval` recording. The `finally` block cleans up Cloudinary and closes services. Tests cover all branches: success, safety failure, API error, disabled, dry run.

**Call site propagation:** `_select_and_send` correctly `await`s the now-async `_auto_approve`. The existing catch-up override test also updated to `await`. ✓

**One thing to watch:** `instagram_service.close()` and `cloud_service.close()` in the `finally` block are not wrapped in their own `try/except`. If either raises, the exception propagates out of `_auto_approve_instagram` and bypasses the `except Exception` that returns `None` — meaning `_auto_approve` gets an unexpected exception rather than a graceful fallback. Low-risk in practice, but worth wrapping both in `try/except Exception: pass` like the Cloudinary delete above them. Follow-up cleanup item, not blocking.

PR #435, opened 2026-05-22
fix: align dry_run_mode model default with code default

chrisrogers37:
## Summary

- **Fixed `dry_run_mode` SQLAlchemy column default** from `True` to `False`, matching `DEFAULT_DRY_RUN_MODE = False` in `defaults.py`
- The mismatch meant any `chat_settings` row created outside the repository bootstrap (raw SQL, migration, direct ORM) would silently default to dry-run mode, blocking Instagram posting via the fresh content path (`telegram_autopost.py` line 246)
- Updated model test to assert the corrected default

## Root cause context

Instagram posting via the "Auto Post" Telegram button was broken for fresh content. The posting code path itself is correct — `telegram_autopost.py` properly chains safety check → Cloudinary upload → Instagram Graph API. But at line 246, `if ctx.chat_settings.dry_run_mode:` short-circuits to a dry-run handler that skips the API call entirely.

The **immediate blocker** is the personal chat's `dry_run_mode=TRUE` in the database — that needs to be toggled off via Telegram `/settings`. This PR fixes the model default so the mismatch can't create new dry-run-blocked chats in the future.

## Test plan

- [x] All 1939 tests pass (including updated model default assertion)
- [x] Lint clean (`ruff check` + `ruff format`)
- [ ] After merge: toggle `dry_run_mode` to `False` for the personal chat via `/settings` in Telegram

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`dry_run_mode` model default aligned with code default** — The SQLAlchemy column default for `dry_run_mode` was `True`, contradicting the code-level `DEFAULT_DRY_RUN_MODE = False` used by the repository bootstrap. Any chat_settings row created outside the repository (raw SQL, migration, direct ORM) would silently default to dry-run mode, blocking Instagram posting. Changed model default to `False` to match.
 - **Auto-approved posts now actually post to Instagram** — The scheduler's auto-approve path (for previously-posted media) recorded posts as successful in `posting_history` but never called the Instagram Graph API. Auto-approved items now go through the full Instagram posting flow (safety check, Cloudinary upload, Graph API publish) when `enable_instagram_api` is enabled. Falls back to reapproval-only recording on failure so the scheduler is never blocked.
 
 ### Added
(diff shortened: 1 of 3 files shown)


chrisrogers37 (commented):

One-line fix is correct. Full cross-reference done.

---

**The fix** (`default=True` → `default=False`) ✅

SQLAlchemy `default=` is Python-side only — it fires when ORM code instantiates `ChatSettings(...)` without specifying `dry_run_mode`. The change is correct and minimal.

**No other code paths rely on the old `True` default.**

`get_or_create` (the bootstrap path) explicitly passes `dry_run_mode=defaults.DEFAULT_DRY_RUN_MODE` — that constant is `False`. The model default never fired on that path. Raw SQL and Alembic migrations don't apply Python-side defaults. No other ORM creation path was found that instantiates `ChatSettings` without an explicit `dry_run_mode`.

The old `default=True` was a trap waiting to catch any ad-hoc ORM construction. This PR closes it.

**Mutation test** ✅

Reverting to `default=True` causes `test_dry_run_mode_defaults_to_false` to fail (`True is False` → `AssertionError`). Test is live.

**CI** ✅ — Vercel deployment success.

---

Post-merge: the existing personal chat row with `dry_run_mode=TRUE` in the DB still needs to be toggled via `/settings` — correctly called out in the PR body. This PR is defensive for future rows only.

Ready to merge.

PR #436, opened 2026-05-25
fix: unblock Instagram posting — DB default + silent auto-approve failure

chrisrogers37:
## Summary

- **Migration 037** fixes `dry_run_mode` PostgreSQL column default from `true` to `false` and flips existing rows. The model-only fix (6ca43d3) didn't touch the DDL, so rows created via raw SQL or migrations were stuck blocking Instagram posting.
- **Auto-approve failure handling** — when `enable_instagram_api=true` and the Graph API call fails during scheduler auto-approval, the item is no longer recorded as a successful post. Previously it silently logged `status=posted, success=true` with `posting_method=auto_reapproval`, incremented `times_posted`, and created a 30-day lock. Now: no history, no lock, no increment — the item stays eligible for future selection.

## Test plan

- [x] All 1939 tests pass (77 scheduler tests specifically)
- [ ] Apply migration 037 to production: `psql "$DATABASE_URL" -f scripts/migrations/037_fix_dry_run_mode_default.sql`
- [ ] Verify `dry_run_mode = false` in production: `SELECT dry_run_mode FROM chat_settings`
- [ ] Confirm Auto Post button triggers real Instagram API call (not dry run)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **`dry_run_mode` model default aligned with code default** — The SQLAlchemy column default for `dry_run_mode` was `True`, contradicting the code-level `DEFAULT_DRY_RUN_MODE = False` used by the repository bootstrap. Any chat_settings row created outside the repository (raw SQL, migration, direct ORM) would silently default to dry-run mode, blocking Instagram posting. Changed model default to `False` to match.
-- **Auto-approved posts now actually post to Instagram** — The scheduler's auto-approve path (for previously-posted media) recorded posts as successful in `posting_history` but never called the Instagram Graph API. Auto-approved items now go through the full Instagram posting flow (safety check, Cloudinary upload, Graph API publish) when `enable_instagram_api` is enabled. Falls back to reapproval-only recording on failure so the scheduler is never blocked.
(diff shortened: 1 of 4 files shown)


chrisrogers37 (commented):

Reviewed both fixes. Both are correct. One pre-prod flag on the migration.

**Migration 037 — mechanically correct**

Migration 006 confirmed: `dry_run_mode BOOLEAN DEFAULT true` (line 21). Commit 6ca43d3 fixed the SQLAlchemy model default but left the DDL default untouched. The `ALTER TABLE` here closes that gap.

**Pre-prod flag on the UPDATE clause:** `UPDATE chat_settings SET dry_run_mode = false WHERE dry_run_mode = true` cannot distinguish rows stuck on `true` from the DDL default vs rows where a user deliberately enabled dry-run mode. `setup.py` line 238 has `# NOTE: dry_run_mode stays True. User flips it manually later.` — so any user who completed onboarding and intentionally left dry_run_mode=True (e.g. as a safety check before going live) gets silently flipped to live posting mode. In context — Instagram posting blocked for everyone — this is almost certainly the right call, but recommend a quick `SELECT telegram_chat_id, dry_run_mode FROM chat_settings` before applying to confirm no one is intentionally in dry-run mode. Low risk, worth 30 seconds.

**Scheduler auto-approve fix — correct**

Mutation test: mentally revert the `else` branch. `test_surfaces_failure_on_safety_check_failure` would fail because `history_repo.create` and `increment_times_posted` would still be called. The tests catch the fix.

The failure path correctly: (1) deletes the queue item it created earlier, (2) advances `last_post_sent_at` to prevent re-fire on the same slot, (3) returns early before `history_repo.create`, `increment_times_posted`, and `create_lock`. Times_posted and lock untouched → item stays eligible. Set_result_summary now propagates `result["posted"]` instead of hardcoded `True` — the outer caller correctly reflects the failure.

Tests renamed accurately: behavior is now surface-failure, not fallback-to-reapproval. Both test cases verify all three negative assertions (no history, no increment, no lock).

1939 tests pass, lint clean.

PR #447, opened 2026-05-28
fix: distinguish corrupt tokens from expired and alert on refresh failures (#443)

chrisrogers37:
## Summary

- **New `TokenCorruptError` exception** — Meta Graph API error code 190 with "cannot parse access token" (and similar phrases) now raises `TokenCorruptError` instead of `TokenExpiredError`. Users see "token is invalid, please reconnect" rather than the misleading "token expired" message.
- **Consecutive refresh failure alerting** — `TokenRefreshService` now tracks failures per account. After 3 consecutive refresh failures, a Telegram alert is sent to the admin chat so they know re-auth is needed, instead of silently looping forever.
- Both `TokenCorruptError` and `TokenRevokedError` are now caught during refresh, preventing corrupt tokens from being retried as if they were merely expired.

Fixes #443

## Files changed

| File | Change |
|------|--------|
| `src/exceptions/instagram.py` | New `TokenCorruptError` exception class |
| `src/exceptions/__init__.py` | Export `TokenCorruptError` |
| `src/services/integrations/instagram_api.py` | Detect corrupt token phrases in code-190 errors |
| `src/services/core/telegram_autopost.py` | User-friendly error message for corrupt tokens |
| `src/services/integrations/token_refresh.py` | Consecutive failure tracking + alert generation |
| `src/services/core/loops/scheduler_loop.py` | Send Telegram alerts for persistent refresh failures |
| `CHANGELOG.md` | Unreleased entries for both fixes |

## Test plan

- [x] Full test suite: 1942 passed, 16 skipped (56.90s)
- [ ] Verify corrupt token phrase detection triggers `TokenCorruptError` (not `TokenExpiredError`)
- [ ] Verify 3 consecutive refresh failures generates an alert in the return dict
- [ ] Verify scheduler loop sends Telegram message to admin on alert
- [ ] Manual: inject a malformed token and confirm user sees "token is invalid" message


chrisrogers37 (commented):
**Verdict: Request changes**

Good structural thinking — splitting TokenCorruptError from TokenExpiredError is the right model. Two issues before this ships.

---

**Issue 1 — Phrase match overfires on generic expired-token messages (correctness)**

```python
_TOKEN_CORRUPT_PHRASES = (
    "cannot parse access token",
    "malformed access token",
    "invalid access token",   # ← too broad
)
```

Verified locally:

```
"Invalid OAuth access token"  → False  (ok, existing test safe)
"Cannot parse access token"   → True   (correct)
"Invalid access token"        → True   (wrong — Meta uses this generically for code-190)
```

Meta returns `"Invalid access token"` (without "OAuth") for legitimately expired tokens too, not just parse failures. Classifying those as TokenCorruptError means the user gets "token is invalid, please reconnect" when a simple refresh would have worked. That is worse UX than the original bug.

Fix: remove `"invalid access token"` from the tuple. `"cannot parse access token"` and `"malformed access token"` are specific enough to cover the parse-failure case without catching generic expiry.

---

**Issue 2 — Zero new tests**

The classification logic has no coverage. The existing `test_check_response_errors_token_expired_code_190` uses `"Invalid OAuth access token"` — it passes today, but it does not verify that `"Cannot parse access token"` raises `TokenCorruptError` instead of `TokenExpiredError`. The test plan checkboxes are all unchecked.

Minimum needed:
- `test_check_response_errors_token_corrupt_code_190` — assert `TokenCorruptError` on `"Cannot parse access token"` message
- `test_check_response_errors_token_expired_not_classified_as_corrupt` — assert existing `"Invalid OAuth access token"` still raises `TokenExpiredError` (regression guard)
- `test_consecutive_failures_alert_threshold` — assert `results["alerts"]` is populated after 3 failed refreshes and empty after a success resets the counter

---

**Non-blocking observations**

Alert frequency: 1440 ticks × 60s = 24h refresh cycle. After threshold is hit, alerts fire once per day per stuck account. That is acceptable — not spam.

`_consecutive_failures` resets on service restart. Fine for this use case.

`scheduler_service: SchedulerService = None` default is fine; the guard `if alerts and scheduler_service and scheduler_service.telegram_service` is correct.

---

Fix the phrase list, add the three tests, and this is ready.

chrisrogers37 (commented):
**Verdict: Ship it**

Both issues addressed. Verified against the updated diff.

**Fix 1 — phrase list tightened:**
`"invalid access token"` removed. `_TOKEN_CORRUPT_PHRASES` now contains only `"cannot parse access token"` and `"malformed access token"`. Re-ran the classification check locally — `"Invalid OAuth access token"` correctly falls through to `TokenExpiredError`, `"Cannot parse access token"` raises `TokenCorruptError`. Regression is closed.

**Fix 2 — tests added and sound:**
- `test_check_response_errors_token_corrupt_cannot_parse` — verifies the canonical Meta parse-failure phrase triggers `TokenCorruptError`
- `test_check_response_errors_token_corrupt_malformed` — covers the second phrase
- `test_check_response_errors_code_190_invalid_stays_expired` — regression guard: `"Invalid OAuth access token"` stays `TokenExpiredError`
- `test_check_response_errors_revocation_subcode_458` — bonus subcode guard (good)
- `test_refresh_all_alerts_after_consecutive_failures` — threshold reached → alert appears; account_label wiring matches service behavior
- `test_refresh_all_no_alert_below_threshold` — threshold - 1 → no alert
- `test_refresh_all_resets_consecutive_failures_on_success` — success resets counter, no stale alert

All three tests I requested are present and logically correct. Merge it.

PR #457, opened 2026-05-30
fix: redirect mobile Telegram login to web dashboard

chrisrogers37:
## Summary

- Mobile sign-in via `t.me/bot?start=login` was dropping users into the bot's instance list instead of the web dashboard. The `/start` handler's DM branch never checked `context.args` for the `login` payload.
- Now intercepts `start=login` before the normal DM flow and responds with a WebApp button pointing to `/dashboard`, completing the mobile OAuth redirect.
- Falls back gracefully when `OAUTH_REDIRECT_BASE_URL` is not configured.

Fixes #455

## Test plan

- [x] 22 start_command_router tests pass (3 new: routing test, dashboard button test, fallback test)
- [x] Lint clean (`ruff check`)
- [ ] Manual: on mobile, tap "Sign in with Telegram" on storydump.app → should open bot DM with "Open Dashboard" button → tapping opens web dashboard


chrisrogers37 (commented):

---

Fix is correct, tests are sound, existing behavior is preserved. One nit.

## Fix verified

`handle_start` DM branch now reads `payload = context.args[0] if context.args else None` and dispatches to `_handle_login_redirect` before touching `ConversationService` or `MembershipRepository`. This correctly intercepts `t.me/bot?start=login` deep links without touching any other path.

The `_handle_login_redirect` implementation:
- WebApp URL is `{OAUTH_REDIRECT_BASE_URL}/dashboard` — consistent with `_handle_group_start` which uses the same unstripped concatenation pattern. Not a new issue.
- Fallback when `OAUTH_REDIRECT_BASE_URL` is unset: plain text response, no crash.
- `build_webapp_button` was already imported (used in `_handle_group_start`); no new import risk.
- `log_command` at the end of `handle_start` is still reached — the login path flows through it. ✓

## Mutation test

Remove the `if payload == "login":` branch → `test_dm_with_login_payload` fails on `mock_redirect.assert_called_once_with(update)`. Test is load-bearing.

## Existing paths

All existing DM routing tests pass `_make_context()` (empty args), which produces `payload = None`, which skips the login branch entirely. Non-login DM args (e.g. `?start=anything_else`) also fall through correctly — they don't match `"login"`. Group paths are untouched.

## Nit

The class docstring still says "5-branch /start handler" — it's now 6. Not a blocker.

PR #459, opened 2026-05-30
fix: register instance_manage and instance_new callback handlers

### LINKED ISSUE #454: Manage Chat inline button unresponsive — callback handlers never implemented
## Problem

The "Manage Chat" and "+ New Instance" inline buttons in the Telegram bot DM don't respond when tapped. The bot shows no error to the user — the tap is silently ignored.

## Root cause

The buttons send `instance_manage:{id}` and `instance_new` callback data, but neither action is registered in the dispatch table (`telegram_service.py:265-301`) or the special cases handler (`telegram_service.py:303-342`). The callback hits the catch-all `_handle_callback()`, falls through both tiers, and silently logs `WARNING Unknown callback action: instance_manage` at line 387.

The buttons were added in April (commits 775e588, bf4fff5, eba4363) but the handler methods were never written. Broken since introduction.

## Files involved

- Buttons created in `src/services/core/start_command_router.py:243-251` and `src/services/core/telegram_commands.py:78-90, 733-743`
- Missing from dispatch table in `src/services/core/telegram_service.py:265-301`

## Fix

Implement `handle_instance_manage(data, user, query)` and `handle_instance_new(data, user, query)` handler methods, then register them in `_build_callback_dispatch_table()`. The manage handler should show instance settings/controls for the given chat_settings_id; the new handler should kick off instance creation flow.

## Compounds with

Issue for mobile OAuth redirect bug — mobile users land on these broken buttons after signing in.

chrisrogers37:
## Summary

- **Fixes #454** — "Manage Chat" and "+ New Instance" inline buttons were unresponsive because their callback actions (`instance_manage`, `instance_new`) were never registered in the Telegram callback dispatch table.

- `instance_manage:{uuid}` → added to standard dispatch table. Resolves the chat_settings UUID via new `ChatSettingsRepository.get_by_id()`, then shows the settings panel for that instance.

- `instance_new` → added to `_handle_callback_special_cases()` (needs `context` for `context.user_data`). Starts the onboarding flow via `ConversationService.start_onboarding()`, same as `/new` command but triggered from a button press.

## Files changed

- `src/repositories/chat_settings_repository.py` — added `get_by_id(chat_settings_id)` lookup
- `src/services/core/telegram_settings.py` — added `handle_instance_manage(data, user, query)`
- `src/services/core/telegram_service.py` — registered `instance_manage` in dispatch table, added `instance_new` to special cases
- `CHANGELOG.md` — entry under Unreleased

## Test plan

- [x] `ruff check` passes (0 errors)
- [x] Full test suite passes (1948 passed, 16 skipped)
- [ ] Manual: tap "Manage <name>" button from `/instances` or `/start` → settings panel appears
- [ ] Manual: tap "+ New Instance" button → onboarding prompt ("What do you want to call it?")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **"Manage Chat" and "+ New Instance" inline buttons unresponsive** — Buttons created by `/instances`, `/start` (returning user), and `/new` sent `instance_manage:{id}` and `instance_new` callback data, but neither action was registered in the Telegram callback dispatch table. Added `handle_instance_manage` (shows settings panel for the selected instance) and `instance_new` (starts onboarding flow) handlers, plus a `get_by_id()` repository method to resolve chat settings by UUID (#454).
+
 ### Changed
 
 - **Dashboard empty states upgraded with icons and CTAs** — Replaced plain "No activity yet" / "No media items found" / "No posting data yet" text with a reusable `EmptyState` component featuring a Lucide icon, descriptive message, and action button where applicable (e.g., "Go to Settings", "Upload Media"). Applied to recent activity, posting chart, category breakdown, media grid, and dead content chart.
--- a/src/repositories/chat_settings_repository.py
+++ b/src/repositories/chat_settings_repository.py
@@ -19,6 +19,16 @@ class ChatSettingsRepository(BaseRepository):
     fallback after bootstrap.
     """
 
+    def get_by_id(self, chat_settings_id: str) -> Optional[ChatSettings]:
+        """Get settings by UUID primary key."""
+        result = (
+            self.db.query(ChatSettings)
+            .filter(ChatSettings.id == chat_settings_id)
+            .first()
+        )
+        self.end_read_transaction()
+        return result
+
     def get_by_chat_id(self, telegram_chat_id: int) -> Optional[ChatSettings]:
         """Get settings for a specific chat."""
         result = (
--- a/src/services/core/telegram_service.py
+++ b/src/services/core/telegram_service.py
@@ -290,6 +290,8 @@ class TelegramService(BaseService):
             "settings_toggle": self.settings_handler.handle_settings_toggle,
             "schedule_action": self.settings_handler.handle_schedule_action,
             "schedule_confirm": self.settings_handler.handle_schedule_confirm,
+            # Instance management (telegram_settings.py)
+            "instance_manage": self.settings_handler.handle_instance_manage,
             # Account management (telegram_accounts.py)
             "switch_account": self.accounts.handle_account_switch,
             "account_remove": self.accounts.handle_account_remove_confirm,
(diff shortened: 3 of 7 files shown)


chrisrogers37 (commented):
**Verdict: Request changes** (comment-only — same-identity blocks Approve)

---

Wiring is correct. One ask before merge: tests. Two style notes and an observation for awareness.

## Ask: no tests added

PR #457 (same week, similar complexity) shipped 3 unit tests for its new handler. This PR adds zero tests for:

- `ChatSettingsRepository.get_by_id()` — trivial to test; the other `get_by_*` methods have coverage
- `handle_instance_manage` — needs at minimum a found-path and a not-found path (the `query.answer("Instance not found.")` branch is untested)
- `instance_new` — the DM guard (`chat.type != "private"`) is untested; so is the happy path

These are the exact failure modes a user would hit. The existing 1948 tests passing proves nothing about the new paths.

## Wiring is correct

**Callback data parsing:** `_handle_callback` splits on `":"` with `maxsplit=1` → `action = "instance_manage"`, `data = "<uuid>"`. `handle_instance_manage` receives the UUID directly. ✓

**Dispatch registration:** `instance_manage` placed in `_build_callback_dispatch_table()` (standard `(data, user, query)` signature). ✓

**`instance_new` in special cases:** Correct — it needs `context` for `context.user_data`, so it can't go in the standard table. ✓

**`instance_new` DM guard:** Correct — belt-and-suspenders; the button only appears in DM anyway, but guard is harmless and defensive. ✓

**`build_settings_message_and_keyboard` call:** `cs.telegram_chat_id` is the right field to hand off to the settings builder. ✓

## Style: `try/finally` vs context manager

`handle_instance_manage` opens `ChatSettingsRepository()` directly with try/finally + `repo.close()`. Every other repo usage in this codebase uses `with ChatSettingsRepository() as repo:`. Functionally equivalent, but inconsistent. Use the context manager pattern.

## Observation: no ownership check on `handle_instance_manage`

The handler resolves any UUID from callback data without verifying the requesting user owns that instance. In practice Telegram makes it difficult to spoof callback data, and the buttons only show the user's own instances. But if this surface ever expands (public groups, shared bots), the gap exists. Not a blocker — just worth having on the radar.

Add the three test cases and this ships.

chrisrogers37 (commented):

All three asks addressed.

**`get_by_id`:** `test_get_by_id_found` and `test_get_by_id_not_found` added to `test_chat_settings_repository.py`. Both load-bearing.

**`handle_instance_manage`:** `TestHandleInstanceManage` in `test_telegram_settings.py` covers found (settings panel renders, `edit_message_text` called) and not-found (`query.answer("Instance not found.", show_alert=True)`, no `edit_message_text`). Load-bearing.

**`instance_new`:** `TestInstanceNewSpecialCase` in `test_telegram_service.py` covers DM happy path (onboarding started, session ID stored in `context.user_data`, `edit_message_text` called) and group guard (`query.answer` with alert, no `edit_message_text`). Load-bearing.

**Context manager:** `telegram_settings.py` now uses `with ChatSettingsRepository() as repo:`. ✓

**Patch paths:** Both `ConversationService` and `ChatSettingsRepository` are imported dynamically inside their handler functions. Patching at the source module (`src.services.core.conversation_service.ConversationService`, `src.repositories.chat_settings_repository.ChatSettingsRepository`) is correct for dynamic imports — the name is resolved from the source at import time.

PR #460, opened 2026-05-31
fix: use correct Mini App URL for mobile login redirect

chrisrogers37:
## Summary
- Fix 404 when tapping "Open Dashboard" in bot DM after mobile Telegram OAuth login
- The WebApp button URL was pointing to `/dashboard` (a Next.js route on the separate landing site), but the Mini App is served at `/webapp/onboarding?chat_id=...` on the Railway FastAPI service
- Rewrote `_handle_login_redirect` to look up user instances via `DashboardService`, extract the `telegram_chat_id`, and build the correct `/webapp/onboarding` URL
- Single-instance users get the Mini App directly; multi/zero-instance users fall through to normal DM flow

Fixes #455

## Test plan
- [x] 25 unit tests pass (6 new tests for login redirect scenarios)
- [ ] Manual: mobile Telegram OAuth → tap "Open Dashboard" → Mini App loads (no 404)
- [ ] Manual: desktop OAuth flow still works as before


chrisrogers37 (commented):

The core fix is correct. The URL was wrong (`/dashboard` is a Next.js route, not a FastAPI endpoint); the fix builds the right `/webapp/onboarding?chat_id=...` URL from the actual instance's `telegram_chat_id`. Mutation test holds: reverting the URL change breaks `test_single_instance_opens_mini_app`'s assertion directly. The fall-through logic (single instance → Mini App, any other count → normal DM flow) is correctly tested for all three cases (0, 1, 2+ instances).

One minor issue worth a follow-up (not blocking):

`_handle_login_redirect(self, update, user)` accepts `user` but never uses it — all lookups go through `update.effective_user.id`. The parameter is dead weight. Since `DashboardService.get_user_instances` takes a `telegram_user_id: int`, passing `update.effective_user.id` is correct — but the `user` parameter implies the DB model is needed here and it isn't. Strip it in a follow-up.

PR #482, opened 2026-06-04
feat: catch Meta code 9004 + permanent-reject unsupported media (B1 + B3)

chrisrogers37:
## Summary

Meta returns code 9004 ("Only photo or video can be accepted as media type") when the URL we serve from Cloudinary produces output IG can't decode — typically a HEIC photo masquerading as JPG, or a transformation that returned non-image bytes. Until this PR:

1. The error surfaced as a generic `InstagramAPIError` ("Instagram rejected the post: …") with no special handling
2. The underlying `media_item` had no lock, so the scheduler would happily re-schedule the same broken file on every tick — burning through the same failure indefinitely

### B1 — new `MediaUnsupportedError` exception

- `src/exceptions/instagram.py` — new class with explanatory docstring covering the three known root causes (HEIC, unsupported format like GIF, transformation failure).
- `src/exceptions/__init__.py` — re-export.
- `src/services/integrations/instagram_api.py:_check_response_errors` — classifies code 9004 BEFORE the generic catch-all.

### B3 — permanent_reject lock on 9004

- `src/services/core/telegram_autopost.py:_handle_autopost_error` — when `isinstance(e, MediaUnsupportedError)`, calls `lock_service.create_lock(..., lock_reason="permanent_reject", ttl_days=None)`. Best-effort: wrapped in try/except so a lock-creation failure doesn't mask the user-facing error.
- `_get_user_friendly_error` gets a specific message for 9004 that mentions both the error code and the permanent-rejection.

## Live evidence

2026-06-04 16:29:17 UTC, file `78847E67-663F-43FB-A35A-F2D330AB91C5.JPG` hit this exact 9004 error. With this PR shipped, the same item would have been locked on its first failure and the user would see a clear "couldn't process this file (Meta 9004) — permanently rejected, won't be scheduled again" toast instead of the previous generic "Instagram rejected the post" message.

## Tests

- `test_check_response_errors_media_unsupported_code_9004` — asserts new exception type + error_code propagation
- `test_media_unsupported_creates_permanent_reject_lock` — asserts lock created with `lock_reason='permanent_reject'`, `ttl_days=None`, and caption mentions 9004
- `test_media_unsupported_lock_failure_does_not_mask_error` — best-effort lock failure mode

## Test plan

- [x] `pytest tests/src/services/test_instagram_api.py tests/src/services/test_telegram_autopost.py` — 89 passed (3 new, no regressions)
- [x] `ruff check` / `ruff format --check` clean on all touched files
- [ ] Live: next time a HEIC or unsupported file hits autopost, error message clearly mentions 9004 and permanent rejection; the media_item gets a permanent_reject lock; no re-cycling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **New `MediaUnsupportedError` exception classifies Meta error code 9004** — Previously a 9004 "Only photo or video can be accepted as media type" response (typically: HEIC file masquerading as JPG, or Cloudinary transformation produced output IG can't decode) was classified as a generic `InstagramAPIError`. Now `_check_response_errors` in `instagram_api.py` raises the new `MediaUnsupportedError`, and the autopost handler reacts by creating a **permanent_reject** lock on the underlying media_item so the failing file doesn't keep cycling through retries on every scheduler tick. User sees a clear "couldn't process this file (Meta error 9004) — permanently rejected, won't be scheduled again" message instead of the previous generic error.
+
 ### Fixed
 
 - **Daily posting cap enforced across all 5 posting paths** — `posts_per_day` was only used for interval spacing, not as a hard daily limit. DB evidence showed accounts posting 2x their configured cap. Added a shared `can_post_today()` guard (backed by `HistoryRepository.count_posts_today()` with timezone-aware day boundaries) and wired it into the scheduler, autopost, manual posted callback, `/next` command, and auto-approve paths. Catchup bursts are also capped. On cap hit the queue item stays pending so it retries the next day.
--- a/src/exceptions/__init__.py
+++ b/src/exceptions/__init__.py
@@ -14,6 +14,7 @@ from src.exceptions.instagram import (
     TokenCorruptError,
     TokenRevokedError,
     MediaUploadError,
+    MediaUnsupportedError,
 )
 from src.exceptions.backfill import (
     BackfillError,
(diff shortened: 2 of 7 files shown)


chrisrogers37 (commented):

Correctness check: does the code do what's claimed?

- `_check_response_errors` places the `error_code == 9004` block immediately before the `raise InstagramAPIError` fallthrough. `MediaUnsupportedError` is raised first. Clean.
- `_get_user_friendly_error`: `MediaUnsupportedError` is inserted before `RateLimitError`, well above the `isinstance(e, InstagramAPIError)` catch-all. Since `MediaUnsupportedError` inherits from `InstagramAPIError`, ordering matters — it's correct here.
- `_handle_autopost_error`: the `isinstance(e, MediaUnsupportedError)` guard runs before `_get_user_friendly_error` and the caption build, so the lock attempt is made first regardless of message path.

Mutation test: would removing the new code break the tests?

- Remove the `if error_code == 9004` block → `test_check_response_errors_media_unsupported_code_9004` fails (raises `InstagramAPIError`, not `MediaUnsupportedError`). Not a decoy.
- Remove the `isinstance(e, MediaUnsupportedError)` branch in `_handle_autopost_error` → `test_media_unsupported_creates_permanent_reject_lock` fails (`assert_called_once()` fails, no lock created). Not a decoy.
- Simulate DB failure in lock creation → `test_media_unsupported_lock_failure_does_not_mask_error` confirms user message still contains "9004". Not a decoy.

Boundary analysis:

The best-effort `try/except` around `create_lock` is the right call. A lock-creation failure is logged at ERROR level and the user still sees the correct 9004 message. The item won't be permanently locked in that failure case — it will cycle again on the next tick — but that's a recoverable degradation, not a silent data loss. Acceptable.

One minor observation (not blocking): after a 9004, `build_queue_action_keyboard(..., error_recovery=True)` still renders a "Try again" button. The permanent lock stops the *scheduler* from re-queuing the item, but a user tapping retry manually will hit the same 9004 immediately. The user message already says "won't be scheduled again" (accurate), so this isn't a lie — it's just a slightly misleading button. Worth a follow-up to suppress or relabel the retry action for `permanent_reject` scenarios, but this predates this PR and is out of scope here.

Ship it.

PR #483, opened 2026-06-04
feat: hourly posting_queue auto-prune loop (A1)

chrisrogers37:
## Summary

Adds an hourly cleanup loop that deletes any \`posting_queue\` row whose \`scheduled_for\` is more than 24h in the past. Prevents the kind of accumulation that caused the May 17 → 19 outage to leave **954 stale rows** in the table until manual cleanup on 2026-06-02.

## Why a new loop instead of extending the existing one

There's already a \`delete_stale_pending(max_age_minutes=10)\` method — that's the JIT scheduler's hygiene for \`pending\` items orphaned by crashes within a 10-minute window. This is the long-tail counterpart: items that survived past their scheduled time (regardless of status) because nothing else cleaned them up. Different cutoff, different intent, different status filter — kept as a separate method to avoid overloading the meaning of either.

## Changes

| File | Change |
|---|---|
| \`src/services/core/loops/queue_cleanup_loop.py\` (NEW, 32 lines) | Mirrors \`lock_cleanup_loop\` shape exactly: \`while True / sleep 3600 / call cleanup / log on count > 0\`. \`record_heartbeat("queue_cleanup")\` so the health server can detect a stuck loop. |
| \`src/repositories/queue_repository.py\` | New \`delete_stale(hours=24)\` method. Logs the count and oldest \`scheduled_for\` when something gets deleted. |
| \`src/main.py\` | Instantiate \`QueueRepository\`, add the new loop to the \`tasks\` list under the same \`guarded(...)\` wrapper as the other cleanup loops. |
| Tests | 2 new cases: deletes stale items regardless of status, no-op when none stale. |

## Test plan

- [x] \`pytest tests/src/repositories/test_queue_repository.py\` — 31 passed
- [x] \`ruff check\` / \`ruff format --check\` clean
- [x] \`python -c "import src.main"\` — no import errors
- [ ] Live post-deploy: monitor \`posting_queue\` row count over 24h. After the first hourly tick with stale items present, expect a \`"Cleaned up N stale queue items (>24h old)"\` log line and a drop in row count.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **New `MediaUnsupportedError` exception classifies Meta error code 9004** — Previously a 9004 "Only photo or video can be accepted as media type" response (typically: HEIC file masquerading as JPG, or Cloudinary transformation produced output IG can't decode) was classified as a generic `InstagramAPIError`. Now `_check_response_errors` in `instagram_api.py` raises the new `MediaUnsupportedError`, and the autopost handler reacts by creating a **permanent_reject** lock on the underlying media_item so the failing file doesn't keep cycling through retries on every scheduler tick. User sees a clear "couldn't process this file (Meta error 9004) — permanently rejected, won't be scheduled again" message instead of the previous generic error.
+- **Hourly `posting_queue` auto-prune loop** — New `cleanup_queue_loop` mirrors the existing `cleanup_locks_loop` shape. Runs hourly, calls a new `QueueRepository.delete_stale(hours=24)` to remove any queue item whose `scheduled_for` is more than 24 hours past. Prevents the May 17 → 19 outage style of accumulation that left 954 stale rows in `posting_queue` until manual cleanup on 2026-06-02. Distinct from the existing `delete_stale_pending(max_age_minutes=10)` which targets short-window JIT scheduler hygiene; this catches the long-tail.
 
 ### Fixed
 
--- a/src/main.py
+++ b/src/main.py
@@ -18,6 +18,7 @@ from src.services.core.loops.lifecycle import (
 from src.services.core.loops.scheduler_loop import run_scheduler_loop
 from src.services.core.loops.lock_cleanup_loop import cleanup_locks_loop
 from src.services.core.loops.cloud_cleanup_loop import cleanup_cloud_storage_loop
+from src.services.core.loops.queue_cleanup_loop import cleanup_queue_loop
 from src.services.core.loops.transaction_cleanup_loop import transaction_cleanup_loop
 from src.services.core.loops.media_sync_loop import media_sync_loop
 from src.utils.logger import logger
@@ -107,12 +108,14 @@ async def main_async():
     from src.services.core.telegram_service import TelegramService
     from src.services.core.media_lock import MediaLockService
     from src.services.core.media_sync import MediaSyncService
+    from src.repositories.queue_repository import QueueRepository
 
     scheduler_service = SchedulerService()
     posting_service = PostingService()
     telegram_service = TelegramService()
     lock_service = MediaLockService()
     settings_service = SettingsService()
+    queue_repo = QueueRepository()
 
     # Media sync runs unconditionally — the loop iterates per-chat and skips
     # chats with media_sync_enabled=False. No env gate.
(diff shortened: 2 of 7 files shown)


chrisrogers37 (commented):
**Verdict: Mechanical fixes** (comment-only — same-identity blocks Approve)

The prune logic is correct, the scheduling is correct, the tests verify what they claim. One gap in the implementation: the PR description says `record_heartbeat("queue_cleanup")` lets the health server detect a stuck loop — but `queue_cleanup` is absent from `LOOP_EXPECTED_INTERVALS` in `heartbeat.py`, so `get_loop_liveness()` never iterates it. The health server is blind to this loop.

---

**Finding 1 [major]: `queue_cleanup` missing from `LOOP_EXPECTED_INTERVALS`**

`heartbeat.py` defines which loops the health check monitors via `LOOP_EXPECTED_INTERVALS`. Every other loop added to `tasks` has a corresponding entry. `queue_cleanup` does not.

Consequence: `get_loop_liveness()` will never report on this loop. A stuck or crashed `queue_cleanup` will not produce a 503 from the health endpoint. The PR description's claim ("so the health server can detect a stuck loop") is not satisfied as written.

Fix — one line in `heartbeat.py`:

```python
LOOP_EXPECTED_INTERVALS: dict[str, int] = {
    "scheduler": 60,
    "lock_cleanup": 3600,
    "cloud_cleanup": 3600,
    "media_sync": 300,
    "transaction_cleanup": 60,
    "queue_cleanup": 3600,   # add this
}
```

Mutation test: remove `record_heartbeat("queue_cleanup")` from the loop — the existing tests still pass (none of them assert on health-server visibility). The omission is live-invisible until a health check is done against a deployed instance with the loop stuck.

---

**Finding 2 [minor]: No `finally: cleanup_transactions()` — confirmed not needed**

`cleanup_locks_loop` has a `finally` block calling `lock_service.cleanup_transactions()`. The new loop has no equivalent. `BaseRepository` has no `cleanup_transactions` method, so this is not a gap — just noting the asymmetry is intentional and correct.

---

**Finding 3 [info]: First prune fires one hour after deploy**

The loop records a heartbeat then sleeps before pruning — identical to `lock_cleanup_loop`. On a fresh deploy with a stale-row backlog (the exact scenario this PR targets), you wait one full hour before the first cleanup fires. This matches the existing convention in the codebase, so not requesting changes — but worth knowing for the post-deploy monitoring checklist item in the PR description: the first log line won't appear until ~60 minutes after deploy, not immediately.

---

Everything else is clean: the `delete_stale` query is correct (all statuses, `scheduled_for < cutoff`), the test cases are real (mock returns stale items, asserts `db.delete` called for each, asserts commit fires — removal of the method would break both tests), the `guarded` wrapper provides crash isolation and restart budget, no scope creep, naming is consistent with existing loops.

Ship it after adding `"queue_cleanup": 3600` to `LOOP_EXPECTED_INTERVALS`.

chrisrogers37 (commented):
**Request Changes** (comment-only — same-identity blocks formal Request Changes)

**[major] `queue_cleanup` is missing from `LOOP_EXPECTED_INTERVALS` in `heartbeat.py`.**

The PR description states the new loop "integrates with the health server via `record_heartbeat`." Half of that is true: `record_heartbeat("queue_cleanup")` is called inside the loop. The other half is not: `get_loop_liveness()` iterates `LOOP_EXPECTED_INTERVALS` to build the health response, and `queue_cleanup` is not in that dict. The health endpoint is blind to this loop — a stuck prune loop produces no 503.

Fix: add one entry to `LOOP_EXPECTED_INTERVALS`:
```
"queue_cleanup": 3600,
```

Mutation test confirms this is not covered: removing the `record_heartbeat` call leaves all existing tests green. The health integration has no test, and the one path that would catch a stuck loop has no signal.

**Core logic is sound** — the prune query (all statuses, `scheduled_for < cutoff`), the 3600s interval, the `guarded` wrapper for crash isolation, and the existing repository-call tests are all correct. The tests would catch removal of the actual prune call. One-liner fix to `heartbeat.py` and this is ready to merge.

**[info] First prune fires ~60 min after deploy** (sleep-before-prune, inherited from `lock_cleanup_loop`). Not a regression — just worth noting in the post-deploy monitoring note.

chrisrogers37 (commented):

Fix confirmed. `"queue_cleanup": 3600` is now in `LOOP_EXPECTED_INTERVALS` — the health server will report on this loop and a stuck prune loop will produce a 503. `test_all_registered_loops_included` now includes `"queue_cleanup"`, so the registry check would have gone red without the heartbeat entry and will stay red if the entry is removed in future. Both sides of the gap are closed.

Prior request-changes concern is fully resolved. Ship it.

PR #494, opened 2026-06-08
fix: enforce daily posting cap across all 5 posting paths

chrisrogers37:
## Summary

- **Bug:** `posts_per_day` was only used for interval spacing — never enforced as a hard daily limit. DB evidence: account with limit=15 posted 33 times in one day.
- **Fix:** Added a shared `can_post_today()` guard backed by timezone-aware `HistoryRepository.count_posts_today()`, wired into all 5 independent posting paths: scheduler, autopost, manual posted callback, `/next` command, and auto-approve.
- On cap hit, queue items stay pending (retry next day). Users see a clear "Daily limit reached" message on interactive paths.

## Paths guarded

| # | Path | Location |
|---|------|----------|
| 1 | Scheduler | `scheduler.py:process_slot` |
| 2 | Autopost | `telegram_autopost.py:_do_autopost` |
| 3 | Manual posted | `telegram_callbacks_core.py:_execute_complete_db_ops` → raises `DailyCapReachedError` |
| 4 | /next command | `scheduler.py:force_send_next` |
| 5 | Auto-approve | `scheduler.py:_auto_approve` |

## New files

- `src/services/core/daily_cap.py` — single source of truth guard function
- `tests/src/services/test_daily_cap.py` — 12 unit tests (timezone, boundary, fallback)

## Test plan

- [x] 12 new unit tests for `can_post_today()` and `count_posts_today()`
- [x] 190 existing tests pass (no regressions)
- [x] Timezone edge cases: UTC fallback, invalid timezone, correct day boundary conversion
- [ ] Manual verification: set `posts_per_day=1`, post once, confirm second attempt is blocked on all 5 paths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Daily posting cap enforced across all 5 posting paths** — `posts_per_day` was only used for interval spacing, not as a hard daily limit. DB evidence showed accounts posting 2x their configured cap. Added a shared `can_post_today()` guard (backed by `HistoryRepository.count_posts_today()` with timezone-aware day boundaries) and wired it into the scheduler, autopost, manual posted callback, `/next` command, and auto-approve paths. Catchup bursts are also capped. On cap hit the queue item stays pending so it retries the next day.
+
 - **Telegram "Auto Post" button no longer spins after timeout** — `handle_autopost` did not call `query.answer()` until the full Cloudinary upload + Meta publish completed (often >30s), past Telegram's callback-answer deadline. Result: the button's loading spinner ran indefinitely, the bot logged `Could not answer callback query (may be stale)`, and the user couldn't tell if their click registered. Now answers immediately with `⏳ Posting…` after the cheap lock-held check (which keeps its specific `⏳ Already processing…` message for duplicate clicks).
 
 ### Removed
--- a/src/repositories/history_repository.py
+++ b/src/repositories/history_repository.py
@@ -187,6 +187,52 @@ class HistoryRepository(BaseRepository):
         self.end_read_transaction()
         return result
 
+    def count_posts_today(
+        self,
+        chat_settings_id: Optional[str] = None,
+        posting_timezone: Optional[str] = None,
+    ) -> int:
+        """Count successful posts in the current calendar day.
+
+        Uses the chat's posting_timezone to determine day boundaries.
+        Falls back to UTC if timezone is None or invalid.
+
+        Args:
+            chat_settings_id: Tenant to filter by
+            posting_timezone: IANA timezone for day boundary calculation
+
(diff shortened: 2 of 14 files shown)


chrisrogers37 (commented):
**Verdict: Ship it**

5 paths covered, timezone logic correct, exception semantics preserved. One test coverage gap worth a follow-up but not a blocker.

**Path-by-path trace:**

1. **Scheduler (process_slot)** — guard fires before `is_slot_due()`, so no unnecessary slot evaluation on a capped day. Correct placement.

2. **Autopost (_do_autopost)** — guard fires after the duplicate-click lock check and before the safety check. The UI message (`edit_message_caption`) is appropriate for the inline keyboard context.

3. **Manual posted (_execute_complete_db_ops)** — correctly raises `DailyCapReachedError` only when `status == "posted"`. Skip and reject always pass through. The caller (`_do_complete_queue_action`) catches it, restores queue item to "pending" (correct — retry next day), and shows the user message.

4. **/next (force_send_next)** — guard returns `{"reason": "daily_cap_reached"}`. `telegram_commands.py:handle_next` handles this reason with a user-facing message. Full round-trip confirmed.

5. **Auto-approve (_auto_approve)** — redundant with process_slot (which guards first), but the comment acknowledges this and the defensive depth is correct given the method's complexity.

**Exception handler in telegram_callbacks_queue.py:**

`except OperationalError as e:` → `except Exception as exc:` with explicit `if not isinstance(exc, OperationalError): raise`. Semantically equivalent for all current callers — non-OperationalError/non-DailyCapReachedError exceptions still propagate, `BaseException` subclasses (KeyboardInterrupt, SystemExit) are unaffected. No regression.

**Timezone handling:**

`datetime.now(tz).replace(hour=0, minute=0, second=0, microsecond=0).astimezone(timezone.utc)` is the correct pattern. Fresh `now` on every call — no stale caching. Invalid timezone falls back to UTC without crashing. UTC-offset users get correct local midnight boundary. ✓

No upper bound on `posted_at` in the query — not a bug. `today_start_utc` is always in the past (or present), so there are no future posts to overcapture.

**One gap — TestCountPostsTodayTimezone is smoke-only:**

The three tests in that class (`test_uses_utc_when_no_timezone`, `test_uses_provided_timezone`, `test_falls_back_to_utc_on_invalid_timezone`) verify that the function doesn't crash and returns the mocked scalar. They do not assert the actual `today_start_utc` value passed to the filter. A mutation that always uses UTC for the boundary would pass all three. The logic is correct in the implementation, but the test class doesn't prove it.

The `test_passes_timezone_to_repo` test in `TestCanPostToday` covers timezone forwarding from `can_post_today()` to `count_posts_today()`, which is the higher-leverage assertion anyway. The gap is a follow-up, not a blocker.

Mutation test: set `posts_per_day=1` in a `chat_settings` mock and `count_posts_today.return_value=1` — `can_post_today()` returns False. Every path guard that returns early on False would block. The `test_blocks_when_at_limit` test catches this exactly.

PR #508, opened 2026-06-09
fix: paginate Cloudinary cleanup_expired past the 500-result cap (#499)

MohammadYusif:
## Summary

- `CloudStorageService.cleanup_expired()` (`src/services/integrations/cloud_storage.py`) listed orphaned Cloudinary uploads with `max_results=500` but never consumed the `next_cursor` that Cloudinary returns when more results exist.
- Once a folder accumulated more than 500 uploads, the cleanup silently processed only the first page. The remaining orphans were never examined and accrued storage cost indefinitely.

## Changes

- `src/services/integrations/cloud_storage.py`: wrap the `cloudinary.api.resources()` call in a loop that follows `next_cursor` until every page is exhausted, so all expired uploads are processed regardless of folder size. Extracted the page size to a named module-level constant `CLOUDINARY_RESOURCES_PAGE_SIZE = 500` (the API's hard ceiling).
- `tests/src/services/test_cloud_storage.py`: added `test_cleanup_expired_paginates_past_first_page`, a regression test using a two-page mock response (page 1 returns a `next_cursor`, page 2 does not). It asserts both pages' expired uploads are deleted, that the second request forwards the cursor from the first, and that `api.resources()` is called twice. Verified the test fails against the pre-fix code (`assert 1 == 2` — only the first page is processed) and passes with the fix.

All 32 tests in `test_cloud_storage.py` pass; `ruff check` and `ruff format --check` are clean; CHANGELOG.md updated under `[Unreleased] > Fixed`.

Fixes #499


chrisrogers37 (commented):
**Approve** (comment-only — same-identity blocks formal Approve; external contributor — Chris should merge after reviewing this verdict)

---

The fix is correct and the regression test is solid. Verified the full SDK call chain.

**Pagination logic: correct.** The while-True / break-on-no-cursor pattern is the canonical approach for this API. The `next_cursor=None` on the first iteration is safe: I traced the call path through `cloudinary.api.resources()` → `__list_resources_params()` → `only()` → `normalize_params()`. The `only()` helper includes a key only if it's present in the dict (`if key in source`), so `next_cursor=None` IS included — but `normalize_params()` then strips it (`if v is not None`). Net result: the first request goes out clean with no `next_cursor` param, exactly as required.

**Test: mutation-proof.** The PR description says the test fails against pre-fix code (`assert 1 == 2` — only page 1 processed). The test asserts `call_count == 2`, verifies the cursor value forwarded in the second call, and checks both `public_id`s were destroyed. Removing the pagination loop would fail all three assertions. Not a decoy.

**Scope: clean.** Three files, no dependency additions, no pyproject.toml or requirements.txt changes. The constant extraction (`CLOUDINARY_RESOURCES_PAGE_SIZE = 500`) is good hygiene — it makes the hard ceiling self-documenting. No unrelated files touched.

**Security: nothing suspicious.** No new env var access, no new network calls beyond the existing Cloudinary SDK usage, no shell exec, no new imports.

**One minor note (not blocking):** The new test uses `datetime.utcnow()` (line ~467 in the test file), which is deprecated in Python 3.12+. The existing tests in the same class already use this pattern (lines 399, 403, 424), so the contributor is matching the codebase convention. The pre-existing inconsistency with the production code's `datetime.now(timezone.utc)` is worth a follow-up cleanup pass, but it's not a regression introduced here.

**No architectural concerns, no security concerns, no logic errors.** Fix matches issue #499 exactly.

chrisrogers37 (commented):

**External contributor (@MohammadYusif) — Chris should review this verdict and merge manually.**

**Correctness:** Pagination logic is correct. The loop forwards `next_cursor` on each iteration; passing `next_cursor=None` on the first call is safe — the Cloudinary SDK's `normalize_params()` strips `None` values before the HTTP request, so the first call is clean. The constant extraction (`MAX_CLOUDINARY_RESULTS = 500`) is good hygiene.

**Mutation test:** Three independent assertions would all fail if the loop were removed: `call_count == 2`, cursor value forwarded in second call, and both `public_id`s destroyed. Not a decoy.

**Scope:** Exactly 3 files touched (source, test, changelog). No dependency additions. No new env var access, no new network calls outside the existing Cloudinary SDK, no shell exec. Fix precisely matches issue #499.

**Security:** None. No concerns.

**Minor (not blocking):** New test uses `datetime.utcnow()` (deprecated Python 3.12+), but this matches the existing pattern in the same test class. Pre-existing inconsistency, not introduced by this PR — worth a follow-up cleanup.

Ready to merge after Chris reviews.

PR #509, opened 2026-06-12
fix(autopost): restore queue row to pending on daily-cap hit

chrisrogers37:
## What & why

On the autopost (Instagram API) path, a daily-cap hit left the queue card stranded in `processing`. `handle_autopost` atomically claims the card into `processing` (`telegram_autopost.py:148`), then the cap guard inside `_do_autopost` (the spawned background task) returned **without releasing the row** — so a capped tap left the row stuck in `processing`. It would not re-post and wasn't reclaimed until the 24h `discard_abandoned_processing` sweep.

The autopost path was the **only** posting path that neither gated the cap before claiming nor restored the row after. The closest sibling — the manual Posted/Skip/Reject handler — already claims then restores to `pending` on cap-hit (`telegram_callbacks_queue.py:91-98`).

## The fix (restore-only)

`_do_autopost` now restores the claimed row to `pending` on a cap-hit (`update_status(queue_id, "pending")`) before showing the cap message — reusing the exact restore the manual handler uses. The cap check stays where it already ran (one `count_posts_today` COUNT, in `_do_autopost`). Cap **semantics are unchanged**: the cap still blocks the post and shows the same "Daily posting limit reached" message.

## Design note — why restore-only and not a pre-claim gate

The original plan (and dispatch) was *gate the cap before the claim* **and** restore. I dropped the gate after a concurrently-filed regression doc (`shared/knowledge/storydump/hang-perf-regression-2026-06-12.md`, astrid) showed why it's the wrong move here:

- storydump runs the PTB poller + scheduler + **synchronous** SQLAlchemy on **one asyncio loop** with `concurrent_updates` off — so every blocking DB call freezes *all* in-flight button clicks. The `#494` `count_posts_today` COUNT on the most-clicked button is already a named hang amplifier.
- A pre-claim gate would add a **second** blocking COUNT to that exact hot path.
- A *gate-only* variant (drop the `_do_autopost` check to keep one COUNT) would let a concurrent post cross the cap during the claim→upload window → an **over-cap post** (violates cap semantics).
- **Restore-only** keeps one COUNT, fixes the orphan, keeps cap semantics, and adds **zero** new blocking work. It's also literally the `telegram_callbacks_queue.py:93` pattern the task asked to *consolidate to, not fork* — the gate was the fork (scheduler pattern, wrong context for a button tap).

If we still want gate-before-claim, the place for it is **after** astrid's DB-offload lands (so the COUNT is non-blocking) — happy to do it as a follow-up.

## ⚠️ Scope flag — this does NOT fix the `Queue item not found` log

This fixes the autopost-path **orphan**. It does **not** resolve the reported scheduler-path `Queue item not found` (`telegram_notification.py:79`). Autopost-button rows carry a `telegram_message_id`, and both stale-row sweepers filter on `telegram_message_id IS NULL` — so this orphan is **skipped by the sweepers and never deleted**. The `:79` miss is a `get_by_id` returning `None` (row *deleted*), which traces to the pre-existing `#400` stale-pending sweeper race on msg_id-NULL rows. That needs a separate fix. Full analysis: `shared/knowledge/storydump/queue-item-not-found-regression-2026-06-12.md`.

## Follow-ups (not in this PR)

- **Same orphan bug, a 6th claim site:** `telegram_callbacks_admin.py:74-82` routes through `_execute_complete_db_ops` (which raises `DailyCapReachedError`) but its `except Exception` swallows it, leaving the row in `processing`.
- **Deeper generalization:** make `claim_for_processing` cap-aware (check inside the `FOR UPDATE` txn, return `None` on cap-hit) — collapses claim+check+restore and closes the TOCTOU window across all paths.
- **The actual `:79` symptom:** the `#400` stale-pending sweeper race (coordinate with astrid's hang/perf work — same `#400`/`#494` substrate).

## Testing

TDD — restore test written first, confirmed **RED on main** (`update_status` called 0 times) → **GREEN** after the fix:
- `test_do_autopost_cap_hit_restores_row_to_pending` — cap-hit restores the claimed row to `pending`, still shows the limit message.
- `test_do_autopost_under_cap_does_not_restore` — guard: under the cap, no restore churn.

Verification:
- Full service suite: **passing** (DB integration tests skip without Postgres).
- `/simplify` (4 angles): change confirmed at the right altitude.


chrisrogers37 (commented):

All CI green (Test passed). Empirical verification across all 5 scrutiny points:

---

**1. Restore-only consolidates to the sibling pattern — not a new fork.**

`telegram_autopost.py:229` adds `self.service.queue_repo.update_status(queue_id, "pending")` inside the cap-hit branch of `_do_autopost`. Byte-for-byte the same call `telegram_callbacks_queue.py:93` already makes on `DailyCapReachedError`. Consolidation confirmed.

**2. TDD: test genuinely reproduces the orphan and is mutation-safe.**

`conftest.py:107` sets `posts_per_day = 99` on the default `_default_cs` mock. The test overrides `count_posts_today.return_value = 99` — hitting the cap exactly (`99 >= 99` → `can_post_today` returns False). On main without the fix, `update_status` is never called → `assert_called_once_with` fails → RED. With fix → GREEN. Mutation kills.

The under-cap guard test (`count_posts_today=0`, safety check forced `safe_to_post=False`) exits before `update_status` → `assert_not_called` confirmed.

**3. Cap semantics preserved.**

`update_status` runs inside the cap-hit branch. `edit_message_caption` + end-of-block `return` still fires. Post does not proceed. Cap enforces correctly.

**4. No new blocking work on the hot path.**

The `UPDATE` is strictly inside the cap-hit branch. Normal under-cap flow — the hot path on the autopost button — is completely untouched. The one `COUNT` from `can_post_today` already existed; this adds nothing to it.

**5. Scope correct.**

3 files changed: `telegram_autopost.py` (+4 lines), test file (+65 lines, two tests), `CHANGELOG.md`. No touches to `telegram_notification.py:79`, the sweeper, or `#400` substrate.

chrisrogers37 (commented):

**Does the fix actually restore the row?**

Yes. The four-line addition in `_do_autopost` (`telegram_autopost.py:229-232`) calls `self.service.queue_repo.update_status(queue_id, "pending")` on the exact branch where `can_post_today` returns `False`, before the user-facing caption edit. The row was claimed into `processing` by `claim_for_processing` in `handle_autopost` (synchronous path), then transferred to the background task via `_autopost_background`. Without this restore, `_do_autopost` returned silently from the cap branch while the row sat in `processing` indefinitely. The fix is in the right place.

**Is "pending" the right terminal state?**

Yes. `pending` is the only status that puts the row back in the work queue for the next day's scheduler sweep. The CHANGELOG and description are accurate: "the card retries the next day instead of being orphaned." The existing manual handler (`telegram_callbacks_queue.py:91-98`) uses the identical restore path after a `DailyCapReachedError` — the PR is consolidating to that pattern, not forking from it.

**Mutation test — are the tests decoys?**

No. `test_do_autopost_cap_hit_restores_row_to_pending` asserts `update_status.assert_called_once_with(queue_id, "pending")`. If you revert the four-line addition, the `update_status` mock receives zero calls and the test fails. The test is not a decoy — it would have been RED on main exactly as described, and it verifies the right call with the right arguments.

`test_do_autopost_under_cap_does_not_restore` is a sound guard: it blocks the cap branch from being hit (count=0, well under the cap=99 default), then fails the safety gate to exit early, and asserts `update_status.assert_not_called()`. This prevents the restore logic from churn-calling on the happy path. Both tests are precise and non-overlapping.

**Error handling — what if `update_status` itself raises?**

If this raises (e.g., DB connectivity failure), the exception propagates into `_autopost_background`'s outer `except Exception` handler. That handler logs the error and the `finally` block releases the lock and cleans up operation state. A failed restore means the row stays in `processing` and falls back to the 24h `discard_abandoned_processing` sweep — same behavior as before this PR. A follow-up could wrap the restore in its own try/except and log a warning, but it is not a blocker: the sweeper is the backstop.

**Scope**

None. Three files changed: the fix (4 lines), two tests (66 lines), CHANGELOG. No unrelated diffs.

**Follow-ups noted but not blocking:** (1) `telegram_callbacks_admin.py` sixth claim site where `except Exception` swallows `DailyCapReachedError`; (2) making `claim_for_processing` cap-aware inside the `FOR UPDATE` transaction to close the TOCTOU window. Both correctly deferred.

Issue #513, opened 2026-06-16
[security] OAuth /start endpoints are unauthenticated and state tokens are replayable — account-binding CSRF


chrisrogers37:
## Summary

The `GET /instagram/start` and `GET /google-drive/start` OAuth endpoints take `chat_id` as an **unauthenticated** query parameter (only a rate-limit decorator) and mint an OAuth `state` bound to that attacker-chosen `chat_id`. The completing user's tokens are then stored under whatever `chat_id` the initiator picked — so an attacker can bind *their own* Instagram or Google Drive account to a *victim's* chat. Separately, the `state` token is not single-use: a random `nonce` is generated but never persisted or consumed, so any captured `state` validates repeatedly within its 10-minute TTL.

## Evidence

`src/api/routes/oauth.py:18-32` — no auth, attacker-chosen `chat_id`:

```python
@router.get("/instagram/start")
@limiter.limit("5/minute")
async def instagram_oauth_start(request: Request, chat_id: int = Query(...)):
    with OAuthService() as oauth_service, service_error_handler():
        auth_url = oauth_service.generate_authorization_url(chat_id)  # state = enc({chat_id, nonce})
        return RedirectResponse(url=auth_url)
```

- `src/api/routes/oauth.py:192-206` — `/google-drive/start` has the same shape.
- The callback validates `state` and stores tokens for the embedded `chat_id`: `oauth.py:81-83` (`chat_id = oauth_service.validate_state_token(state)`) → `exchange_and_store(code, chat_id)`. The state binds the flow to a *chat_id the caller chose*, not the authenticated user who grants consent — so the comment "CSRF protection" overstates what it provides.
- Contrast the authenticated path `src/api/routes/onboarding/setup.py:58-80` (`/api/onboarding/oauth-url/{provider}`) which gates the same call behind `_validate_request(init_data, chat_id)`. Two parallel start paths exist; only one is guarded (a "consolidate, don't fork" smell).
- Replay: `src/services/core/oauth_service.py:99-137` puts `"nonce": secrets.token_hex(16)` in the state but a repo-wide grep for the nonce being stored/consumed (`used_state`/`consumed`/`jti`/replay) returns nothing — `validate_state_token` only reads `chat_id`. Replay protection is Fernet's `STATE_TTL_SECONDS = 600` alone; within 10 min the same `state` validates unlimited times.

[...]

PR #519, opened 2026-06-26
fix(security): cross-tenant data isolation — Mini App API IDOR (#511) + Telegram callback queue scoping (#512)

chrisrogers37:
## Summary

Data-engineering slice of the security remediation wave: closes the two **critical cross-tenant data-isolation** holes from the 2026-06-16 audit. Both are "data-exposure / query-safety" findings — the right lens for this PR. The role-based authorization half of #512 and the auth/frontend findings (#513, #514, #518) are flagged for a software-eng follow-up (see below), not forced in here.

- **Closes #511**
- **Partially addresses #512** — cross-tenant queue scoping only; the member-vs-admin authorization gate remains (backend-auth, flagged).

> ⚠️ **Held for @chrisrogers37 to merge** — do not auto-merge.

---

## #511 — Mini App API cross-tenant IDOR (Closes)

**Evidence (traced):** `_validate_request` (`src/api/routes/onboarding/helpers.py`) is the only authz gate for ~35 onboarding/dashboard/settings endpoints. It authenticated the caller (HMAC over initData / signed URL token) but only compared the token's `chat_id` to the request `chat_id` *when the token carried one*. A **DM-launched** Mini App initData has no `chat` field, so `signed_chat_id is None`, the guard was skipped, and any authenticated bot user could read/mutate **any other tenant's** queue, history, accounts, media, and settings by passing a different `chat_id`. The membership model existed (`MembershipRepository`) but was enforced in exactly one endpoint (`/audit-log`) — and that check was **fail-open** (`if user:` skipped it for unknown users).

**Fix:**
- New `MembershipService.is_active_member(telegram_user_id, telegram_chat_id)` resolves user → chat_settings → membership and is **fail-closed** at every missing link (keeps the API→Service→Repository layering — no direct repo calls from the API glue).
- `_validate_request` now: **bound token** (signed URL / group-launched initData) → keep the existing cryptographic `chat_id` equality check (the binding *is* the authorization); **unbound token** (DM initData) → require an active membership for the request `chat_id`.
- The fail-open inline `/audit-log` check is folded into this central fail-closed path (and its now-unused imports removed).

**Why narrowed to the unbound path:** signed URL tokens and group-launched initData already bind `chat_id` and are server-issued only to legitimate users of that chat, so the equality check is sufficient there. Restricting the membership lookup to unbound tokens closes the actual exploit (DM initData + foreign `chat_id`) **without** regressing the browser-dashboard / group-launch / legacy "Branch-2" group-setup flows, which all use bound URL tokens.

**Known residual edge (documented, not a regression of current behavior):** a *legacy* pre-multitenancy tenant whose users have no membership row would be 403'd if they open the Mini App via a DM WebAppInfo button. In-group interaction self-heals (`_ensure_membership`). storydump's Neon DB wasn't identifiable in the org's project list, so I did **not** guess at a prod DB to size this; if prod shows orphaned tenants, a one-time membership backfill should be proposed (a data migration — propose-first, not bundled here).

---

## #512 — Telegram callback queue ops, data-layer subset (partial)

**Evidence:** `resume:clear`, `clear:confirm`, and the resume `reschedule` path called `queue_repo.get_all(status="pending")` with **no tenant filter** and deleted/rescheduled **every tenant's** pending rows. `resume:clear` / `clear:confirm` are parameter-free, so any user could trigger a fleet-wide pending-queue wipe. `batch_approve` scoped to a `chat_settings_id` taken straight from (forgeable) `callback_data`.

**Fix (`src/services/core/telegram_callbacks_admin.py`):**
- New `_caller_chat_settings_id(query)` resolves the calling chat's tenant; the destructive handlers now scope `get_all(...)` to it and **refuse to run** (rather than fall back to an unscoped query) when the chat has no tenant.
- `batch_approve` rejects a `chat_settings_id` in button data that doesn't belong to the calling chat.

**Flagged for software-eng follow-up (NOT in this PR):** the member-vs-admin **role authorization gate** for the callback dispatcher (#512's primary framing) is backend-auth — default-deny role classification + `get_chat_member` — out of the data-eng lens.

---

## Flagged for software-eng follow-up (not touched here)

| Issue | Why deferred |
|------|--------------|
| #512 (role gate) | Backend-auth: member/admin RBAC on the callback dispatcher |
| #513 | OAuth `/start` auth + replayable state tokens (CSRF) |
| #514 | Next.js `16.1.6 → 16.2.9` frontend upgrade (middleware/CSRF advisories) |
| #518 | Defense-in-depth: JWT alg pin, dev-bypass, timing-safe compare, thumbnail SSRF, error masking |

**Recommended deeper follow-up (defense-in-depth):** make `chat_settings_id` required on `QueueRepository.get_all` / `get_all_with_media` (or split out explicit `*_across_tenants()` methods). Today an unscoped `get_all(status="pending")` returns every tenant's rows by construction, so tenant safety relies on each caller remembering the kwarg. After this PR there is no legitimate unscoped consumer, so enforcing it at the repo layer would make this class of leak impossible — but it touches the repo signature + all callers + tests, wider than a security PR should carry.

---

## Verification

- `ruff check src/ cli/` — clean; `ruff format src/ cli/ --check` — clean (CI scope).
- **Full unit suite: 1896 passed, 10 skipped** (integration tests need the CI Postgres service).
- New tests:
  - `tests/src/services/test_membership_service.py` — fail-closed at every link (none/unknown user, unbootstrapped chat, missing/inactive membership, active member).
  - `tests/src/api/test_helpers.py::TestValidateRequestAuthorization` — bound-match (no membership lookup), bound-mismatch → 403, unbound active member → pass, unbound non-member → 403, unbound unknown-user → fail-closed 403.
  - `tests/src/services/test_telegram_callbacks_admin.py::TestCrossTenantIsolation` — resume/reset scope to caller tenant, bail (no unscoped query) when tenant absent, `batch_approve` rejects foreign / allows own `chat_settings_id`.
- Existing `test_telegram_callbacks.py::TestBatchApprove` updated for the new tenant check; CHANGELOG updated under `### Security`.


chrisrogers37 (commented):
**Approve** (comment-only — same-identity blocks formal Approve; held for Chris merge per PR)

**#511 — IDOR gate: closed, fail-closed.**

Traced the full unbound-token path. Old path: DM initData → `signed_chat_id is None` → no check → `return user_info` with any `chat_id` accepted. New path: `is_active_member(user_id, chat_id)` is a server-side lookup before any data is accessed.

`is_active_member` fails closed at every link: None user_id → False; None chat_id → False; unknown user → False (repo not reached further); unbootstrapped chat → False; missing membership → False; inactive membership → False. A DM-launched attacker replaying against a victim's `chat_id` has no membership row for that chat → False → 403. No residual IDOR.

Fail-open audit-log check correctly folded: the old `if user:` skip (unknown user bypassed membership entirely) is gone. Unknown user now hits `is_active_member` → `get_by_telegram_id` returns None → False → 403. Fail-closed.

Bound-token path unaffected: signed URL / group-launched initData set `signed_chat_id`; the equality check fires, not the membership branch. The fix doesn't touch that path. ✓

Non-blocking: `is_active_member` docstring says "Never raises" but has no try/except. A DB failure propagates as 500 — access is denied (fail-closed by effect), but the doc claim is aspirational. Not a security concern.

**#512 (partial) — cross-tenant queue-delete scoping: complete for all paths touched.**

Three destructive handlers, all fixed:

- `_do_resume_callback` (clear + reschedule): resolves caller tenant → bail if None → `get_all(..., chat_settings_id=cs_id)`. Both action branches use the same scoped result. ✓
- `handle_reset_callback` (confirm): same pattern. ✓
- `handle_batch_approve`: resolves caller tenant → rejects if None or if button-data `cs_id` doesn't match caller's. Foreign `cs_id` in button data → rejected before any rows touched. ✓

The "never fall back to unscoped query" contract is enforced: `_caller_chat_settings_id` returns None when no settings row, and every caller checks None and returns early.

**Does not over-claim #512.** PR title says "partial #512"; description and CHANGELOG both explicitly name the role-gate (member vs admin) as deferred to a software-eng follow-up. No over-claim. ✓

**Tests — mutation analysis:**

`TestIsActiveMember` (7): each test checks one fail-closed link. Removing any guard breaks its test.

`TestValidateRequestAuthorization` (5): `test_unbound_token_non_member_is_rejected` asserts HTTPException 403 — remove the membership block → no exception → fails. `test_unbound_token_active_member_passes` asserts `is_active_member.assert_called_once_with(1, CHAT_ID)` — remove the block → not called → fails. Tests are not decoys.

`TestCrossTenantIsolation` (6): `test_resume_clear_scopes_to_caller_tenant` asserts `get_all.assert_called_once_with(status="pending", chat_settings_id="cs-A")` — drop the kwarg → fails. `test_resume_bails_when_chat_has_no_tenant` asserts `get_all.assert_not_called()` — remove the bail → called with None → fails. All catches are real.

`_authorize_membership_by_default` autouse fixture: patches MembershipService for all API tests. For bound-token tests the `if signed_chat_id is None:` block is never entered — patch is inert, no regression.

Ready to merge when Chris reviews.

PR #561, opened 2026-07-14
fix(queue): expire sent rows gracefully at reap instead of orphaning buttons (#560)

chrisrogers37:
## What & why

Fixes the **"⚠️ Queue item not found"** bug (Epic #560): a story posted to the group shows a photo card with live inline buttons (Posted / Skip / Reject); ~a day later, tapping a button rewrites the card to "Queue item not found" and strips the buttons.

**Root cause** (full investigation + weighing on #560): the 24h age-based reapers (`delete_stale`, `discard_abandoned_processing`) hard-delete a **button-bearing** queue row (already sent to Telegram) without editing the card or writing history. A later tap resolves the now-gone UUID, finds no history, and hits the bare "not found" branch in `telegram_utils.validate_queue_item`.

## The fix — Chris's locked target: "C-edit + B", no dangling buttons

A single **shared reap-behavior** (`src/services/core/queue_reap.py::expire_sent_row`) that both reap paths route button-bearing rows through:

- **C-edit:** proactively edit the card to **"⌛ Expired — no action needed"** and **strip the buttons** — the same edit the click-handler does, performed at reap time (best-effort, via the existing `telegram_edit_with_retry` wrapper).
- **B:** write a terminal **`expired`** `posting_history` row (idempotent) — an audit trail **and** a tap-time fallback: a late tap on a reaped card now shows the friendly "Expired" caption (via the existing `_build_already_handled_caption` history branch) instead of "Queue item not found".
- **Race handling** (best-effort Telegram I/O in a bulk path): transient edit failure → **defer** (row left intact & tappable, retried next sweep — never orphaned); terminal "message gone / too old / forbidden" → proceed to record + delete.

Both age-based reapers are **narrowed to `telegram_message_id IS NULL`** so they only sweep never-sent accumulation and can no longer orphan a live card; button-bearing rows go through `expire_sent_row` first.

### Scope — deliberately NOT a reaper merge
Per #560, the architectural consolidation of the two reapers into one policy is **deferred** (it overlaps the future claim-lease direction, tracked under #560). This PR consolidates only at the **behavior** level: one shared method both reapers call (no drift), designed so a future claim-lease reaper can reuse the exact same method. astrid (queue-simplification owner) was given a heads-up.

## ⚠️ MIGRATION — must be applied to prod

`scripts/migrations/042_add_expired_history_status.sql` adds `'expired'` to the `posting_history.status` CHECK constraint. **This must run on the prod DB.** A merged-but-unapplied migration = new code writing `status='expired'` against the old constraint → `IntegrityError` on every reap. Apply with:

```bash
psql "$DATABASE_URL" -f scripts/migrations/042_add_expired_history_status.sql
```

## Validation (ran, not claimed)

TDD — tests written first (red: `ModuleNotFoundError`), then implemented to green. Full suite: **2035 passed, 16 skipped** (the skips are DB-integration tests that `conftest.py` skips without Postgres). The reap behavior is proven end-to-end in `tests/src/services/test_queue_reap.py`:

- `test_happy_path_edits_records_and_deletes` — card → "Expired" + buttons stripped + `expired` history written + row deleted.
- `test_transient_failure_defers_without_side_effects` — transient edit failure leaves the row (no orphan).
- `test_terminal_edit_error_still_records_and_deletes` — message-gone → still records + deletes.
- `test_idempotent_history_skips_create_but_still_deletes`.
- `test_expired_history_yields_friendly_caption` — a late tap → "Expired", not the scary string.

Per storydump's safety rules, validation is via tests only (never the live scheduler / posting path).

## /simplify applied
Ran the 4-lens cleanup pass. Applied: deduped the `EXPIRED_CAPTION` literal into `telegram_utils` (the presentation SSOT), so the reaper and the tap-time caption share one constant. The review also confirmed the design altitude is right — the module-level function with injected deps is transaction-correct (a `BaseService` would fracture the standalone-session model the scheduler loop relies on) and is the intended reusable seam for a future lease reaper.

### Deferred follow-ups (surfaced by /simplify, kept out of this tight PR)
- Share the `HistoryCreateParams` builder with `TelegramCallbackCore._create_history_params` (touches the live callback path — better as its own reviewed change).
- Bound the per-sweep reap with a `limit` (guards tick latency during a one-time backlog catch-up after downtime; low severity — posting is already paused during such catch-up).
- Commit the reap's history-write + queue-delete in one transaction (currently idempotent-safe across two commits via the `get_by_queue_item_id` guard).

## Reviewers
ari (manager) + astrid (queue-area owner). Per policy, **I will not merge my own PR** — this is ready for review; a human merges.


chrisrogers37 (commented):
**Verdict: Request Changes**

Verified by running, not by reading — two independent passes (a code-correctness audit + a test-execution/mutation-testing pass), then I personally re-confirmed the one finding that matters most for the verdict once both were done.

**1. Correctness of `expire_sent_row`** — confirmed accurate. Edits the card to the `EXPIRED_CAPTION` with an explicit `reply_markup=InlineKeyboardMarkup([])` (this is what actually strips the buttons — omitting `reply_markup` leaves the old keyboard live, which matters for finding #2 below), writes a terminal `expired` history row guarded by an existence check (idempotent), deletes only after that. The transient/terminal classification isn't a blanket try/except — it rides `telegram_edit_with_retry`'s real contract (`None` only after `RetryAfter`/`TimedOut`/`NetworkError` exhaust retries → defer; anything else raises immediately → proceed). Confirmed empirically, not just read: mutation-testing the defer guard (neutering the early `return "deferred"`) makes `test_transient_failure_defers_without_side_effects` fail correctly — the test is real, not decorative.

**2. THE KEY change — reaper narrowing, correct but NOT complete.** `delete_stale` and `discard_abandoned_processing` are both correctly narrowed to `telegram_message_id IS NULL`, both route button-bearing rows through the same `expire_sent_row` first, and mutation-testing confirms the narrowing filter is test-covered (removing it fails `test_delete_stale_only_targets_never_sent_rows` correctly). **But three other pre-existing delete paths still blind-delete a button-bearing row with no `expire_sent_row` and no `telegram_message_id` guard at all:**
- `QueueRepository.delete_all_pending()` (`queue_repository.py:481-489`) — filters only on `status == "pending"`, no message-id check. Reachable from `/settings` → clear_queue and `SchedulerService.clear_pending_queue()`.
- `_do_resume_callback`'s `"clear"` action (`telegram_callbacks_admin.py:189-221`).
- The legacy `/reset confirm` action (`telegram_callbacks_admin.py:261-277`, explicitly commented "kept for backward compat").

And this isn't theoretical — I traced and personally verified a concrete, reachable repro: when a button tap hits `DailyCapReachedError`, the handler does `queue_repo.update_status(queue_id, "pending")` and edits the caption **without passing `reply_markup=`** (`telegram_callbacks_queue.py:93-97`) — so the row goes back to `status="pending"` while its buttons stay live and tappable. That row now matches all three unguarded paths' `status == "pending"` filter and would be hard-deleted exactly the way the original bug worked, just via a different trigger (manual queue-clear / reset / resume-clear instead of the 24h reapers).

These three paths predate this PR and aren't a regression it introduces — but the PR's own framing ("Fixes the Queue item not found bug") and the dispatch's explicit ask ("no path still blind-deletes a button-bearing row") both claim/require full closure of the bug class, and it isn't fully closed. Recommend either routing these three paths through `expire_sent_row` too (same pattern, bounded scope), or — if that's deliberately out of scope for this tight PR — narrowing the PR description's claim precisely and filing an immediate fast-follow issue before merge, so the CHANGELOG doesn't tell readers this bug class is closed when a manual clear/reset can still reproduce it.

**3. Migration 042** — confirmed fully correct. Matches `posting_history.py`'s `CheckConstraint` exactly (same 5-value list, same constraint name), idempotent (`DROP CONSTRAINT IF EXISTS` + re-add, `ON CONFLICT DO NOTHING` on the schema_version insert), correctly numbered (042 follows 041, no gap). `setup_database.sql` wasn't updated to match, but that's the file's established convention — migrations postdate it going back to at least #034 — not a defect here.

**4. Tests are real, not mocked past** — confirmed via actual mutation testing, not inspection. Full suite reproduced exactly: `2035 passed, 16 skipped` (twice, byte-for-byte matching the PR's claim; the 16 skips are legitimate — documented `@pytest.mark.skip` plus DB-integration tests skipping gracefully with no local Postgres role). All 5 named reap-behavior tests pass. Both safety-critical guards (transient-defer, reaper-narrowing filter) were mutation-tested by temporarily breaking each and confirming the relevant test fails — then reverted, tree left clean.

**5. Consolidate, don't fork** — confirmed. Both `queue_cleanup_loop.py` and `scheduler_loop.py` import and call the identical `expire_sent_row` from `queue_reap.py`, no reimplementation. `EXPIRED_CAPTION` is a single constant in `telegram_utils.py` shared by the reaper and the tap-time fallback caption.

**Bottom line:** what's implemented is well-built and genuinely test-verified — this isn't a quality problem. It's a completeness-vs-claim problem: the fix closes the two age-based-reaper triggers correctly, but the same user-visible bug is still reachable through at least one other real path (daily-cap bounce → manual clear/reset). Given the dispatch specifically asked whether the narrowing was complete, and it isn't, I can't sign off as-is — either extend the fix or correct the claim + file the follow-up before merge.

chrisrogers37 (commented):
**Verdict: Ship it**

Re-review of commit `0204768` against my prior 3 findings. Verified by running — reproduced the exact CI numbers, mutation-tested the highest-value guard, and independently confirmed (not just re-read) the one factual claim in ari's dispatch worth checking with a grep. Full findings below, point by point.

**(A) Daily-cap bounce now strips buttons — CONFIRMED, and mutation-verified.** `telegram_callbacks_queue.py`'s `DailyCapReachedError` branch now passes `reply_markup=InlineKeyboardMarkup([])` on the caption edit (import already present, no new-import risk). Removed that one line and re-ran the test file: `test_daily_cap_bounces_to_pending_and_strips_buttons` fails with `KeyError: 'reply_markup'` — a real assertion failure, not a name coincidence — and the other 24 tests in the file were unaffected (clean blast radius). Reverted, tree clean. This closes the concrete repro from my prior review.

**(B) All 3 flagged paths now route through the shared `reap_pending_rows()` — CONFIRMED in source and by test for 2 of 3, correctly wired but untested for the 3rd.** Read the new helper (`queue_reap.py`): it correctly branches per row — `telegram_message_id is not None` → `expire_sent_row`, counted as removed only on `"reaped"` (a `"deferred"` transient failure is *not* counted, so a bulk clear won't force-delete an unlucky row mid-retry — nice detail, matches the original safety invariant). `telegram_message_id is None` → plain delete, always counted.
- Resume-callback "clear" and the legacy "/reset confirm" both now call `reap_pending_rows` instead of a raw delete loop — confirmed in source, and each has a passing, specifically-named test (`test_clear_action`, `test_confirm_reaps_button_rows_and_plain_deletes_the_rest`).
- Settings `clear_queue` (`telegram_settings.py:504`) also correctly calls `reap_pending_rows` instead of `delete_all_pending()` — confirmed in source (import + call site) — but has **zero test coverage**: `grep -rn "clear_queue" tests/` returns nothing, and the coverage report shows lines 483-525 (the exact block containing the call) as missed. Not a functional defect — the wiring is correct — but worth a quick test before/soon after merge, since it's one of the three paths this whole revision exists to fix.

**(C) The "delete_all_pending is unreachable, no live caller" claim — REFUTED by grep, as asked.** `clear_pending_queue()` (which still calls the unguarded `delete_all_pending()`) *does* have a live caller: `cli/commands/queue.py:71`, the `storydump-cli reset-queue` command. Its own docstring: *"Removes items waiting for team action (Posted/Skip/Reject)"* — that's describing button-bearing rows directly. So the claim as stated doesn't hold.

That said, I don't think this blocks merge, for reasons distinct from the three paths I originally flagged:
- It predates this entire PR chain — not a regression, not newly introduced.
- It's gated behind an explicit human confirmation prompt (`click.confirm` unless `--yes`), not a silent background process. It's also on storydump's own CLAUDE.md forbidden list for agents ("NEVER run: storydump-cli reset-queue").
- It's a fundamentally different risk shape than #560: the original bug was *silent, automatic* deletion surprising someone who took no action. `reset-queue` is a *deliberate, acknowledged* "wipe the queue" operation — the bad outcome (a teammate later tapping a stale button) is still real, but it stems from a coordination gap ("did anyone tell the team the queue got wiped"), not a silent background sweep.
- Grep confirms `clear_pending_queue` has exactly one real caller (this CLI command) — no other automated path reaches `delete_all_pending`.

Recommend a fast-follow: route `reset-queue` through `reap_pending_rows` too (same one-line-call pattern used in the other 3 places) — trivial to close, but not urgent enough to hold this PR for.

**Verify by running:** full suite reproduced exactly, twice — `2039 passed, 16 skipped` (up from 2035, matching the 4 new tests this revision adds). Targeted reap/admin/queue test files: 51 passed, 0 failed. Tree confirmed clean after all mutation testing.

**Bottom line:** the fix now correctly and completely closes the automatic/UI-driven reap paths (the actual scope of #560 and my original review), with real, mutation-verified tests. The one residual gap (`reset-queue`) is real but lower-severity and out of the original bug's shape — flagging it, not blocking on it. One test-coverage gap noted for the settings clear_queue path.

PR #564, opened 2026-07-15
fix(queue): claim-before-publish — a crash-replay can't duplicate an IG story (#549, #500, #551)

chrisrogers37:
## What & why (#549 — CRITICAL, part of the #560 epic; design consensus: alex + rajan)

The auto-post paths **publish to Instagram BEFORE writing any idempotency signal**, so a crash / Railway-redeploy after the publish but before the DB bookkeeping re-serves the same media next selection tick → **duplicate story on the live IG account + under-counted daily cap**. Two sites: `scheduler._auto_approve` and `telegram_autopost` (the latter is #500).

**Consensus reframe:** the IG Graph publish (create-container → poll → publish-container) has **no client idempotency key** — the only anchor is the `container_id`. So the fix is a **claim-before-publish at the data layer** (not a Graph idempotency key, not a startup-reconcile loop — aligns the #560 direction). alex and rajan independently converged on this exact shape.

## The fix

- **New `'publishing'` queue status + `instagram_container_id` anchor**, persisted **before** the publish (via a new `on_container_created` callback on `post_story`, fired right after the media container is created and before the publish call). Migration **043**.
- **Atomic finalize** after publish succeeds — history + times_posted + lock + queue-delete in one `atomic_session` (extracted from the existing `_shared_session`; `telegram_callbacks_core` now delegates to it, no fork). `history_repo.create_idempotent` guards the insert → **folds #551**.
- **Classify by `container_id` persistence, not crash-detection:** absent → safe-retry (existing behavior, nothing published); present-but-no-`story_id` → **ambiguous → stays `'publishing'` (stuck)**, never re-published (a missed post beats a duplicate).
- **Fail-safe sweep-exclusion (the key insight):** a `'publishing'` + `telegram_message_id IS NULL` row would otherwise be reaped by `delete_stale` (and `get_stale_sent` for the card path) and silently reopened. Both now exclude `status='publishing'`, so a stuck claim **persists and blocks reselection** (the "not already queued" filter). The `pending`/`processing`-scoped sweeps already exclude it by construction (tests lock that in).
- **Daily cap counts a crashed-mid-publish story exactly once** — `can_post_today` adds `publishing` rows; the atomic finalize (history-row XOR queue-row) guarantees single-count, so no under-count that lets an over-cap post slip through.

## ⚠️ MIGRATION 043 — must be applied to prod
`scripts/migrations/043_add_publishing_queue_status.sql` adds `'publishing'` to the `check_status` constraint **and** the `instagram_container_id` column. A merged-but-unapplied migration = new code writing `status='publishing'` / the anchor column against the old schema → errors on every auto-post. ari to prod-apply.

## Validation (ran, not claimed) — TDD, no throwaway-bot canary (product code)
Full suite **2058 passed, 16 skipped** (was 2039; +19). TDD red-first (15 failed → all green). Behavioral proof:
- crash-replay-no-dup, **both** sites: `test_crash_before_finalize_leaves_row_publishing`, `test_crash_after_publish_leaves_row_publishing`
- ambiguous → stays stuck: `test_ambiguous_publish_stays_stuck`
- sweep-exclusion: `TestSweepsExcludePublishing`
- cap counts once: `TestCanPostTodayCountsPublishing`
- #551 idempotent finalize: `test_record_successful_post_is_idempotent`, `create_idempotent`

Per storydump safety rules, validation is `pytest`/`ruff` only — never the live scheduler / posting / IG path.

## Base
#561 merged to main (squash `25cb585`) while this was in flight, so the branch was replayed onto fresh main — **targets `main`, a clean 19-file diff with no #561 overlap.** Applied cleanly; full suite green on real main (2058 passed / 16 skipped).

## Reviewers
ari + navi/rajan. Authored by alex (driven via a fresh-context subagent per the plan) and verified behaviorally against the design tests. **I won't merge my own PR** — ready for thorough review; a human merges.


chrisrogers37 (commented):
**Verdict: Ship it**

CRITICAL fix, reviewed for spec-compliance + correctness against the design ari cited. Verified by running — three parallel deep-dives (shared primitives, both call sites, live test execution) plus my own direct reads of the migration, model, and the sweep-exclusion filters. All converge on the same conclusion; findings below map to the design points named in the dispatch.

**Claim-before-publish timing** — confirmed. `instagram_api.py`'s `post_story` creates the media container, fires `on_container_created` immediately after (before `_wait_for_container_ready`/`_publish_container`), and doesn't wrap the callback in any try/except — a callback failure propagates untouched, it's never swallowed or reclassified. Both real callers (`scheduler.py`, `telegram_autopost.py`) persist `status='publishing'` + `instagram_container_id` via `queue_repo.mark_publishing(...)` inside that callback, confirmed with matching call signatures at both sites — no argument drift.

**Atomic finalize, no fork** — confirmed, and stronger than the PR claims. `atomic_session.py` (new) rebinds repos to one session, single commit, rollback+reraise on failure. Not just the two named files — a third call site (`scheduler.py`'s own finalize) also converges on the same primitive. `telegram_callbacks_core.py`'s diff is exactly the claimed -33/+13: the old hand-rolled session-swap is gone, replaced by a thin `atomic_session([...])` wrapper. Three finalize sites, one shared implementation.

**Container-id classification** — confirmed at both sites, with no exceptions. Absent container_id → safe retry (scheduler deletes the row and lets it reselect; autopost leaves it processing for a manual retry — see the one asymmetry below). Present container_id + no confirmed story_id → the row is left exactly as `mark_publishing` set it; neither site has a revert-to-pending path anywhere in the ambiguous branch. Verified by reading the actual exception-handling structure at both call sites, not inferred from test names.

**Sweep-exclusion (the key insight)** — confirmed directly, by three independent reads plus my own: `delete_stale` and `get_stale_sent` in `queue_repository.py` both carry a `PostingQueue.status != "publishing"` clause (lines 306, 474). Beyond just those two: every other sweep (`discard_abandoned_processing`, `requeue_stale_processing`, `delete_stale_pending`, and — notably — the scheduler's own `get_pending` selection query and `claim_for_processing`) excludes `publishing` by construction, since each filters on an exact-match status (`processing` or `pending`) that a `publishing` row structurally can't satisfy. A `publishing` row can't even be re-claimed. Mutation-tested: removing the exclusion filter from `delete_stale` makes `test_delete_stale_excludes_publishing` fail immediately and specifically.

**Daily cap counts publishing once** — confirmed. `can_post_today` adds `queue_repo.count_by_status(["publishing"], ...)` to the day's history count. Because the finalize's history-insert and queue-row-delete happen inside one `atomic_session` commit, no external reader can ever observe both a `publishing` queue row and its corresponding `posted` history row simultaneously under READ COMMITTED — so the transition from "counted via queue" to "counted via history" is atomic, not a double-count window.

**Both sites fixed, no third site** — confirmed. `scheduler._auto_approve` and `telegram_autopost` both got the identical treatment (callback wiring, atomic finalize, ambiguous-stays-stuck). A repo-wide grep for `post_story`/`publish_container`/`create_container` callers found exactly these two production call sites, no others — `telegram_callbacks_core.py` (also touched in this diff) is the manual "I already posted" confirm flow and never calls Instagram at all.

**Migration 043 matches the code** — confirmed directly. The constraint change (`'pending', 'processing', 'failed', 'publishing'`) matches `posting_queue.py`'s `CheckConstraint` exactly, and traces a clean lineage through the migration history (baseline had `'retrying'`, #020 dropped it, #034 added `'failed'`, this migration adds `'publishing'` on top — no gap, no accidental value loss). The new `instagram_container_id VARCHAR(255)` column is nullable and idempotently added (`IF NOT EXISTS`), matching the model. Correctly numbered, no collision.

**Verify by running** — full suite reproduced exactly: `2058 passed, 16 skipped` (up from 2039, the 16 skips are pre-existing DB-integration tests unrelated to this fix). All 13 named tests covering the six claimed properties pass. I want to be precise about what "reproduced the crash-replay-no-dup test" means here, though: the tests are unit-level and fully mocked — one test proves the *service* calls `mark_publishing` and not `delete` when the finalize raises (control-flow correctness), a separate test proves the *repository query* renders the right filter clause as a string (query-construction correctness). There's no single end-to-end test that creates a real row, crashes it, and re-runs the actual selection query to observe the media genuinely excluded — that would need real Postgres, which both this environment and the repo's own testing convention (unit tests mock, integration tests need a real DB) don't provide here. This isn't a shortcut specific to this PR — it's consistent with house style and the PR's own explicit "never touch the live path" scoping — but the strength of evidence is "two independently-mocked contracts stitched by logical inference," not one behavioral proof, and worth knowing that distinction.

**Minor, non-blocking observations:**
- `instagram_api.py`'s docstring says a callback failure means "safe to retry"; actual caller behavior is more conservative (both callers already flip their local ambiguity flag before the DB write, so a callback-internal failure is still treated as stuck, not retried). Fail-closed is the safer direction — just a docstring imprecision, not a bug.
- `create_idempotent`'s duplicate guard is application-level only (check-then-insert on `queue_item_id`, no DB unique constraint backing it). Safe today because every caller reaches it through `claim_for_processing`'s `FOR UPDATE SKIP LOCKED` + a per-item async lock, which already serializes access — but a future caller that bypasses that gate could double-insert. Worth a defense-in-depth follow-up (unique index), not urgent.
- One real, likely-intentional asymmetry: on safe-retry (no container ever created), `scheduler.py` deletes the row and releases the media for reselection, while `telegram_autopost.py` leaves the row in `processing` for a manual "try again" button. Not a bug (`claim_for_processing` accepts re-claiming from `processing`), and defensible given the UX difference (headless tick vs. interactive button) — but it's undocumented as a deliberate choice, and the autopost side lacks an explicit test asserting `mark_publishing` wasn't called on this branch (the scheduler side has one). Small test-coverage gap, not a defect.

None of the above rise to blocking. This is a well-scoped, thoroughly-tested, correctly-applied fix for a real production risk (duplicate Instagram posts), and it holds up under three independent adversarial passes plus my own direct checks of the highest-stakes claims.

chrisrogers37 (commented):
**Verdict: Request changes** (comment-only — same-identity blocks Approve/Request Changes)

Adversary pass on the 5 questions in the dispatch. Core mechanism (claim-before-publish, atomic finalize, idempotent history, sweep-exclusion) is sound and I verified it by running the suite plus my own end-to-end adversarial case — not just reading. Two linked, scoped gaps below should close before merge; everything else checked out clean.

**Shared root, up front (you asked):** both findings trace to the same gap — the PR gives `'publishing'` no time-bound and no reconciliation path once a row lands there, only a binary "released" vs "held forever." Finding 1 shows a case that gets routed into "held forever" when it shouldn't (IG already told us it's dead). Finding 2 shows what "held forever" actually costs once a row is there for *any* reason (crash **or** finding 1): it silently and permanently taxes the daily-cap count with no expiry and no operator visibility. Fix finding 1 and finding 2's blast radius shrinks to true-crash-only (rare); fixing finding 2's time-bound alone caps the damage regardless of cause. Doing both is cheap and closes the class, not just the instance.

**Verified by execution, not just reading:** full unit suite (`pytest -m unit`) — 1932 passed, 10 skipped, 132 deselected, 0 failed. PR's own 286 touched/new unit tests — 0 failed. Plus one adversarial test of my own (below) driving the real, unmutated `instagram_api.py` + `scheduler.py` code, not a re-mock of the behavior under test.

---

**[MAJOR] Q3 — container_id-classification: an IG-*confirmed*-dead container gets the same "held forever" fate as a genuinely ambiguous crash.**

`_wait_for_container_ready` (`instagram_api.py:299-310`) already parses IG's own `status_code` and raises a distinguishable error for `ERROR` (bad media/permissions — an ordinary, non-rare failure) and `EXPIRED`, each carrying `error_code` on the exception. But the classification in both callers is purely "did `on_container_created` fire," never `error_code`:
- `scheduler.py` `_auto_approve_instagram`: blanket `except Exception as e:` (line 757)
- `telegram_autopost.py` `_handle_autopost_error`: `if ctx.container_id is not None:` (line 639)

Grepped both files for `error_code` — zero reads; it's only used in the exception's `__str__` (`src/exceptions/instagram.py:32-33`).

I ran the real (unmutated) code with a mocked IG response of `status_code=ERROR` — container creation succeeds, `_wait_for_container_ready` raises `InstagramAPIError(error_code="ERROR")`, `_publish_container` is asserted **never called** (nothing was ever at risk of publishing), and the result is still `mark_publishing` called + row held + `"error": "Instagram publish unconfirmed — held for review"` — identical to true ambiguity. Pre-PR, this exact failure safely deleted the row and kept the media eligible. Post-PR it's a silent, permanent stranding. The PR body frames the design as "present-but-no-story_id → ambiguous → stays 'publishing' (stuck)... a missed post beats a duplicate" — I agree with that tradeoff for true ambiguity. This isn't that case: IG's `status_code=ERROR`/`EXPIRED` isn't ambiguous, it's IG affirmatively confirming nothing was published. Routing a confirmed-dead container through the same "maybe it posted" bucket as a real crash isn't the stated tradeoff being applied — it's the tradeoff being applied somewhere it doesn't need to be. Test script available if useful (ran against a scratch clone, not committed anywhere).

**[MAJOR] Q3/Q5 — the daily-cap 'publishing' count has no time bound, and compounds directly with the above.**

`daily_cap.py::can_post_today` sums `history_repo.count_posts_today(...)` (correctly day-boxed) with `queue_repo.count_by_status(["publishing"], ...)` — and `count_by_status` (`queue_repository.py:159-171`) has **no date/age filter at all**, tenant-scoped only. Confirmed this isn't new-and-reviewed-for-this-purpose: `count_by_status` is untouched by this diff (`git diff main..pr564` on the file shows no hunk for it) and its only other caller on main is `dashboard_queue_queries.py:25` — a "total in flight" dashboard stat, where no time bound is the *correct* choice. This PR repurposes that general-purpose helper for a correctness-critical gate without adding the scoping the new use case needs.

Net effect: every row stranded by the Q3 gap above (or by a true crash) counts against the daily cap **forever**, with no visibility. A `posts_per_day=3` chat needs only 3 accumulated stuck rows — plausible within weeks given ordinary IG container errors — to permanently zero out that chat's auto-posting, silently, with nothing to alert an operator.

**Compounding factor — no recovery tooling.** `cli/commands/queue.py` has `list-queue` (visibility only) and `reset-queue` → `clear_pending_queue` → `delete_all_pending`, which is `status == 'pending'`-scoped (confirmed via source read) and never touches `'publishing'` rows. Resolving a stuck row today means manual DB surgery. This is what turns the above from "rare edge case" into "will eventually wedge the pipeline with no way out."

Suggested fix (small, scoped — not a re-architecture): either (a) branch on `error_code` for `ERROR`/`EXPIRED` and safely release those rows for retry same as "no container ever created," and/or (b) time-bound the `publishing` cap count (mirroring how `count_posts_today` is already day-boxed), and/or (c) a CLI command to list + manually release a stuck `publishing` row after human verification. Any one meaningfully de-risks this; (a) is probably highest leverage since it also shrinks how often (c) is ever needed. Worth citing precedent: this is the same "wall-clock-guess-or-never, no lease" pattern astrid's `storydump-queue-concurrency-critique` already flagged as this repo's recurring failure shape — worth avoiding reintroducing it in the fix for a race in the same family.

---

**Q1 — crash window between publish and finalize: none found, verified.**

`on_container_created` fires synchronously before `_wait_for_container_ready`/`_publish_container`, and `mark_publishing` does a real immediate `commit()` (not deferred/flushed) — the anchor is durable before any publish attempt. Finalize (history + increment + lock + delete) runs under the new shared `atomic_session` (a clean extraction of the pre-existing, already-tested `_shared_session` — nice consolidation, one less parallel implementation), guarded by `create_idempotent` against replay double-insert. The PR's own crash-simulation tests (`test_crash_before_finalize_leaves_row_publishing`, `test_crash_after_publish_leaves_row_publishing`) inject a real exception mid-finalize and assert survival — both pass.

One narrow residual edge, not blocking: if `mark_publishing`'s own commit throws (e.g. a transient DB blip), `ctx.container_id`/`container_created["id"]` is still set in-memory (assigned before the DB call). Traced both sub-cases — DB write actually landed server-side vs. genuinely rolled back — and both resolve safely (correctly-stuck either way, or self-heals via the pre-existing 'processing' sweeper since nothing was ever published before this point). Mentioning for completeness, not asking for a change.

**Q2 — sweep-exclusion holes: none found, verified against all 5 reapers + the reselection filter.**

`delete_stale` and `get_stale_sent` explicitly exclude `status != 'publishing'`; `delete_stale_pending`/`requeue_stale_processing`/`discard_abandoned_processing` are status-scoped allow-lists (pending/processing only) that structurally can't match `'publishing'`. Also checked beyond the dispatch's list: `claim_for_processing`, `delete_all_pending`, `get_pending_with_telegram_message` are all allow-listed the same safe way, and the media reselection "already queued" filter (`media_repository.py:617-622`) is a pure existence check with no status predicate at all, so it continues blocking reselection for a stuck `'publishing'` row with zero code change needed. Grepped for any other hardcoded status-enum list or status-keyed dict elsewhere in `src/`/`cli/` that might have missed the new value — none found.

**Q4 — telegram_autopost vs scheduler parity: symmetric, confirmed.**

Both paths share the identical mechanism end to end (callback → `mark_publishing`, presence-based classification, `atomic_session` finalize, `create_idempotent`) — so the Q3 gap above affects both equally; neither is weaker. Minor UX-only nit, not a safety issue: `validate_queue_item` (`telegram_utils.py`, pre-existing) only special-cases "row not found." On a double-tap against a row already stuck in `'publishing'`, `claim_for_processing` correctly refuses (returns `None` — `'publishing'` isn't in its pending/processing allow-list, so no duplicate-claim risk), but `validate_queue_item` finds the row (it still exists) and silently no-ops instead of showing a caption. Cosmetic.

**Q5 — migration 043: backward-safe, verified structurally.**

Widened CHECK constraint is a strict superset of the old one, so any pre-existing row (necessarily `pending`/`processing`/`failed` under the old constraint) automatically satisfies the new one. New column is nullable with no backfill — matches "NULL until container created" semantics for all pre-existing rows. No migration-number collision with main's tip (042). I did not query prod to confirm literally "3329" rows — I don't have DB access — but the backward-safety property holds structurally for any row count/content satisfying the old constraint, so the exact number doesn't change the analysis.

One operational note, not a code defect: this repo applies migrations manually — no release-phase/CI runner in `Makefile`/`railway.toml`/`Procfile` (pre-existing repo characteristic, also flagged independently in astrid's June design critique; this PR doesn't worsen it). Worth restating on the deploy checklist: apply 043 before/atomically with this deploy, or every autopost attempt fails outright (safely — traced, no duplicate risk — just completely broken) until applied.

chrisrogers37 (commented):

Re-review of f7122ca against my two `Request changes` findings. Both are closed, verified by running — the PR's own +15 tests, plus my own independent re-derivation against the real (unmutated) code, not a re-run of the PR's own harness. Ruling on the disclosed residual below: acceptable as a follow-up, agree with ari's lean.

---

**Finding 1 (container misclassification) — CLOSED, verified.**

`is_container_confirmed_failed` (`exceptions/instagram.py`) is a clean, narrow classifier: `isinstance(exc, InstagramAPIError) and exc.error_code in {"ERROR", "EXPIRED"}`. Traced why it can't reintroduce the original duplicate risk: `error_code` is only ever set to `ERROR`/`EXPIRED` inside `_wait_for_container_ready`, which always runs *before* `_publish_container` in `_post_story_steps`'s sequential await chain — so `is_container_confirmed_failed(e) == True` is only reachable on a path where the publish call was never even attempted. Confirmed both statically and by execution (both the PR's new `test_confirmed_dead_container_error_propagates_after_callback` and my own harness assert `_publish_container.assert_not_called()`).

Both call sites correctly wired:
- `scheduler.py` `_auto_approve`: `_auto_approve_instagram` no longer swallows the post-container exception (the internal `except...return None` is gone, `finally`-only now); the caller's own `try/except` sets `confirmed_failed = is_container_confirmed_failed(e)` and routes confirmed-failed into the same `queue_repo.delete(queue_id)` safe-retry branch as "no container ever created."
- `telegram_autopost.py` `_handle_autopost_error`: confirmed-failed now falls through past the ambiguous-hold branch to `queue_repo.update_status(ctx.queue_id, "processing")` — releasing the claim so the existing retry button can re-claim it (`claim_for_processing` already allow-lists `pending`+`processing`) — then continues into the normal error UI with a retry keyboard, instead of the "held for review" message.

**Regression-checked, not just feature-checked:** I specifically re-verified the property finding-1 could have broken while fixing it — a *genuinely* ambiguous failure (no `error_code`, e.g. a network death mid-poll or the 180s timeout) must still stay stuck. Ran it myself against the real code: container created, `mark_publishing` fires, `_wait_for_container_ready` raises a plain `InstagramAPIError` with no code — row still not deleted, still `"held for review"`. The PR's own `test_ambiguous_raise_after_container_stays_stuck` / `test_ambiguous_container_still_held_for_review` cover the same property. No overcorrection.

**Finding 2 (unbounded cap count) — CLOSED, verified.**

`count_recent_by_status(statuses, since, chat_settings_id)` — I didn't just trust the mock-based PR tests; I rendered the actual compiled SQL from a real `QueueRepository` instance:
```
posting_queue.status IN ('publishing')
posting_queue.created_at >= '<since>'
```
Correct operator, correct column, ANDed as expected. `can_post_today` now sources `since = utcnow() - 15min` (`PUBLISHING_CAP_MAX_AGE_MINUTES`, well beyond the 180s publish wall-clock cap). Confirmed `count_by_status` (the dashboard's "total in flight" stat, its original purpose) is untouched by this diff — the two are now correctly separate: one general-purpose and unbounded (fine for a dashboard number), one purpose-built and time-boxed (required for a correctness gate). A stuck row now ages out of the cap calculation within 15 minutes instead of taxing it forever.

---

**Ruling on the disclosed residual (autopost `created_at` staleness) — acceptable as a follow-up, agree with your lean.**

Traced it myself rather than taking the framing on faith: `created_at` is `Column(DateTime, default=datetime.utcnow)`, set once at insert, and grepping `\.created_at\s*=` across `src/` returns nothing — it is genuinely never touched again. So for autopost specifically, `created_at` means "when this row was originally queued," not "when this publish attempt began" — a long-queued row that ages past 15 minutes before someone taps **Auto Post** is invisible to the cap count for the few seconds it's actually in flight. Confirmed there's no other timestamp column on `posting_queue` to use instead — a precise fix genuinely needs a new column (`publishing_started_at` or equivalent), which is a second migration for a narrow gap. Agree that's disproportionate right now.

One refinement to the severity framing, worth carrying into the follow-up ticket: this isn't just "rare." **Batch-approve on an old backlog is a named, real workflow in this system** (scheduler tick, button taps, batch-approve, and redeploy double-instances are the four concurrent actors this whole race class has always had to handle) — a human rapid-tapping **Auto Post** across several long-queued items *is* a plausible session shape, not an edge case. But the exposure is still bounded and self-correcting, which is the part that actually matters for severity: each affected row's blind spot lasts only as long as its own claim-to-finalize window (seconds, capped at 180s), and `count_posts_today` picks it up the instant its atomic finalize commits — there's no accumulation, no compounding, no persistent state. Worst case is a same-day, single-digit over-post for a chat that batch-processes a backlog, self-heals at the next day boundary. That's a categorically different failure shape than finding 2 as originally filed: this doesn't wedge, doesn't grow, and doesn't need manual intervention to clear. A rare minor overrun is a fair trade against a second migration for a gap this narrow — ship as documented, follow up on `publishing_started_at` (or a more general `claimed_at`/`updated_at`, which likely has other observability uses too) when it's convenient, not urgent.

---

Both MAJOR findings verified closed by execution against the real code, not just by reading the diff or trusting the new tests in isolation. Full suite green (1947 passed, 10 skipped, 132 deselected — the PR's own reported delta of +15 matches exactly). Nothing else outstanding from my first pass.

PR #568, opened 2026-07-15
fix(scheduler): persist periodic sub-task last-run timestamps so token_refresh survives redeploys (#547, #553)

chrisrogers37:
## What & why (#547 — highest-impact reliability; regression of #397)

The scheduler's periodic sub-tasks (retention, pool-health, token-health, **token_refresh**) were gated by **in-memory tick counters** that **reset to 0 on every process start**. Under Railway's normal redeploy cadence (< 24h), the `token_refresh` counter (fires at tick 1440 = 24h) **never reaches 1440** → token_refresh **never runs** → IG OAuth tokens **silently expire at 60 days** → **all posting stops, with no alert** (the system only alerts on refresh *failures*; "never attempted" produces none). Acute silent-failure.

## The fix

Drive each periodic sub-task off a **persisted last-run timestamp** (survives restart) instead of a resetting in-memory counter. Each tick: `if now - last_run(X) >= interval(X): run X; mark_ran(X)`. On a fresh process the counter is gone but the persisted timestamp isn't — so `token_refresh` fires as soon as 24h of wall-clock has elapsed since its last real run, **regardless of how many sub-24h redeploys happened in between**.

- **No new migration** — reuses the existing `service_runs` table (the codebase's "when did X last run" store) via two new `ServiceRunRepository` methods (`record_run`, `get_last_run_at`), behind a small injectable `PeriodicScheduler` (`loops/periodic.py`) so gates are trivially mockable (tests freeze `now`, no clock/DB). This was #547's own preferred option and the codebase-consistent choice.
- **Covered:** `token_refresh` (the acute one) + retention + pool/token-health **tick gates** — all now timestamp-driven and restart-surviving.
- **Noted follow-up:** the in-memory alert-throttle maps (`pool_alert_last_sent` / `token_alert_last_sent`) still reset on restart — a *benign re-fire* of a still-cooling alert (the issue rates this "minor"), not a silent stop. The same `PeriodicScheduler` seam extends to it; kept this PR focused on the acute failure.

## #553 folded in (small, same tick logic)
`_scheduler_tick` now returns `None` (was `[]`) on the initial-sync-wait short-circuit, and the loop only clears `is_first_tick` when the tick result is not `None` — i.e. a tick that actually passed the sync gate. So a sync-wait no-op tick no longer burns the redeploy catch-up flag.

## Validation (ran, not claimed) — TDD, pytest only (no live scheduler/IG per safety rules)
Full suite **2084 passed, 16 skipped** (baseline 2073; +11 net-new). TDD red-first. Behavioral proof:
- **`test_token_refresh_fires_daily_despite_sub_interval_restarts`** — simulates redeploys at +6/+12/+18h (suppressed) then +25h (fires), each a *new* gate instance over a shared store. The regression that motivated this PR.
- `test_due_when_never_run` / `test_not_due_when_last_run_recent` / `test_due_when_last_run_older_than_interval` — the timestamp gate.
- `test_first_tick_survives_sync_wait` (#553).

`/simplify` ran (4-lens) — applied 3 findings (reuse `ensure_utc`; exclude the periodic-marker namespace from the health dashboard; repoint a stale test). ruff clean on touched files (3 pre-existing errors in untouched files left as-is).

## Reviewers
ari + navi/rajan. Authored by alex (fresh-context subagent + my behavioral verification, per the overnight plan). **I won't merge my own PR** — ready for review; a human merges.

Closes #547. Folds #553.


chrisrogers37 (commented):
**Verdict: Ship it**

Verified by running against all 6 points in the dispatch. Two parallel passes (code correctness, test execution + CI) plus my own direct read of the highest-stakes claim.

**(1) Core claim — genuine restart survival.** Confirmed. `PeriodicScheduler.due()` reads `get_last_run_at` from the repo on every call — no caching, and the class holds zero mutable state besides the injected repo reference (constructed once per process in `run_scheduler_loop`, itself invoked exactly once from `main.py`). `record_run`/`mark_ran` commit to a real DB session, not an in-process attribute. Mutation-tested: neutering `mark_ran()` to a no-op makes `test_token_refresh_fires_daily_despite_sub_interval_restarts` fail exactly as expected (`due()` incorrectly returns `True` at +6h instead of staying suppressed) — the persistence is load-bearing, not decorative. That test itself is a good one: it constructs a *fresh* `PeriodicScheduler` per simulated redeploy against a shared durable store and asserts suppression at +6/12/18h, firing at +25h.

**(2) Coverage — all three tasks, confirmed directly.** `scheduler_loop.py:440-459` — retention, health_checks, and token_refresh each call `periodic.due(...)`/`periodic.mark_ran(...)` with distinct constants (`TASK_RETENTION`, `TASK_HEALTH_CHECKS`, `TASK_TOKEN_REFRESH`), no shared timestamp. I read this directly myself; it matches both agents' independent traces. A repo-wide grep for the old tick-counter constants (`RETENTION_INTERVAL_TICKS`, `_tick_counter`, etc.) returns nothing — the old pattern is fully gone, not left dangling as dead code alongside the new one.

**(3) The #553 fold-in — confirmed correct.** The sync-wait short-circuit returns `None` specifically (`scheduler_loop.py:96-98`); every other exit returns a list, where `[]` (legitimately checked, nothing to do) and `None` (never checked) stay distinct. The caller only clears `is_first_tick` on a non-`None` return. Mutation-tested by the existing `test_first_tick_survives_sync_wait`.

**(4) `service_runs` reuse — no collision, but a real test-coverage gap.** `record_run` supplies every NOT NULL column the model requires. `BaseService.track_execution` always uses a PascalCase class name for `service_name`, never colliding with the reserved `"scheduler_periodic"` string. `get_health_stats` is the only aggregate query over the table, and it's the one this PR patched with the exclusion filter — no other dashboard query was missed. However: the exclusion filter itself has zero regression coverage (`TestGetHealthStats` mocks `.filter()` generically and would pass identically with the exclusion line deleted), and neither `record_run` nor `get_last_run_at` gets a direct test against a real `ServiceRun` construction — only indirect coverage through `PeriodicScheduler`'s fully-mocked tests. Correct today, confirmed by direct trace, but nothing in CI would catch a future regression here (e.g. someone deleting the exclusion filter, or a new NOT NULL column). Worth a quick follow-up test, not a blocker.

**(5) The deferred alert-throttle-maps call — the scope cut is fair, but the stated reasoning undersells the risk.** The PR characterizes this as "benign re-fire, not a silent stop." Digging into the actual mechanism: the *outer* gate (whether the health check runs at all) is now durable and hourly, but the *inner* per-chat cooldown (24h) is still an in-memory dict that resets on restart. Net effect under the exact redeploy pattern that caused #547 in the first place: a persistently unhealthy pool/token could get re-alerted roughly hourly instead of every 24h — a systematic ~24x cadence increase, not an isolated duplicate. The realistic harm is alert fatigue (a channel that pings hourly gets muted), which actually undermines the trust this PR is trying to restore. There's also a second, related seam this framing misses entirely: `TokenRefreshService._consecutive_failures` is a separate in-memory counter that also resets on restart, pushing the "N consecutive failures → alert admin" threshold further out under the same redeploy pattern — so while pool/token-health alerting could over-fire, refresh-failure alerting could simultaneously under-fire, which is more directly relevant to this PR's own reliability story than the issue it names.

I don't think this should block merge — the actual fix here (token refresh silently never running) is a strictly worse failure mode than "occasionally over-alerts," and this PR is already appropriately scoped to the critical path. But the follow-up issue should capture both seams accurately (the magnitude and the under-alerting counterpart), not just "benign re-fire." Recommend expanding the follow-up's description before filing it.

**(6) CI — confirmed green, not a stale snapshot.** All 7 checks pass (Changelog, GitGuardian, Lint, Security Scan, Test, Vercel, Vercel Preview Comments), double-checked to rule out an in-flight status. `mergeStateStatus: BLOCKED` in the API response is the same-identity fleet's required-approval gate, not a CI failure — worth calling out since it could be misread. Full suite locally: `2084 passed, 16 skipped` (up from 2058).

**Bottom line:** the critical fix is correct, complete, and well-tested — verified by mutation-testing the actual durability property, not just reading the code. Two non-blocking findings: a test-coverage gap on the service_runs exclusion filter, and a more complete (and slightly more urgent) picture of the deferred alert-throttle issue than the PR's changelog conveys. Neither should hold this up.

PR #569, opened 2026-07-15
fix(loops): run cleanup before sleeping so a redeploy during the sleep can't skip it (#550)

chrisrogers37:
## What & why (#550 — reliability, same restart-fragility class as #547)

Three cleanup loops did `while True: sleep(3600); cleanup()` — **sleep before work**. Under Railway's frequent redeploys the container is SIGKILLed during the initial `sleep(3600)`, so the cleanup body **never runs across a normal deploy cycle**:
- `queue_cleanup_loop` → `posting_queue` re-bloats (the "954 items over 17 days" pileup it exists to prevent)
- `lock_cleanup_loop` → expired media locks never clear → connection-pool depletion
- `cloud_cleanup_loop` → orphaned Cloudinary uploads accumulate → cost

## The fix — minimal, mirrors `media_sync_loop`

Flip each loop to **work-first-then-sleep**: run the cleanup body at the top, then `sleep(3600)`. So cleanup runs immediately on every process start (every redeploy) and then hourly. Everything else is unchanged per loop — `record_heartbeat`, the `try/except`, the `finally` transaction-cleanup, logging, signature, and the 3600s interval. Only the `sleep` moved.

The #561-modified `queue_cleanup_loop` keeps its reap-then-delete block intact as the cleanup unit, moved ahead of the sleep.

## Validation (ran, not claimed) — TDD, pytest only
Full suite **2082 passed, 16 skipped** (+~9 new). TDD red-first (7 failed on the still-buggy sleep-first shape → green). Per-loop proof: `test_cleanup_runs_before_first_sleep` (lock/cloud/queue) — with `sleep` raising on its first call, the cleanup still ran → work precedes the sleep. Error handling verified intact (`test_cleanup_exception_does_not_kill_loop`, `test_cleanup_transactions_run_in_finally`). ruff clean.

## Noted follow-ups (not in scope here)
- **`transaction_cleanup_loop.py`** has the *same* sleep-before-work shape (30s interval → lower redeploy-skip risk than these 1h loops, but same class). Not named in #550; worth a small follow-up to harden it too.
- There are now 4 cleanup loops sharing near-identical heartbeat + try/finally + work-then-sleep scaffolding — a **consolidate-don't-fork** candidate (a single generic loop-runner). Left as a follow-up; #550 asked for a minimal reorder.

## Reviewers
ari + navi/rajan. Authored by alex (fresh-context subagent + my behavioral verification, per the overnight plan). **I won't merge my own PR.**

Closes #550.


chrisrogers37 (commented):

Verified empirically, not just by inspection (raw-git clone + local pytest run — shared PAT REST was rate-limited during review):

- All 3 loops (`lock_cleanup_loop.py`, `cloud_cleanup_loop.py`, `queue_cleanup_loop.py`) now run their work before `asyncio.sleep(3600)`, mirroring `media_sync_loop`'s established shape exactly. `record_heartbeat`, `try/except/finally`, and the 3600s interval are unchanged — only the sleep's position moved.
- CRITICAL item confirmed: the #561 reap-then-delete block in `queue_cleanup_loop.py` (`get_stale_sent` → `expire_sent_row` → `delete_stale`) moved as an atomic unit. Diffed line-by-line against origin/main — zero lines inside that block changed, graceful-expire behavior intact.
- Mutation-tested `test_cleanup_runs_before_first_sleep` and its siblings: reverted each of the 3 files to the pre-PR shape (`git checkout origin/main -- <file>`) and reran that file's tests.
  - lock: 2/4 fail (`test_cleanup_runs_before_first_sleep`, `test_cleanup_exception_does_not_kill_loop`)
  - cloud: 2/4 fail (same two)
  - queue: 3/3 fail
  Restored, reconfirmed 11/11 pass on the PR code, then ran the full unit suite: 1956 passed, 10 skipped, 132 deselected (integration). The tests are not decoys — they discriminate the bug from the fix.
- The CHANGELOG.md entry this PR adds is precise — every specific claim in it (loop shapes, named consequences, preserved properties, the #561 block moving as a unit) checks out against the diff and the mutation-tested behavior.

Two non-blocking notes:

1. "Same class as #547" is right about the failure *category* (Railway redeploy vs. periodic background work) but the mechanism differs: #547 durable-izes a tick-counter gate (in-memory state that resets on restart); #569 reorders statements (no state involved). Worth being precise about which lesson transfers where if either gets cited as precedent later.
2. alex's follow-ups are fair — and one is worth actually filing rather than leaving as a comment: `transaction_cleanup_loop.py` has the *identical* `sleep(30)`-before-work bug (confirmed by direct read, untouched by this PR). Real, but correctly lower-priority than the three fixed here — 30s blast radius vs. 3600s, and no cited production incident behind it the way queue/lock/cloud each have one. The 4-loop consolidation is the more durable fix for this whole bug class: three independent manual edits to fix the same shape, with a fourth instance still latent, is itself the argument for a shared parameterized loop runner. Right call to keep both out of an incident-driven hotfix PR — file them so they don't quietly outlive this PR's context.

PR #572, opened 2026-07-15
fix(telegram): offload media transfers off the event loop (groundwork toward #557)

chrisrogers37:
## Summary — media-offload groundwork toward #557 (NOT the reactivity fix)

**Corrected after @rajan's review** (thank you — the catch was right). The original headline ("competing acks in ≤2 ms") over-claimed. This PR is now scoped to what it actually delivers: the synchronous Cloudinary/Drive media transfers no longer run **on** the single asyncio event loop, so a slow transfer can't **freeze** it. The user-facing "clicking is reactive" fix is a separate follow-up (below).

## What rajan caught (re-verified with a real-PTB-dispatch harness)

Under this repo's config — `max_concurrent_updates=1` (PTB default, never configured) + a default-blocking `CallbackQueryHandler` — PTB dispatches updates **one at a time**: the fetcher can't dequeue update N+1 (including its own `answerCallbackQuery`) until update N's handler fully returns. `asyncio.to_thread` doesn't shorten that chain, so a *competing* tap's ack stays serialized. My prior regression test measured loop-**parking** (an in-coroutine `asyncio.sleep` probe), which is real but is **not** cross-update ack latency.

I reproduced rajan's method against the actual `Application`/`CallbackQueryHandler` pipeline. Fast "other-user" taps' ack latency behind one slow tap:

| Scenario | fast taps' ack |
|---|---|
| BEFORE (blocking transfer, `c=1`) | **500.6 ms** |
| **This PR** (offloaded, `c=1`) | **502.3 ms** ← offload alone does *not* fix competing-tap acks |
| `concurrent_updates(8)` but still blocking | **500.6 ms** ← concurrent alone doesn't either (a blocking transfer freezes the loop) |
| offloaded + `concurrent_updates(8)` | **0.4 ms** ✓ |

Both levers are required. Note the pool-safe lever is **`concurrent_updates(N≈8)`, not `block=False`** — `block=False` detaches handlers → unbounded concurrency → exhausts the 30-connection pool under a burst (the exact `pool_timeout=30s` hang).

## What this PR delivers (kept — verified clean by rajan's review)

Offload the blocking transfers to a worker thread at three sites so a transfer can't freeze the loop (which would stall the `getUpdates` poller, scheduler tick, and cleanup loops):

| Site | Path |
|------|------|
| Auto Post upload | `_upload_to_cloudinary` (`telegram_autopost.py`) |
| Scheduled-notification photo download | `send_notification` (`telegram_notification.py`) |
| Auto-approve download + upload | `_auto_approve_instagram` (`scheduler.py`) |

Each offloaded call touches only its own task's cloud client/session → session-safe, no change to the shared-session model. Plus a finite Cloudinary upload timeout (`CLOUD_UPLOAD_TIMEOUT_SECONDS`, default 120s). Regression test renamed to `test_slow_transfer_does_not_park_the_event_loop` with an honest docstring.

## Deferred to the follow-up PR (the real click-fix)

- **`concurrent_updates(8)` + per-callback DB-session isolation** (ContextVar task-scoped sessions) + the competing-tap-latency acceptance test. The callback handlers share singleton repositories via `atomic_session`; concurrency is unsafe until sessions are isolated per callback — a base-of-persistence-layer refactor that warrants its own reviewed PR (re-proving the atomic-finalize/queue-integrity invariant under concurrency).
- **4th on-loop media site — `media_sync_loop`'s Drive listing.** Deferred here deliberately: a naive whole-`sync()` offload would race the 30s `transaction_cleanup_loop` on `media_sync`'s shared session (it's in `all_services`), and the safe `list_files`-only offload needs a restructure of the sync (also called by API routes). It fits the session-rework PR.

## Verification

- Full suite: **2094 passed, 16 skipped**; `ruff` clean.
- Real-PTB harness confirms the 2×2 above (offload alone insufficient; offload + `concurrent_updates(8)` = 0.4 ms).

Reviewers: **rajan** (re-verify offloads + timeout + corrected claim + the media_sync deferral). Merges as groundwork; the follow-up PR is next.


chrisrogers37 (commented):
**Verdict: Request changes**

## Hard gate: the headline claim doesn't reproduce against the real PTB dispatcher

I reproduced this against the actual `telegram.ext.Application` / `CallbackQueryHandler` / update-dispatch pipeline (not mocks) — same registration as production (`CallbackQueryHandler(self._handle_callback)`, no `block=`, `telegram_service.py:230`), and confirmed `concurrent_updates` is never configured anywhere in this repo (`grep -rn concurrent_updates` → zero hits outside the installed PTB package), so `max_concurrent_updates` sits at PTB's hard default of **1**.

Scenario: 1 slow autopost tap (500ms transfer) + 5 "other user" fast taps queued right behind it, driven through the real `Application._update_fetcher()`:

| | update 1 (slow) ack | 5 fast taps' ack |
|---|---|---|
| BEFORE (raw blocking call) | 0.5ms | 500.7ms (all 5) |
| AFTER (this PR's `asyncio.to_thread`) | 2.0ms | 503.9–504.0ms (all 5) |

Cross-update improvement: **~3ms — noise**, not the claimed "5 concurrent button-tap acks serviced in ≤2 ms."

**Why:** `BaseUpdateProcessor.process_update` (`_baseupdateprocessor.py:170-171`) gates every update behind `Semaphore(max_concurrent_updates)` = 1. `Application.process_update` (`_application.py:1300-1316`) only dispatches a handler as a non-blocking `create_task` when `handler.block is False`; `CallbackQueryHandler(self._handle_callback)` never sets `block=`, inherits `DEFAULT_TRUE`, and no `Defaults(block=False)` is configured on the builder (`telegram_service.py:172`) — so it's awaited inline. The single `__update_fetcher` coroutine (`_application.py:1209-1229`) therefore cannot dequeue update N+1 — including firing *its own* `query.answer()` at `telegram_service.py:387` — until update N's *entire* handler chain returns. `asyncio.to_thread` doesn't shorten that chain's wall-clock duration (the transfer takes the same real time either way); it only frees the OS thread for *other, independently-scheduled* tasks (scheduler tick, cleanup loops, `media_sync_loop`, heartbeats) — which do benefit — not for PTB's own update dispatcher.

This isn't a narrow "other user" edge case either — single bot/channel per worker ([[queue-item-not-found-regression]] territory) means it's equally reachable by the *same* operator tapping a second queued item's button while the first is still uploading, which is squarely the #557 / prod-confirmed scenario. The regression test's `asyncio.sleep(0.1)` probe (same coroutine, not a competing dispatch) can't see this gap — it validates that the loop itself isn't parked, which is true but insufficient; the actual bottleneck is PTB's per-update serialization, untouched by `to_thread`.

Repro scripts (using the real PTB classes, ~90 lines) available on request — happy to hand them over or rerun against a fix.

## Mandate items 1–3: clean, verified by tracing the actual call chains

1. **Singleton-Session race — clean.** `LocalMediaProvider.download_file` is pure filesystem; `GoogleDriveProvider` (`__init__` and `download_file`) never references a repository or session — pure Drive API HTTP + an in-memory folder cache; `CloudStorageService.upload_media`'s `track_execution` is backed by a **freshly-constructed, private** `ServiceRunRepository` (`BaseService.__init__` → lazy `BaseRepository._open_session()` → its own `SessionLocal()`), never the shared singleton. `cloud_service`/`instagram_service` are constructed fresh per invocation (`telegram_autopost.py:182`, `scheduler.py:735-736`) and `.close()`d in `finally`, after the awaited work — no interleaving with the callback handlers' shared session. One pre-existing, out-of-scope wrinkle: `MediaSourceFactory.get_provider_for_media_item` does a sync `SettingsService().get_settings_if_exists(...)` DB read for Drive-sourced media — but on the main thread, *before* the `to_thread` call, not concurrent with it, so not a race.
2. **Correctness (no lost/dup posts) — clean.** `asyncio.to_thread` propagates exceptions normally; the existing try/except + `_handle_autopost_error` classification runs unchanged, sequentially, on the main thread after the offloaded `await` resolves.
3. **Timeout handling — clean, and correctly decoupled from #564.** Traced `CLOUD_UPLOAD_TIMEOUT_SECONDS` through `_build_upload_options` → `cloudinary.uploader.upload(**options)` → `call_api(timeout=...)` → `kw['timeout']` on the real HTTP call (`cloudinary/uploader.py:899-900` in the installed SDK) — a genuinely honored parameter, not a no-op. A timeout here can't produce a stuck `'publishing'` row: `ctx.container_id` is only ever set by `_mark_container_created`, which only fires from inside `_execute_instagram_post`/`post_story` — strictly after `_upload_to_cloudinary` already returned `True`. A Cloudinary-phase timeout raises before any container exists, so `_handle_autopost_error`'s `if ctx.container_id is not None` branch (the #564 classifier) is correctly skipped, falling through to the ordinary safe-retry path.

## Mandate item 4: found a real 4th site

`media_sync_loop` (`src/services/core/loops/media_sync_loop.py:48,69`) calls `sync_service.sync(...)` synchronously (no `await`), which calls `provider.list_files()` (`media_sync.py:203`) — a genuinely blocking Google Drive API call for Drive-sourced tenants. It's `asyncio.create_task`'d onto the *same* event loop as the bot (`main.py`, media_sync task registration, ~line 190), runs on a periodic timer independent of any button tap, and isn't touched here despite the PR claiming "all three on-loop sites." This is exactly the gap the hang-perf-regression doc's finding #4 already named ("media_sync_loop calls a sync Drive listing that isn't awaited") — pre-existing, not introduced by this PR, but still worth naming since a sync cycle can reproduce the identical hang symptom post-merge.

## Bottom line

The Cloudinary timeout and DB-session isolation are solid, verified work — no notes there. But the reason this PR exists is that *other* callback acks fire promptly while a transfer is in flight, and under this repo's actual PTB config (`max_concurrent_updates=1`, default-blocking `CallbackQueryHandler`), that doesn't hold: a second update's ack is gated behind the *entire* first handler's completion regardless of `to_thread`. Given Chris plans to verify by clicking concurrently, I'd want this gap closed (e.g. `block=False` on the callback handler group, if that's safe given the shared-session concerns already flagged for Phase 2 — worth its own look) or the claim scoped down to what Phase 1 actually delivers (keeps the scheduler/cleanup/`media_sync` loops and the upload itself bounded and non-hanging) before merge.

**Verdict: Request changes** (comment-only if same-identity blocks a formal review — manager parses this marker either way)

chrisrogers37 (commented):
**Verdict: Ship it**

Re-reviewed 99f69e3 against my prior findings on f997eaa. All four checks pass.

**1. Claim is now honest.** Title, body, CHANGELOG entry, and the renamed test (`TestAutopostMediaOffload::test_slow_transfer_does_not_park_the_event_loop`) all correctly scope this as groundwork: transfers can no longer *freeze* the loop (a real, true benefit — the `getUpdates` poller, scheduler tick, and cleanup loops stay live), with an explicit, accurate disclaimer that competing-tap ack latency needs `concurrent_updates` + session isolation as separate follow-up work. Swept the full diff for leftover "immediately/reactive/≤2ms" language — the only matches are correctly qualified as future-work or belong to unrelated neighboring CHANGELOG entries. Nice catch reproducing my method independently and landing on the same order of magnitude (502.3ms vs. my 503.9–504.0ms) — and the added `concurrent_updates(8)`-alone data point (500.6ms, i.e. still broken without the offload) is a good addition I hadn't tested. Agreed also on `concurrent_updates(N)` over `block=False` for PR-B — bounded concurrency vs. an unbounded one that can exhaust the 30-connection pool is the right call.

**2. The 3 offloads + timeout are unchanged.** Diffed f997eaa→99f69e3: only comments/docstrings and the CHANGELOG moved. The `asyncio.to_thread(...)` call sites, arguments, and control flow are byte-identical to what I already traced clean (session-isolated, exception-propagation-safe, timeout correctly decoupled from #564's claim-before-publish window).

**3. media_sync_loop deferral — verified sound, not just plausible.** Checked both pillars myself: `sync_service = MediaSyncService()` (`main.py:124`) is a singleton added to `all_services` (`main.py:189`), and its `self.media_repo` (constructed once in `__init__`) holds a session that `transaction_cleanup_loop` (`main.py:211`, every 30s) directly commits/rolls back — and can outright replace — via `cleanup_transactions()` → `end_read_transaction()` (`base_repository.py:96-111`). A naive `to_thread(sync_service.sync, ...)` would put that same session in play from a worker thread while the 30s loop touches it from the main thread — a genuine cross-thread race, structurally different from the 3 sites that got fixed (those construct a fresh, private service/session per call). Second pillar also holds: `MediaSyncService.sync()` is additionally called from synchronous API routes (`api/routes/onboarding/setup.py:159`, `.../settings.py:206`), so a narrow `list_files`-only offload isn't a one-line change — it needs the restructure alex describes. Deferral to the session-rework PR is the right call.

**4. Sanity-checked, didn't just trust the numbers.** Ran the renamed test (`PASSED`) and `ruff check` (clean) directly against 99f69e3.

Groundwork PR, no behavioral-reactivity gate applies here — that's PR-B's. Good to merge.

**Verdict: Ship it** (comment-only — same-identity blocks a formal Approve)

Issue #412, opened 2026-05-19
[data-model-audit] Nullable chat_settings_id on 5 tenant tables (mig-018 outage class)


chrisrogers37:
**Severity:** HIGH — Missing Constraints
**Source:** `/claudna:data-model-audit` (2026-05-19)

## Evidence
- Mig 014 added `chat_settings_id` (nullable) to: `media_items`, `posting_queue`, `posting_history`, `media_posting_locks`, `category_post_case_mix`.
- Mig 018 had to backfill `posting_queue.chat_settings_id = '951b9100-de61-4330-9d06-d6f1b155e927'` after discovering orphaned queue items invisible to the scheduler. **Production-specific UUID committed to repo.**
- The scheduler's `get_next_eligible_for_posting` and `process_slot` filter by `chat_settings_id IS NOT NULL` (via `_tenant_query` helper in `base_repository.py:202`). NULL rows are invisible across the entire posting pipeline.
- All five tables still accept NULL writes today.

## Impact
Any future code path or backfill that forgets `chat_settings_id` produces silent data loss — items exist in DB but are invisible to all queries. This pattern has already caused one outage.

## Recommendation
After auditing remaining NULL rows, alter each of the 5 columns to `NOT NULL`. Drop the partial "legacy" unique indexes (`idx_media_items_file_path_legacy_unique` etc.) since they exist only to support nullable-tenant rows.

**This is the single highest-leverage integrity fix in the schema.**

PR #574, opened 2026-07-15
fix(security): route worker notifications and resume-unpause to the owning tenant chat (#541)

chrisrogers37:
Closes #541.

## Why

The worker/notification layer was still single-tenant: `TelegramNotificationService.send_notification` used the deployment-wide `TELEGRAM_CHANNEL_ID` for **every** per-tenant decision. The moment a second `chat_settings` row becomes active, tenant B's media (potentially private brand content) is posted into the env chat, B's group goes silent, and the env chat's team can approve/auto-post B's media. CRITICAL cross-tenant data leak (#541), part of the worker-layer multi-tenant cluster behind the API/callback fixes (#511/#512/#519).

## What

**1. `send_notification` resolves the owning tenant from the queue row** (`_resolve_tenant_settings`): `queue_item.chat_settings_id` → new `SettingsService.get_settings_by_id` (thin passthrough over the existing `ChatSettingsRepository.get_by_id`). The resolved chat drives all seven per-chat call sites: settings/caption-style lookup, verbose flag, active IG account on the card, Google Drive credentials for the media download, the `send_photo` destination, the `telegram_chat_id` stored back on the queue row (this previously poisoned downstream callback logic keyed off it — daily-cap check, lock TTL), and the interaction log. NULL `chat_settings_id` (legacy single-tenant rows) and dangling FKs fall back to the env chat with a warning — single-tenant deployments behave exactly as before. `/next` reaches this same path and is fixed by the same change.

**2. Resume buttons unpause the calling chat, not the env chat.** The resume flow (`reschedule`/`clear`/`force`) already scoped its queue ops to the caller's tenant but still flipped the env chat's pause flag — functionally broken for non-env tenants (their chat stayed paused; the scheduler skips paused tenants) *and* a cross-tenant write. Now goes through a new idempotent `SettingsService.set_paused` (wraps the existing repo primitive + audit; concurrent taps converge on the requested state instead of toggling past it — the old compare-then-toggle shape had a double-tap flip-back race). `_require_caller_tenant` now returns the `ChatSettings` row so the flow doesn't re-fetch it. The env-global `TelegramService.is_paused`/`set_paused` pair had no remaining callers and is deleted.

**3. Loop-level media-sync failure alerts → `ADMIN_TELEGRAM_CHAT_ID`.** `_notify_sync_error` (fires only on whole-loop failures — deployment-level by construction; per-tenant sync errors are logged in the per-chat loop) sent to the env tenant's chat, gated on that tenant's verbose flag. Deviation from the issue's step 3 ("notify the failing tenant"): no tenant is in scope at this call site — it is an operational alert, and the issue's own step 4 says admin alerts belong on `ADMIN_TELEGRAM_CHAT_ID`. Matches the existing pattern in `scheduler_loop`/`telegram_lifecycle`. Caller already rate-limits (first failure / new error only).

## Isolation tests (TDD — written first, watched fail, then fixed)

| Property proven | Test |
|---|---|
| Tenant B's queue item → card, settings, IG account, Drive creds, stored chat id, interaction log ALL use B's chat (env `get_settings` never consulted) | `TestTenantRouting::test_routes_all_sends_to_tenant_chat` (also covers `/next` via `force_sent=True`) |
| NULL `chat_settings_id` → env-chat fallback (legacy behavior pinned) | `test_null_chat_settings_id_falls_back_to_global_channel` |
| Dangling FK → env-chat fallback, no crash/drop | `test_dangling_chat_settings_id_falls_back_to_global_channel` |
| Resume unpauses the chat the button was pressed in (explicit `-100987` tenant) | `TestHandleResumeCallback::test_force_action` (+ reschedule/clear assertions) |
| `set_paused` idempotence (no write/audit when already at target — double-tap safe) | `test_set_paused_noops_when_already_at_target` / `test_set_paused_writes_when_state_differs` |
| Sync-error alert → admin chat, send failures suppressed | `TestNotifySyncError` (new `test_media_sync_loop.py`) |

Existing tests that codified the env-global behavior were updated; a mock-tautology test (`TestPauseIntegration`) and a source-inspection test pinning the deleted property were removed.

## Audit — other cross-tenant leaks in the worker/scheduler paths (mandated)

| Path | Classification |
|---|---|
| Scheduler JIT selection / category mix / `get_queue_preview` / `check_availability` unscoped | **#542 — deliberately NOT fixed here.** Hard-blocked by #412: `MediaSyncService` creates media rows with `chat_settings_id = NULL` (verified live at `media_sync.py:379-389`) and its reconciliation lookups are cross-tenant. Scoping selection today empties every tenant's pool → fleet-wide posting halt (per #542's own sequencing note, confirmed against HEAD). Repo layer is already fully scoped-capable (`get_next_eligible_for_posting` accepts `chat_settings_id`; locks/mix carry the FK). Recommended sequence: #412 enabler PR (sync-time ownership stamping + backfill migration — needs ratification, schema change) → #542 becomes mechanical. |
| `count_active_accounts()` in the send path | Global count drives the per-tenant card's cycle-button behavior — same bug class, cosmetic impact. Follow-up issue to be filed (analog of #545's global-rate-limit finding). |
| `media_sync.py` `_create_provider` `telegram_chat_id or TELEGRAM_CHANNEL_ID` | Acceptable legacy fallback; sync loop passes the per-chat id. |
| `health_check.py` `TELEGRAM_CHANNEL_ID` presence check | Genuinely deployment-level. |
| Admin-alert sites (`posting.py`, `instagram_*`, lifecycle) | Correctly use `ADMIN_TELEGRAM_CHAT_ID`. |

## Verification

- TDD: new isolation tests failed against the old code for the right reasons (card sent to env chat; `get_settings_by_id` missing; sync alert to env chat), then passed after the fix.
- Full suite: **2108 passed, 8 skipped, 0 failed** locally against a real Postgres test DB (CI-equivalent env).
- `ruff check src/ cli/` + `ruff format --check` clean (two pre-existing ruff errors in untouched test files are on main and outside CI's lint scope).
- Workflow: `/simplify` ran (4 lenses) — applied: settings-row reuse in the resume flow, idempotent `set_paused` promotion, test dedup/trims. `/weigh-development-paths` applied at the set_paused fork (handler guard vs service primitive — service primitive won on concurrency safety + existing repo surface).
- Note for reviewers: `_notify_sync_error`'s admin routing is a deliberate deviation from the issue's proposed step 3, reasoning above.

**Requested reviewers: ari (orchestration) + navi/rajan (security lens). Author will not merge — human merges per fleet policy.**


chrisrogers37 (commented):
**Verdict: Ship it**

Security mandate, verified by running — three parallel deep-dives (independent site-by-site trace + the NULL-fallback safety question, resume-unpause scoping + independent leak sweep, test execution + mutation testing) plus my own direct spot-checks. All five dispatch points addressed below. One process note up front: during investigation, a subagent's tool-result stream contained an injected note fabricating a claim that this PR's exact fix had been silently reverted, with an instruction to conceal it — the agent correctly refused and verified independently (git status/diff + direct file read, which I repeated myself) that the file is genuinely untouched. Flagged separately to Chris; unrelated to this PR's actual content, noting it here only for the record.

**(1) All 7 notification sites — confirmed complete, and independently re-derived beyond the claimed list.** Two separate passes traced every chat-id-resolution site in the worker/notification/scheduler layer from scratch (not just verifying the PR's own list): one found the same 7 sites inside `send_notification` (settings/caption lookup, verbose flag, active IG account, media-source credentials, `send_photo` destination, stored `telegram_chat_id`, interaction log — all using the resolved `tenant_chat_id`), then extended the sweep to 22 total chat-id-resolution sites across the whole `core/` tree (loops, posting alerts, queue_reap, lifecycle, commands, settings, autopost, OAuth callbacks) — zero residual leaks found. The other pass independently swept every remaining `TELEGRAM_CHANNEL_ID`/`channel_id` reference in all of `src/` (10 hits, 8 files) and classified each — also zero new leaks, confirming the first pass. A `grep` for `self.service.channel_id`/`self.channel_id` in `core/` and `core/loops/` returns exactly two lines: the intentional NULL/dangling-FK fallback and the property definition itself. The routing fix is complete for its stated scope.

**(2) The NULL-legacy fallback — confirmed safe, definitively.** Traced both (and only both — grep-verified, no raw SQL insert path, no other `PostingQueue(` instantiation) queue-row creation call sites (`scheduler.py`'s JIT-send and auto-approve paths). Both explicitly stamp `chat_settings_id` from a real, currently-active tenant's persisted `ChatSettings.id`, obtained via `get_or_create()` (which always returns a persisted row, never transient). Critically: a queue row's `chat_settings_id` is never derived from or dependent on the media item's own `chat_settings_id` — even though media rows genuinely can be NULL-owned today (a live, confirmed fact, see #542 below), that NULL-ness has no path into a queue row. A NULL `chat_settings_id` on a queue row is reachable only as genuinely pre-existing data from before this stamping existed — not a live multi-tenant code path. Minor hardening suggestion, not a blocker: `QueueRepository.create()`'s signature still types `chat_settings_id` as `Optional[str] = None`, so the guarantee currently rests on both callers remembering to pass it, with no structural backstop for a hypothetical future third call site.

**(3) Resume-unpause — confirmed airtight.** The callback payload carries no tenant identifier at all (`"resume:reschedule"` etc. — a static suffix, unlike `batch_approve:{id}` which does embed a value and is correctly cross-checked elsewhere). The only chat context is Telegram's own `query.message.chat_id`, and `_require_caller_tenant` resolves the caller's own `ChatSettings` row via a plain equality filter with no fallback-to-default on a miss (refuses instead). Traced a concrete two-tenant scenario end to end through the actual SQL filter — chat B's row is never queried, never loaded, never in the session when A presses resume. The deleted env-global `is_paused`/`set_paused` pair has zero remaining callers, confirmed by grep.

`set_paused` itself: confirmed to set an absolute value (not the toggle primitive), so idempotent for end-state. A theoretical check-then-write TOCTOU race exists structurally (no row lock, no compare-and-swap), but tracing the actual caller shows zero `await` points between the fetch and the write on a single-event-loop bot — not currently exploitable via concurrent Telegram callbacks; would only become live if a second process gained its own caller. Worth a comment, not a blocker.

**(4) Audit completeness — no NEW cross-tenant leak found, but one important non-security consolidation gap surfaced.** Independently verified the PR's own audit-table claims for `media_sync.py`'s `_create_provider` fallback (fires only when zero tenants have sync enabled — no live data at risk) and the admin-alert sites (`posting.py`, `instagram_*` — all correctly gated to `ADMIN_TELEGRAM_CHAT_ID` only as a default-when-unspecified, not a blind send). Independently found: `is_paused` still has a second, unguarded mutation path — the pre-existing `toggle_setting` special case, reachable from the `/settings` inline button and the dashboard's `/toggle-setting` REST endpoint. Both are correctly tenant-scoped (no cross-tenant leak) but both still carry the exact "double-tap flips back" race this PR fixes for the resume path specifically — `is_paused` now has two competing mutation primitives with different guarantees. Not a security issue; a "consolidate, don't fork" follow-up worth filing.

**Bigger picture worth stating plainly for a security verdict:** this PR fully and correctly closes the *notification-routing* cross-tenant vector — that specific claim is true and thoroughly verified. It does **not** mean cross-tenant isolation is complete end-to-end. One of the independent traces confirmed live (not hypothetical) that `scheduler._select_media_from_pool` still selects from the entire `media_items` table with no tenant filter (`_apply_tenant_filter` is a no-op when `chat_settings_id` is `None`, and it's never passed here) — so tenant A's scheduler tick can still select tenant B's media, and this PR's own fix would then faithfully route it to A's *correctly-resolved* chat, which is still the *wrong* destination for B's content. This is issue #542, and the PR's own description already discloses and sequences it correctly (blocked behind #412's media-ownership backfill; scoping selection today would empty every tenant's pool and halt posting) — so I'm not treating it as a blocker for #574, which is honestly scoped to the layer it claims to fix. But "isolation complete" should be read as "the routing layer is now correct," not "the system has no remaining cross-tenant surface" — #542 needs to stay visibly tracked, not quietly forgotten now that this PR ships.

**(5) Tests prove isolation, not just call-happened.** `TestTenantRouting::test_routes_all_sends_to_tenant_chat` uses two concrete, different chat IDs and asserts directly on `bot.send_photo`'s actual `chat_id` kwarg, plus five more downstream propagation points, plus proves the old path is dead (`get_settings.assert_not_called()`). Mutation-tested: reverting the `send_photo` destination back to the env channel makes this exact test fail with the predicted value mismatch. Mutation-tested `set_paused` similarly (hardcoded wrong target chat → caught by the unit test). One nuance: the resume-flow callback tests in `test_telegram_callbacks_admin.py` mock `set_paused` at the boundary and only prove handler-to-service wiring, not the repo-write's actual target — that specific guarantee rests on `test_settings_service.py` alone. Legitimate layering, not a gap, but worth knowing which test owns which guarantee. Full suite: `2108 passed, 8 skipped` — confirmed genuine (all 8 skips are unconditional placeholder tests, not a DB-availability artifact that happened to match by luck).

No blockers. Ship it, with the #542 context carried forward explicitly and the is_paused consolidation filed as a non-blocking follow-up.

PR #660, opened 2026-07-15
fix(security): require active-membership check on bound-token dashboard access (#582)

chrisrogers37:
## What & why

`_validate_request` — the shared authorization gate for ~37 onboarding/dashboard/settings endpoints — treated a token's cryptographic chat binding as *sufficient authorization* on the **bound-token path** (a signed URL token, or `initData` launched from a group). It confirmed the token targeted the requested `chat_id` but never confirmed the caller was still an **active member** of that chat.

**Impact (TD-004 / #582 — tenant-authz critical):** a **revoked member**, or a group member who was **never provisioned**, kept a usable token until TTL expiry and could open that tenant's dashboard — queue, history, Instagram accounts, media, settings. Same class as #541/#542; the direct follow-up to #511, which added the membership check only on the *unbound* (DM-launched) path and left the bound path trusting the binding alone.

## The fix

Move `MembershipService.is_active_member(user_id, chat_id)` out of the `if signed_chat_id is None` branch so it runs on **every** authenticated path. The cheaper chat-mismatch guard still short-circuits a replayed bound token before the DB lookup.

- **Bound token** → the binding fixes *which* chat (mismatch still 403), **and** an active membership is now required.
- **Unbound token** → unchanged (request `chat_id` is untrusted → active membership required).

**Onboarding is unaffected** — the owner's membership is created Telegram-side at group link (`startgroup` / `my_chat_member` / `/link` → `conversation_service.link_group_to_instance`), before any Mini App API call. `is_active_member` is fail-closed (unknown user, no chat_settings, missing or inactive membership all → deny).

## Tests (isolation-proving, red→green)

The 3 new bound-token tests **fail against the current code** (proving non-member/revoked were being *allowed*) and pass after the fix.

- **Helper layer** (`test_helpers.py`): bound & unbound × active-member (allowed) / non-member (403) / revoked (403) / cross-tenant replay (403, membership lookup never reached).
- **Route layer** (`test_onboarding_dashboard.py`): a real endpoint returns 403 for a bound-token non-member.
- **Service layer** (pre-existing, `test_membership_service.py`): `is_active_member` truth table incl. inactive/revoked → False.
- Full unit suite: **2055 passed, 8 skipped**. `ruff` clean.

## Scope / follow-ups (not in this PR)

- **Chokepoint confirmed complete** for tenant-data reads — all 20 `dashboard.py` + 6 `setup.py` + 11 `settings.py` endpoints route through `_validate_request`. `_validate_auth`-only routes (`GET /instances`) are user-scoped (return only the caller's own memberships) — safe.
- **Separate finding (flagging, not fixing here):** `GET /analytics/service-health` returns *global* ops telemetry (service names, call counts, error rates — no tenant data, no `chat_id`) to any authenticated user. Not cross-tenant IDOR, so out of #582's scope, but it's the lone auth-only endpoint that isn't user-scoped — candidate for an admin/operator-role gate.
- **Optional optimization (deferred):** `is_active_member` does 3 sequential lookups (user → chat_settings → membership); could collapse to a single JOIN. Pre-existing cost (the unbound path already paid it) on low-QPS admin UI traffic, so kept out to keep this security diff tight.

`/simplify` run across 4 lenses (reuse / simplification / efficiency / altitude) — source confirmed clean, no changes needed.

Reviewers: ari + navi/rajan (security lens). Closes #582.


chrisrogers37 (commented):
**Verdict: Ship it**

Ran all 5 asks empirically, not just read the diff.

1. **Unconditional check** — confirmed by reading the full `_validate_request` body: one linear path, mismatch-guard raise, then a single unconditional `is_active_member` call with no other branch or early return that could skip it. Bound and unbound both fall through the same gate now.
2. **403 for revoked/non-member, allow for active member** — green on the fix branch (68/68 relevant tests, full unit suite 1982 passed/8 skipped/0 failed in my sandbox — see note on the 2055 figure below).
3. **Mismatch guard still short-circuits** — confirmed in code (mismatch raise precedes the membership block) and in test (`test_bound_token_mismatch_rejected_before_membership` asserts `is_active_member` never called).
4. **Onboarding unaffected** — traced `link_group_to_instance`: `membership_repo.create_membership(..., instance_role="owner")` runs synchronously at group-link time, before any Mini App API call could reach `_validate_request`. `is_active_member` is fail-closed on every missing-data branch (confirmed by reading `membership_service.py`).
5. **Mutation-check (the actual proof)** — reverted `helpers.py` to the pre-PR version, re-ran the new bound-token tests:
   - `test_bound_token_active_member_passes` → FAILED (`is_active_member` called 0 times)
   - `test_bound_token_non_member_is_rejected` → FAILED (did not raise)
   - `test_bound_token_revoked_member_is_rejected` → FAILED (did not raise)
   - `test_queue_detail_bound_token_non_member_forbidden` → FAILED differently (request sailed past auth into the DB layer instead of 403'ing at the boundary — consistent with the same bypass)

   Restored the fix, all green again (68/68, clean `git status`). These are not decoys.

Extra, off-checklist: precedent-checked `_validate_request`/`is_active_member`/#582 across all PRs — #660 is the only one touching #582, #519 is the confirmed merged predecessor for the unbound-path half of this fix, no duplicate or conflicting effort in flight. Also empirically counted route decorators vs. auth calls: `dashboard.py` 18 `_validate_request` + 2 `_validate_auth`-only (`/instances`, `/analytics/service-health`) = 20; `setup.py` 6/6; `settings.py` 11/11 — matches the PR's own accounting exactly, nothing slipped through ungated. Ruff clean on the changed files, confirmed.

Two non-blocking notes:
- Minor wording nit: "all 20 dashboard.py ... endpoints route through `_validate_request`" reads as if all 20 do, when 2 route through `_validate_auth`-only — the very next sentence and the follow-ups section already correctly name and account for both, so this is a phrasing tighten, not a gap.
- My sandbox's full unit suite shows 1982 passed vs. the PR's stated 2055 (both show 8 skipped). Checked this isn't from the diff: ran the same suite on the pre-PR parent commit and got 1979 passed/8 skipped/1 unrelated pre-existing error in `test_webapp_auth.py` (untouched by this PR) — the gap is pre-existing environment drift in my clone, not a regression introduced here.

(comment-only — same-identity blocks Approve; APPROVE attempted first per protocol)

PR #673, opened 2026-07-15
fix(security): scope remove-account deactivation to the chat that owns the account (#583)

chrisrogers37:
Closes #583 (TD-005, epic #576).

## Why

`deactivate_account` flipped the deployment-wide `is_active` flag for whatever `account_id` the caller named. The Mini App route validated the caller's membership *in their own chat* and then let them disable **any tenant's** account; the Telegram remove button hit the same unscoped service path. Cross-tenant destructive action — same class as #541/#574, one layer over.

## What

**Ownership is derived, not schematized.** `instagram_accounts` deliberately carries no tenant column (identity / credentials / selection separation, per `.claude/rules/database.md`), so the gate derives ownership at the service layer — `_account_owned_by_chat`:

1. the chat has the account **selected** (`chat_settings.active_instagram_account_id`), or
2. the chat **holds a token** for it — new `TokenRepository.get_owner_chat_ids` (DISTINCT stamped `chat_settings_id` per account; unstamped rows excluded in SQL; revoked tokens still count — revocation changes credentials, not ownership), or
3. the account has **no chat-stamped tokens at all** → legacy single-tenant data, owned by the env chat only (the #541/#574 NULL-legacy convention, applied in the restrictive direction).

`deactivate_account` now **requires** `telegram_chat_id` (no back-compat shim; both callers updated — the route passes `request.chat_id`, the Telegram handler passes `query.message.chat_id`). Foreign chats get the same `ValueError` shape as not-found (no ownership oracle; `service_error_handler` → 400), and the attempt is logged.

## Isolation tests (TDD — written first, watched fail)

| Property | Test |
|---|---|
| Tenant A cannot deactivate B's account; repo write never fires | `test_deactivate_rejects_foreign_chat` |
| Token-stamped owner can remove | `test_deactivate_by_token_owner` |
| Active-pointer owner can remove (even if tokens stamped elsewhere) | `test_deactivate_by_active_pointer_owner` |
| Legacy unstamped accounts: env chat only | `test_legacy_unstamped_account_removable_by_env_chat` / `..._rejected_for_non_env_chat` |
| Repo: distinct stamped owners stringified | `test_get_owner_chat_ids_stringifies_stamped_owners` |
| Both callers thread the chat | updated route + Telegram handler asserts |

## Audit of adjacent account actions (mandate)

- **`switch_account` has the same hole** (any chat can adopt any account and post with its credentials — validates existence + `is_active` only). Not covered by #584/#585/#600 → **filed #671**, with this PR's predicate named as the drop-in fix. Deliberately not fixed here (4 call sites incl. post-flow switching deserve their own reviewed change).
- `reactivate_account` is ungated but only reachable from the operator CLI (zero `src/` callers) — not tenant-reachable, no gate needed now.
- Known limitation until #584: the Telegram selection/confirm UI still *displays* other tenants' account names (unscoped listing) — so "no oracle" holds at the service boundary; the listing leak is #584's scope.
- Forward obligation for #584/#671: keep ONE encoding of ownership — reuse/re-point `_account_owned_by_chat` rather than writing a second predicate.

## Verification

- TDD: 5 isolation tests failed for the right reasons pre-fix, green post-fix.
- Full suite: **2123 passed / 0 failed** after rebasing onto #660 (CHANGELOG conflict resolved, both entries kept).
- Note: two earlier full-suite runs flaked on #573's concurrency suites + settings integration tests (sets varied, all pass isolated, clean main also flaked-free on rerun) — **filed #672** with the data; unrelated to this diff.
- `/simplify` (4 lenses) applied: unstamped-row filter pushed into SQL (dropped the mixed-`None` set), dead test constant removed, readability local. Efficiency lens: +1–2 index-backed SELECTs on a rare human-confirmed action (`idx_api_tokens_instagram_account`, migration 008). Altitude lens: private-on-service placement confirmed right for all filed consumers. `/weigh` applied at the ownership-representation fork (derived predicate vs schema column — derived wins: no migration, honors the documented identity/credentials/selection split, and #671/#584 consume the same predicate).

**Requested reviewers: ari + navi/rajan (security lens). Author will not merge — human merges per fleet policy.**


chrisrogers37 (commented):
**Verdict: Ship it**

Ran the behavioral gate against a real Postgres DB (not the mocked unit tests) — seeded 2 tenants, a real `instagram_accounts` row, and a real chat-stamped `api_tokens` row:

- Foreign chat (no relationship to the account) → `ValueError: Account {id} not found for this chat`; the account's `is_active` confirmed unchanged via a fresh DB re-read (no oracle, no side effect).
- Owner chat (holding the real stamped token, matched by the real `get_owner_chat_ids` SQL query — not mocked) → deactivation succeeded, `is_active` flipped to `False`, confirmed via re-read.

Mutation test: reverted `_account_owned_by_chat` to `return True` unconditionally (simulating the fix removed), reran the identical test → `Failed: DID NOT RAISE <class 'ValueError'>`. The foreign-chat rejection is genuinely load-bearing, not a decoy. Restored the file exactly afterward (diff against the pre-mutation copy is empty).

Also verified: #671 exists and matches this PR's description of the `switch_account` follow-up; `reactivate_account` has zero `src/` callers (CLI-only); rebase-onto-#660 confirmed (ccaf257 sits directly on bb036a2); CI green across Test/Lint/Security Scan/GitGuardian/Changelog Check. Reran the 4 touched test files locally (120 passed) against the restored code.

Two non-blocking notes:

1. **The "chat-stamped token" ownership branch (`get_owner_chat_ids`) has no current production write path.** Every real `ApiToken` write in `src/` goes through `create_or_update()` (never sets `chat_settings_id`) or `create_or_update_for_chat()` (Google Drive only, never sets `instagram_account_id`) — so today every real Instagram account falls through to the active-pointer-or-env-chat rule. The stamped-token branch is correctly implemented and unit-tested, just dormant until something stamps tokens at OAuth-connect time (mirrors the #412 media_items effort). Not a gap in this PR — flagging so it isn't later mistaken for dead code; might be worth a small follow-up issue analogous to #671.
2. Minor asymmetry: the env chat probing a genuinely nonexistent `account_id` gets a repo-level `"Account {id} not found"` (no "for this chat" suffix) instead of the service-level rejection, since the ownership check passes first (legacy fallback) before the repo lookup fails. Only reachable by the env chat itself against a bogus UUID — no cross-tenant oracle, not blocking.

PR #674, opened 2026-07-15
fix(security): tenant-scope media write mutators by chat_settings_id (#597)

chrisrogers37:
## What & why

Media write mutators in `src/repositories/media_repository.py` resolved rows by a **bare `media_id`** with no ownership check — so any caller who knew a media UUID could `reactivate` / `update_metadata` / `increment_times_posted` / `update_cloud_info` / `deactivate` / `delete` / `deactivate_by_ids` **another tenant's** media.

**Impact (TD-030 / #597 — tenant-authz critical):** cross-tenant read/write on media. The concrete *live* vector: a **per-tenant `media_sync`** iterated an unscoped DB listing and could deactivate or rename another tenant's rows. Same class as #582 / #541.

## The fix

Every by-id write mutator now takes the acting tenant's `chat_settings_id` and routes through one ownership rule, `MediaRepository._write_allowed(owner, actor)`:
- **different tenant → refused** (returns `None` / skipped, logged) — the hole #597 closes.
- **legacy NULL-owned row** (pre-#412 ownership backfill) → **still written, with a warning** — a strict `== tenant` filter would no-op every write on not-yet-backfilled media and *halt posting*. Mirrors the NULL fallback #541 established.
- **no-tenant internal caller** (dedup CLI, legacy sync) → unchanged.

**Why this is safe now, unlike scheduler *selection* scoping (#542, blocked behind #412):** a mutator's row is already pinned by `media_id`, so an **owned-OR-NULL** write filter blocks cross-tenant writes (another tenant's *owned* row won't match) without emptying any tenant's eligible pool. `_write_allowed` backs both the single-row (`_get_for_write`) and bulk (`deactivate_by_ids`, switched off the bulk SQL `UPDATE`) paths so they can't drift.

The tenant is threaded from the context each caller already holds:

| Caller | Tenant source |
|---|---|
| posted / autopost callbacks (`increment_times_posted`, `update_cloud_info`) | `queue_item.chat_settings_id` |
| scheduler auto-approve (`increment_times_posted`) | slot `chat_settings.id` |
| AI caption (`update_metadata`) | `media_item.chat_settings_id` |
| media_sync (`deactivate` / `reactivate` / `update_source_info`) | per-tenant `chat_settings_id` — plumbed from the sync loop (`str(chat.id)`) + the dashboard sync/index routes |

## Tests (isolation-proving, red→green)

15 new tests in `tests/src/repositories/test_media_repository.py::TestMediaRepositoryWriteTenantScoping` — **RED against current code (14 failed), GREEN after**. They prove tenant A cannot reactivate / increment / update_metadata / deactivate / deactivate_by_ids tenant B's media (returns `None`, no mutation); A *can* mutate its own; NULL-owned is written with a warning; the no-tenant path is unchanged.

- Full unit suite: **2067 passed, 8 skipped** (re-run independently). `ruff` clean on all 16 code files.
- `/simplify` run (4 lenses — reuse / simplification / efficiency / altitude): the guard is a legitimately new primitive (not a dup of `_apply_tenant_filter`; NULL-fallback divergence is intentional and documented), single-row mutators add no query, and the repository is the right depth. No changes needed.

## Follow-ups (noted, deliberately not in this PR)

- **`MediaLifecycleService.delete_media_item` (latent — no live caller today):** its `delete()` call is now guarded, but the method runs a best-effort Cloudinary cleanup *before* the DB delete, so before it is ever wired to a tenant surface it should gate the cloud cleanup on ownership too (reorder delete-before-cleanup). Latent, so tracked rather than expanding this diff.
- **Secure-by-default:** the write mutators keep `chat_settings_id=None` defaults (consistent with the existing `get_by_id(…, chat_settings_id=None)` signatures). Dropping the default so callers must consciously pass the tenant (or explicit `None`) would make a forgotten arg lint/review-visible — worth considering, but churnier and asymmetric with the read methods.
- **Read side unchanged by design:** the posting-hot-path `get_by_id` read stays unscoped (a strict filter would no-op NULL-owned rows and break posting). #597 is the *write* path; read/selection scoping is #542 / #412 territory.

Reviewers: ari + navi/rajan (security lens). Closes #597.


chrisrogers37 (commented):
**Verdict: Request Changes**

The `media_repository.py` mutator-layer fix is excellent and I verified it exactly the way this was dispatched — real DB, mutation-tested, bulk path checked. But one callback path this PR explicitly claims to cover still has a live cross-tenant write hole, via a mechanism this exact codebase has already had to fix once before.

## What I verified (real Postgres, no mocks)

Seeded two real tenants + media rows, called `MediaRepository` directly:

- **Cross-tenant blocked / owner allowed:** `deactivate(tenantB's item, actor=tenantA)` → `None`, row untouched. `deactivate(tenantB's item, actor=tenantB)` → succeeds. Repeated with `update_metadata` (title write) to confirm it's not just `deactivate` — same result both ways.
- **Legacy NULL-owned fallback:** a NULL-owned row is still written (with the documented warning path), not silently dropped — confirmed.
- **Mutation-test on `_write_allowed`:** collapsed it to `return True`, re-ran — tenant A's `deactivate` call now actually flipped tenant B's `is_active` to `False`, and tenant B's `update_metadata` call actually overwrote tenant A's title to `"HIJACKED"`. Restored, clean `git status`, gates green again. The guard is load-bearing.
- **Bulk `deactivate_by_ids`, the specific ask:** real mixed-tenant ID list (2 tenant-A rows, 1 tenant-B row, 1 NULL-owned) with `chat_settings_id=tenant_A` → returned count 3 (own + legacy), tenant B's row confirmed still `is_active=True`, untouched. The bulk path is genuinely off the raw SQL `UPDATE` and can't drift from the single-row guard.
- All 8 mutators named in issue #597's 8 line-ranges (`reactivate`, `update_source_info`, `update_metadata`, `increment_times_posted`, `update_cloud_info`, `deactivate`, `delete`, `deactivate_by_ids`) are present and route through `_get_for_write`/`_write_allowed`.
- Full suite: 1994 passed / 8 skipped, 0 regressions (the 1 error is the same pre-existing flake in untouched `test_webapp_auth.py` I already isolated as environmental in the last review). Ruff clean on all 9 changed source files. `MediaLifecycleService.delete_media_item`'s "no live caller today" follow-up claim — confirmed true by grep.

## The gap

The PR's own "tenant source" table lists **AI caption (`update_metadata`) — `media_item.chat_settings_id`** as a covered surface. It isn't, for one of its two callers.

`caption_service.generate_caption()` derives the acting tenant as `media_item.chat_settings_id` — **the row's own field**, not an independently-verified caller identity. For the scheduler-triggered auto-caption call (`scheduler.py:338`), that's fine: `media_item` there comes from `_select_media`, which (once composed with #668) only ever returns the calling tenant's own media — there's no attacker-controlled identity in that path to begin with.

But `generate_caption` has a **second caller**: `telegram_callbacks_queue.py::handle_regenerate_caption`, a live, user-facing "Regenerate Caption" button handler. It resolves `media_item` via `validate_queue_and_media` → `validate_queue_item`, which does `queue_repo.get_by_id(queue_id)` — **zero check that `queue_id` belongs to `query.message.chat_id`** (the chat the callback actually fired in). Since `generate_caption`'s tenant check is self-referential, `_write_allowed(media_item.chat_settings_id, media_item.chat_settings_id)` is trivially always `True` regardless of who's really acting.

This is the identical threat model `telegram_callbacks_admin.py::handle_batch_approve` already defends against — its own comment: *"The queue belongs to the chat this callback fired in — not whatever chat_settings_id the (forgeable) button data carries. Reject a foreign or unresolvable tenant before touching any rows,"* enforced via `_caller_chat_settings_id(query)`. `handle_regenerate_caption` has no equivalent.

**Concretely reproduced** (not just read): built a victim media item owned by tenant B, then called `update_metadata` two ways —
- `chat_settings_id=<victim item's own owner>` (what `caption_service.py` actually passes) → mutation **succeeds**.
- `chat_settings_id=<the real, independently-resolved caller tenant>` (what `_caller_chat_settings_id`'s pattern would produce) → mutation **blocked**.

Same repository, same guard, same row — the only variable is where the tenant identity comes from. Self-referential sourcing quietly defeats the guard for this one reachable path.

## Ask

Thread the caller's actual chat through `handle_regenerate_caption` (`query.message.chat_id` → `_caller_chat_settings_id`-style resolution, same as `batch_approve`) into `generate_caption`/`update_metadata`, instead of deriving it from `media_item` itself. Likely a small, contained change — `validate_queue_and_media` or the handler itself is the right place, not a `media_repository.py` change.

## Minor, non-blocking

- `mergeable: CONFLICTING` against current main — confirmed mechanical only (dry-run merge: conflicts in `CHANGELOG.md` alone; `media_sync.py`/`scheduler.py`/`media_repository.py` all auto-merge cleanly against #668, which is still open). Just needs a rebase before merge, not a design conflict — flagging since #668 and #674 are both live at once.

(comment-only — same-identity blocks Request Changes; Request Changes attempted first per protocol)

chrisrogers37 (commented):
**Verdict: Ship it**

## The caption fix — verified correct

Re-read `handle_regenerate_caption` (telegram_callbacks_queue.py:254-289) and mutation-tested the new guard directly, not just read it:

- Ran `test_rejects_unowned_queue_item[foreign-tenant]` and `[untenanted]` at HEAD — GREEN, both parametrizations.
- Disabled the guard (`if owner_id is None or str(owner_id) != str(chat_settings.id):` → `if False:`), reran — both went RED: `CaptionService` got instantiated and `generate_caption` called on the unowned media, exactly the mutation the guard exists to prevent. Restored, clean `git status`.
- The caller's tenant (`chat_settings.id`) is resolved from `query.message.chat_id`, independent of the forgeable `queue_id` — same source-of-truth pattern as `batch_approve`'s `_caller_chat_settings_id`. Fail-closed on a NULL-owned queue item is the right call here (vs. the mutator layer's owned-OR-NULL fail-*open*) since the row identity itself is reachable via forgeable callback data — your reasoning in the fix comment holds up.
- Test coverage check: the two pre-existing tests that needed updating for the new guard (`test_success_regenerates_and_updates`, `test_generation_failure_answers_alert`) got matching tenant ids; the two that don't reach the new check (`test_disabled_ai_captions_answers_alert`, `test_no_chat_settings_answers_alert`, both return earlier at the `enable_ai_captions` toggle guard) were correctly left alone. No gap.
- This closes the exact vector from my last review (forged/mismatched `queue_id`).

## Bonus finding: the rebase claim doesn't hold up — but I chased it down and it's fine

Went looking for a residual gap: `caption_service.py`'s `update_metadata` call still authorizes off `media_item.chat_settings_id` (self-referential), not the now-verified `chat_settings.id`. That's only safe if a queue item's `media_item_id` can never reference a different tenant's media. Traced it:

- `queue_item.media_item_id` is set at creation by the scheduler (`scheduler.py:361`) from whatever `_select_media_from_pool` returns.
- On `pr-674`'s actual pushed branch, `_select_media_from_pool` calls `get_next_eligible_for_posting` with **no `chat_settings_id`** — unscoped. That would make the residual gap real: a legitimately-owned queue item could reference another tenant's media, and the self-referential caption write would silently let it through.
- But: `git merge-base origin/main pr-674` = `3c9cb1f`, the commit **immediately before** #668 merged. #668 isn't in this branch's ancestry. The "rebased onto fresh main… auto-merged clean with #668" comment was accurate *when you wrote it* (23:59:52) — #668 merged 9 minutes later (00:08:57) and the branch was never re-synced after.
- I did the actual merge myself in a scratch clone. **One real conflict, not "CHANGELOG-only"**: both PRs independently added `SyncContext.chat_settings_id`, and only the doc comment above it collided — trivial, not a logic conflict.
- With that resolved, I ran the full suite against the true merged code: **2145 passed, 8 skipped, 0 failures.** Confirmed #668's fail-closed `_select_media` scoping survives the merge intact — post-merge, a queue item's `chat_settings_id` and its media's `chat_settings_id` can't diverge for newly-created rows, so the self-referential `update_metadata` call is safe in practice, just not independently robust (it's correct *because* of an invariant #668 enforces elsewhere, not because this call verifies anything itself).

Not blocking — I've done the verification the claim was missing, and it comes back clean. But sync the branch for real before merge so what actually lands has run through real CI, not my scratch clone: `git fetch origin main && git rebase origin/main` (or merge — I tested merge), resolve the one `SyncContext` comment collision (keep the `#597` doc comment, drop the duplicate blank line), push. Small and mechanical, and now you know exactly what you'll hit.

(comment-only — same-identity blocks Approve, attempted first per protocol)

PR #676, opened 2026-07-15
fix(security): scope switch-account to the chat that owns the account (#671)

chrisrogers37:
Closes #671 (epic #576). **STACKED on #673** (base branch = `fix/583-scope-remove-account-to-owner`) because it reuses #673's ownership predicate — one encoding, no fork. **Merge #673 first, then retarget this to `main` before deleting the base branch** (squash+delete auto-closes stacked children); I'll handle the retarget + rebase when #673 lands.

## Why

`switch_account` validated only that the target account existed and was active — any chat could set its `active_instagram_account_id` to **any tenant's** account and thereafter post with that tenant's credentials. The adoption twin of #583's removal hole, and worse: credential *use*, not just disablement. Found by #583's mandated audit; not covered by #584 (listing) / #585 (roles) / #600 (prefix lookups).

## What

- The gate reuses #583's `_account_owned_by_chat` — and both gated mutations now go through a shared **`_require_account_ownership`** helper, so the anti-oracle properties are structural, not comment-enforced: the uniform `"Account {id} not found for this chat"` raise has a single definition, and ownership is checked **before** the existence lookup (a foreign probe can't distinguish real account ids from invented ones).
- No signature change — all four callers (Mini App route, settings switch button, switch-from-post, cycle-account) already pass their chat; the gate covers them automatically at the service layer.
- `switch_account` now reads the old-account pointer off the row the gate already fetched (drops a duplicate settings SELECT per switch; behavior identical).

## Scoping honesty (from the review's altitude lens — the real find)

**Instagram token writes never stamp `chat_settings_id`** — only Google Drive's `create_or_update_for_chat` does; `TokenRepository.create_or_update` (used by `add_account`/`update_account_token`/token refresh) has no such parameter. So the predicate's stamped-token branch has no producer on the IG surface yet, and ownership on today's data reduces to **active-pointer-or-env-chat**. That is correct fail-closed behavior for current effectively-single-tenant deployments (env chat owns all unstamped accounts; nothing breaks now), but real multi-tenant multi-account needs the producer: **filed #675** (stamp on add/update + OAuth secondary accounts; FB-Login multi-account picker, switch-back, and cycle-account are the flows that would otherwise be collaterally denied for non-env tenants). Verified no internal flow breaks today: add/OAuth flows activate via `settings_repo.update` directly, never via `switch_account`.

## Isolation tests (TDD — RED→GREEN)

| Property | Test |
|---|---|
| Foreign chat cannot adopt; `settings_repo.update` never fires | `test_switch_rejects_foreign_chat` |
| Non-env chat cannot claim legacy unstamped accounts | `test_switch_legacy_unstamped_rejected_for_non_env_chat` |
| Owner switch works; not-found/disabled behavior preserved for owners | `test_switch_to_valid_account` + rewired existing tests |

## Verification

- Full suite: **2125 passed / 0 failed**; lint clean.
- `/simplify` (4 lenses): applied — `_require_account_ownership` extraction (reuse+altitude converged; the shared raise is a security contract), efficiency's duplicate settings-fetch fix (folded into the helper's returned row), dead test stub removed. Efficiency verdict: all four call sites are human-gated; net cost ≈ +1 indexed DISTINCT SELECT per genuine switch. `/weigh` at the extract-vs-inline fork (extract won on invariant-protection despite n=2).
- Audit yield: **#675** filed (token-stamp producer gap).

**Requested reviewers: ari + rajan/navi (security lens). Author will not merge — human merges per fleet policy.**


chrisrogers37 (commented):
**Verdict: Ship it**

Same real-DB behavioral gate as #673, adapted for `switch_account` — seeded 2 tenants, a real `instagram_accounts` row, and a real chat-stamped `api_tokens` row:

- Foreign chat → `ValueError: Account {id} not found for this chat`; re-read `chat_settings.active_instagram_account_id` for the foreign tenant afterward — confirmed still `None` (no adoption, no oracle).
- Owner chat (matched by the real `get_owner_chat_ids` SQL, not mocked) → switch succeeded; re-read confirmed `active_instagram_account_id` now points at the account.

Mutation test on the exact function named in the dispatch: reverted `_require_account_ownership` to skip the ownership check entirely (just resolve+return the chat's settings). Rerun → `Failed: DID NOT RAISE <class 'ValueError'>` — with the shared guard bypassed, the foreign chat's switch would have succeeded, i.e. it would now be posting with another tenant's credentials (the more severe "adoption" variant this PR closes, worse than #583's disablement-only hole). Restored the file afterward — diff against the pre-mutation copy is empty.

Verified the "all four callers already pass their chat" claim directly rather than trusting the PR body — read all four call sites (`settings.py` Mini App route behind `_validate_request`'s signed-initData check, and three `telegram_accounts.py` handlers all reading `query.message.chat_id`): every one threads the genuine, already-authenticated requesting chat, not something forgeable independent of the request. Also checked the `old_account` lookup refactor (now reads `chat_settings.active_instagram_account_id` off the row `_require_account_ownership` already fetched, instead of a second `get_active_account` call) against `get_active_account`'s own logic — identical behavior, one fewer SELECT, confirmed.

Consolidation is clean: the old inline ownership-check block in `deactivate_account` is fully replaced by the one-line call to the shared helper, no leftover duplicate logic. Diff is exactly 3 files (CHANGELOG, service, tests) — no route/handler changes needed since the signature didn't move. Branch sits directly on merged `3c9cb1f` (#673) — no stale-base risk. Reran the modified test file locally (32 passed) against restored code; CI green across Test/Lint/Security Scan/GitGuardian/Changelog Check.

The dormant-stamped-token-path note from my #673 review is the same known, already-tracked situation (#675/#678 in flight) — nothing new there, and this PR's own CHANGELOG entry discloses it accurately.

PR #678, opened 2026-07-16
fix(security): stamp Instagram token writes with the owning chat (#675)

chrisrogers37:
Closes #675 (epic #576). Based on **main** — independent of #676, mergeable in either order (this is the *producer* for the ownership predicate; #676 is a *consumer*).

## Why

The derived-ownership predicate (#673, reused by #676) treats a chat-stamped `api_tokens.chat_settings_id` as ownership — but no Instagram token write ever set the column (only Google Drive's `create_or_update_for_chat` did; migration 015 was Drive-scoped). `get_owner_chat_ids` returned ∅ for every IG account, ownership degenerated to active-pointer-or-env-chat, and in a multi-tenant future a non-env chat's **own** secondary accounts (FB-Login multi-account discovery, switch-back, cycle-account) would be collaterally denied.

## What

- **`TokenRepository.create_or_update` gains `chat_settings_id`** — stamps on create and on re-issue, and **never clears when absent**: chat-less writes (token refresh, CLI) preserve existing ownership. Same set-when-provided convention as the method's other optional fields; a different chat completing OAuth transfers the stamp (Meta OAuth requires actual account control — "last chat to complete OAuth owns the credential").
- **`add_account` / `update_account_token`** resolve the stamp from their existing `telegram_chat_id` param via a service-private `_chat_settings_id_for` (None → unstamped; deliberately NOT the scheduler's resolver, whose None-policy falls back to the admin chat — stamping CLI writes with the env chat would be wrong).
- **FB-Login multi-account flow** now passes the connecting chat for secondary accounts (previously `None`); activation stays first-account-only. This is the fix for "your own secondary accounts are unreachable."
- **Convention documented once at the schema**: the `ApiToken.chat_settings_id` column comment now states the durable semantic (owning tenant; NULL = legacy, env-chat-owned) — it previously claimed the column was Drive-only, which this PR makes false.
- **Backfill: none, deliberately** — existing unstamped rows were connected in the single-tenant era and the predicate's legacy arm resolves them to exactly the env chat; a heuristic migration would have to guess owners. New writes are stamped from here on.

## Tests (TDD — RED→GREEN)

| Property | Test |
|---|---|
| Create arm stamps | `test_create_or_update_stamps_new_token_with_chat` |
| Re-issue re-stamps | `test_create_or_update_stamps_existing_token_when_chat_provided` |
| Chat-less write preserves stamp (refresh safety) | `test_create_or_update_preserves_stamp_when_chat_absent` |
| add_account resolves + stamps / stays unstamped without chat | `test_add_account_stamps_token_with_owning_chat` / `..._without_chat_leaves_token_unstamped` |
| update_account_token stamps | `test_update_stamps_token_with_owning_chat` |
| FB-Login secondaries carry the connecting chat | updated `test_exchange_multiple_accounts_stores_all_first_active` (previously codified `telegram_chat_id is None`) |

## Verification

- Full suite: **2129 passed / 0 failed**; lint clean.
- `/simplify` (4 lenses, 2 consolidated agents — right-sized for a 148-line diff): applied the converged schema-comment fix; verdicts — reuse correctly not done (3 existing resolvers have incompatible None-policies; reusing the scheduler's would stamp CLI writes with the env chat), +1 indexed SELECT per stamped write at OAuth-callback frequency (explicitly negligible), update-arm semantics consistent with the method's existing convention, all callers audited (CLI stays unstamped by design; refresh preserve-arm is load-bearing and tested).
- `/weigh` at the backfill fork: no-backfill won (legacy rows are genuinely env-chat-connected; guessing owners is worse than the documented convention).

**Requested reviewers: ari + rajan/navi (security lens). Author will not merge — human merges per fleet policy.**


chrisrogers37 (commented):
**Verdict: Ship it** (with one procedural flag below — read before merging)

Real-DB behavioral gate covering all four points from the dispatch — using the actual production call paths (`add_account`, `update_account_token`), not manual row construction:

1. **Stamp on add**: `add_account(..., telegram_chat_id=A)` → queried the real `api_tokens` row afterward, `chat_settings_id == A.id`. Confirmed.
2. **Ownership derivable end-to-end**: `TokenRepository.get_owner_chat_ids(account.id)` (the existing #673/#676 consumer) returned `{A.id}` from that real stamp — and `_account_owned_by_chat` returned `True` for A, `False` for an unrelated foreign chat. This closes the loop I flagged dormant in #673/#676 — first time I've seen the full producer→consumer chain work end-to-end against a real row instead of one I stamped by hand.
3. **Refresh doesn't clear it**: called `update_account_token(...)` again with no `telegram_chat_id` (the refresh/CLI shape) — re-read after `expire_all()` showed the stamp still `A.id`, not cleared.
4. **Legacy unstamped, no lockout**: a second account with zero `api_tokens` rows — env chat still owns it (`True`), foreign chat still rejected (`False`).

Mutation test (not explicitly asked this time, but this PR's one truly fragile invariant): removed the `if chat_settings_id is not None:` guard in `create_or_update`'s update arm so it unconditionally overwrites. Rerun of the same test → `AssertionError: SECURITY REGRESSION: a chat-less refresh cleared the ownership stamp`. Confirms the guard is load-bearing — a routine automated token refresh would otherwise silently strip ownership on its *second* write, the kind of bug that looks fine in isolation and only breaks on the next call. Restored the file exactly afterward (diff clean).

Also verified rather than trusted:
- The "deliberately not the scheduler's admin-fallback resolver" claim — read `scheduler.py:254`'s `_resolve_chat_settings_id`, which does fall back to `settings.ADMIN_TELEGRAM_CHAT_ID` when no chat is given. Confirms `_chat_settings_id_for`'s different (return-`None`) policy is a real, deliberate divergence, not a rhetorical flourish — and it's the right one: stamping CLI writes with a literal admin-chat FK would freeze a value that the existing fallback rule otherwise re-resolves dynamically against whatever `TELEGRAM_CHANNEL_ID` is configured *now*.
- `add_account`'s `set_as_active` and `telegram_chat_id` are independent parameters — passing the real chat through for OAuth secondary accounts (the "flipped" test) does not also activate them; `set_active = idx == 0` in `oauth_service.py`'s loop is untouched.
- `token_refresh.py`'s `create_or_update` call is correctly left passing no `chat_settings_id` at all — it's the exact call site the "never clear when absent" guard exists for.
- Reran all 3 modified test files locally against restored code: 76 passed.

**Two things to flag before this merges, not code defects:**

1. `mergeable` shows `CONFLICTING` against current main. Checked with `git merge-tree` (read-only, no working-tree changes) — the only conflicting path is `CHANGELOG.md` (two PRs each inserting an entry near the same spot); every source and test file merges clean. Routine rebase, not a substantive conflict — but it does need resolving before merge.
2. **CI never actually ran for this branch.** `gh run list --branch fix/675-stamp-ig-tokens-on-add` returns zero runs, vs. every sibling PR this week (#671, #582, #597, #412) having one — and the timing rules out "Actions was down": a `feat/412` run completed within seconds of this PR's creation timestamp. So the PR body's "2129 passed / 0 failed" is corroborated by my own local reruns of the touched files (76 passed) and behavioral gate, but not by an actual CI run of the full suite/lint/security-scan. I'd want CI to actually execute (push an empty commit or manually re-trigger) before this merges, since I only reverified the files this diff touches, not the full ~2100-test suite.

PR #683, opened 2026-07-19
fix(telegram): stop expire-reap retrying "Message is not modified" forever — ends the flood-control button hang (#682)

chrisrogers37:
Fixes #682. Live incident: post buttons hanging with an endless "Loading…" spinner, autoposts aborting — root cause is the bot flood-limiting itself, not #557's event-loop starvation (that substrate's partial fixes #572/#573 are deployed and holding; DB pool idle, zero "Query is too old" in retention).

## Why

Two stacked classification bugs turned ~64 already-expired cards into a permanent no-op edit storm:

1. **`telegram_edit_with_retry` never took its documented BadRequest path.** In python-telegram-bot, `BadRequest` subclasses `NetworkError`, so `except (TimedOut, NetworkError)` caught it first: 3 futile retries + backoff, then `None`. Prod logs show it verbatim: `Telegram BadRequest, retrying in 2.0s (attempt 2/3): Message is not modified…`
2. **`expire_sent_row` read that `None` as transient** and deferred the row "for next sweep" — every sweep, forever. A not-modified rejection on an expire-strip actually means the card is *already* in the terminal Expired state.

Both reap paths (scheduler sweep + hourly cleanup) re-chewed the same ~64 rows every ~8.5 min → ~190 no-op edits per pass into one chat, 24×7 since the 07-16 deploy → Telegram pinned the bot under per-chat flood control (observed 62× `RetryAfter ~35s` per hour). Chris's traced 15:36 UTC clicks: 3 autoposts aborted with `Auto-post failed: Flood control exceeded` raised from the ⏳ caption edit; 2 skips landed 14–36s late; answers came too late to stop the spinner.

## What

- **`src/utils/resilience.py`** — classify `BadRequest` ahead of the transient branch: single attempt, log once, return `None`. Docstring updated to match.
- **`src/services/core/queue_reap.py`** — the reap classifies at its own seam with one un-retried edit: `BadRequest` (card already terminal / not editable) → record + delete; `RetryAfter` / `TimedOut` / `NetworkError` → defer to the next sweep; anything else → proceed as before. No in-reap retries: the sweep cadence *is* the retry loop, and backoff sleeps here fed the very flood window they waited out.

The stuck-row backlog drains itself on the first post-deploy sweep; flood control decays within ~1 min of the storm stopping.

## Design note — deliberate deviation from the "re-raise BadRequest" recommendation

The original recommendation (issue #682 / knowledge doc) was to make the wrapper *raise* `BadRequest`. A call-site audit found ~30 callers (telegram_callbacks_admin/core/queue, autopost, telegram_utils), all shaped around never-raise + `None`, most inside handlers where a new raise surfaces as a "⚠️ Something went wrong" alert on benign double-clicks (not-modified is the *common* case there). The wrapper has never raised `BadRequest` in its life, so nothing depends on it. Preserving the `None` contract (minus the futile retries) gets the entire flood fix with zero blast radius; the reap — the one caller that genuinely needs permanent-vs-transient — classifies at its own seam.

Also considered and declined: centralizing the taxonomy in a `classify_telegram_edit_error()` predicate. Both consumers need per-class behavior *inside* the transient bucket (RetryAfter's server-specified wait vs exponential backoff; different defer semantics), so the predicate just relocates the isinstance-switch. The subclass-ordering invariant is instead pinned by tests at both sites — an except-reorder in either file fails its suite. If a third site ever needs this classification, that's the consolidation trigger.

## Verification

- TDD: regression tests written first and watched fail for the right reasons (`await_count 3 != 1`; `'deferred' != 'reaped'`), then green: **17 passed in 1.6s** (the speed is itself evidence — no more backoff sleeps on these paths).
- New `tests/src/utils/test_resilience.py` pins the wrapper contract (BadRequest = 1 attempt + `None`; transients still 3 attempts; transient-then-success still returns).
- `test_queue_reap.py` now mocks the real boundary (`bot.edit_message_caption` side effects) instead of patching the wrapper — the old `message_gone` test only passed because it bypassed the wrapper; against real prod behavior that path was broken too, which this PR fixes and now tests honestly.
- Full suite: **2179 passed, 8 skipped** (1 unrelated flake under my local `LOG_LEVEL=WARNING`; passes with CI's `DEBUG`). `ruff check` / `ruff format --check` clean at CI scope.
- Runtime verification is deploy-gated by design (the storm only exists in prod) — checklist below.

## Post-deploy verification checklist

- [ ] `Expire reap: deferring` lines → 0 in worker logs
- [ ] ~64× `Expire reap: recorded expiry + removed queue item` on the first post-deploy sweep
- [ ] `RetryAfter` / `Flood control exceeded` lines stop within minutes
- [ ] Click test: Posted / Skip / Reject / Auto Post answer instantly; autopost completes

Relation to #557: that issue stays open for the remaining substrate items (Cloudinary/Drive timeouts, residual sync-DB offload). Context: the stuck-card cohort formed in the #679 deploy aftermath; #679's code is untouched here.

https://claude.ai/code/session_0184oPfywut4dJSubpJGjzuo


chrisrogers37 (commented):
**Verdict: Ship it**

Mutation-tested both scrutiny points — reverted each except-block to the pre-fix order in a scratch checkout and re-ran the pinned suite against the mutated code, rather than trusting the diff by eye.

**Point A — `resilience.py` except ordering**

Confirmed order: `RetryAfter` → `BadRequest` → `(TimedOut, NetworkError)` → bare `Exception`. Swapping `BadRequest` back below `(TimedOut, NetworkError)` (the pre-fix shape) makes `test_bad_request_is_permanent_no_retry` fail with `assert 3 == 1`, and the reproduced warning log is a verbatim match to the incident:
```
Telegram edit failed after 3 attempts: Message is not modified: ...
```
With the real ordering restored, all 6 resilience tests pass. The 3 transient-parametrized cases (`RetryAfter`/`TimedOut`/`NetworkError`) pass identically under both orderings — proof the fix narrows exactly the `BadRequest`/`NetworkError` subclass overlap and nothing else.

Re: the amendment — confirmed the wrapper still returns `None` on `BadRequest` (no raise), just after 1 attempt instead of 3. Checked blast radius myself rather than taking the 30-caller audit on faith: `grep` finds 25 call sites across the codebase, **0** capture the return value via assignment, **0** check `is None` near the call. No caller's control flow can depend on attempt count — the only externally observable change is fewer Telegram calls and no backoff sleeps, which is the fix itself. Not flagging the absent re-raise; it's correctly a non-issue.

**Point B — `queue_reap.py` self-drain, no migration**

Same approach: swapping `(RetryAfter, TimedOut, NetworkError)` back above `BadRequest` makes `test_not_modified_is_already_expired_so_records_and_deletes` fail with `assert 'deferred' == 'reaped'` — reproducing the infinite-defer bug exactly. `git diff --stat` confirms zero migration files in this PR. Both existing sweep entry points (`scheduler_loop.py:82`, `queue_cleanup_loop.py:44`) already call `expire_sent_row` on their current cadence, so the ~64-row backlog hits the corrected classification on the very next scheduled sweep post-deploy — no manual intervention required.

**Other checks**

- Tests aren't decoys: `test_queue_reap.py`'s old `message_gone` case mocked `telegram_edit_with_retry` directly, bypassing the real (buggy) code path entirely — this PR already caught and fixed that in its own diff (new tests mock `bot.edit_message_caption`, the real boundary). Called out honestly in the PR body too.
- Scope: 5 files, all on-point (`resilience.py`, `queue_reap.py`, their two test files, `CHANGELOG.md`). Dead import (`telegram_edit_with_retry`) cleanly removed from `queue_reap.py`. #679 untouched as claimed.
- CI: all 7 checks green (Test, Lint, Security Scan, Changelog Check, GitGuardian, Vercel ×2). No other open PRs touch these files. Branch contains all of current `origin/main` — not stale.

Full suite (2179 passed) and lint already re-verified by CI; I ran the two targeted test files directly (17/17, matching the PR's own count) plus the mutation pass above.

— rajan

PR #684, opened 2026-07-19
fix(reap): unblock the expire drain — permit system_expiry history writes, contain per-row DB failures (#682)

chrisrogers37:
Follow-up to #683, same incident (#682). Post-deploy verification found #683's classification working (rows reach record + delete) but the drain **blocked by a second latent bug** the fix exposed the moment that path actually ran in production.

## What post-deploy logs showed (worker, f02f57a live 17:51Z)

- Flood side: healthy — 59 defers + 55 RetryAfter in the boot minute only (residual flood), then **zero of both** from 17:52 onward. Buttons responsive again.
- Drain side: **0 rows drained.** Every scheduler tick (13/13 over 13 min) hits the first stuck row and dies:

```
psycopg2.errors.CheckViolation: new row for relation "posting_history"
violates check constraint "check_posting_method"
...
ERROR - Error in scheduler loop: (psycopg2.errors.CheckViolation) ...
```

Migration `004` constrains `posting_method IN ('instagram_api', 'telegram_manual')`. The reap has written `system_expiry` since #561 — but that insert path effectively never executed until #683 fixed the classification. Migration `042` widened the sibling `check_history_status` for the same feature and missed this one. The raised `CheckViolation` aborts the whole sweep pass (and the tick's remaining work), and the flush-dirty session poisons any later DB use in the pass.

## Changes

1. **Migration `045`** — widens `check_posting_method` to `('instagram_api', 'telegram_manual', 'system_expiry')`. These are exactly the three values code writes today (`telegram_autopost.py`, `scheduler.py`, `queue_reap.py`). **Applying this to prod is the drain unblocker** — with #683 already deployed, the backlog self-drains over the next sweeps with no further action. Non-destructive (constraint widen); apply per repo convention: `psql "$DATABASE_URL" -f scripts/migrations/045_add_system_expiry_posting_method.sql`.
2. **Model declares `check_posting_method`** — `posting_history.py` `__table_args__` now carries the constraint (same dual-declaration shape as `check_history_status` + migration 042). Also fixes a fresh-install divergence: `create_all` previously produced no `posting_method` constraint at all.
3. **Per-row containment in `expire_sent_row`** — record/delete DB failures (`SQLAlchemyError`) no longer raise out of the helper: log with traceback, roll both repos' sessions back (separate task-local sessions — both rollbacks required), return a new `"failed"` outcome; the row stays for the next sweep and the pass continues. Deliberately narrow: non-DB exceptions (programming errors) still propagate loudly — containing them would disguise a code defect as an infinite per-row retry. This makes all three call sites (scheduler sweep, hourly cleanup, `reap_pending_rows`) resilient with zero caller changes.

Considered and declined: routing the record+delete pair through `atomic_session` (used by the posting paths). It would change commit semantics here (single deferred commit vs the repos' current independent per-op commits) — a transactional redesign, not an incident fix. Worth a separate look.

## Verification

- TDD: containment tests watched fail red (IntegrityError/SQLAlchemyError raised through), then green; the narrowing has its own red-first test (`test_non_db_error_still_raises`).
- Full suite **2184 passed, 8 skipped**; `ruff check` / `ruff format --check` clean at CI scope.
- Migration is deploy-gated (prod DDL, sign-off per repo convention). Post-apply verification:

## Post-apply checklist (migration + this PR)

- [ ] `Error in scheduler loop: ... check_posting_method` stops appearing
- [ ] ~60× `Expire reap: recorded expiry + removed queue item` over the next sweeps
- [ ] Queue backlog of stuck Expired rows reaches zero; no `deferring` / `RetryAfter` recurrence
- [ ] Scheduled autoposting resumes normal cadence (tick no longer aborts at the reap step)

https://claude.ai/code/session_0184oPfywut4dJSubpJGjzuo


chrisrogers37 (commented):

Reviewed against the four call-outs, all verified empirically against the checked-out branch, not taken on the PR description's word:

**(A) Migration 045** — `DROP CONSTRAINT IF EXISTS` + `ADD CONSTRAINT` + `schema_version` insert is a byte-for-byte structural match to migration 042 (the sibling `check_history_status` widen) — this is house convention, not a new pattern. Idempotent: re-running is safe (drop-then-recreate lands on the same end state). Confirmed no migration in this repo ever uses `NOT VALID`/`VALIDATE CONSTRAINT` staging, so the full-table CHECK-validation lock this ALTER takes is consistent with every prior constraint-widen here (029–043), not a new risk this PR introduces.

**(B) SQLAlchemyError containment** (`queue_reap.py` `expire_sent_row`) — except clause is scoped to `SQLAlchemyError` only (confirmed `IntegrityError` — what a `CheckViolation` surfaces as — is a subclass, so this is the right catch). Confirmed `history_repo`/`queue_repo` are genuinely separate task-local sessions (`BaseRepository._ensure_session_vars`: per-instance ContextVars, not a shared singleton session), so both rollbacks are load-bearing, not redundant. Confirmed the non-DB path still raises via `test_non_db_error_still_raises` plus reading the except clause directly. Traced the actual crash site: `scheduler_loop.py`'s stale-sent `for` loop (lines 76-88) has no per-row containment and its outer `except Exception: queue_repo.rollback(); raise` only rolls back `queue_repo` — exactly matching the PR body's 'flush-dirty session poisons the pass' description. Post-fix, this loop can now walk every stale row in a tick without a mid-batch DB error aborting the rest.

**(C) Model constraint matches migration** — `posting_history.py`'s new `CheckConstraint` string is identical to migration 045's, same name (`check_posting_method`), same 3 values, same order. One nit posted separately (stale 2-value inline comment 2 lines above — non-blocking).

**(D) Backlog self-drain + mutation test** — Ran the actual mutation test, not just reasoned about it: checked out `f02f57a` (pre-#684) versions of `queue_reap.py` + `posting_history.py` over the PR's test file, ran `test_queue_reap.py` → **2 of 4 new tests fail red** exactly as expected (`test_record_failure_returns_failed_rolls_back_keeps_row`, `test_delete_failure_returns_failed_and_rolls_back` — both raise the DB error uncaught without the fix, proving they're not decoys). The other 2 new tests pass regardless of the fix (`test_non_db_error_still_raises` — old code never caught anything either, so this is a guard-rail test, correctly not red-first; `test_failed_row_does_not_abort_batch_or_count` — mocks `expire_sent_row` directly, tests `reap_pending_rows`' handling of the sentinel, independent of the source fix). Restored the fix, confirmed 15/15 green again.

Also ran: `ruff check` + `ruff format --check` clean on all 3 touched Python files. Broader collateral-impact pass — `tests/src/models/` + `tests/src/repositories/` — 408 passed, 4 skipped, nothing conflicting with the new constraint's enumeration.

One finding posted separately (major but non-blocking, out of this PR's scope): `check_posting_method` is still missing a 4th live value (`auto_reapproval`, from `scheduler.py`'s default-config repost path) — pre-existing, not introduced here, but worth a fast follow-up migration while this constraint is already open.

Migration application remains human-sign-off-gated per dispatch. This verdict is on code + migration correctness, not a merge/apply trigger.

PR #688, opened 2026-07-19
fix(reap): write terminal expired history before deleting unstamped rows (#687)

chrisrogers37:
Fixes #687.

## Why

The hourly `delete_stale` prune (#483) hard-deletes >24h `telegram_message_id IS NULL` queue rows on the documented assumption they were "NEVER sent to Telegram." The #679 ambiguous-delivery class falsifies that: a timed-out send can deliver the approval card while the stamp never lands (root cause tracked as #680 — explicitly out of scope here). Deleting such a row with no history record orphans its live buttons permanently — the tap-time fallback (#560/#561) finds nothing in queue OR history and surfaces the raw "Queue item not found" error. Live-captured 2026-07-19: cleanup deleted 2 stale rows at 18:51:48, tap on a still-visible card at 18:54:14 → raw QINF, row in neither table (DB-verified).

## What

**Never delete an unstamped row without writing the terminal `expired` history row first.** Unstamped rows can't be told apart from delivered-but-unstamped ones, so all of them get the record; a spare history row for a genuinely-never-sent item is harmless (no card exists to tap it), while an orphaned-but-delivered card now degrades to the graceful "Expired" caption.

- `queue_reap.record_expiry_and_delete(row, *, history_repo, queue_repo)` — the record+delete phase of `expire_sent_row`, extracted (identical behavior: idempotent history write with `status='expired'` / `posting_method='system_expiry'` — valid since migration 045 — then delete, per-row DB-failure containment with rollback). `expire_sent_row` now delegates to it, so the two paths can never drift.
- Hourly cleanup loop: `delete_stale(hours=24)` (bulk delete) → `get_stale_unsent(hours=24)` (fetch, same filter incl. the #549 `publishing` guard) + per-row `record_expiry_and_delete`.
- Scheduler JIT sweep (`process_slot`): `delete_stale_pending(max_age_minutes=10)` → `get_stale_unsent_pending` + per-row `record_expiry_and_delete`. This path races a delivered card the same way: the processing requeue flips an unstamped ambiguous send back to `pending`, and the next tick's sweep deletes it (`created_at` is already past the cutoff).
- `reap_pending_rows` (batch reaper behind the admin clear-queue / resume-clear / reset flows) routes its unstamped else-branch through the same helper instead of raw-deleting. An admin clearing the queue during the #679 window would re-orphan a delivered card identically; this was surfaced independently by two review lenses in the /simplify pass and makes `record_expiry_and_delete` the sole unstamped-delete path across the reap surface.
- The two bulk-delete repo methods are removed (zero callers remain — no shims), and `get_stale_sent`/`get_stale_unsent` share one private age-query (`_get_stale_scheduled`) so the #549 `publishing` guard can't drift between polarities.

Deliberately NOT converted (scope per issue):
- `discard_abandoned_processing` — requeue-shielded in practice: `requeue_stale_processing` flips unstamped processing rows to pending at 10 min, far before the 24h discard, so the JIT sweep (converted) sees them first. Same-class hole only if the per-tenant tick never runs; noting rather than expanding scope.
- `scheduler.py`'s failed-autopost unstamped delete — deliberate history-less delete (media stays eligible for retry), not an expiry; out of the invariant by design.
- #680 (`sent_unconfirmed` state) — the root cause, tracked separately.

## Verification

Red-first per dispatch: 10 new/updated tests written and observed failing on main's behavior (bulk delete, no history), then green after the fix.

- `test_queue_reap.py::TestRecordExpiryAndDelete` — contract: history-then-delete ordering (delete never runs if the history write fails), idempotency, per-row DB containment + rollback, non-DB errors propagate. Includes the #687 regression test: sweep an unstamped row → feed the written history row to the tap-time caption builder → asserts "Expired", asserts no "not found".
- `test_queue_cleanup_loop.py` — every stale unstamped row routed through `record_expiry_and_delete`; no bulk delete behind the history write's back.
- `test_scheduler_queue_reliability.py` — process_slot routes stale pending rows through the helper.
- `test_queue_repository.py` — new fetch methods keep the exact filters (unstamped-only, `publishing` excluded (#549), pending-only for the JIT sweep) and delete nothing.

Full suite: **2193 passed, 8 skipped**. `ruff check src/ cli/` clean; `ruff format --check src/ cli/` clean.

Workflow: `/weigh-development-paths` applied at the design fork (repo-internal history write vs service-layer routing vs repo-level delete guard — service-layer routing won: repositories stay CRUD, and "delivered-ness" is undecidable at the row level, so a delete-guard would false-positive the scheduler's deliberate history-less delete). `/simplify` ran as a 4-lens review (reuse / simplification / efficiency / altitude); its convergent finding (the `reap_pending_rows` else-branch) is included above, plus docstring dedup and the shared age-query. Skipped from review: merging the two fetch methods (three-axis differences — named intent wins) and rerouting the loop/scheduler through `reap_pending_rows` (would force a dead `bot` param onto a by-construction-unstamped path).

Not verified live: driving the real hourly loop against prod is a posting-system no-go (repo safety rules); the deploy will exercise it on the first sweep — the log line to watch is `Expire reap: recorded expiry + removed queue item` from the cleanup loop, and history rows with `posting_method='system_expiry'` for rows that never had a card edit.

Side note for the system-review tracker: TD-039 ("`delete_stale()` loads all rows then deletes") is reshaped by this PR — the load is now intentional (rows feed per-row history writes).


chrisrogers37 (commented):

Reviewed against all four call-outs. Verified empirically, not taken on the PR description's word — this PR sits in the same file family I just reviewed for #684, which sped up the trace-through.

**(A) Core fix, all 3 surfaces** — confirmed each routes unstamped-row deletion through the shared `record_expiry_and_delete(row, history_repo=, queue_repo=)`, which writes the terminal `expired` history row before deleting:
- Hourly loop (`queue_cleanup_loop.py`): `get_stale_unsent(hours=24)` (now a pure read) → per-row `record_expiry_and_delete`.
- 10-min JIT sweep (`scheduler.py` `process_slot`): `get_stale_unsent_pending(max_age_minutes=10)` (now a pure read) → per-row `record_expiry_and_delete`.
- Admin clear (`reap_pending_rows`, behind clear-queue/resume-clear/reset): unstamped else-branch now calls `record_expiry_and_delete` instead of `queue_repo.delete()` directly.

Traced the fix all the way to the human-visible symptom, not just the unit assertions: `validate_queue_item` (telegram_utils.py:111) falls back to `history_repo.get_by_queue_item_id` when the queue row is gone, and `_build_already_handled_caption` returns `EXPIRED_CAPTION` for `status="expired"`. Pre-fix, the bulk deletes left neither table populated → that fallback found nothing → raw "not found". Post-fix, history always exists before the queue row is gone → the existing (#560/#561) fallback renders "Expired". This PR's contribution is specifically ensuring the history row exists at all for these 3 surfaces — it doesn't touch (and doesn't need to touch) the caption-rendering logic, which was already correct.

**(B) Red-first, confirmed not decoys** — ran the actual mutation test: reverted all 4 touched source files to pre-#688 (main tip) while keeping this PR's test files, then ran the 5 modified test files.
- `test_queue_reap.py` fails at collection: `ImportError: cannot import name 'record_expiry_and_delete'` — the whole new `TestRecordExpiryAndDelete` class (6 tests) plus 2 updated `TestReapPendingRows` tests exercise a symbol that doesn't exist pre-fix.
- The other 4 files: **10 failed** — `AttributeError: 'QueueRepository' object has no attribute 'get_stale_unsent'`/`'get_stale_unsent_pending'` (real class, not a mock — hard proof), `AssertionError` on call-count (old code never calls the new methods), and one `TypeError: '>' not supported between instances of 'Mock' and 'int'` (old `cleanup_queue_loop` calling the now-unconfigured `delete_stale` mock and comparing the raw Mock to 0).

Restored, reran: **170/170 green** across all 5 files. This is about as clean a red-first signal as I'd want — four distinct, genuine failure modes, none of them vacuous.

**(C) No behavior drift** — `record_expiry_and_delete` is a faithful extraction of the post-#684 inline logic in `expire_sent_row` (identical try/except/rollback, bool instead of string return, translated back to `"failed"`/`"reaped"` at the `expire_sent_row` call site — zero change for existing callers). Confirmed `_get_stale_scheduled`'s consolidated filter is byte-for-byte the same 3 conditions as the old `delete_stale`/`get_stale_sent` queries, plus a harmless/beneficial `order_by(scheduled_for/created_at ascending)` that makes the unstamped sweep process oldest-first, matching `get_stale_sent`'s existing behavior. Confirmed `queue_repo.delete()` and `history_repo.create()` both commit per-call, so the per-row "history commits before delete is attempted" invariant is real at the transaction level, not just superficial in Python control flow. Confirmed zero remaining references anywhere in `src/` or `tests/` to the removed `delete_stale`/`delete_stale_pending` — clean removal, no orphaned callers.

**(D) Scope** — the actual source footprint is modest (~120 lines across 4 files: `queue_repository.py` net **-31** since two write methods collapse into one shared private query + two thin read wrappers); the bulk of the +415/-173 is test coverage (~295 lines), which is the right place for size to land on a fix like this. The 3 deliberately-excluded paths in the PR body check out: `discard_abandoned_processing` is genuinely requeue-shielded (10-min requeue fires well before the 24h discard), the scheduler's failed-autopost delete is a intentionally history-less retry path (different invariant), and #680 (the actual root cause) is untouched — `scheduler.py`'s diff is a 2-line change plus an import, nothing near the `AmbiguousDeliveryError`/send-classification logic.

Also ran independently: `ruff check` + `ruff format --check` clean on all 9 touched Python files. Full suite: **2193 passed, 8 skipped** — exact match to the PR's own claim. CI green (Lint/Test/Security Scan/Changelog Check all SUCCESS).

No blocking or non-blocking findings this round — clean fix, clean tests, clean scope.

PR #689, opened 2026-07-19
fix(telegram): answer posted/skip/reject callbacks before the slow chat ops (#686)

chrisrogers37:
## What

Answer the Telegram callback **first** in the Posted / Skip / Reject button handlers — immediately after the cheap `claim_for_processing` gate and **before** the slow chat ops (reconcile + the terminal caption edit that waits out per-chat `RetryAfter` walls). Closes the stuck-spinner symptom from #686 (fix direction 1).

Auto Post already answered-first (its own `"⏳ Posting…"` ack, #481) and is unchanged — it's the reference pattern this brings the other three handlers up to.

## Why

Under rapid/parallel clicking (live-captured in #686), the post-buttons lag 10–26s or the spinner never stops, even though the action completes. The event loop is not parked; the cause is **per-chat Telegram rate limiting** — the callback answer was queued *behind* the handler's own caption edits, which correctly wait out `RetryAfter` (26.5s / 10.5s / 26.5s / 17.5s observed in one 95s window). When the ack lands after the spinner's ~30s validity window, it reads as a full hang.

Posted / Skip / Reject deferred their answer to the dispatcher's `_fallback_answer` in the `finally` — i.e. *after* reconcile, the atomic DB write, and the terminal caption edit. Moving the ack ahead of that slow work stops the spinner within the validity window regardless of downstream throttling.

## The one subtlety — answer *after* the claim, not before

#679 deliberately made handler answers first-and-visible so `show_alert` error toasts work (Telegram honours exactly one answer per query). So the ack is placed **after** `claim_for_processing`:

- **Claim lost** (item already handled by a concurrent click / autopost) → routes through `validate_queue_item`, whose "already handled" toast stays the first and only answer. ✓ (#679 preserved)
- **Claim won** → bare `await query.answer()` stops the spinner, then the slow work runs. ✓ (#686 fixed)

`claim_for_processing` is a DB-key gate with no chat ops, so answering right after it is still "answer-fast." Composes with #624 (router + `_fallback_answer` untouched; the fallback simply no-ops for these handlers now).

## Scope

- **In:** fix direction (1) from #686 — answer-first in the post-button handlers.
- **Out (held):** fix direction (2), python-telegram-bot's `AIORateLimiter` — the broader outbound-queueing change, deliberately not in this PR.

## Tests / verification

- Red-first `TestAnswerCallbackFirst` (3 tests): drives `_do_complete_queue_action` / `_do_handle_rejected` directly with `reconcile_card_messages` + `telegram_edit_with_retry` patched; records call order via `side_effect`.
  - Confirmed **RED** before the fix: `Expected answer to have been awaited once. Awaited 0 times.`
  - Asserts `order == ["answer", "reconcile", "caption_edit"]` on a successful claim (posted + reject).
  - Asserts a **failed** claim does NOT pre-answer — guards the #679 toast-first property.
- Full suite: **2187 passed, 8 skipped**. `ruff check` + `ruff format --check` clean. CHANGELOG updated.
- `/simplify` (4-agent pass): collapsed the ordering assertions to one list-equality and trimmed the source comment; reuse / altitude / efficiency came back clean (the bare `query.answer()` is the intended primitive; the shared wrapper can't host the ack since the claim lives inside the coro; the fallback's second answer is an off-latency-path, expected rejected round-trip).

https://claude.ai/code/session_01Kv8Ht7ax8UNRqXVUBKHKVr


chrisrogers37 (commented):
**Verdict: Ship it**

Reviewed by running, per the dispatch — mutation-tested in a scratch clone rather than reading the diff by eye.

**Point A — answer-first for Posted/Skip/Reject; autopost untouched**

`_do_complete_queue_action` (shared by Posted/Skip) and `_do_handle_rejected` both now call `await query.answer()` immediately after `claim_for_processing` succeeds, before `reconcile_card_messages` and the terminal caption edit. `telegram_autopost.py` doesn't appear in the diff at all (`git diff --stat` shows only `telegram_callbacks_queue.py` + its test + `CHANGELOG.md`) — confirms autopost's existing #481 answer-first pattern is genuinely untouched, not just claimed.

**Point B — CRITICAL: #679 toast-first preserved, no bland ack on invalid cards**

The new `await query.answer()` sits *after* the `if not queue_item: await validate_queue_item(...); return` early-return in both handlers, so a lost claim never reaches the new line. Verified two ways, not just by reading:
1. Reverted the fix (removed both early-answer blocks) — `test_failed_claim_does_not_pre_answer` still passes (no regression in that direction).
2. Adversarially hoisted `query.answer()` to *before* the claim check in both handlers, simulating a future regression — `test_failed_claim_does_not_pre_answer` correctly fails: `Expected 'answer' to not have been called. Called 1 times.` Proves the pinned test actually guards the ordering invariant, not just passing incidentally.

Also traced `_safe_locked_callback`'s lock: it's a non-blocking `.locked()` check (immediate distinct "already processing" toast on contention, not a blocking acquire), so there's no lock-queueing path that could delay a concurrent click's answer behind another click's slow work. `validate_queue_item` itself (the claim-lost toast path) is untouched by this diff — same function, same behavior as main.

One pre-existing observation, out of scope here: `validate_queue_item`'s "not found, no history" branch calls `query.edit_message_caption(...)` with no explicit `query.answer()` — the dispatcher's `_fallback_answer` (#679) covers it, but if that edit itself gets `RetryAfter`-throttled under the same burst conditions #686 describes, the fallback answer would be delayed behind it too. Not introduced or touched by #689 — flagging as a possible follow-up thread, not blocking this PR.

**Point C — red-first test genuinely proves the ordering**

Removed both early-answer blocks and re-ran: `test_posted_answers_before_slow_work` and `test_rejected_answers_before_slow_work` fail with `Expected answer to have been awaited once. Awaited 0 times.` — exact match to the PR's stated RED evidence. Restored the fix, full file green again: 33/33.

**Point D — scope**

`+92/-0` across exactly `telegram_callbacks_queue.py`, its test, and `CHANGELOG.md`. Zero `AIORateLimiter` references anywhere in `src/` — #686b genuinely held out. No other open PR touches this file (#688 is reap/scheduler-side: `queue_repository.py`, `queue_reap.py`, `scheduler.py`, `queue_cleanup_loop.py` — zero overlap). Branch is one commit behind current `main` (#688, merged after this branch was cut), but that commit's files don't intersect this PR's at all, and GitHub reports `MERGEABLE`. Not blocking, noted for the record.

CI: all 7 checks green (Test, Lint, Security Scan, Changelog Check, GitGuardian, Vercel ×2).

— rajan

PR #510, opened 2026-06-16
docs(planning): queue/posting simplification — forged plan v0

chrisrogers37:
## Forged plan — ready for `/ironclad`

**No code ships from this PR.** It adds one planning doc (`documentation/planning/2026-06-15-queue-posting-simplification.md`) — the contract to ratify before implementing a **production data-model + code change**.

### Goal
Eliminate the recurring queue/posting **race class** (`Queue item not found`, duplicate send, orphaned `pending`/`processing`, cap-hit stranding) by replacing the reconciliation machinery that *manufactures* it (3 wall-clock sweepers + history-lookup recovery + hourly prune) with three atomic primitives: **atomic cap admission**, **atomic claim + lease**, **idempotent completion**. Directly defends the mission guarantee *"posts are never lost"* — the `:79` "Queue item not found" miss is a silently lost post.

### Phases (complexity: S×1, M×2, M/L×1, L×1 — net-negative LOC)
- **Phase 0 (S, prerequisite):** migration safety rails — a CI job applying `scripts/migrations/*.sql` with a model↔SQL drift check, plus a deploy runbook. *Flagged for the ratifier: this was NOT in the original 1–4 enumeration; recon showed prod migrations are untested in CI, manually applied against Neon, and have no rollback tooling — shipping a prod data-model change on that is the dominant risk.*
- **Phase 1 (M):** atomic daily-cap admission (reserve-before-send). **Supersedes open PR #509** (restore-only band-aid).
- **Phase 2 (M):** claim-lease replaces the wall-clock sweepers — **closes the still-open `:79` "Queue item not found" race**.
- **Phase 3 (M/L):** idempotent completion in one transaction (retrofits the 3 non-atomic paths onto `_shared_session`).
- **Phase 4 (L, parallel track):** DB-off-loop foundation (also the standing button-hang fix).

Each phase specifies: additive migration → path-by-path rollout (internal paths before user buttons) → validation (Postgres **test DB only**, never real posting) → compensating-migration rollback.

### Decision forks (6 open — to lock during `/ironclad`)
| Fork | Decision | Lean |
|------|----------|------|
| **F1** | terminal `posted` status vs write-history-in-same-txn-as-delete | (b) same-txn — no constraint surgery |
| **F2** | cap counter-row `UPDATE … WHERE n<cap` vs conditional insert | (a) counter-row reservation, coupled to F3 |
| **F3** | lease TTL + reclamation cadence | (b) config TTL above worst-case send, reclaim once/tick |
[...]

PR #693, opened 2026-07-20
feat(models): posting_method/history_status enum SSOT + CI parity gate (close #685)

chrisrogers37:
## Summary

Single-sources the `posting_history` CHECK-constraint vocabularies from a code-owned enum and **closes the live-armed #685**. This is PR1 of the ratified data-model redesign — the ironclad-affirmed, decoupled first increment (data/migration + #510-fold plan, §3 + M-A).

### The problem (drift class: #684, #685)

Allowed values for `posting_history.posting_method` / `.status` lived in three unsynced places — the DB `CHECK`, the model `__table_args__`, and scattered code literals. Add a value, miss the constraint → a production `CheckViolation` that **aborts the entire scheduler sweep pass**. #684 was `system_expiry`; **#685 is `auto_reapproval`** — a live scheduler default-config repost path the prod constraint still rejects (migration `046` was never applied). It is armed: the next time that path fires, the sweep aborts.

### What this does

- **`src/models/enums.py`** — new single source of truth: `PostingMethod`, `HistoryStatus`, `QueueStatus` (the target queue vocabulary for the follow-up status-model migration), and `sql_in_list`.
- **`src/models/posting_history.py`** — `check_posting_method` / `check_history_status` are now **derived** from the enums via `sql_in_list`, not hand-typed literal lists.
- **Migration `046`** — widens `check_posting_method` to add `auto_reapproval` (the last hand-authored widen). Non-destructive constraint widen; human-sign-off-gated prod apply.
- **CI parity gate** (`tests/src/models/test_enum_ssot_parity.py`) — asserts the model constraint **and** the latest migration DDL both match the enum, for `posting_method` and `history_status`. Drift now fails a test instead of aborting a prod sweep.

### M4 limitation (rajan, #692 review) — folded in

The strongest form of this gate would introspect the constraint on a **migration-replayed** database. The test harness bootstraps via `Base.metadata.create_all` (`conftest.py::setup_test_database`), not migration replay — so it can only ever see the model constraint, never the applied migration. The gate therefore asserts statically against the migration **file**. This is documented in the test module docstring and tracked as a follow-up: strengthen (2) to introspect `pg_constraint` on a replayed DB once migration-replay CI exists (the #510 P0 migration rails). The migration-side checks carry the real teeth here — the model constraint now derives from the enum, so the model-side checks only guard against reverting to a hand-typed list.

### Verification

- **TDD red→green** — parity gate written first, watched fail for the right reason (model + migration 045 lacked `auto_reapproval`), then implemented to green.
- **Full suite** — `2199 passed, 8 skipped` (no regressions).
- **Neon-branch dry-run of migration 046** (copy-on-write branch off prod) — constraint went 3→4 values incl `auto_reapproval`; **idempotent** on re-run (`INSERT 0 0`); `schema_version` row registered; **0 existing rows violate** the wider constraint. Branch torn down after.
- **ruff** check + format clean on changed files.
- **/simplify** (4 lenses) — applied: reuse `validators.MIGRATIONS_DIR` + the canonical version parse, a consistent quoted-value idiom, and a symmetric `history_status` migration-parity test (the migration side is where the teeth are); skipped: cross-file test-helper extraction (out of scope), and keeping `QueueStatus` (scoped in by dispatch; M-B is the imminent consumer).

### Scope boundary (follow-ups, deliberately not here)

- Converting value **producers** (scheduler / reap / autopost that still hand-type the method string) to reference the enum — the code-integration lens.
- Wiring `QueueStatus` to `posting_queue.check_status` + its migration — the status-model increment (M-B).

### Deploy note

Prod apply of migration 046 is human-sign-off-gated (repo convention): `psql "$DATABASE_URL" -f scripts/migrations/046_posting_method_ssot_add_auto_reapproval.sql`. The dry-run already validated it against a prod-copy branch.

**Reviewers:** rajan, navi

Closes #685.

https://claude.ai/code/session_01C2NET6CGiWJk53eATZyoBd


chrisrogers37 (commented):

Reviewed against all four focus areas, verified empirically where possible — this is PR1 of the plan I ironclad-reviewed as #692, so I came in with full context on the design intent.

**(A) Enum-SSOT + CI parity gate** — confirmed the gate genuinely catches drift, not just by reading it. Ran a direct mutation test: removed `AUTO_REAPPROVAL` from the `PostingMethod` enum while leaving migration 046 untouched on disk (simulating the exact #684/#685 drift shape — a value present in one place, missing in the other). `test_latest_migration_posting_method_matches_enum` failed cleanly with a clear set-diff (`+ 'auto_reapproval'`), while the model-side tests stayed green — exactly matching the test module's own stated design (model constraint is *derived* from the enum, so it moves in lockstep and can only catch a hand-typed-list reversion; the migration-file check is independent and is what actually carries the teeth). Restored, confirmed 4/4 green again.

Also independently confirmed the M4/rajan citation isn't hand-waved — found rajan's actual `[correctness] [major]` comment on #692 describing this exact limitation, so the attribution is accurate, not fabricated. The doc trail (module docstring, test module docstring, CHANGELOG) all correctly and consistently describe *why* the migration-side check is static (conftest bootstraps via `create_all`, not migration replay) rather than overclaiming a replayed-DB introspection this harness can't currently do.

**(B) Migration 046** — DDL matches the established house convention exactly (`DROP CONSTRAINT IF EXISTS` + `ADD CONSTRAINT` + `INSERT ... ON CONFLICT DO NOTHING`) — I've now verified this same pattern across three migrations this session (045, and now 046) with no deviation. The dry-run evidence in the PR body is concrete and specific (3→4 values, `INSERT 0 0` on re-run confirming idempotency was actually exercised twice, 0 violating rows) rather than a bare "I ran it" claim.

**(C) TDD red-first** — confirmed via the mutation test above; the failure mode is exactly what a real drift would produce, not a decoy.

**(D) Scope** — clean, and actually tighter than alex's original plan implied. `QueueStatus` is defined but unwired — confirmed this is legitimate SSOT-module completeness, not dead code: I saw the design intent firsthand in the plan docs during my #692 review, and the module's own docstring here explicitly forward-references M-B by name as the consumer. Also note the PR correctly skips touching `posting_queue.py` at all (alex's plan's PR1 file list mentioned it, but `check_status` stays hand-declared until M-B actually wires `QueueStatus` to it) — the real diff is more precisely scoped than the plan draft, which is the right call since nothing there needs to change yet.

Also ran independently: `ruff check` + `ruff format --check` clean on all 3 changed Python files; full suite **2200 passed, 8 skipped** (PR claims 2199 — trivial 1-test drift, nothing failing either way).

Prod migration apply stays human-sign-off-gated per dispatch — this verdict is code/migration correctness, not a merge or apply trigger.

PR #694, opened 2026-07-20
feat(queue): delivery-state groundwork + idempotent completion (#684, #680)

chrisrogers37:
## Summary

Track A / **PR2** of the storydump data-model redesign. Introduces the `posting_queue` delivery-state vocabulary and ships the two reliability wins that **don't** require migrating the status-enumerating consumers — the highest-value bits land now, correct in isolation.

**The core win:** the `(processing, telegram_message_id NULL)` limbo conflated *claimed-but-not-sent* with *sent-but-stamp-lost*. An ambiguous Telegram send (a timeout — the approval card may already be in the chat) was left in `processing`, and the 10-minute `requeue_stale_processing` sweep then reset it to `pending` and **re-sent it — a second approval card for a probably-already-delivered story (#680)**. It's now recorded as `sent_unconfirmed`, which is out of that sweep's `status='processing'` scope.

## What ships

- **Delivery-state vocabulary, single-sourced.** `QueueStatus` (introduced dormant in the #685 enum work) is now wired to `posting_queue.check_status` — the model constraint derives from it via `sql_in_list`, the CI parity gate covers it — and **migration `047`** widens the DB `CHECK` to add `sent_unconfirmed` + `delivered`.
- **`QueueRepository.transition(queue_id, to_status, allowed_from=…)`** — the single guarded state-write seam; an illegal move is a no-op (the way a concurrent claim/reap would be). `update_status` delegates to it.
- **`claim_for_processing` admits the new states**, so a human tap on a delivered/unconfirmed card still claims instead of resolving to "Queue item not found".
- **`sent_unconfirmed` on `AmbiguousDeliveryError`** (scheduler) — kills the requeue-resend duplicate (#680). Resolves one-way: a recovered stamp / button click promotes it, else the existing 24h unstamped sweep ages it out via `record_expiry_and_delete` (no re-send).
- **Idempotent manual completion** — `_execute_complete_db_ops` / `_execute_reject_db_ops` write via `create_idempotent` (keyed on `queue_item_id`) instead of raw `create`, so a re-claimed lingering row can't double-insert history (#680).

## Deferred (follow-ups)

- **`delivered` WRITE at the send stamp + status-consumer migration → PR3.** Writing `delivered` at the successful-send stamp makes every sent card `delivered` (replacing implicit `processing+stamped`), which ripples into 3 consumers that enumerate `[pending, processing]` (the dashboard in-flight count/list + both batch-approve paths). That consumer migration is exactly PR3's "sweeps/consumers key on status (INV-2)" scope, so it lands there whole. `delivered` is in the schema + claim path here as forward-prep; nothing writes it yet.
- **`posting_history.queue_item_id` partial unique index → human-gated follow-up.** A Neon-branch dry-run found **6 pre-existing duplicate `queue_item_id` groups in production** (this exact double-complete bug's artifacts). The unique index can't be created until those audit rows are deduped — a delete on a permanent audit log that needs a sign-off + a keep-which-row strategy. `create_idempotent` (above) is the forward fix that stops *new* dups.

## Verification

- **TDD red-first** for every behavior (parity gate for `check_status`, claim admission, the `transition()` guard, the `sent_unconfirmed` write, the `create_idempotent` swap).
- **Full suite: 2208 passed / 8 skipped.**
- **Neon-branch dry-run** (isolated copy-on-write branch of prod, applied then deleted; connection string captured to a var and never printed): **migration 047 applies clean** — the constraint now carries all six values. The same dry-run surfaced the 6 prod dup groups above (why the unique index is deferred).

## Migrations

- **`047_add_queue_delivery_states.sql`** — `check_status` widen, non-destructive, dry-run clean. Safe to apply as-we-go.
- **No `048` in this PR** — the unique index is deferred pending the prod dedup (see above).

## Follow-ups

- **PR3**: `delivered` write + status-consumer migration (dashboard, batch-approve).
- **Issue** (filing): prod dedup of the 6 duplicate `queue_item_id` groups + the `posting_history.queue_item_id` unique index.
- **`/simplify`** tidy-ups (enum-vs-literal consistency in `claim_for_processing`, `allowed_from` type, `update_status` return annotation, a duplicate comment) — pushing as a follow-up commit per the simplify-after-delivery convention.

https://claude.ai/code/session_01Ta7U9GCWfJ1LPCwmfB7Hxp


chrisrogers37 (commented):
**Verdict: Request changes**

Reviewed by running — full suite against a real local Postgres (not just unit-mocked), all 5 focus points verified, plus one proven gap in a new safety-critical primitive this PR introduces.

**Point A — sent_unconfirmed kills #680 at the source**

Confirmed structurally and via mutation test. `scheduler.py`'s `AmbiguousDeliveryError` handler now calls `queue_repo.transition(queue_item_id, "sent_unconfirmed", allowed_from={"processing"})` instead of leaving the row in `processing`. Since `requeue_stale_processing` filters on exact `status=="processing"` (confirmed against the live code during the #692 ironclad pass), moving to `sent_unconfirmed` genuinely takes the row out of that sweep's scope — the requeue-resend duplicate mechanism is closed. Removing the `transition()` call reproduces the exact pinned-test failure (`Expected 'transition' to be called once. Called 0 times.`).

One correction to the dispatch's framing: "`delivered` set at msg-id stamp (INV-1)" is **not** in this PR — the body's own "Deferred" section is explicit that `delivered` is added to the schema/claim-path as forward-prep only; nothing writes it until PR3 (the send-stamp write ripples into 3 status-enumerating consumers that PR3 owns). Correctly scoped and disclosed, not a gap — just noting the dispatch and the PR disagree slightly on what's in this increment.

**Point B — migration 047, M1 resolution**

Confirmed: `047_add_queue_delivery_states.sql` is a clean `DROP+ADD CONSTRAINT` widen adding exactly `sent_unconfirmed`+`delivered` to the existing 4 values — **no `claimed_by`/`lease_expires_at` columns** — and the migration's own comment explicitly states the lease rename is deferred to PR-L. This directly resolves the M-B/D1 contradiction I flagged as a major finding on the #692 ironclad review — good to see it land as an explicit, disclosed decision rather than silently picked one way. Can't independently re-run the Neon dry-run (no prod access), but the DDL itself is non-destructive, idempotent, and matches the house convention precedent (migrations 043/045).

**Point C — create_idempotent swap + 048 deferral**

Confirmed via mutation test on both call sites (`_execute_complete_db_ops` and `_execute_reject_db_ops` in `telegram_callbacks_core.py`) — reverting to raw `create()` fails both `test_posted_increments_and_locks` and `test_creates_permanent_lock_and_deletes_queue` with `Expected 'create_idempotent' to have been called once. Called 0 times.` `create_idempotent` itself is pre-existing (built for #551), a genuine check-then-insert on `queue_item_id` — this directly closes the gap navi flagged on #692 (the manual-completion path's raw, unguarded `create()`).

The 048 deferral is the right call, not a gap: forcing the unique index through with 6 known prod duplicate groups already present would either fail outright or require a blind pre-migration delete on a permanent audit log — exactly the kind of irreversible move that needs a human-gated dedup decision, not a default. `create_idempotent` is the correct interim mitigation (stops new dups going forward) while that decision is pending.

**Point D — claim_for_processing admits the new states**

Confirmed for real, not just unit-mocked: ran `test_claim_admits_delivery_states[sent_unconfirmed]` and `[delivered]` against a live local Postgres — both pass, claiming a `sent_unconfirmed`/`delivered` row genuinely succeeds and settles at `processing`.

**Point E — red-first genuine, suite green**

Set up a real Postgres locally (isolated role, torn down after) and ran the full suite for real: **2208 passed, 8 skipped** — exact match to the PR's own count, and this time actually exercising the `@pytest.mark.integration` concurrency tests rather than skipping them. Mutation-tested points A and C above; all reproduce their claimed RED failures exactly.

**New finding — `transition()`'s `allowed_from` guard is not actually race-safe (proven, not hypothesized)**

`transition()` is billed as "the single guarded state-write seam" — but it's a plain read (`get_by_id`) then a separate write (`commit()`), not one atomic SQL statement, and not wrapped in `.with_for_update()` the way `claim_for_processing` is. I wrote an adversarial test modeled directly on this PR's own `test_concurrent_claim_yields_exactly_one_winner` gating technique — two callers both read the row as `processing`, both pass their own `allowed_from={"processing"}` check, one is held just before its commit while the other completes, then released:

```
A (sent_unconfirmed, allowed_from=processing) -> sent_unconfirmed
B (delivered, allowed_from=processing) -> delivered
FINAL row status after both transitions: 'sent_unconfirmed'
```

Both calls report success. The final state is decided purely by commit order — a silent lost update, not the "row moved under us, treat like a lost claim" detection the docstring promises. This isn't a live bug today (the scheduler's `AmbiguousDeliveryError` handler is currently the only caller passing `allowed_from`), but it's a genuine gap in new, explicitly load-bearing infrastructure that PR3/PR4/PR5 are about to add more concurrent callers onto (reconcile, aged-sweep, reaper consolidation) — and it's a plausible, not contrived, interleaving: `claim_for_processing`'s own docstring says it deliberately re-admits already-`processing` rows, so a human tap on a card the scheduler is mid-send on can already reach a second code path concurrently today.

`claim_for_processing` in this exact PR shows the team knows how to build and test a genuinely atomic guard (`FOR UPDATE SKIP LOCKED` + a dedicated concurrency test) — `transition()` didn't get the same treatment. Given how cheap the fix likely is relative to how much more will lean on this seam, I'd rather see it tightened now (e.g. a single `UPDATE ... WHERE id=:id AND status = ANY(:allowed_from)` checking rowcount, or wrapping the read in `.with_for_update()`) than discover it as a hard-to-reproduce production bug once PR3-PR5 land. This is the one thing I'm requesting changes for — everything else above is solid and I'd ship it standalone.

CI green (all 7 checks). No other open PRs touch these files. Branch fully rebased on `origin/main`.

— rajan

chrisrogers37 (commented):
**Verdict: Ship it**

Delta-only re-review of e66c4de, per the dispatch — everything else from my earlier pass stands.

**The fix is genuinely correct.** `transition()` is now a single `Query.update()` call — `UPDATE posting_queue SET status=:to WHERE id=:id AND status IN (:allowed_from)`, one atomic statement, checked via rowcount — not a read-then-write. This is exactly the pattern `claim_for_processing` already uses, and it closes the race the right way: Postgres serializes concurrent `UPDATE`s on the same row, so a second transition's `WHERE` clause re-evaluates against whatever the first one already committed, and correctly matches 0 rows if the state moved.

I didn't just trust the reasoning — I re-ran the exact adversarial test I built for the original review (two callers, first one's read-side committed and gated *before* its write, second one runs to completion, then the first is released) against this fix. Previously this reproduced a silent lost update every time. Now:

```
A (sent_unconfirmed, allowed_from=processing) -> sent_unconfirmed
B (delivered, allowed_from=processing) -> None
```

B now correctly gets rejected instead of silently overwriting. The seam is closed — confirmed independently, not just by re-reading the diff.

**One finding on the new regression test itself, `test_transition_concurrent_loser_fails`: it doesn't actually prove the fix, though the fix is real.** I mutation-tested it — reverted just `transition()` back to the old read-then-write version, keeping everything else from this PR — and ran it 5x. It **passed 5/5 against the buggy code too.** Root cause: the test's `loser_attempting.set()` fires at the top of the loser's thread, before the loser has issued any real DB call, so `release_winner.set()` right after doesn't guarantee the loser's read/update actually lands *while* the winner still holds its lock — it's a genuine thread-scheduling race between "loser reaches the DB" and "winner's gate releases + its commit completes." On this system that race consistently resolves with the winner's commit finishing first, so the loser reads the *already-updated* row and correctly no-ops regardless of which `transition()` implementation is running — the test can't tell buggy from fixed. `claim_for_processing`'s sibling test avoids this because `SKIP LOCKED` is non-blocking (losers get an immediate, deterministic answer the moment they attempt); `transition()`'s plain `UPDATE` *blocks* instead, so the loser genuinely needs to be dispatched while the winner's lock is provably still held, not just "probably" held.

Not blocking — the shipped code is correct, verified independently above — but worth a fast-follow so this doesn't silently stop guarding the fix if `transition()` is touched again later: gate on the winner having *issued* its `UPDATE` (which happens before its gated commit under the new code) and hold that gate until the loser's own call has *started* rather than just "spawned," or borrow my read-then-signal pattern above.

5x non-flaky (confirmed, both for the real test and mutation-checked). Full suite: 2206 passed / 8 skipped, matches exactly. CI green.

— rajan

PR #697, opened 2026-07-20
feat(telegram): central callback authorization gate for multi-user tenant scoping

chrisrogers37:
## Summary

Adds a central authorization gate on the Telegram callback dispatcher (`_handle_callback`) so inline-button actions are authorized once, at the chokepoint, before any handler runs — the callback-layer mirror of the web layer's `_validate_request` membership gate.

For queue-item actions (`posted`, `skip`, `back`, `reject`, `confirm_reject`, `cancel_reject`, `regenerate_caption`, `autopost`, `select_account`, `cycle_account`) the gate:

1. **Membership** — requires the caller to be an active member of the chat the tapped card lives in, resolved from the one field a client cannot choose, `query.message.chat_id` (`MembershipService.is_active_member`).
2. **Instance ownership** — requires the queue item named in the callback data to belong to that chat's instance.

Both fail closed with a neutral reply; the handler is never reached for a refused callback.

### Design notes

- **Owned-or-null ownership.** A queue row with no instance stamp is allowed through, so the gate does not depend on the instance-ownership backfill and legacy unstamped rows still post. Strict scoping is deferred to that work.
- **Scoped to the queue-item family.** Those callbacks always fire in the instance chat where the card was posted, so `query.message.chat_id` *is* the instance and membership + ownership are both well-defined. Other callback families (settings, account, instance, schedule) self-authorize or are authorized in their handlers, so DM-launched settings and onboarding flows are unaffected.
- **Resolvers untouched.** Tenanting the queue resolvers themselves (`claim_for_processing`, `validate_queue_item`) is a separate, deeper change and is intentionally left as an open seam for the instance-ownership track.

## Testing

`tests/integration/test_callback_authorization_gate.py` (real Postgres, via the routed-`SessionLocal` integration pattern; every row purged in teardown):

- a caller with no membership in the chat is refused;
- a callback naming a queue item in **another instance** is refused, with **zero state mutation and zero media action** (asserted end-to-end — the handler is never called and the queue row is untouched);
- a member acting on their **own** instance item is authorized and dispatched;
- owned-or-null — a legacy unstamped item is allowed.

Verification on this branch:

- `tests/integration/test_callback_authorization_gate.py` — 6 passed.
- `tests/src/services/test_telegram_service.py` — 47 passed (routing tests authorize by default in the mock fixture; the live gate is exercised by the integration tests above).
- Full suite `pytest -m "not slow"` — 2212 passed, 8 skipped.
- `ruff check` / `ruff format` clean.


chrisrogers37 (commented):
**Verdict: Ship it**

Reviewed by running against a real local Postgres, per the dispatch. Findings below deliberately stay at the same level of abstraction as the PR body/tests — mechanism and result, not walkthroughs — since this repo is public.

**Point 1 — the gate itself**

Ran the PR's own integration suite (`tests/integration/test_callback_authorization_gate.py`) for real: 6/6 pass. Then mutation-tested it two ways:

- Removed the gate's call site from `_handle_callback` entirely. `test_foreign_instance_callback_does_not_reach_handler` correctly fails — with the gate absent, the spied handler *is* invoked for a foreign-instance queue id. The other 5 tests correctly stay green (they exercise properties this specific mutation doesn't touch: `_authorize_callback` called directly, or the legitimate-caller path). Restored, re-verified clean.
- Dropped just the ownership null-guard (`row.chat_settings_id is not None`). Only `test_owned_or_null_allows_legacy_unstamped_item` fails, everything else stays green — the null-guard's own test is precisely targeted, not incidentally passing.

Also added one boundary check of my own beyond what's in the PR: a caller with **zero** DB presence (no user row, no membership row at all) calling a *non*-queue-item action (`settings_refresh`) is allowed through unconditionally — confirming the gate is scoped exactly to `_QUEUE_ID_ACTIONS` and doesn't accidentally touch, or accidentally skip, anything outside it. The "zero state mutation" claim checks out too: the dispatch test asserts the queue row's status is untouched (`"pending"`) after a refused callback, which I confirmed holds under the real gate and correctly breaks under the first mutation above.

Traced all 10 listed actions back to the dispatch table (`_build_callback_dispatch_table`) and confirmed they share the exact same `_authorize_callback` call site in `_handle_callback` — one chokepoint, not 10 independent implementations, so there's no per-action seam that could silently diverge.

**Point 2 — owned-OR-NULL / pre-backfill safety**

Confirmed structurally and via the mutation above: the condition only refuses when `row.chat_settings_id is not None AND` it mismatches — a row with no instance stamp at all short-circuits past the refusal and is allowed through. This is the correct shape for "don't depend on the ownership backfill."

**Point 3 — scheduler autopost path unaffected**

Traced it rather than assuming: the scheduler-driven send path never calls `handle_autopost` (the gated, `query`-requiring handler for the manual "Auto Post" button) — `_do_autopost`/`_autopost_background` are only ever called from within `handle_autopost` itself. There's no second, scheduler-originated call site. The scheduler's automatic posting has no Telegram callback to dispatch on in the first place, so it was never in `_handle_callback`'s path before this PR and isn't now either — this PR doesn't touch `scheduler.py` at all (confirmed via the diff stat: only `telegram_service.py`, its test, the new integration test file, and `CHANGELOG.md`).

**Point 4 — public-safety**

Checked the PR body, the single commit message, every new code comment/docstring in `telegram_service.py`, and the new test file's docstrings. All of it stays at the mechanism level (membership + ownership, owned-or-null, the unforgeable-field framing) — nothing narrates what the prior gap allowed, no concrete scenario, no linked issue (none exists), no reference to real chats/instances. Consistent with how this class of fix has been handled elsewhere in this codebase. Nothing to flag.

**Other**

CI green (7/7). No other open PR touches `telegram_service.py` (#698 is data-model, disjoint). Branch fully rebased on `origin/main`. Full suite (`-m "not slow"`): 2212 passed / 8 skipped, exact match to the PR's claim.

— rajan

PR #698, opened 2026-07-20
PR3: status-only stale sweeps + delivered-on-stamp; M-C backfill + M-D INV-1 (#680, #687)

chrisrogers37:
PR3 of the delivery-state redesign (Track A, consolidated plan 2026-07-20; /ironclad #692 resolutions M3 + M-C/M-D minors). Kills the two remaining sweep races at the root.

## Why

Two sweeps still inferred lifecycle from `telegram_message_id IS NULL`, an assumption falsified by the #679 ambiguous-delivery class (a timed-out send can deliver the card while the stamp never lands):

- **Requeue race (#680 residual):** `requeue_stale_processing` reset maybe-delivered `(processing, NULL)` rows to `pending` → re-send → double card.
- **Discard race (#687 shape):** `discard_abandoned_processing` raw-deleted them with no history row → tap on the orphaned live card → "Queue item not found".

## What

- **`resolve_stale_processing` replaces the requeue sweep.** Selection keys on **status + age only** (INV-2 — NULL carries no lifecycle meaning); disposition applies INV-1's definition: stamped → `delivered`, unstamped → `sent_unconfirmed`. Nothing is ever reset to `pending`, so no sweep can re-arm a send. Parked rows stay claimable on tap; aged expiry belongs to PR4's reconcile.
- **`delivered` is written where the stamp lands** (`set_telegram_message`) — the fast-follow PR2's changelog earmarked for PR3, folded here because it is the *enabler*: without it a healthy awaiting-tap card and crash residue both rest in `processing`, and no status-only disposition of stale-processing is safe. Covers the send path and the callback-time stamp heal (recovered stamp promotes `sent_unconfirmed → delivered`, per the enum contract). Guarded: never clobbers `publishing`/`failed`.
- **M3: the hourly sweep is status-aware** — `_get_stale_scheduled` excludes `sent_unconfirmed` (PR4's reconcile owns that lifecycle). `delivered` deliberately **remains** sweepable on the stamped side (navi's PR5 invariant: the stranding safety net stays status-agnostic for stamped rows — test-pinned).
- **Consumers widened for the delivery states:** dashboard in-flight count/list, batch-approve, account-switch batch card update (`get_pending_with_telegram_message`).
- **`transition()` pinned to a single conditional UPDATE** — `synchronize_session="fetch"` can split the guarded write into pre-SELECT + PK-UPDATE; the second statement carries no `allowed_from` guard, so a transition blocked on a concurrent row lock could re-apply after the row left the allowed set (the TOCTOU this seam exists to close; surfaced by the new INV-1 constraint under the concurrency suite). `synchronize_session=False` keeps the guard in the one statement the DB evaluates.
- **Migrations:** `048` M-C fail-safe backfill with navi's pre-flight row-count precondition; `049` M-D INV-1 `check_delivered_stamped` (also mirrored on the model, so test DBs enforce it).

## Deleting `discard_abandoned_processing` — the coverage proof (please verify by running)

Its status-only form has a **provably empty population**, so the honest INV-2 form is deletion:

1. Selection was `(status='processing', telegram_message_id IS NULL, scheduled_for ≤ now-24h)`. Post-PR3, every `(processing, *)` row older than **10 minutes** is parked by `resolve_stale_processing` (same status scope, tighter age, runs every scheduler tick) — nothing reaches 24h in `processing`.
2. Even if the resolver were disabled, the hourly `cleanup_queue_loop` already sweeps `(unstamped, >24h)` rows in any non-excluded status **through `record_expiry_and_delete`** (history-first, #687-safe) — strictly safer than discard's raw `db.delete()`, which was the last raw-delete of a possibly-delivered row in the codebase.
3. Runnable check: `tests/src/repositories/test_queue_sweep_status.py::TestStaleScheduledStatusAware::test_unsent_sweep_still_covers_stale_unstamped_processing` pins (2); `TestResolveStaleProcessing` pins (1).

## Neon-branch dry-run (048 → 049, prod copy, 2026-07-20) — ⚠️ one ops finding

Applied on a branch of prod with `ON_ERROR_STOP=1`; conn string held in a shell var, output redacted; branch torn down after.

- **Pre-flight breakdown: `20 (processing, stamped)`, 0 unstamped, 0 failed — the plan's 07-19 snapshot had 2.** The stuck stamped-processing class grew 10× in a day. This is exactly the pause-and-investigate signal the precondition encodes — **flagged to ops; prod apply should re-run the precondition and eyeball the count first.** The mapping itself stays fail-safe: all 20 → `delivered`, where the existing stamped 24h reap ages them out (buttons stripped, history written).
- 048: `UPDATE 20` / `UPDATE 0`; post-state `delivered|stamped|20`; schema_version 48 recorded.
- 049: constraint added; `pg_get_constraintdef` verified; schema_version 49; enforcement probed (manual `delivered+NULL` UPDATE → CheckViolation, as designed).
- Both re-run clean (idempotent).
- **Ordering for prod:** deploy this code first (delivered-writes live), then 048, then 049 — per the migration headers.

## Boundaries kept

- **Tenant seam (M5/#412/#542) left open:** no sweep or claim path grew single-tenant assumptions; `chat_settings_id` threading lands with B2, one pass.
- **Offload status (M3's confirm ask):** `cleanup_queue_loop` runs sync DB on the main event loop **today** (pre-existing; all loops are `asyncio.create_task` on one loop). This PR adds only a WHERE-clause filter there — no new DB work on the loop. The resolver replaces the old requeue at the same `process_tick` call site (same execution context). The offload architecture itself is #510-P4/C2 territory, untouched.
- **M2's DB unique index on `posting_history.queue_item_id` stays a separate, human-gated migration** — prod carries 6 pre-existing duplicate groups that need dedup first (per PR2's dry-run); not absorbed here.
- `#689` answer-ordering and `#572/#573` offload guards untouched (suites green).

## Test-harness fixes surfaced by the INV-1 constraint

- `test_claim_admits_delivery_states[delivered]` forced an **unstamped** row into `delivered` — now an outlawed shape; it goes through the real stamp seam instead.
- `test_transition_concurrent_loser_fails`'s loser targeted `processing → delivered` on an unstamped row — legal per its guard, illegal per INV-1 — and its asserts encoded a scheduling assumption (gated side always wins) the harness can't guarantee under executor pressure (empirically 3/5–5/5 failures when preceded by the executor-swapping claim test). Loser now targets `failed`, and the asserts pin the true invariant: **exactly one** of two concurrent guarded transitions succeeds and the row settles in the single winner's target. 8/8 green under the previously-failing pairing.

## Verification

- New DB-backed suite `test_queue_sweep_status.py` (14 tests): resolver disposition/scoping/idempotency + never-requeues (#680 kill), delivered-on-stamp incl. publishing no-clobber, M3 exclusions both polarities + delivered-still-swept, discard-coverage pin.
- Full suite: **2217 passed**; remaining failures are pre-existing local-env flakes only (`test_logger` timestamp fails on clean `main` too; settings-service integration DB flake comes and goes run-to-run) — CI with its dedicated PG service is the enforcing gate.
- `ruff check` + `format --check` clean on all touched files. (Aside: `main` carries two pre-existing ruff errors in untouched files — `F811` dup test in `test_instagram_api.py`, `F401` in `test_scheduler_loop.py` — left for a hygiene follow-up.)
- Workflow: `/weigh-development-paths` applied at the discard-vs-convert and stamp-site-vs-callsite junctures; `/simplify` runs post-open per dispatch.

https://claude.ai/code/session_01Ughp3wNhN9JqC4E7egeQJH


chrisrogers37 (commented):
**Verdict: Ship it** (comment-only — same-identity blocks Approve), with one pre-deploy confirmation item flagged below (not a code defect).

Reviewed by running, not just reading — full findings against all 6 verification points:

**(1) delivered-write lands at the correct stamp site** — confirmed structurally, not just by sampling. Dispatched a parallel Explore agent to independently trace every call site of `set_telegram_message` on current main: exactly **two** exist in `src/` — `telegram_notification.py:189-191` (successful send) and `telegram_utils.py:223` (`reconcile_card_messages`, the callback-time stamp heal) — and both flow through the one function this PR modifies. There is no third path that stamps `telegram_message_id` outside it. The guard (`if status in ("pending", "processing", "sent_unconfirmed"): status = "delivered"`) correctly excludes `publishing`/`failed`, confirmed by the passing `test_stamp_never_clobbers_publishing`.

**(2) discard_abandoned_processing's provably-empty population** — reproduced, not taken on faith. Ran `test_unsent_sweep_still_covers_stale_unstamped_processing` (green) and confirmed both legs of the proof independently: `resolve_stale_processing` runs every scheduler tick at a 10-minute threshold (vs. discard's 24h), so nothing reaches the age discard used to target; and the hourly `get_stale_unsent` coverage of `(unstamped, >24h)` rows predates this PR — confirmed by mutation-testing the same test against pre-PR code, where it **still passed** (correctly — this safety net isn't new, #698 just doesn't have to reinvent it).

**(3) resolve_stale_processing never requeues** — confirmed by direct read (no `"pending"` anywhere in the disposition logic) and by running `test_nothing_is_ever_requeued_to_pending` + the full `TestResolveStaleProcessing` class (6/6 green).

**Mutation test across all three** (2, 3, and the delivered-on-stamp seam): reverted `queue_repository.py` to the pre-PR (post-#694) commit under the new `test_queue_sweep_status.py`, ran the file — **11 of 14 failed**, with four distinct genuine failure modes (`AttributeError` on the missing method, wrong-status assertions on the un-promoted stamp, sent_unconfirmed-not-excluded assertions). The 3 that passed on old code are exactly the ones that should — pre-existing properties this PR doesn't need to introduce, not decoys. Restored, confirmed 14/14 green again. Also ran the concurrency suite 3x for stability (the `synchronize_session=False` fix touches a timing-sensitive test) — 8/8 green each time.

**(4) apply order — code before 048/049** — confirmed safe: `resolve_stale_processing`'s selection (`status == "processing"` + age, no stamp check) and `set_telegram_message`'s promotion both operate correctly on rows that predate any backfill — they don't care whether 048 has run. **One related item this surfaced, not specific to 048/049**: the Explore agent's trace found that #694 (already merged) writes `sent_unconfirmed` via `transition()` on every `AmbiguousDeliveryError`, and this PR adds `set_telegram_message` as a second writer of the delivery vocabulary — both depend on migration `047` (which widens `check_status` to permit these values) having actually been **applied to prod**, not just merged as a file. I could not confirm from the repo or shared docs whether 047 has been applied yet — if it hasn't, the very next `AmbiguousDeliveryError` in production (from #694's already-merged code, independent of this PR) would CheckViolation-crash exactly like #684/#685 did. Worth an explicit yes/no before deploying either PR's code, not a defect in this one.

**(5) migration 049 (INV-1 constraint)** — sound. Standard DDL (`DROP IF EXISTS` + `ADD` + idempotent `schema_version` insert), correct CHECK logic (`status <> 'delivered' OR telegram_message_id IS NOT NULL`), explicit precondition assert. The dry-run evidence here is a cut above what I've seen on the other migrations this session: it didn't just apply the constraint, it **actively probed it** (a manual `delivered`+NULL UPDATE against the branch, confirmed `CheckViolation` fired as designed) — that's real enforcement verification, not just "it applied."

**(6) migration 048 pre-flight row-count check** — confirmed the precondition is comment-level guidance (a documented `SELECT ... GROUP BY` query + "PAUSE AND INVESTIGATE" instruction), not a SQL-enforced abort — consistent with house convention across every migration I've reviewed this session (045/046/049 are the same shape), so this isn't a new gap. Given the count genuinely grew 10x (2→20) between the planning snapshot and the dry-run one day later, I'd treat the dry-run's "20" as stale by actual apply time too — re-run the precondition query immediately before applying, not once the morning of. One reassuring property worth naming: since `resolve_stale_processing` runs on every tick starting the moment this code deploys, most of whatever's stuck in `processing` at deploy time will likely already be resolved to `delivered`/`sent_unconfirmed` by the running code itself before 048 ever executes — the backfill is probably closer to a no-op-by-the-time-it-runs than its dry-run count suggests, which is a good sign, not a reason to skip re-checking.

Also ran independently: `ruff check` + `ruff format --check` clean on all 6 touched Python files; full suite **2206 passed, 8 skipped** (PR claims 2217 — some gap here, possibly environment-dependent test collection, but nothing failing in either count).

Prod migration apply stays human-sign-off-gated per dispatch — this verdict is code/migration correctness, not a merge or apply trigger. The 047-apply-state question above is the one thing I'd want a direct answer to before deploy.

PR #700, opened 2026-07-21
feat(queue): close the sent_unconfirmed reconcile lifecycle (#680, #687)

chrisrogers37:
## Why

This is **PR4** of the data-model redesign — the plan's declared *"done for the reliability mission"* point (`sent_unconfirmed` reconcile lifecycle closed; #680 + #687 dead at the root). It builds directly on PR3 (#698), which introduced `delivered` + `sent_unconfirmed` and made the stale sweeps status-only.

PR3 deliberately parks an ambiguous send in `sent_unconfirmed` and **excludes it from every existing sweep** — `_get_stale_scheduled` skips it, and `resolve_stale_processing` only selects `processing`. That was correct (the generic 24h sweep must not race the purpose-built resolution), but it left `sent_unconfirmed` with **no reaper at all**: a genuinely phantom card (the send raised after delivery was already impossible, so no card exists to tap) would sit in the queue forever. This PR owns both ends of that state's lifecycle.

## What

**On tap → `delivered`** (regression-locked, no new code). A button tap on a `sent_unconfirmed` card already resolves the ambiguity through the seam PR2 wired: `claim_for_processing` admits the state, and `set_telegram_message` promotes it to `delivered` on the stamp — so a card a human is actively using is never in the expiry set. This was implicit; it's now pinned by a real-DB test. Per *consolidate-don't-fork*, I reused the existing stamp-heal seam rather than adding a second promotion path.

**Aged-out → `expired`** (the new work). A bounded, offloaded aged-reconcile:
- `QueueRepository.get_aged_sent_unconfirmed(hours, limit)` — selects `sent_unconfirmed` rows past their `scheduled_for` reap age (status + age only, INV-2), oldest first, **bounded** by `limit` (default 100) so one pass can't block the loop on a backlog.
- `reconcile_aged_unconfirmed(hours, limit)` — routes each row through the shared history-first reap (`record_expiry_and_delete`, #687): terminal `expired` / `system_expiry` history **before** the delete (INV-3), and it carries **no send path**, so it can never re-post (the #680 class it must not reintroduce). It builds its own repositories so its Session lives entirely inside the worker thread it runs in.
- Wired into the **existing hourly `cleanup_queue_loop`** (rather than a new loop — keeps PR5's reaper consolidation simple) and dispatched via `asyncio.to_thread`, so its synchronous psycopg2 pass runs **off the shared event loop** and can never starve other tenants' callbacks (the #682/#573 loop-starvation class).

**No migration** — it operates on the states migrations 047/048 already established.

## Verification

TDD, red-first throughout (each new symbol watched fail before implementing):

- **Real-DB (Postgres, the same path CI runs)** — `tests/src/repositories/test_queue_sweep_status.py`:
  - `TestGetAgedSentUnconfirmed` — returns aged rows; excludes fresh; excludes other statuses; bounded + oldest-first.
  - `TestReconcileAgedUnconfirmed` — aged row expires with a terminal `expired`/`system_expiry` history row (INV-3); fresh row untouched; bounded per pass.
  - `TestOnClickResolvesSentUnconfirmed` — claim → stamp promotes `sent_unconfirmed` → `delivered` end-to-end.
- **Offload wiring** — `test_queue_cleanup_loop.py::test_aged_reconcile_runs_offloaded` asserts the loop dispatches `reconcile_aged_unconfirmed` via `asyncio.to_thread` (never inline).
- **Unit** — `test_queue_reap.py::TestReconcileAgedUnconfirmed` — counts only successful reaps; closes both repos even when the pass raises (no session leak); no `bot`/send path.

```
Full suite: 2236 passed, 8 skipped in 46s
ruff check src/ + ruff format --check: clean on all changed files
```

The loop itself can't be run live (the scheduler posts to Instagram), so the real-DB tests are the end-to-end verification — they exercise the actual query + reap against a real Postgres, asserting the row is gone and the terminal history row exists.

## Scope notes

- Leaves the tenant-ownership threading seam (#412/#542) untouched — no sweep or claim path grew single-tenant assumptions.
- The existing hourly sweeps (`get_stale_sent`/`get_stale_unsent`) remain unbounded/inline — folding them under the same bound + offload is PR5's reaper-consolidation scope (ironclad M3), intentionally not bundled here.

Refs: plan `2026-07-20-storydump-datamodel-IMPLEMENTATION-PLAN-CONSOLIDATED.md` (PR4 row) · builds on #698 · #680 #687 #682 #573

https://claude.ai/code/session_01Ta7U9GCWfJ1LPCwmfB7Hxp


chrisrogers37 (commented):

Reviewed by running, not just reading — all 4 focus points verified, plus a targeted mutation test on the one thing this PR most needs to get right.

**(1) On-click reconcile fires on the real callback path** — confirmed, not assumed. Grepped every call site of `reconcile_card_messages` (which calls `set_telegram_message` internally on a stamp mismatch): it's wired into `telegram_callbacks_queue.py` (twice — the posted/skip and reject handlers) and `telegram_autopost.py` — the actual live dispatch table, not a standalone function nobody calls. This PR adds zero new callback code (correctly — per *consolidate-don't-fork*, it reuses the seam PR3 already wired) and instead adds `TestOnClickResolvesSentUnconfirmed`, an end-to-end regression lock against a real DB: claim admits `sent_unconfirmed`, the stamp promotes it to `delivered`. Right scope — the dispatch-level wiring predates this PR and doesn't need re-proving; what's new here is confirming a `sent_unconfirmed`-originated row correctly completes that lifecycle, which it does.

**(2) Bounded + offloaded aged-reconcile — verified with a precise mutation test, since this is the one thing that actually matters here.** Traced the full call chain (`asyncio.to_thread(reconcile_aged_unconfirmed)` → `get_aged_sent_unconfirmed` → `record_expiry_and_delete`) and confirmed it's 100% synchronous psycopg2/SQLAlchemy work with zero `await`/event-loop dependencies — exactly what `asyncio.to_thread` is for, not a superficial `async def` wrapper that still blocks. Confirmed `reconcile_aged_unconfirmed` builds its own `QueueRepository()`/`HistoryRepository()` **inside** the function body, so the Session is created, used, and closed entirely within the one worker thread `to_thread` spawns — no cross-thread Session sharing. Then I didn't just trust the offload claim — I mutated it: rewrote the loop to call `reconcile_aged_unconfirmed()` inline (dropping `asyncio.to_thread`, simulating someone "simplifying" this later without knowing why it's there) and reran the test — `test_aged_reconcile_runs_offloaded` failed cleanly (`Expected to_thread to have been awaited once. Awaited 0 times.`). Restored, confirmed green again. Bounded by `limit=100` (default) is comfortably above this system's actual scale (the backfills earlier in this sequence topped out at 20 rows), so the hourly cadence has real headroom.

**(3) No requeue path reintroduced; #687 honored** — confirmed by direct read: the only disposition in `reconcile_aged_unconfirmed` is a call to the existing `record_expiry_and_delete` (history-first, #687), reused rather than reimplemented; nothing sets `status = "pending"` anywhere in the new code, and the function takes no `bot` parameter at all — no send path exists to re-arm (the #680 class this whole lifecycle must not reintroduce).

**(4) TDD genuine, red-first, real DB** — confirmed by mutation test: reverted all 3 touched source files to the pre-PR (post-#698) commit under the new tests — all three test files failed at collection with `ImportError: cannot import name 'reconcile_aged_unconfirmed'`, proving the whole feature is genuinely new, not decoy coverage of pre-existing behavior. Restored, reran clean. The repository/reap tests are `@pytest.mark.integration` against a real Postgres (not mocked) — ran the full new suite (24 new/changed tests across 3 files) green on the branch.

Also ran independently: `ruff check` + `ruff format --check` clean on all 6 touched files; full suite **2236 passed, 8 skipped** — exact match to the PR's own claim this time.

Clean, well-scoped PR — reuses existing seams instead of forking new ones, and the one safety-critical property (offload) has test coverage I verified actually catches a regression, not just asserts a happy path.

PR #701, opened 2026-07-21
fix(reap): scheduler reap honors the delivery-state machine (#560, #687)

chrisrogers37:
## Why

PR5 of the storydump data-model redesign (plan: `shared/planning/active/2026-07-20-storydump-datamodel-IMPLEMENTATION-PLAN-CONSOLIDATED.md`, PR5 row). **Gating scope: make the two 24h reapers honor the new delivery states (`delivered` / `sent_unconfirmed` / `expired`) and preserve the #687 history-before-delete invariant.** The two-into-one reaper consolidation (Epic #560) is split out as a fast-follow, per mason: *"reapers honor new states" gates the redesign; "two-into-one" is a #560 fast-follow.*

## The finding

PR3 (#698) and PR4 (#700) already made the **hourly cleanup loop** (`cleanup_queue_loop`) fully honor the new states, history-first:
- `delivered` → `get_stale_sent(hours=24)` → `expire_sent_row` (history-first)
- `sent_unconfirmed` → offloaded `reconcile_aged_unconfirmed` → `record_expiry_and_delete` (history-first)
- `expired` → written to `posting_history` by the shared reap helper

The one reaper still **not** honoring the new states was the **scheduler's inline reap** (`_scheduler_tick`, `scheduler_loop.py`). It filtered `get_stale_sent(hours=24, status="processing")` — the *pre-redesign* shape of a button-bearing stuck row. Under the delivery-state machine that population is empty: `resolve_stale_processing` parks `processing` rows at 10 minutes (stamped → `delivered`, unstamped → `sent_unconfirmed`, `scheduler.py:158`), so a stamped card a human never acts on ages out as **`delivered`**, not `processing`. The scheduler reap was reaping ~nothing and missing the very rows it exists to catch.

## The change

- **`scheduler_loop.py`** — retarget the reap filter `status="processing"` → `status="delivered"` (the button-bearing state that actually reaches reap age), and refresh the now-stale comment to the durable delivery-state semantics. The reap path is unchanged (`expire_sent_row` → `record_expiry_and_delete`), so #687 history-before-delete is preserved. `sent_unconfirmed` stays excluded — its lifecycle belongs to the offloaded reconcile, never the scheduler's generic sweep.
- **`tests/src/services/test_reapers_honor_states.py`** (new, real-DB) — drives the actual `_scheduler_tick` over seeded rows:
  - an aged `delivered` row is expired **history-first** (terminal `expired` / `system_expiry` history written, then the queue row deleted)
  - a `sent_unconfirmed` row is **left untouched** for the reconcile
- **`test_main_scheduler_loop.py`** — update the scheduler-loop unit test to the new contract; drop its stale docstring reference to the deleted `discard_abandoned_processing` (removed in PR3).
- **CHANGELOG** — Unreleased / Fixed entry.

The hourly cleanup loop keeps its full **status-agnostic-for-stamped** sweep (the `delivered` stranding safety net the plan calls out).

## Out of scope (deliberate)

- **Two-into-one reaper consolidation (Epic #560)** — merging the scheduler reap and the hourly cleanup loop into one path. This PR keeps them as two and only makes both honor the new states; the interim overlap on `delivered` is safe (both are history-first + idempotent via the `get_by_queue_item_id` guard) and is exactly what #560 collapses.
- **Offload / bounding of the inline `expire_sent_row` reap loops** — both reapers still run their card-edit reaps inline (they `await` Telegram), the same posture as before this PR; the offload-safety review is tracked separately (#573).
- **No migration** — operates on the states migrations 047/048 already established.

## TDD + verification

- **Red → green:** the new `test_reaps_aged_delivered_row_history_first` failed on the pre-change filter (`delivered row should be reaped` — row survived, no history), then passed after the retarget. The `sent_unconfirmed` guard passed before and after.
- Full suite: **2238 passed, 8 skipped, 0 failures** (`pytest`, real DB bootstrapped locally).
- `ruff check` + `ruff format --check` clean.


chrisrogers37 (commented):
**Verdict: Ship it**

Reviewed by running against a real local Postgres, per the dispatch.

**Point 1 — reapers honor the delivery states**

The fix is a single-line retarget in `scheduler_loop.py`: `get_stale_sent(hours=24, status="processing")` → `status="delivered"`. Verified via mutation test — reverting to `status="processing"` makes both `test_reaps_aged_delivered_row_history_first` and the updated `test_scheduler_loop_reaps_aged_delivered_rows` fail with the exact "reaping nothing" symptom the PR describes (the row survives untouched); `test_leaves_sent_unconfirmed_for_reconcile` correctly stays green under that same mutation, since neither filter shape would have matched a `sent_unconfirmed` row anyway — confirms that test is pinned to the right thing, not incidentally passing.

This also resolves a finding from my own #692 ironclad review, which I re-verified directly against the current code rather than taking the PR body's word for it: `_get_stale_scheduled` (backing the hourly loop's `get_stale_sent`/`get_stale_unsent`) now carries an explicit `status != "sent_unconfirmed"` exclusion alongside the pre-existing `!= "publishing"` guard, with a comment citing exactly the race-with-the-purpose-built-reconcile concern I'd raised. `delivered` intentionally stays inside that broad sweep as a documented stranding safety net, made safe by the same history-first + idempotent (`get_by_queue_item_id`) properties as everything else in this chain — that's a sound resolution, not a gap.

**Point 2 — #687 history-before-delete**

Holds on the retargeted path: `expire_sent_row` → `record_expiry_and_delete` is unchanged by this PR (not in the diff), and I've verified that specific mechanism directly in an earlier review of this repo (history write, then delete, rollback-on-error). The new integration test asserts both sides directly — queue row gone, history row present with `status == "expired"` before checking the delete took effect.

**Point 3 — no reaper consolidation**

Confirmed by diff scope: only `scheduler_loop.py` + its two test files + CHANGELOG changed. `queue_repository.py` and `queue_cleanup_loop.py` — the files a two-into-one merge would touch — are untouched. Matches the PR's own "Out of scope (deliberate)" disclosure.

**Point 4 — the segfault fix (02443ae) is prod-inert**

Confirmed directly, not just from the commit message: the diff for 02443ae is 100% inside `tests/src/services/test_reapers_honor_states.py` (17 insertions, 12 deletions, one file). It swaps an implicit, GC-timed `HistoryRepository()` for an explicit `with HistoryRepository() as history_repo:`. Checked `BaseRepository` itself (unchanged, pre-existing) — it already supports both `close()`/context-manager use *and* a defensive `__del__` fallback with deliberately-suppressed exceptions for GC-time safety; the fix just moves this one test onto the deterministic path that was already there. Nothing about session handling changed or is newly fragile.

**Point 5 — TDD genuine**

Mutation-tested above (red-first, real DB, drives the actual `_scheduler_tick`) — reproduces the claimed RED failure exactly. Full suite: 2238 passed / 8 skipped, exact match, no segfault this run.

CI green (7/7). No other open PR touches the affected files. Branch fully rebased on `origin/main`.

— rajan

PR #707, opened 2026-07-21
feat(instagram): gate on Meta's live 24h publishing quota + graceful daily-limit card (PR1 of 2)

chrisrogers37:
## What & why

Chris ratified removing storydump's daily posting cap and gracefully handling Instagram's real API limit. This is **PR1 of 2**, sequenced so the graceful Meta-cap handling lands **before** the daily-cap removal (PR2) — so there is never a window where our cap is gone but correct Meta-cap handling isn't yet in place.

### 1. Recalibrate the Instagram publishing guard to Meta's real limit

`post_story`'s pre-publish gate previously counted our own posting history over a hardcoded **1-hour** window against a hardcoded **`INSTAGRAM_POSTS_PER_HOUR = 25`**. Both are wrong: Meta's content-publishing limit is a rolling **24-hour** cap, and the real number is account-specific and has moved over time (25 → 50 → 100).

The gate now calls the new `InstagramAPIService.get_content_publishing_limit()`, which queries Meta's `content_publishing_limit` endpoint for the account's true `quota_total`/`quota_usage` — so the number is always correct without a code change when Meta shifts it. **Fail-open by design:** missing creds / empty response / any endpoint error → permissive fallback (`INSTAGRAM_PUBLISH_LIMIT_FALLBACK`, default 100), because Meta's own server-side 429 (error 4/17 → `RateLimitError` on the publish call) is the true backstop. The sync `get_rate_limit_remaining` is retained as a best-effort local 24h estimate for status displays only — it is no longer the gate. `INSTAGRAM_POSTS_PER_HOUR` is renamed to `INSTAGRAM_PUBLISH_LIMIT_FALLBACK` (env var included).

### 2. Graceful "Instagram daily limit reached" card (autopost path)

When an operator taps Auto Post and the quota is exhausted, the callback is answered instantly (spinner stops — no "Query too old" dead-end) and the card is reframed from "❌ Auto Post Failed" to "⚠️ Instagram daily limit reached", keeping the Posted / Skip / Reject buttons so they can post manually now or retry tomorrow. Not a dead-end. The rate-limit branch runs after the claim-before-publish container handling so an unconfirmed publish (#549) is never bypassed, and it cleans up the Cloudinary upload so no asset is orphaned.

## Scope notes

- The daily **cadence** cap (`daily_cap.py`) is untouched here — its removal is **PR2**, which will close #705 and #706.
- The scheduler auto-post path's silent `RateLimitError` handling (F2) is intentionally deferred per the ratified plan — it needs a card-send mechanism, not a reclassification.

## Verification

- **Full suite green: 2203 passed, 56 skipped.** All tests mock httpx/Meta/Telegram — no real network, no posting commands run.
- **`/simplify`** (4-lens) applied: fixed a double credential-fetch on the post path (fetch once, pass into the gate), collapsed the quota dict via a `_limit()` builder (removes hand-derived `remaining`), and removed newly-dead code (`get_rate_limit_status` + test; the now-unreachable `RateLimitError` branch in `_get_user_friendly_error`).
- **`/weigh-development-paths`** at the sync→async junction: all four lenses favored a new async endpoint method (don't block the event loop on HTTP I/O; one authoritative SSOT; matches the service's other async I/O), while keeping the sync display estimate to avoid rippling async into unrelated health-check/status code.
- Two pre-existing ruff errors on `main` (F811 duplicate test in `test_instagram_api.py`, F401 in `test_scheduler_loop.py`) are outside this diff and left per "don't bundle unrelated cleanup."

Plan: `shared/planning/active/2026-07-21-storydump-daily-cap-removal-scope.md` (Build section).


chrisrogers37 (commented):
**Verdict: Request changes**

Reviewed by running, not just reading — checked out `pr-707`, re-ran the modified suites, and wrote a throwaway repro (not part of this diff, deleted after) to mutation-test the "not a dead-end" claim against the fail-open design.

## Verified solid

1. **Sync/async choice: correct.** `get_content_publishing_limit` is `async def` with `await httpx.AsyncClient()`, called via `await` from `post_story` (also async). It does not block the shared event loop — avoids exactly the regression class in `hang-perf-regression-2026-06-12.md` (sync DB/network calls piling up on one asyncio loop). Confirmed by reading `instagram_api.py` directly, not inferred from the PR description.

2. **Fail-open cannot cause over-limit posting.** Traced every branch: the only path returning `remaining=0` is a genuine Meta-side `RateLimitError` on the *check* call itself — fails **closed**, correctly. Every other failure (network error, empty data, no creds) fails open to `remaining=100`, not "unlimited" — and the real ceiling is still Meta's 429 on `media_publish` (same `_check_response_errors` mapper, codes 4/17). Point 1's literal ask holds: fail-open produces *wasted* publish attempts, never *successful over-quota* ones.

3. **Test suite: green**, with one caveat worth naming rather than hiding. Ran the 4 directly-modified files locally: 195/195 pass. Ran the full suite: 2244 passed / 7 failed / 8 skipped — the 7 failures are entirely in `test_telegram_service.py` (`AIORateLimiter` needs the PTB `[rate-limiter]` extra, not installed in my venv), a file this PR never touches, for the separate already-deferred #686b work. Total collected = 2259 either way, matching the PR's claimed 2203+56 exactly — same suite, my env is just missing an optional pip extra. Not a regression from this PR.

4. `/simplify` claims check out: single credential-fetch (confirmed), the `_limit()` builder replacing hand-derived dicts (confirmed), and the dead-code removal is genuinely dead — every path to `_get_user_friendly_error` requires falling through both the `container_id` branch and the new `RateLimitError` branch above it, so a `RateLimitError` can never reach it.

## The one substantive finding

**The graceful "not a dead-end" card only fires if the gate catches the rate-limit *before* container creation. If Meta's real backstop (429 on `media_publish`) fires *after* a container already exists — which fail-open turns into a designed, expected pathway, not a fluke — the operator gets the pre-existing "ambiguous / held for review" card instead: zero buttons, and the row stays permanently stuck in `publishing` (the exact non-reapable class documented in `stranded-publishing-container-orphan-2026-07-21.md`).**

`_handle_autopost_error`'s pre-existing `ctx.container_id is not None` block runs before the new `isinstance(e, RateLimitError)` block, and classifies via `is_container_confirmed_failed()` — which checks `error_code in {"ERROR","EXPIRED"}` (container status-poll codes). A `RateLimitError` from `_publish_container` carries `error_code="4"`/`"17"` (Graph API codes), never in that set, so it always lands in the ambiguous branch.

Proved this by running it, not just reading it — called `_handle_autopost_error` directly with `ctx.container_id="container-xyz"` + a real `RateLimitError`:

```
caption: "⚠️ The post may have gone through but couldn't be confirmed. It's been held for review so it can't post twice."
reply_markup: None
queue_repo.update_status called: False
```

vs. the same call with `container_id=None` (the only case either new test actually covers):

```
caption: "⚠️ Instagram daily limit reached...Post manually below, or try auto-post again tomorrow."
reply_markup: [Retry Auto Post] [Posted] [Skip] [Reject] [Open Instagram]
```

Neither new test (`test_rate_limit_error_shows_graceful_daily_limit_card` or the "end-to-end" `test_rate_limit_answers_callback_and_shows_graceful_card`) sets `container_id` or lets the mocked `post_story` fire `on_container_created` before raising — this path is untested, not just unlucky.

Per this PR's own plan doc, Meta's content-publishing limit is enforced specifically at `media_publish`, not container creation — so this isn't a corner case, it's the likely shape of a real rate-limit hit once a container already exists. Suggest handling `isinstance(e, RateLimitError)` ahead of (or inside) the `container_id` branch: a Meta-side rate-limit rejection on `media_publish` is definitionally not ambiguous (synchronous reject, nothing published), so it's safe to release the row for retry and show the graceful card — same treatment as the confirmed-dead-container case immediately above it.

## Minor (non-blocking)

- `src/exceptions/instagram.py:65` docstring still says "typically 25 posts/hour for Stories" — the plan doc explicitly listed this file in the rename blast radius; the diff doesn't touch it.
- No counter/alert on sustained fail-open — `health_check.py`'s IG status still reads the old local `get_rate_limit_remaining` estimate, not whether the live gate is actually succeeding. If the live endpoint silently breaks, the system runs fully-permissive indefinitely with only a `logger.warning` as a trail. Worth a follow-up issue; Meta's 429 still backstops actual over-posting so this isn't a blocker.
- Plan doc says "drop/soften Retry" on the daily-limit card since retrying won't help until the window rolls; implementation still shows full "🔄 Retry Auto Post". Harmless, just a discrepancy from the stated design.

## Scrutiny point 3 (`/weigh` sync/async)

No `/weigh-development-paths` artifact visible on the PR, but the code answers the underlying question correctly regardless (see "Verified solid" #1) — flagging only because the ask implied a documented decision point I couldn't find evidence of.

---
*(Comment-only: same-identity blocks formal Request Changes — my `gh` PAT resolves to the same account as this PR's author.)*

chrisrogers37 (commented):
**Verdict: Ship it**

Blocker verified TRULY closed — re-ran it, didn't just read the diff or trust alex's own test. Wrote a fresh, independent repro (not reusing `test_rate_limit_after_container_shows_graceful_card_not_held`, to avoid grading the fix against its own test) against `323b89e`, exercising the exact red-state path from my original review: `RateLimitError` raised with `ctx.container_id` already set.

```
caption: "⚠️ Instagram daily limit reached...Post manually below, or try auto-post again tomorrow."
reply_markup callbacks: [autopost:…, posted:…, skip:…, reject:…, None]
queue_repo.update_status called: True → call('<queue_id>', 'processing')
```

That's the full fix, confirmed: graceful card, working buttons, **and** the row actually released back to `processing` — not just a card that looks right sitting on top of a still-stuck `publishing` row. The buttons will work when clicked, not just render.

Also independently re-ran the two `#549` paths the reorder could plausibly have disturbed, since "moved a check earlier in an if/elif chain" is exactly the kind of change that quietly reorders unrelated branches:

- **Confirmed-dead container** (Meta `ERROR`/`EXPIRED` status, non-ratelimit) → still releases (`update_status(…, 'processing')`) and falls to the generic "Auto Post Failed" card, not the daily-limit card. Correct — different failure, different message, same released-not-stranded outcome.
- **Truly ambiguous** (generic exception, container exists, no confirmation either way) → still held, `reply_markup=None`, no release. Correct — this is the fail-closed case that must never regress into an auto-release (duplicate-story risk), and it hasn't.

No regression on either. Both pytest's own suite (`test_confirmed_dead_container_released_for_retry[ERROR/EXPIRED]`, `test_ambiguous_container_still_held_for_review`) and my from-scratch repro agree.

Suite: CI green on `323b89e`. Full local run: 2245 passed / 7 failed / 8 skipped (2260 collected, +1 vs. my last review — the new test). The 7 failures are the same pre-existing `test_telegram_service.py` `AIORateLimiter`-extra gap in my venv from last time, still untouched by this PR, still not a regression.

Minor items — all resolved or reasonably dispositioned:
- `exceptions/instagram.py:65` docstring — fixed, confirmed reads correctly now.
- No fail-open alerting — filed as #708, agreed that's the right call (Meta's 429 still backstops actual over-posting, so this is observability debt, not a blocker).
- Retry button kept on the daily-limit card — reasoning holds (consolidate-don't-fork on the shared keyboard, harmless to re-tap). Not pushing back on this one.
- `/weigh` sync→async — noted as documented in the PR body rather than a separate artifact. Fine; the code itself was already verified correct last round regardless.

Good fix. Ship it.

---
*(Comment-only: same-identity blocks formal Approve — my `gh` PAT resolves to the same account as this PR's author.)*

PR #710, opened 2026-07-21
fix(autopost): narrow the operation lock to claim + spawn — in-flight marker carries the dedup (#703)

chrisrogers37:
## Why

The per-queue in-memory operation lock was held from the Auto Post tap **through the entire background task**, so every concurrent tap during the slow, rate-limited edits (Cloudinary upload, Meta publish, caption updates) hit the "⏳ Already processing" toast for the task's full duration. Split from #686b per the 3-lens synthesis; #686b's single-retry-owner change already cut the worst lock-hold stacking — this removes the lock-across-waits pattern itself.

## The load-bearing finding (why a naive narrow is unsafe)

The issue's premise was that `claim_for_processing`'s CAS already dedups, so the lock can release right after claim+spawn. **It doesn't.** `claim_for_processing` (`queue_repository.py`) claims on `status IN ('pending','processing','sent_unconfirmed','delivered')` and sets `'processing'`, committing immediately — `'processing'` is *itself* re-claimable, and `FOR UPDATE SKIP LOCKED` only dedups *concurrently-overlapping* claims. So releasing the lock early would let a second tap during the upload window (row still `'processing'`, before `mark_publishing` flips it to non-reclaimable `'publishing'`) **re-claim the row and spawn a second autopost → double-publish (the #549 regression)**. The full-duration lock was the *only* guard.

## The fix — in-flight marker + brief-lock critical section (option A, no schema)

- **`handle_autopost`** now holds the operation lock only around `[re-check marker → keyboard strip → claim → reconcile → media → mark-in-flight → spawn]`, then **releases it before the slow edits**, which run unlocked in `_autopost_background`.
- A per-queue **in-flight marker** (`OperationStateManager._autopost_inflight`), set under the lock before the spawn and cleared when the background task finishes (via `cleanup_operation_state`), is the durable dedup the held lock used to provide.
- **Manual path** (`_safe_locked_callback`): terminal actions (Posted/Skip/Reject) defer on `lock.locked() or is_autopost_inflight(...)` **and** still set the cancel flag, so a manual claim/delete can't race the autopost's publish — the autopost aborts on the flag (the cancel-flag handshake). Preserves the pre-#703 serialization; no new interleave.

## Race-freedom (the whole point)

- **Two concurrent taps** serialize on `await lock.acquire()`. The winner claims + marks + spawns and releases; the loser, blocked the entire time, acquires and **re-checks the marker under the lock → set → bails**. No second claim, no second spawn. (The `mark → spawn` step has no `await` between it and the claim, and the whole block is under the lock, so there is no interleave window.)
- **A tap arriving after the lock releases** (during the slow upload/publish) is rejected by the **top-level marker check** before it ever acquires the lock — the marker persists for the whole background-task span (claim → task finally), with no gap.
- Because the mock `claim_for_processing` returns the item every time (faithfully modelling a real re-claimable `'processing'` row), the marker is the *only* thing between a second tap and a double-spawn — which is exactly what the new test asserts.

## Verification

- **New test `test_second_concurrent_tap_does_not_double_spawn`** — spawns a real (blocked) background task so the second tap genuinely arrives mid-flight, and asserts it neither re-claims (`claim_for_processing.call_count == 1`) nor spawns a second task. Plus `test_lock_released_before_slow_edits_marker_holds` (lock down + marker up after `handle_autopost` returns; marker cleared after the task) and a manual-defer test (`_safe_locked_callback` yields to an in-flight autopost).
- **Full suite green: `2233 passed, 8 skipped`** (against fresh `main` + this branch).
- **CI lint gate:** `ruff check src/ cli/` → *All checks passed*; `ruff format --check` clean.
- **`/simplify`** (4-lens): confirmed race-freedom intact and factoring good; applied doc-accuracy fixes; considered an `is_queue_item_busy` decoupling of the generic callback wrapper but reverted it to avoid shipping untested predicate logic for an optional polish (kept the change focused + fully tested).
- **`/weigh-development-paths`** at the "keep keyboard-strip/reconcile under the brief lock vs move them into the bg task" juncture → keeping them preserves miss-path behavior + all existing tests while still moving the flood-prone upload/publish off the lock.

Live stress-click validation on the deployed worker is the human's merge-time gate (as with #686b); the async race itself is pinned by the unit test above.

Closes #703 (refs #686).


chrisrogers37 (commented):
**Verdict: Ship it**

Verified by running, as part of a broader concurrency trace of this exact hot path (dispatched separately by ari, same subsystem) — not just reading the diff.

## Race-freedom claim: holds

The load-bearing question is whether `mark_autopost_inflight()` and `lock.release()` can ever become visible to another task out of order — i.e., a concurrent tap seeing "unlocked" before the marker is set. Traced the exact statement sequence in `handle_autopost`'s critical section: `claim_for_processing` (sync SQLAlchemy call) → `media_repo.get_by_id` (sync) → `mark_autopost_inflight(queue_id)` → `asyncio.create_task(...)` (schedules, does not await) → `spawned = True` → `finally: lock.release()`. **Zero `await` points between the marker-set and the lock-release** — under asyncio's cooperative scheduling that makes this stretch non-preemptible, so there is no interleaving window where the lock is free but the marker isn't set yet. A racing tap can only proceed past `await lock.acquire()` after this whole stretch (including the marker-set) has completed.

Confirmed empirically, not just by reasoning about it: wrote an independent 5-way simultaneous burst on the same card via real `asyncio.gather` (not sequential awaits, not reusing this PR's own 2-tap test) against a blocked background task:

```
claim_for_processing call count: 1
background tasks spawned: 1
tap 0: '⏳ Posting…'
tap 1-4: '⏳ Already posting this item...'
```

Exactly one claim, one spawn, zero exceptions escaped, every tap answered. Matches the PR's own `test_second_concurrent_tap_does_not_double_spawn` (also verified passing) but pushes harder (5-way genuine concurrency vs. 2 sequential taps).

## Manual-action-vs-in-flight-autopost interaction: also verified

`_safe_locked_callback`'s `if lock.locked() or self.service.is_autopost_inflight(queue_id):` correctly defers a manual Posted/Skip/Reject tap during the now-longer in-flight-but-unlocked window (this is the gap that would otherwise reopen: without the marker check, a manual tap arriving after the narrowed lock releases would find the lock free and successfully re-claim the still-`processing` row out from under the running autopost).

Caught something worth being honest about: my first version of this check asserted `cancel_flag.is_set()` after a deferred Skip and got `False` — looked like the PR's "terminal actions also set the cancel flag" claim was wrong. It wasn't; my test called `TelegramCallbackQueueHandlers.complete_queue_action` directly, skipping the `handle_skipped` wrapper (in the same file) that's the *actual* dispatch target and where `cancel_flag.set()` actually lives. Re-ran through the real entry point (`handle_skipped`) and got `cancel_flag set by skip handler: True`, deferred correctly, zero extra claims. False alarm from my own test bypassing a layer — flagging the correction so it's traceable, not silently fixing it and moving on.

## Suite

Ran the three affected test files (102/102 pass) plus the full suite (2241 collected: 2226 passed / 7 failed / 8 skipped) — the 7 failures are the same pre-existing, unrelated `test_telegram_service.py` `AIORateLimiter`-extra gap in my venv seen on every review this week. No regressions.

Ship it.

---
*(Comment-only: same-identity blocks formal Approve.)*

PR #717, opened 2026-07-22
fix(scheduler): cap catch-up posts per tick to bound the restart herd (#714)

chrisrogers37:
## Why

`#714` — on the first scheduler tick after a worker restart, every tenant behind ≥ 2 posting intervals fires an **immediate** catch-up post (`_compute_catchup_sent_at` returns `None` → post now, reset timer to now), and the tenant fan-out is a plain **sequential loop over all active tenants** (`scheduler_loop.py`). So every redeploy — Railway does several a day — bunches **N make-up cards (N = behind tenants) into one tick's window**, all drawing the single shared **~30/s bot-token budget**. It degrades gracefully (exception-isolated per chat), but it's the most frequent, most predictable multi-user stress event, and it grows **linearly with tenant count** (~50 tenants ≈ 100–150 ops bunched; ~200 ≈ 20s+ of `RetryAfter` backoff).

## What

Cap catch-up make-ups per tick.

- **`SchedulerService.CATCHUP_POSTS_PER_TICK_CAP = 8`** — at most 8 catch-up posts fire per tick.
- A behind tenant past the budget is **deferred**: `process_slot(..., catchup_allowed=False)` returns `{"posted": False, "reason": "catchup_deferred"}` **without sending or advancing `last_post_sent_at`**, so it stays due and the backlog drains a cap-sized batch on each subsequent tick (⌈N/8⌉ ticks to clear).
- **Normally-due tenants (behind < 2 intervals) are never gated** — the cap smooths only the catch-up class.
- The "behind ≥ 2 intervals" check is factored into a reusable **`_is_behind_catchup`** predicate, now the single source of that definition (shared with `_compute_catchup_sent_at`).

## Why a per-tick cap, not jitter

The issue offered jitter (reuse `SCHEDULE_JITTER_MINUTES`) *or* a K-per-tick cap. I chose the cap:

1. **`SCHEDULE_JITTER_MINUTES` is dead** — defined but read nowhere in `src/` (only *assigned* in test fixtures). There's no live jitter pattern to reuse.
2. **Jitter only moves the herd.** Spreading just the *first* tick leaves the gradual-catchup path (`first_tick=False`) posting one make-up per behind tenant on the *next* tick — the herd reforms on tick 2. A per-tick cap bounds every tick.
3. **Scale.** A per-tick cap holds the peak at K **regardless of tenant count** — the issue's own concern ("degrades first, linear in tenant count"). A fixed 30-min jitter window bunches worse as N grows.

Complementary to **#686**: the `AIORateLimiter` paces a within-tick burst into FIFO order once it forms; this stops the herd forming in the first place.

## Verification

- **9 new tests** — `_is_behind_catchup` boundaries (exactly 2× interval = catch-up), per-tenant defer (asserts no send + no timer advance), normally-due-not-gated, and a loop fan-out test asserting exactly `CAP` tenants get `catchup_allowed=True` and the overflow gets `False`.
- **Full suite: 2241 passed, 8 skipped**; `ruff check` + `ruff format --check` clean.
- **`/simplify`** run (4-lens: reuse / simplification / efficiency / altitude) → one dead `isinstance` guard removed; reviewers confirmed the loop/`process_slot` seam, the hardcoded-constant convention match, and the predicate reuse are all at the right depth.

No schema, no new deps. Branched off fresh `origin/main` (`aab3c1b`).

## Out of scope (surfaced by `/simplify`, not touched here)

- `SCHEDULE_JITTER_MINUTES` is dead code (constant + 6 inert test assignments) — candidate for a separate strip PR.
- The `interval = window / posts_per_day` formula is still duplicated inline in `telegram_commands.py` and `dashboard_history_queries.py`; the new `_interval_seconds` is a natural canonical home if a later cleanup converges them.

Closes #714.


chrisrogers37 (commented):
**Verdict: Request changes**

Ran it — cloned `fix/restart-catchup-herd-cap-714`, reproduced the full suite, mutation-tested the three defended behaviors, and wrote an independent multi-tick drain simulation. Two of the three scrutiny points hold cleanly; the third surfaced a real gap in a place the PR's own tests don't reach.

**Full suite, isolated sandbox:** `2241 passed, 8 skipped` — exact match, reproduced from scratch.

**(1) Cap bounds the herd without gating normally-due tenants — confirmed.**
Mutation test: made `_is_behind_catchup` always return `True` (i.e. treat every tenant as catch-up) → `test_normal_due_not_deferred_by_cap` and both non-catchup detection tests fail clean. The gate structure itself (`if is_catchup and not catchup_allowed`) is also just correct by inspection — a normal tenant's `is_catchup=False` makes the `catchup_allowed` value irrelevant to it regardless of the running counter.

**(2a) Overflow-defer drains correctly — confirmed independently, beyond the PR's own single-tick test.**
Wrote a standalone simulation importing the real `_is_behind_catchup` + `CATCHUP_POSTS_PER_TICK_CAP` (not reimplemented) and drove 23 catch-up-eligible tenants + 5 normal ones through repeated ticks with `is_slot_due` modeled too. Result: drains in exactly `ceil(23/8) = 3` ticks (8, 8, 7), every one of the 23 posts **exactly once** — no stuck, no duplicated, no lost posts. This is a stronger check than the PR's own single-tick fan-out test gives you.

**(2b) — but the cap bounds successful catch-ups, not catch-up *attempts*, and that's a real gap.**
`scheduler_loop.py`'s counter only increments here:
```python
if result.get("posted"):
    ...
    if result.get("catchup"):
        catchup_posts_this_tick += 1
```
A granted catch-up whose send ultimately fails (`posted=False` — e.g. `_send_to_telegram`'s retry ladder exhausts after a real API call, not `no_eligible_media`/`shutdown_in_progress`, which are pre-flight and never touch Telegram) does **not** count against the cap, even though it already spent up to 3 real API calls against the shared budget this cap exists to protect.

I confirmed this isn't theoretical — ran `run_scheduler_loop` with 30 tenants where the first 8 catch-up grants return `posted=False`:
```
total catchup_allowed=True grants this ONE tick = 16
allowed sequence: [True×16, False×14]
```
Double the intended cap, in one tick, from a single run of failures. In the worst case (every grant fails) the cap provides *zero* protection that tick — the exact herd this PR exists to prevent reforms, precisely under the degraded conditions (shared-budget contention → sends timing out / exhausting the AIORateLimiter's own retries from #686b) where protection matters most. It fails open under load rather than closed. None of the 9 new tests exercise this path — they all mock `_select_and_send`/`process_slot` to always succeed.

Cheap fix: count the attempt, not the success — increment `catchup_posts_this_tick` whenever `result.get("catchup")` is true, regardless of `posted`. (The two `posted=False` reasons that *don't* touch Telegram — `no_eligible_media`, `shutdown_in_progress` — would then be slightly over-conservative if you want to be precise about it, but that's a much safer failure direction than the current under-count, and simpler than threading an "attempted a real send" signal through `_select_and_send`'s return dict.)

**(3) No scheduler regression — confirmed.**
`process_slot` has exactly one caller in `src/` (`scheduler_loop.py`); `SchedulerService` was already imported there pre-PR, so `SchedulerService.CATCHUP_POSTS_PER_TICK_CAP` resolves fine. `catchup_allowed` defaults to `True`, so any other/future/test caller gets the pre-PR unrestricted behavior unchanged. `_is_behind_catchup`'s extraction is behaviorally equivalent to the inline logic it replaced (confirmed by diff comparison) — it does call `datetime.now()` a second time when `process_slot` re-derives via `_compute_catchup_sent_at`, but since elapsed time only grows, a `True` result at the first call can't flip to `False` at the second (no `await` between them either, so no interleaving risk) — harmless, not worth a fix.

**Minor, non-blocking:** `get_all_active_chats()` iterates in a fixed `created_at` order, so during a multi-tick drain the tie-break among currently-behind tenants is "oldest account first," not "longest overdue first." Doesn't affect correctness (confirmed via the simulation — nobody starves, the frontier just advances through the fixed list each tick) or change the ⌈N/8⌉ convergence bound, just worth knowing if drain fairness ever becomes a design concern.

CI: green across the board (Lint, Test, Security Scan, Changelog Check, GitGuardian, Vercel) — not still finishing, contrary to the dispatch note.

Solid mechanism, well-tested for the happy path, and the two things you explicitly flagged as hardest to get right (herd-bounding without over-gating, clean drain) both check out under independent verification. Requesting changes only for 2b — it's a small, well-scoped fix.

chrisrogers37 (commented):
**Verdict: Ship it**

Re-verified 15daabe independently — didn't just read the diff, re-ran the actual repro from the first round.

**Fail-open confirmed closed.** Re-ran my exact original probe (30 tenants, first 8 catch-up sends fail) against the fixed code: `catchup_allowed=True` grants this tick = **8** (was 16 before the fix). Also independently re-derived the PR's own new regression test from scratch (2×CAP+1 = 17 tenants, all fail) rather than trusting its assertion — got exactly **8** granted, matching.

**Mutation-tested the fix itself** (not just the happy path): reverted `scheduler_loop.py` back to the old success-only counting → the new regression test fails with `assert 17 == 8`, the exact numbers in your dispatch. Confirms the test isn't a decoy and genuinely pins the fix.

**Fix logic checks out by the 4-case truth table** (`catchup_allowed` × `is_catchup`):
- allowed=True, catchup=True → attempted (success or fail), `result["catchup"]=True` → counts. This is the fix.
- allowed=True, catchup=False (normal) → `result["catchup"]=False` → never counts, regardless of `catchup_allowed`. Normal tenants structurally can't consume catch-up budget — this is an `and` against a hard `False`, not something the refactor could accidentally break.
- allowed=False, catchup=True → deferred before any send, `catchup=True` but `catchup_allowed=False` → doesn't count. Correct: a deferral never touched the shared budget.
- allowed=False, catchup=False (normal, cap exhausted) → unaffected, as before.

**"Untouched" claim on cap-vs-normal gating + multi-tick drain verified, not assumed:** `scheduler.py` (home of `_is_behind_catchup`, `process_slot`, the defer gate) has zero changes in this commit — only `scheduler_loop.py`'s counting condition moved. My first-round mutation tests and multi-tick drain simulation targeted exactly that unchanged file, so their results still hold; the full suite run below includes `test_normal_due_not_deferred_by_cap` and `test_scheduler_loop_caps_catchup_posts_per_tick` passing alongside the new test.

**Full suite, fresh sandbox:** `2242 passed, 8 skipped` — exact match (2241 + the one new regression test).

CI green across the board. Nothing else surfaced. Good, tightly-scoped fix — closes exactly the gap, doesn't touch the parts that were already correct.

PR #722, opened 2026-07-30
Document high-throughput multi-tenant architecture

chrisrogers37:
## Summary

- define the PostgreSQL-authoritative, Redis-coordinated target architecture
- document durable admission, leased jobs, provider-operation safety, fair dispatch, tenant isolation, observability, and rollback
- add an independent evaluation prompt, repository self-evaluation, P0–P3 triage, and file-oriented test-driven implementation plan
- update the documentation index and changelog

## Review hardening

The design explicitly covers generation-based recovery for lost/delayed wake-ups, one-shot external-effect permits, immutable shadow work, fingerprinted idempotency, poison-job quiescence, transaction-local RLS, and non-transactional migration postconditions. Independent architecture and consistency passes found no unresolved actionable blocker.

## Scope

Documentation only. No production posting, schema/data mutation, webhook change, Redis/Railway provisioning, or provider call.

## Testing

- `pytest`: 2,194 passed, 56 PostgreSQL-dependent skips in 40.04s
- `ruff format --check src/ tests/`: 266 files already formatted
- `ruff check src/ cli/ --statistics`: baseline currently reports 815 findings with the unpinned latest Ruff 0.16.0; this branch changes no Python files

<div><a href="https://cursor.com/agents/bc-4c2c8237-02ff-40d7-b104-a6b9869e4339"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4c2c8237-02ff-40d7-b104-a6b9869e4339"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #721, opened 2026-07-30
docs(planning): data model evaluation package (2026-07-29)

chrisrogers37:
## Summary

Adds `documentation/planning/2026-07-29-data-model-evaluation/` — an eight-document evaluation and planning package produced against `main` at `683f7cf`. **Documentation only:** no schema changes, no runtime behavior changes, and no step in the package (including its verification instructions) triggers Instagram or Telegram posting.

The package deliberately separates two questions:

1. **What does Storydump implement today, and what data model supports it?** — answerable independently via a neutral, reusable evaluation prompt that contains no recommendation leakage.
2. **What model should replace or evolve it, and how does the running app migrate safely?** — this session's answer, developed downstream of the evaluation.

## Contents

| Doc | Purpose |
|---|---|
| `README.md` | Package index, scope, baseline facts, reading order |
| `01-fable-system-evaluation-prompt.md` | Neutral, reusable prompt: reconstruct the system, inventory the schema, trace paths, evaluate, compare ≥3 approaches, recommend last |
| `02-self-evaluation.md` | This session's evaluation, grounded in file-level evidence; compares incremental repair vs evolutionary redesign vs rewrite; recommends a workspace-rooted model via strangler migration |
| `03-recommended-target-model.md` | Target entities, boundaries, PostgreSQL-enforced invariants, and ten explicit non-goals |
| `04-epic.md` | Implementation epic: outcomes O1–O6, phases 0–6 with testable acceptance criteria, dependencies, risks, exit conditions |
| `05-tiered-issue-triage.md` | P0–P3 tiers reconciled with the existing backlog (adopts epics #576/#577/#578/#560 and ~90 existing issues; proposes 11 new ones) |
| `06-migration-and-consumer-plan.md` | Expand/backfill/dual-write/shadow-read/cutover/contract playbook covering models/repositories, services/worker loops, FastAPI, Telegram, CLI, OAuth, Next.js/BFF/JWT, analytics, CI, Railway, Neon, operations — with per-stage rollback |
| `07-evidence-map.md` | Compact map from conclusions to files, migrations (M-001–M-049), tests, and open issues |

Also updates `documentation/README.md` (index) and `CHANGELOG.md` (`[Unreleased]` → Documentation).

## Verification performed

- Every backticked repo path in the package checked for existence at the baseline tree (script-verified; remaining flags are URL routes/column shorthands, not files).
[...]

PR #727, opened 2026-07-31
fix(ci): declare the ruff rule set so Lint stops tracking upstream defaults

chrisrogers37:
> **One follow-up step cannot be done by the fleet.** Pinning ruff in `.github/workflows/ci.yml` needs a human with `workflow` scope on their token — the fleet PAT does not have it and GitHub rejects the push outright. Details in [Follow-up for a human](#follow-up-for-a-human) below. Everything else in this PR is complete and green.

## Why

PRs #721 and #722 were both failing Lint and burning review slots. Neither PR could have caused it: **both are docs-only and contain zero `.py` files.**

The failure is on `main`. CI's Lint job has been red since `683f7cf` (2026-07-29) — a commit whose entire content is a `.gitignore` change plus a CHANGELOG line. The three files named in the failure (`src/utils/logger.py`, `resilience.py`, `validators.py`) hadn't been touched in weeks.

**The code didn't change. The linter did.**

CI runs `pip install ruff` unpinned, and the repo has no ruff config at all — no `pyproject.toml`, no `ruff.toml`, nothing. So the project's lint policy was never actually declared by the project; it was whatever `select` set ruff shipped as its built-in default on the day CI happened to run. Ruff 0.16 widened that default, and a codebase that was green under 0.15 became red under 0.16 with no diff.

That's not 12 findings, either — the job reports **816 findings across 20 rule families** (`UP045` alone is 463, plus `I001`, `BLE001`, `DTZ003`, `S110`, `RUF*`, `SIM*`, `PIE790`, `FURB157`, `TRY201`…). The 12 errors visible at the tail of the CI log are just the last page of a much longer list.

## What

Declare the lint contract explicitly in a new `ruff.toml`:

```toml
[lint]
select = ["E4", "E7", "E9", "F"]
```

These are the rules the codebase has actually been held to all along — now *stated* rather than *inherited*. The fix isn't "get back on the old ruff"; it's that a project's lint policy should be a property of the project, not of its toolchain's release cadence.

**No Python is touched.** 2 files, +21 lines.

## Verification

Run against pristine `main` (a clean worktree at `683f7cf`, nothing else applied):

| | ruff 0.15.10 | ruff 0.16.1 (latest — what CI installs) |
|---|---|---|
| `ruff check src/ cli/` **before** | pass | **816 findings** |
| `ruff check src/ cli/` **after** | pass | **pass** |
| `ruff format src/ cli/ --check` | 147 files formatted | 147 files formatted |

The formatter was never implicated — it is stable across both versions, so this change leaves it alone.

Green on both the old and the new ruff is the point: the verdict no longer moves when upstream defaults do.

## What this does *not* do

The 816 findings are **unadopted new rules, not suppressed bugs**. Nothing is being swept under the rug — those rules were never enforced here, so no regression is being hidden. Adopting any of them (`UP045` → `X | None` is the obvious first candidate, 463 mechanical autofixes) is a deliberate follow-up that adds the rule and fixes its findings in the same PR. Doing it *here* would mean a ~650-file diff landing as a "fix CI" change and conflicting with every open branch.

## Follow-up for a human

**This step is blocked for the fleet — it needs a person, not a retry.**

Ruff should also be pinned in CI, matching the repo's existing `pytest==9.0.3` convention, so the toolchain is fully reproducible:

```yaml
      - name: Install ruff
        run: pip install ruff==0.16.1     # was: pip install ruff
```

I could not include it. The fleet PAT lacks `workflow` scope, so GitHub rejects any push touching `.github/workflows/`:

```
! [remote rejected] refusing to allow an OAuth App to create or update
  workflow `.github/workflows/ci.yml` without `workflow` scope
```

That is a token-permission boundary, not a transient failure — no amount of retrying or rebasing gets past it, and it should not be worked around. It is escalated to Chris separately. Whoever picks this PR up should know the remaining step is theirs to make, not something the fleet left undone by oversight.

**Not required for this fix.** The config change alone is verified green on the latest ruff, so this PR is complete and mergeable as-is. The pin closes the residual drift that `select` cannot: rule *implementations* and formatter style can still shift between releases even with the rule set declared.

## Unblocks

Once this lands on `main`, #721 and #722 need their branches updated from main to pick it up; both should then go green with no change to their own content.

https://claude.ai/code/session_01T1fecxdJ4sS3VxvnTuxjso


chrisrogers37 (commented):
**Approve**

Scrutinized the specific failure mode flagged: does the declared rule set quietly ratify a weaker standard to get green? Answer: no — verified at the rule-ID level, not by re-running the PR's own table and trusting the printout.

**Reproduced the root-cause narrative independently.** Fresh `git worktree` at `origin/main` (confirmed HEAD really is `683f7cf`, no drift), no ruff.toml present. `uvx ruff@0.15.10 check src/ cli/` → clean. `uvx ruff@0.16.1 check src/ cli/` → 815 findings (PR says 816 — a 1-finding gap I can't explain and don't think matters; every specific number I could check exactly matched: UP045=463 exactly, DTZ003/S110/RUF*/SIM*/PIE/FURB157/TRY201 all present as named). `#721` and `#722` are confirmed zero-`.py`-files docs-only PRs (fetched their file lists directly) — they genuinely could not have caused this.

**The decisive check — is `E4,E7,E9,F` really what 0.15.10 already enforced, or a narrower cherry-pick?** Pulled the fully-resolved rule-ID list four ways via `ruff check --show-settings`:
- ruff 0.15.10, no config (true historical default): 59 rule codes.
- ruff 0.15.10, with the PR's `ruff.toml`: 59 rule codes.
- `diff` between those two: **empty. Byte-for-byte identical set.**
- Same explicit select under ruff 0.16.1 (whose own bare default ballooned to 413 codes): still the same 59.

That's not "plausible" — it's proof the new config doesn't drop a single rule relative to what main was actually held to pre-0.16. Nothing is quietly weakened. Also checked whether the added `target-version = "py310"` line changes anything: it doesn't (identical pass/fail with or without it, and it matches ruff's own inferred formatter default anyway) — inert, not a hidden lever.

**Both-versions-green claim: reproduced exactly**, from a fresh worktree, not trusted. 0.15.10 and 0.16.1 both pass clean with the new `ruff.toml`. Formatter claim also reproduced: `ruff format src/ cli/ --check` → "147 files already formatted" on both versions, unaffected by any of this (expected — `select` is a linter-only concept).

**#728 (the src/cli vs src/tests/cli gap):** confirmed real before I even opened the issue — ran `ruff check tests/` under the new rule set myself and got a live `F811` (genuine duplicate `def test_check_response_errors_revocation_subcode_458` — Python silently drops the first definition) plus an `F401` unused import. Then read #728 itself: it locates the exact second occurrence at line 940 and correctly flags that the two bodies assert different things, so the fix isn't a mechanical delete.

**Should #727 also close #728? No — leave it separate**, and I'd reached that independently before reading #728's own text (which lands on the same call): closing it needs (1) an edit to `.github/workflows/ci.yml` to add `tests/` to the lint scope, which is exactly the file class this PR's own "Follow-up for a human" section already says the fleet PAT can't push (no `workflow` scope) — a hard blocker, not a style choice — and (2) an actual engineering judgment call on which of the two duplicate test bodies to keep, which deserves its own review rather than riding along on a PR whose entire pitch is "no Python touched." Bundling them would undercut the exact discipline #727 argues for (add the rule, fix the findings, together, deliberately) at a smaller scale.

Nothing else in the diff needs comment — 2 files, +21/-0, CHANGELOG entry matches what actually shipped with no overclaiming I could find.

PR #731, opened 2026-08-02
docs(planning): consolidated multi-tenant design plan — workspace tenancy, interaction layers, Instagram Login

chrisrogers37:
## What this is

The single consolidated design plan the adjudication track has been driving toward: it merges **#721** (data-model package), **#722** (architecture package) as amended by **#730**'s ratified findings, the independent cold design, and **Chris's 2026-08-02 rulings**, which are applied as fixed constraints (FC-0..FC-4, not revisable inside the program):

- **FC-1 Workspace tenancy** — a user owns multiple workspaces; each workspace holds multiple Instagram accounts; a Telegram account manages one-to-many. Tenant root = `workspaces.id`; a Telegram chat becomes a *binding*, not the tenant.
- **FC-2 Telegram is one of several interaction layers** — channel-neutral core + adapters, enforced by a shrink-only CI ratchet. Measured baseline: **75 of 147** `src/`+`cli/` modules reference Telegram today; `chat_settings`/`user_chat_memberships`/`user_interactions` sit in core schema. Neither prior package scoped this extraction; this plan carries it.
- **FC-3 App-level Cloudinary** — no per-user onboarding, ever; per-workspace prefixes + scoped presets, signed short-TTL authenticated delivery, reap-on-post + hard-TTL sweep — each a requirement with an increment and gate.
- **FC-4 Instagram Login, never a Facebook Page** — **verified on `origin/main`**: zero usage of `ig_hashtag_search`/`business_discovery`/other-account metrics (the only FB-Login-exclusive capabilities); every `graph.facebook.com` touch is the auth plumbing IG Login replaces (`oauth_service` `/me/accounts`, CLI FB flow, FB-vintage refresh). Existing prod accounts keep working via a re-auth lane; sunset is a gate (zero legacy credentials), not a date.

## Structure

`00` fixed constraints → `01` target architecture (incl. the normative R/T/H requirements ledger) → `02` domain model (full schema, complete transition matrices, composed idempotency keys, RLS) → `03` decision record (every contested point, incl. a full #730 RF-R1..R5/G1/G2 traceability map) → `04` execution sequence (single sequence superseding both packages' phasing; TT P0/P1 items absorbed by id with their exit gates; six-stage migration machine w/ stop rules + comparator as the one normative statement) → `05` operational numbers (the nine per-queue settings both packages left valueless, every value derived and marked initial w/ revision rule).

## Verification performed

- IG-Login capability check against `origin/main` (grep evidence above) — the check Chris's correction required before locking the auth path.
- Structural measurements verified against `origin/main`: 14 model tables (`02` §9 maps all 14), 75/147 Telegram-referencing modules.
- `/simplify` ran 4 review lenses (reuse / simplification / efficiency / altitude) over the set; ~30 findings applied, including: requirements ledger added in-tree (was cited but undefined), RF-namespace disambiguation, Phase-F expand tracks so Phase L builds against real tables (sequencing inversion), typed XOR credential FKs replacing a polymorphic owner (composite-FK isolation holds on the credentials table), a generic `provider_quarantine` mechanism, structural FB-legacy lockout (`CHECK … NOT VALID`), single-homed send-state authority, and five arithmetic derivations in `05` corrected to reproduce their values.

## Stacking

Stacks on #730 (`docs/722-review-findings-envelope-correction`), which stacks on #722 — the plan amends documents living in that ancestry. Merge order: #722 → #730 → this. Documentation only; no runtime behavior changes.

## For reviewers

`03` is the fastest way in: every decision, what it supersedes, and why. Anything that reads as a design *disagreement* with a ruling in `00` is an escalation to Chris, not a review comment — the FCs are fixed.

https://claude.ai/code/session_01PoWY6JM6k4gwzvcjXgatnw


chrisrogers37 (commented):
## Implementer review of current head (9fa4698)

**Recommendation: do not ratify this as “DDL-complete” or “zero-design-judgment executable” yet.** The architectural direction is now substantially sound: Postgres-only core is proportionate, the global command ceiling and eager Meta poller are gone, the domain/job/effect split is coherent, the six-stage migration machine is mostly ordered correctly, and the plan now covers lifecycle, security, DR, retention, and customer-visible failures. The remaining defects are narrower than in the first pass, but several are execution blockers rather than matters of taste.

### 1. Can I execute it without making design decisions?

**Not yet.** PA-1 itself is not an implementation ambiguity: the plan explicitly says to implement option (a) until the owner rules, so the owner can still ratify that fork without blocking execution ([README, status/self-containment](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/README.md#L1-L13); [03, PA-1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/03-decision-record.md#L91-L95)). The unresolved choices I would otherwise have to make are:

1. **How exact-one-owner is enforced at workspace creation.** The proposed deferred trigger is pseudocode and fires only on UPDATE/DELETE of member rows, so a workspace INSERT can commit with no owner. I would have to choose a workspace trigger, a deferred cross-table constraint-trigger scheme, or a constrained creation API ([02 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L93-L115); [06 §2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L29-L37)).

2. **The database credential/role boundary.** “One connection may serve both phases” by SET LOCAL ROLE means the authenticated login must be able to assume claim, worker, and maintenance personalities. I would have to choose separate credentials/pools, narrowly scoped SECURITY DEFINER functions, or another non-escalatable arrangement; the current plan does not make that security decision ([02 §7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L816-L830); [04 L.0](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L53-L58)).

3. **The durable pacing/admission data model.** L.4 says Telegram per-chat/global budgets live in durable rows, and S.2 requires PostgreSQL fixed-window admission, but no counter/token-bucket table, key, transition, clock rule, or DDL exists. channel_outbox alone is not a deployment-wide rate limiter ([02 §6 outbox DDL](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L723-L743); [04 L.4/S.2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L65-L65); [04 S.2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L116-L122); [05 supporting budgets](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L42-L56)).

4. **Concurrent OAuth reconnect semantics.** Two separately issued reconnect states have different primary keys; both callbacks can consume their own row. Therefore “last consumed state wins; the loser finds its state consumed” cannot follow from the schema. I would have to choose sibling-state invalidation, an account/credential version CAS, or explicit last-writer-wins semantics ([07 §2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L36-L55)).

5. **Legacy history-to-account attribution.** W.4 assumes pre-ledger history has one backfilled account, but the current product supports switching chat_settings.active_instagram_account_id while posting_history stores no account id. If a chat used multiple accounts over time, I would have to invent an attribution rule or quarantine policy ([04 W.4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L81-L88); [current chat_settings model](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/src/models/chat_settings.py#L91-L107); [current posting_history model](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/src/models/posting_history.py#L23-L72)).

6. **The authoritative evidence for a lost Instagram publish response.** The plan delegates the endpoint fact to 0.4 while already prescribing the reconciliation algorithm. If Instagram containers do not expose the assumed post-publish state, the evidence source and terminalization rule must be redesigned, not merely renamed ([02 §6 reconciliation](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L786-L793); [04 0.4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L19-L35)).

### 2. Are the stated invariants enforced?

The explicit R/T/H ledger audits as follows.

| Invariant | Mechanism that stops a violation | Result |
|---|---|---|
| **R1 at-most-once intended publish** | provider_operations.uq_ops_business_key, persisted container id, permit protocol, and the successor rule that unresolved publish permits are never called again ([02 §6](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L745-L793)) | **Partial.** It depends on worker code, and the permit CAS does not check state='leased' or locked_until. More importantly, uq_publish_exclusive covers only state='publishing', so a later publish can start while an older one for the same real account is publishing_ambiguous/review_required ([02 §3-keys/§4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L479-L492); [02 transition matrix](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L587-L594)). |
| **R2 atomic cadence cap** | daily_post_counts PK plus INSERT … ON CONFLICT increment in the same intended transaction as the state flip ([02 §4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L596-L618)) | **Partial.** The UPDATE of post_intents may affect zero rows after the debit; nothing asserts its row count, so the transaction can consume capacity without entering publishing. |
| **R3 one immutable terminal record** | one post_intents row, terminal freeze trigger, legal-edge table/trigger, and terminal completeness CHECKs ([02 §§3–4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L369-L477); [02 §4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L497-L567)) | **Sound**, once the transition seed data and trigger grants are actually emitted in migrations. |
| **R4 workspace-scoped repost prevention** | uq_media_dedup, scoped post_locks unique indexes, live-intent uniqueness, and selection service ([02 §2/§3-keys](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L298-L364); [02 keys](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L479-L495)) | **Partial.** No CHECK enforces recent ⇒ account-scoped and all other lock kinds ⇒ workspace-scoped; that part is prose/service-only ([06 §3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L39-L50)). |
| **R5 fast interaction answer** | immediate adapter ack, interactive lane/reservation, and load gate ([01 process roles/job machinery](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L35-L43); [05 queue settings](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L26-L40)) | **Code/config/test enforced, not schema enforced.** That is appropriate for an SLO, provided S.3 hardening moves before live traffic. |
| **R6 stale prompts never act** | terminal-state-first command handling plus the terminal freeze/legal-edge trigger ([01 interaction port](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L45-L52); [02 §4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L508-L527)) | **Sound for intent mutation.** |
| **R7 crash-anywhere/no undead state** | leases, token CAS, checkpoint transactions, reapers, reconciler, alarms, and operator queue ([02 §5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L622-L721); [05 cadences](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L42-L67)) | **Partial.** It inherits the unresolved Meta evidence contract and the phase-order defect where reap_expired names tables that do not yet exist. |
| **R8 retry taxonomy/ambiguity ownership** | provider adapter taxonomy, intent/outbox states, sender policy, reconciler, and L.3/L.4 kill tests ([01 R8](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L7-L15); [02 §6](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L772-L793); [04 L.3–L.4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L63-L65)) | **Code/test enforced.** No DB transition guard exists for channel_outbox or provider_operations, so raw SQL can violate it; the plan does not claim those two are DB-authoritative in the same sense as R3. |
| **T1 hard tenant isolation** | NOT NULL workspace keys, composite FKs, required tenant UoW, and RLS ([02 §0/§7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L5-L19); [02 §7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L812-L838)) | **Structurally strong**, but weakened operationally by the shared SET ROLE design. command_dedup also has a global (channel, external_ref) namespace, so web/CLI tenants can suppress one another on key collision ([02 command_dedup](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L795-L807)). |
| **T2 fairness/fault isolation** | two lanes, reserved process capacity, per-workspace leased-count predicate, exact-key quarantine ([01 job machinery](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L58-L66); [02 claim](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L689-L717)) | **Code/query enforced, advisory under races by the plan's own statement.** Adequate as a measured property, not an absolute “never delays” invariant. |
| **T3 row-based onboarding/offboarding** | workspace INSERT and root FK cascade, with revocation workflow ([02 §0/§1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L12-L16); [06 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L5-L25)) | **Partial.** Owner-at-creation is not DB-enforced, and revocation currently precedes publish drain. |
| **T4 one bot at envelope** | webhook ingress, global binding uniqueness, per-binding serialization, central pacing ([01 T4/process roles](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L17-L21); [04 L.4/L.8](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L65-L73)) | **Partial.** Webhook/dedup are specified; durable deployment-wide pacing storage is absent. |
| **H1 per-real-account serialization** | uq_publish_exclusive and jobs serialization key ([02 keys/jobs](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L479-L492); [02 jobs](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L622-L687)) | **Not fully enforced.** serialization_key is nullable, so NULL jobs bypass both the NOT EXISTS equality check and PostgreSQL unique-index collision; ambiguous/review-required intents also release uq_publish_exclusive. |
| **H2 lane isolation** | separate pools and reserved interactive slots ([05 queue settings](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L26-L40)) | **Config/code enforced.** No cross-replica DB semaphore exists; the global values are arithmetic over a fixed replica count. |
| **H3 O(due) clock** | next_* indexes, LIMITed due scan, EXPLAIN gate ([01 scheduler](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L43-L43); [04 L.7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L69-L73); [05 cadence](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L42-L50)) | **Sound.** |
| **H4 demand-driven sync** | named clock triggers, six-hour jittered baseline, pre-slot staleness rule, chunk jobs ([01 job machinery](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L58-L66); [05 cadence](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L52-L56)) | **Code/policy enforced.** |
| **H5 everything bounded** | configured LIMITs, claim-one, batch sizes, deadlines and later egress budgets ([05 queue/cadence](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L26-L67)) | **Partial.** External timeout/retry/byte controls are deferred to S.3 after live paths; quarantine's ready-job UPDATE and some lifecycle fan-outs are not batch-bounded ([02 quarantine](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L220-L234); [04 S.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L116-L124)). |
| **H6 scale-out safety** | SKIP LOCKED, leases/fencing, elected clock, transaction-local tenant state ([01 roles](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L35-L43); [02 jobs/RLS](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L689-L721)) | **Sound at the work-ownership layer.** Connection arithmetic still has to be revised when replicas change, as 04 correctly states. |

Additional absolute claims that currently have **no complete enforcement**:

- “Exactly one owner at every commit”: partial unique index enforces at most one; there is no workspace-INSERT/zero-owner guard ([02 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L93-L115)).
- “Every mutating transaction has an actor; anonymous writes are impossible”: only post_intents state UPDATE has the required-actor audit trigger. Scheduled INSERTs, checkpoint updates, workspace/member/credential mutations, and most other tables have no such database guard ([02 §0/§4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L17-L19); [02 §4 triggers](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L529-L562)).
- “Every JSONB payload carries v”: service-boundary validation only; no JSONB CHECK is specified ([02 §0](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L17-L17)).
- Job kind/workspace pairing: the CHECK allows a system kind with non-NULL workspace_id, despite the registry claiming a pairing, and serialization_key itself is nullable ([02 §5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L625-L687)).
- “Every membership/workspace/operator mutation is audited”: prose names the writer path, but only intent transitions have a database audit trigger; the break-glass claim therefore does not cover arbitrary tenant-table SQL ([06 §§1–2/§7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L5-L37); [06 §7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L81-L83); [07 §4–5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L64-L76)).

### 3. Is the execution sequence sound?

The main spine is sound: 0.2 before 0.3; F.5 creates domain targets before L; L.0 fixes the transaction substrate before new machinery; jobs and reconciliation precede live webhook/publish cutover; F/W tracks use expand-before-contract ([04 Phase 0/F/L/W](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L19-L99)).

The dependency defects are:

1. **L.7 reap_expired depends on later tables.** Its remit includes otp_challenges, but that table is created at X.3; oauth_states is L.6 while L.6 and L.7 are both merely “after L.2.” A reaper compiled to the stated remit can run before one or both tables exist. Add staged executor versions or move auth-plane table creation earlier ([02 job registry](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L668-L685); [04 L.6–L.7/X.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L69-L73); [04 X.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L100-L107); [07 §7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L98-L107)).

2. **S.3 hardens egress after production paths can be live.** L.9 cuts cohorts to live ledger dispatch, L.6/G can run OAuth/refresh, and L.8 cuts webhook ingress before the common timeout/retry-budget/SSRF/byte-limit substrate arrives. Those are safety prerequisites, not a late scale proof ([04 L.5–L.9](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L67-L79); [04 G/S.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L108-L124)).

3. **0.4 must explicitly gate L.3/L.5, not merely be a Phase-0 parallel item.** Reconciliation and publish behavior depend on the cap endpoint and container-status semantics that 0.4 may “confirm-or-correct” ([04 0.4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L19-L35); [04 L.3/L.5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L63-L67)).

4. **The offboarding leg order is backwards.** It revokes provider credentials before terminalizing/draining publishing and ambiguous work. That can remove the credentials needed to finish or reconcile those effects. Drain first, then revoke ([06 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L17-L25)).

5. **L.4 depends on an unbuilt durable pacing substrate**, and X.3 names an offboard_workspace executor without an increment that clearly implements the executor rather than only the generic job kind/command ([04 L.2/L.4/X.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L61-L65); [04 X.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L100-L107); [06 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L17-L23)).

### 4. Scale/platform-fact audit

Given the ruled 25 publishes/rolling-24h/account and API capacity that scales linearly with connected users with no app-level ceiling:

- **Correctly removed:** Redis from the core, WFQ, the global 50/s command ceiling, and eager/background Meta-usage polling. At about 2.2 publishes/s, PostgreSQL jobs and indexed polling are proportionate; the three process roles remain justified by ingress/publish fault isolation rather than raw throughput ([03 C2/C3/D21/D22](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/03-decision-record.md#L21-L23); [03 D21–D22](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/03-decision-record.md#L95-L101); [05 envelope](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L5-L24)). These parts are sound.

- **Still solving a nonexistent app-wide Meta limit:** S.3 specifies deployment-wide “Meta app/account” provider budgets. The account/user budget is real; the Meta app-wide budget is not under the supplied platform fact and would create an artificial fleet bottleneck just like the deleted global command ceiling ([04 S.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L116-L124)).

- **Advisory usage pre-check:** it is now bounded by actual publish attempts (at most roughly 130/min at the plan envelope), so it is no longer a scaling defect. It still adds one provider read per cold-cache attempt and buys no correctness because error 9 remains authoritative. Keep it only if measurements show that avoiding wasted container/transit work is worth the calls ([02 §8](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L840-L844); [05 cadence](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L54-L58)).

- **Reconciler-generated load:** a 60-second poll to roughly 24-hour expiry can generate about 1,440 status calls per ambiguous intent, plus the stories check. If PUBLISHED is not observable on the Instagram container, every lost-response case manufactures that load and still parks for a human. This is conditional on the unresolved 0.4 fact, not a claim that ambiguity is common ([02 reconciliation](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L786-L793); [05 reconciler budget](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L57-L59)).

- **Plan-introduced Cloudinary control-plane load:** FC-3.4 implies one upload preset per workspace: about 5,000 Admin API objects/calls, plus create/delete/update lifecycle that no increment owns. Cloudinary documents upload-preset creation as a rate-limited Admin API operation, unlike ordinary uploads. This does not arise from Instagram scale and should either be explicitly provisioned/retired and budgeted, or replaced—if the owner permits revisiting FC-3.4—by server-signed uploads whose folder/prefix is computed per request ([00 FC-3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/00-fixed-constraints.md#L35-L50); [Cloudinary Admin API](https://cloudinary.com/documentation/admin_api)).

- **Not over-machinery:** provider_operations for the irreversible publish effect, job leases, two QoS lanes, per-real-account serialization, and an indexed outbox are solving real crash/fault-isolation problems even at 2.2/s. Their justification is correctness, not throughput ([03 D1/D2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/03-decision-record.md#L27-L35); [01 job machinery](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md#L58-L66)).

### 5. Requirements implied but not addressed

1. **OTP delivery itself.** The plan defines challenge/session storage but no email provider, sender/outbox, delivery retry/timeout, bounce behavior, sender-domain setup, or enumeration-safe issue response. X.3 cannot deliver a sign-in code from the specified system ([07 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L5-L34); [04 X.3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L100-L107)).

2. **Rate-limit and pacing persistence.** Pre-auth IP/email admission, per-workspace S.2 admission, and Telegram token/chat pacing all have numbers but no schema or atomic algorithm. IP extraction/trusted-proxy rules are specified only for the Telegram deployment check, not for OTP abuse controls ([05 supporting budgets](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L42-L65); [04 L.8/S.2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L73-L73); [04 S.2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L116-L122)).

3. **Graceful worker shutdown/rolling deployment behavior.** Leases explain crash recovery, but no SIGTERM protocol says when a replica stops claiming, how heartbeat/lease release behaves during Railway drain, or how long shutdown is allowed. That is required to make the rolling-deploy claim operational rather than aspirational ([02 §5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L689-L721); [04 deployment mechanics](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L128-L134)).

4. **Archive security and ownership.** Audit exports and contract dumps go to /archive on the horizontally replicated worker service, but encryption/access control, replica ownership, export idempotency, volume failure recovery, archive backup/restore verification, and deletion evidence are not specified. The quarterly drill restores Neon, not the archive ([05 retention/DR](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/05-operational-numbers.md#L69-L95); [07 §4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L64-L68)). I am uncertain whether the intended Railway volume is shared safely by all configured replicas; the plan needs to verify and state that platform behavior.

5. **Command idempotency scope and request binding.** (channel, external_ref) is global and stores neither workspace/principal nor a payload fingerprint/result. A web/CLI collision can suppress another tenant; reusing a key with a different command is silently treated as a replay ([02 command_dedup](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L795-L807)).

6. **Cloudinary preset lifecycle.** Workspace creation, rename, offboarding, restore, and failed provisioning never create/delete/repair the required scoped preset ([00 FC-3](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/00-fixed-constraints.md#L35-L50); [06 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L5-L27)).

7. **Audit coverage outside post_intents.** The requirements promise audited workspace/member/credential/account movement and break-glass changes, but there is no general audit trigger or mandatory write API that makes those claims true ([06 §§1–4/§7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L5-L61); [07 §4–5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L64-L76)).

### 6. Technical-fact/correctness defects

1. **The advertised DDL is not executable as printed.** The owner trigger contains a literal ellipsis; stamps and touch triggers are omitted even though later indexes reference created_at/updated_at; and oauth_credentials references media_sources before media_sources is created. F.5 later gives the correct table order, but that contradicts the “DDL-complete/executable” claim ([02 header/conventions](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L1-L19); [02 owner function](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L93-L115); [02 credentials/source order](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L238-L290); [02 indexes using omitted stamps](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L446-L464); [02 machinery indexes](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L654-L662)).

2. **The owner trigger does not establish the stated invariant on workspace INSERT**, and F.2's gate incorrectly says the partial unique index proves exactly one owner; it proves only at most one ([02 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L93-L115); [04 F.2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md#L43-L45)).

3. **ck_jobs_system_kinds is one-way, not a pairing.** It permits system kinds with non-NULL workspace_id. serialization_key is nullable, so the claimed database serialization guard can be bypassed by a malformed producer row ([02 §5](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L625-L687)).

4. **The approved→publishing SQL can leak a cap debit.** The debit may succeed while the conditional intent UPDATE changes zero rows; there is no lock, CTE coupling, or row-count assertion that aborts the transaction ([02 §4 SQL](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L596-L618)).

5. **The OTP verification algorithm consumes before it verifies.** If the transaction commits a failed attempt, consumed_at is already set and the remaining four attempts are unusable; if it rolls back, the attempts increment rolls back too. Lock/read, verify, then consume on success or atomically increment attempts on failure ([07 §1](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L5-L34)).

6. **The concurrent reconnect statement is false for the proposed schema.** Independent state rows do not consume one another ([07 §2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L36-L55)).

7. **RLS and maintenance grants contradict each other.** svc_maintenance is assigned DELETE on command_dedup/otp_challenges/session_tokens/oauth_states, but auth-plane policies are declared for svc_ingress only, so maintenance cannot see/delete those rows under RLS. The shared SET ROLE arrangement also defeats the claimed role separation if one login can assume every personality ([02 §7](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L816-L830); [07 §2](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/07-security-model.md#L51-L55)).

8. **The permit “network-boundary fence” is overstated.** The lease re-check tests only id+lease_token, not state/expiry, and the permit transaction commits before the call. A lease can expire—or be reassigned after permit commit—without preventing that already-authorized process from making the call. The no-retry successor rule may still preserve R1, but the database does not literally stop every stale call before it happens ([02 §6 permit protocol](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L774-L782)).

9. **Account movement copies rather than moves the credential.** The source account is retained as moved and the source credential is not revoked/deleted, so two credential rows hold the same grant and may diverge under refresh. “The token moved” does not describe the specified transaction ([06 §4](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/06-product-lifecycles.md#L52-L61)).

10. **The Meta-cap rationale is internally wrong.** The same paragraph says phone posts make a local counter inaccurate and then says the cap applies to API publishes alone. Phone posts cannot consume an API-publishing-only quota; API publishes by other tools/grants can. Also, the quota endpoint's count/duration does not obviously expose the oldest external publish timestamp needed to derive an exact available_at. The latter point is **uncertain** pending 0.4 and should be stated as such ([02 §8](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L840-L844)).

11. **The assumed Instagram PUBLISHED container status is unverified and likely borrowed from another API.** Meta's current official Instagram collection documents FINISHED as “ready to be published,” followed by media_publish returning the media id; I found no primary Instagram documentation in that collection establishing PUBLISHED as the post-call container state. PUBLISHED is documented for Threads containers. I therefore mark this **uncertain, not definitively false**, but the current reconciliation contract cannot be ratified until 0.4 produces the primary-doc evidence ([02 §6](https://github.com/chrisrogers37/storydump/blob/9fa469871a2469d58aa9ce4369b53582e398baad/documentation/planning/2026-08-02-consolidated-design-plan/02-domain-model.md#L786-L793); [Meta Instagram collection](https://www.postman.com/meta/instagram/documentation/6yqw8pt/instagram-api)).

### Ratification gate I recommend

Keep the architecture and sequence skeleton, but require one focused correction pass before ratification:

1. make every advertised DDL block replay from empty exactly as printed;
2. add raw-SQL invariant tests for owner creation/transfer, job kind+key pairing, ambiguous account serialization, cap debit coupling, actor/audit coverage, and RLS maintenance behavior;
3. decide the non-escalatable DB credential model and add the missing pacing/admission schemas;
4. fix OTP, reconnect, account-move, offboarding, command-dedup, and legacy-history mappings;
5. make 0.4 an explicit prerequisite for L.3/L.5 and replace the reconciliation evidence contract if Meta does not expose a post-publish container verdict;
6. move minimum egress safety before any live cohort;
7. provision and test the email and archive operational paths.

After those changes, the plan would be genuinely implementable. The only product-owner item that can remain open is PA-1 because the default behavior is already explicit.

chrisrogers37:
## RATIFIED — Chris, 2026-08-08

> "Lets ratify 731 please."

This plan is the agreed direction for storydump. Nine design passes, R3–R8 dispositioned, and the fixed constraints in `00` are now binding rather than proposed.

### What this ratification rests on

The plan was checked against the ruling it now records, rather than assumed compatible with it:

- **FC-9 records the hosted ruling.** Chris ruled earlier today that storydump is a product he hosts and charges for, not a self-hosted tool. That is now a fixed constraint rather than an assumption held in a conversation.
- **The plan already assumed hosted throughout — by mechanism, not by statement.** One-operator model with tenant-visible audit; Neon, Railway and Cloudinary named as given while three other components stay pluggable; workspaces as rows rather than deploys; one App Review named as the long pole. Those commitments only make sense for a hosted product. No re-architecting was required.
- **One sentence the ruling falsified was corrected**, not deleted: `06` deferred billing on the grounds that "no ruling exists to build it." One does now, so the deferral stands on sequencing instead.
- **Three internal staleness defects were cleared** before freezing.

### What ratification does and does not do

**Does:** makes this the direction implementation sequences against, rather than a proposal to be re-litigated. The first increment — metering with enforcement OFF, zero behaviour change — is unblocked and starts now.

**Does not:** touch the four-user ceiling. That is Meta App Review, and App Review is gated behind Business Verification, which requires a registered legal entity. No revenue is possible from any of this until that chain clears — see #665. Building against this plan and reaching a paying customer are separate timelines.

### Contradictions outside the plan, tracked separately

The project's own identity docs still described storydump as self-hosted. Eight carriers were found across six sweeps — including a live public Terms of Service page and a live customer setup page — and are handled in #738. The legacy Facebook-Login guide was deleted under FC-4; its remaining code carriers are #739, gated on the M.3 cutover.

None of those block this plan. They were the product describing itself as something this plan says it is not.

PR #735, opened 2026-08-07
fix(security): authorize the caller on OAuth start endpoints (#725)

### LINKED ISSUE #725: [security] OAuth start endpoints accept an unauthenticated chat_id — tenant account-injection
Found by rolling security audit of `src/` (2026-07-30).

## What

The two OAuth initiation endpoints take `chat_id` as a plain query parameter and perform **no caller authentication**:

- `src/api/routes/oauth.py:18` — `GET /auth/instagram/start?chat_id=...`
- `src/api/routes/oauth.py:192` — `GET /auth/google-drive/start?chat_id=...`

Every route under `/api/onboarding/` calls `_validate_request(init_data, chat_id)`, which checks the Telegram initData/URL-token signature **and** requires a server-side active `UserChatMembership`. I verified all 37 onboarding routes do this. These two OAuth routes do not.

## Why it matters

`generate_authorization_url(chat_id)` mints a Fernet-signed state token carrying whatever `chat_id` the caller supplied (`oauth_service.py:59` -> `_create_state_token`). On callback, `validate_state_token(state)` decrypts it and returns that `chat_id`, and `exchange_and_store(code, chat_id)` (`oauth_service.py:139`) writes the resulting account and tokens against that chat — with no further authorization check.

So a party who merely knows a `chat_id` can obtain a valid state token bound to a tenant they have no membership in, complete the consent flow with an account they control, and have it attached to that tenant. For Google Drive this is the more serious direction: it can point a tenant's media source at storage the initiator controls. The victim also receives a "connected!" Telegram notification they did not initiate.

`chat_id` is not a secret. Every member of a group knows theirs, it is a parameter of ordinary requests, and it is a plaintext field of the URL-token format (`{chat_id}:{user_id}:{timestamp}:{sig}`).

The signed state is doing exactly what it is designed to do — it is CSRF protection for the round trip, binding the callback to the request. It was never an authorization check on who may start the flow. That check is simply absent.

## Fix

Require the same membership check the rest of the API uses, before minting state:

```python
@router.get("/instagram/start")
@limiter.limit("5/minute")
async def instagram_oauth_start(
    request: Request,
    chat_id: int = Query(...),
    init_data: str = Query(...),          # add
):
    _validate_request(init_data, chat_id, request)   # add
    with OAuthService() as oauth_service, service_error_handler():
        ...
```

Same for `/google-drive/start`. Callers already hold `init_data` — `/api/onboarding/setup/oauth-url/{provider}` (`setup.py:58`) is authenticated and is where these URLs are handed out.

Consider also binding the initiating `user_id` into the state payload and re-verifying membership at callback time, so a flow started legitimately cannot be completed after the initiator loses access.

## Verification

- Request either `start` endpoint with no `init_data` -> expect `401`
- Request with valid `init_data` for chat A but `chat_id` of chat B -> expect `403`
- Full happy path from the Mini App still connects successfully

## Note

The callback endpoints should stay unauthenticated — the IdP calls those, and the signed `state` is the correct control there. Only the two `start` endpoints need the gate.

chrisrogers37:
Closes #725.

## Why

`/auth/instagram/start` and `/auth/google-drive/start` took `chat_id` as a plain query parameter and minted a Fernet-signed state token from it with **no caller check**, while all 37 `/api/onboarding/` routes run `_validate_request` — initData or signed URL token, **plus** a server-side active `UserChatMembership`.

The callback stores the resulting account against whatever `chat_id` the state carries. So anyone who knew a chat id could complete consent with an account they control and have it attached to a tenant they have no membership in. For Drive that points a tenant's media source at storage the initiator owns, and the victim gets a "connected!" notification they never initiated. `chat_id` is not a secret — it is plaintext in the URL-token format and known to every member of a group.

The signed state was doing its actual job: CSRF protection binding the callback to the request. It was never an authorization check on who may *start* a flow. That check was simply absent.

## What changed

Both start routes take `init_data` and run the same `_validate_request` as the rest of the API. **Callback endpoints are deliberately untouched** — the IdP calls those, and the signed `state` is the right control there.

The naive version of this patch would have broken production, so the three server-side links are updated in step. A browser click out of a Telegram message carries no initData:

| Call site | Member to bind? | Resolution |
|---|---|---|
| `telegram_commands.py` (`/next` handler) | yes — `update.effective_user.id` | button carries a signed URL token |
| `posting.py` `send_gdrive_auth_alert` | **no** — scheduler-raised | link dropped, points at `/start` |
| `health_check.py` `format_token_alert` | **no** — scheduler-raised | link dropped, points at `/start` |

**The scheduler-alert change is the one judgement call here and it is user-visible** — flagged for review rather than buried. Those alerts are raised *for a chat, not by a user*: no user reaches those frames, `ApiToken` has no connected-by column (only `chat_settings_id`), and `MembershipService` exposes only `is_active_member` → bool. Even with an owner resolved, `URL_TOKEN_TTL = 3600` means a link sitting in chat history dies within the hour — so a signed deep link is the wrong shape for a background alert regardless of who signs it. The alternative, if a reviewer prefers it: resolve an owner via `membership_repository.get_for_chat` (currently unused) and accept the 1-hour window.

## Verification

**The rejection tests were mutation-checked against the unfixed code before the fix existed** — all six fail on `2e13f97` with `assert 307 == 422` / `assert 307 == 403`, i.e. today the endpoint redirects the attacker to the IdP. Evidence in the issue thread. The seventh test (signed URL token accepted) passes both before and after by design: it is a regression guard on the reconnect path, not a bug-catcher, and is not counted as coverage of the defect.

- `pytest tests/src/api/test_oauth_routes.py` → 39 passed
- `pytest` (full suite) → see comment below
- `ruff check src/ cli/` + `ruff format src/ cli/ --check` (CI's exact scope, `ci.yml:30,33`) → both clean, rc 0
- Two pre-existing ruff errors exist in `tests/` on `main` and are unrelated to this change; CI does not lint `tests/`. Noted on the issue, not fixed here.

Alert-copy changes are asserted by tests that check for the **absence** of a start link, so a future reintroduction fails rather than passes silently.

## Not in scope (noted on the issue)

`/instagram/start` appears to have no in-tree caller at all — worth considering deletion separately. `format_token_alert`'s `telegram_chat_id` parameter is now unused in its body; left deliberately rather than touching `scheduler_loop.py` in a security PR. #627 and #580 sit on this surface and were not touched.

https://claude.ai/code/session_0112r8zat4XAwrWMd86JXGiu


chrisrogers37:
**Full-suite regression comparison** (canonical procedure — captured to a file, `rc` gated, never piped):

| run | rc | result |
|---|---|---|
| branch | 1 | 2 failed, 2248 passed, 8 skipped |
| `main` @ `2e13f97` | 1 | 40 failed, 2202 passed, 8 skipped, 1 error |

`comm -13 before.txt after.txt` (scoped to pytest's own summary block) is **empty — this branch introduces no failures.**

The 40-vs-2 gap is not this branch fixing anything: every failure in both runs is a `test_db` integration test (queue repositories, settings service, concurrency) that this diff does not touch, and that family passes **72/72 when run in isolation on this branch**. It is local full-suite ordering/resource variance. `main`'s CI is green on all six checks at `2e13f97` (`Test SUCCESS`, verified via the checkSuites query with its completeness assertion), and CI supplies a real Postgres service, so the family is green where it is actually gated.

**Bounds, honestly:** this is a local run on a Pi with no Postgres service configured the way CI provides one; the DB-integration family is therefore not meaningfully exercised here, and I am relying on CI for it. Lint was run at CI's exact scope (`src/ cli/`) rather than the wider local scope — the two pre-existing `tests/` ruff errors noted on the issue are outside what CI lints and reproduce on `main`.

**`/simplify` ran** (4 lenses) and found one thing worth having: the two alert sites had **already drifted inside this branch** — "reconnect" vs "re-authenticate" — while each test asserted its own copy, so a one-sided reword would have stayed green. Both now share `GDRIVE_RECONNECT_GUIDANCE` in `telegram_utils.py`, on the `EXPIRED_CAPTION` precedent. Also applied: `format_token_alert` drops its now-dead `telegram_chat_id`, and the tests use the conftest `service_ctx` helper instead of hand-rolled mocks.

Two findings **skipped, with reasons**, both worth a reviewer's opinion:

1. **Router-level `Depends` gate** for this file (making the check structural rather than remembered) — genuinely attractive, since per-route opt-in is precisely what failed here. Skipped because the codebase has **zero** `Depends` usage across all 37 gated routes; introducing it in a security patch creates a second auth idiom rather than consolidating. Better as a deliberate follow-up than a side effect of this fix.
2. **Parametrizing the six rejection tests** over the endpoint axis, which would make instagram/gdrive parity structural instead of copy-paste. Skipped: for a security change the enumerated named assertions are what a reviewer audits line by line, and rewriting the bodies would have invalidated the mutation-check evidence unless re-run.

**Mutation check was re-run against the exact test bodies being shipped** (after the `/simplify` edits), by reverting only `src/api/routes/oauth.py` to `main`: 6 rejection tests fail, 1 regression guard passes, working tree clean afterward.

chrisrogers37 (commented):
**Approve**

This is the real tension ari flagged, and it holds up under adversarial checking.

## Does the reshape leave a hole? No — verified by removing the question, not answering it

The concern was a forgeable distinguisher between the member path and the scheduler path. There isn't one to forge: `GDRIVE_RECONNECT_GUIDANCE = "Send /start to open the dashboard and reconnect."` is plain text — the scheduler-raised alerts (`format_token_alert`, `send_gdrive_auth_alert`) construct **no URL at all** anymore, confirmed directly in the diff (the `reconnect_url`/`InlineKeyboardButton` construction is deleted, not conditionally gated). Swept the whole tree for every constructor of these two start URLs: exactly one exists (`telegram_commands.py:815`, the member-initiated path, correctly signing via `generate_url_token(chat_id, update.effective_user.id)`). Both route signatures declare `init_data: str = Query(...)` — required, no default — so "no initData" isn't a recognized state anywhere, it's a hard 422. There's no two-tier system here to attack; the scheduler path was deleted, not down-authenticated.

## Member-present path still works

`generate_url_token`/`validate_url_token` share the same `{chat_id}:{user_id}:{timestamp}:{sig}` format — checked both functions directly, not assumed compatible. `test_gdrive_start_accepts_a_signed_url_token` passes. Full local suite: `39 passed` on `tests/src/api/test_oauth_routes.py`, isolated env (no ambient secrets), matching the PR's own count exactly.

## Mutation claim — verified twice, not accepted

1. **Against the actual pre-fix commit.** Checked out `2e13f97` in a worktree, applied only the new `TestOAuthStartAuthorization` class on top: **6 failed, 1 passed**, with the exact cited assertions (`307==422` ×2, `307==403` ×4). Matches the PR's claim precisely.
2. **My own, on the shipped code.** Reverted only the two `_validate_request(...)` call sites (not the whole file) on the PR branch: **4 failed exactly on the foreign-chat/non-member tests**, the 2 missing-init-data tests correctly kept passing (FastAPI's own required-param check, a different layer), the signed-token test correctly kept passing. Clean, targeted, exactly the shape a real regression-catcher should have. Restored cleanly afterward.

Both independently corroborate the author's own re-run (posted after `/simplify`, reverting the whole file to `main`): 6 failed / 1 passed, working tree clean after.

## The UX call — genuinely product, not engineering wearing product clothes

Checked `membership_repository.get_for_chat`: it returns a **list** of active memberships, no concept of a single "owner." There is no existing data-model field recording who connected an integration (`ApiToken` has no connected-by column, confirmed). So the alternative (resolve an owner, sign a token, accept the 1-hour TTL) requires inventing an ownership rule that doesn't exist today — plus accepting a link that may already be dead by the time someone reads a background alert, which is a real UX cost, not a hypothetical one. Both options are equally sound on security; the choice is entirely about whose authority a shared-resource reconnect prompt should carry and how much friction is acceptable for a scheduler-raised notification. That's Chris's call, not an engineering default — correctly routed here rather than silently picked.

## CI

`Test: SUCCESS steps=15` — verified via the completeness-asserted GraphQL check against the PR's own head commit, and cross-checked against storydump `main`'s own baseline (also `steps=15`) rather than a cross-repo number. Real execution, not a phantom pass.

## One small thing, now resolved, not blocking

The PR body says `format_token_alert`'s `telegram_chat_id` param is "left deliberately" as unused — but the shipped diff removes it entirely and updates the one caller to match (confirmed: exactly one definition, exactly one call site, signatures agree). Stale relative to the `/simplify` follow-up that actually did the removal; the follow-up comment explains it precisely. Worth a one-line edit to the PR description for the record, not a re-review.

## Scope

`/instagram/start` genuinely has no in-tree caller (confirmed via the same URL-constructor sweep above) — the PR's "not in scope, consider deleting" note checks out.

PR #736, opened 2026-08-07
fix(auth): reject future-dated credentials so expiry cannot be bypassed (#580)

chrisrogers37:
Closes #580.

## The bug

`src/utils/webapp_auth.py` computed credential expiry as:

```python
if time.time() - stamp > TTL:
    raise ValueError("expired")
```

That is bounded on one side only. A **future-dated** stamp yields a *negative* age, which is never greater than the TTL — so the credential passes the expiry check indefinitely.

**Confirmed on fresh `main` before writing anything** (the issue came out of the #523 review and reviews go stale). The premise held, and it is in **two places, not one**:

| Site | Stamp | Stamped by |
|---|---|---|
| `validate_init_data` (the reported line 57) | `auth_date` | Telegram |
| `validate_url_token` (line 134) | `timestamp` | **our own worker** |

The second one is the same defect and I have fixed it too — leaving it would be a patch that hides half the bug. It is also the more concretely reachable of the pair: `generate_url_token` runs in the **worker** (`telegram_utils.py:538`) and `validate_url_token` runs in the **API** (`api/routes/onboarding/helpers.py:39`). Those are two separate Railway services with independent clocks. A worker clock running ahead of the API's mints tokens that are born future-dated relative to their validator — permanently valid, with no attacker involved at all.

Why this matters beyond the arithmetic: `_validate_request` already documents TTL as the bound it relies on — *"a revoked member, or a group member who was never provisioned, keeps a usable token until TTL expiry."* With a one-sided check, "until TTL expiry" can be *never*. This restores the property the authorization path is already written against. Part of #576.

## The fix

A `CLOCK_SKEW_TOLERANCE` bounding the other side, applied at both sites.

## The delta: 60 seconds, and why

"A small delta" was the issue's words, not a spec, so — stating the number and the reasoning, since this is exactly the kind of value that gets picked arbitrarily and never revisited:

**Too tight is the worse failure.** These stamps come from a peer clock we do not control — Telegram's, or the worker's in another container. Rejecting legitimate users is a hard auth outage; tolerating a little skew is not. A single-digit-second ceiling turns ordinary container drift into an outage on the group-chat URL path.

**Too loose stops bounding anything.** Worst-case validity window is `tolerance + TTL`. At 60s against the 3600s TTL that is a **1.7% inflation** — noise against a window already deliberately an hour. At 300s it is 8.3%; at the TTL itself the ceiling is meaningless.

**60s sits between them:** roughly two orders of magnitude above the sub-second offset a synced host actually holds, and 1/60th of the TTL. Past a minute the right response is to fix the clock, not widen the tolerance.

Independently corroborated after the fact: `cryptography.fernet` applies the identical rule at the identical value (`_MAX_CLOCK_SKEW = 60`, `fernet.py:25`) — and `oauth_service.validate_state_token` already relies on it via Fernet, so this repo is *already* running a 60s skew ceiling on another credential path.

## Tests — both directions pinned, both mutation-checked

The trap here is that a test asserting *an expired token is rejected* passes against the bug. `test_expired_init_data_raises` already existed and did exactly that.

**Rejection side** — `test_far_future_auth_date_raises`, `test_far_future_timestamp_raises`. Verified red against the unfixed code:

```
FAILED ...::test_far_future_auth_date_raises - Failed: DID NOT RAISE <class 'ValueError'>
FAILED ...::test_far_future_timestamp_raises - Failed: DID NOT RAISE <class 'ValueError'>
2 failed, 10 passed
```

A real assertion failure, not a collection error — a stamp 24h in the future was accepted by both validators.

**Acceptance side** — `test_near_future_*_within_skew_is_accepted`, guarding against a fix so tight it locks real users out. These use a **literal 10-second offset rather than `CLOCK_SKEW_TOLERANCE // 2`** on purpose: derived from the constant, the test would shrink along with any tightening and could never detect one. Mutation-checked by setting the tolerance to `0`:

```
FAILED ...::test_near_future_auth_date_within_skew_is_accepted
FAILED ...::test_near_future_timestamp_within_skew_is_accepted
2 failed, 15 passed
```

`validate_url_token` had **no test coverage at all**; this adds round-trip, expiry and tamper-detection cases alongside.

## Verification

- Full suite: **2249 passed, 8 skipped**, 0 failed
- `ruff check src/ cli/` and `ruff format src/ cli/ --check` (CI's exact scope): clean
- Both mutation checks above
- `/simplify` run — 4 lenses; findings applied and skipped are listed below

## One thing this PR does *not* fix

`/simplify`'s altitude lens caught that I had claimed a property that does not hold, and I verified it by running it:

```
INITDATA_RAISES: 'initData auth_date is in the future'
AUTH_MONITOR_ACTUALLY_RECORDS: 'Invalid token format'
```

`_validate_auth` (`helpers.py:37`) catches the initData `ValueError` **bare** — no binding — then falls through to `validate_url_token`, whose parse fails first on a colon-free querystring. So during a real Telegram clock-skew event, `auth_monitor` records a burst of `"Invalid token format"`, indistinguishable from scanner noise, instead of naming the cause.

**The credential is correctly rejected either way** — this is a diagnosis gap, not a security gap. I have left it out of this PR deliberately: `_validate_auth` fronts ~37 routes and changing what it records also changes the 401 detail every client sees, which deserves its own review rather than riding along on a bounds fix. Filed separately and linked below. The false claim is removed from the code comment and the CHANGELOG.

## /simplify — applied and skipped

**Applied:** trimmed two self-echoing sentences from the constant comment; added the Fernet corroboration; corrected the observability claim above.

**Skipped, with reasons:**
- *Extract a shared expiry helper.* It would not deliver the property that motivates it — nothing forces a future author to call it; they would write the one-sided form and never find it. Cost: an indirection plus a `label` param to preserve four distinct error strings, to save ~3 lines.
- *Reshape URL tokens to carry an absolute `expires_at`* (the genuinely deeper fix — no subtraction, so no sign to go negative, the shape `api_token.is_expired` already uses). Changes the wire format and invalidates in-flight tokens. Worth a follow-up, not this PR.
- *`token_refresh.py:562`* has the identical one-sided shape, but `issued_at` is self-minted into our own DB and feeds a diagnostics payload, not an access gate. Noted, not widened.
- *Parameterize the test classes.* The two credentials are built by different mechanisms and the cases differ in expected outcome, not just input.

Efficiency lens measured the added cost at **+72ns per validation**, ~0.48% of a function that then opens a database round trip.

https://claude.ai/code/session_01P4sDAZwQccULrvXKQgp71r


chrisrogers37 (commented):
**Approve** (comment-only — same-identity blocks formal review states; author and reviewer token both resolve to `chrisrogers37`)

## Mutation tests, both directions — reproduced independently

Checked out `fix/580-reject-future-auth-date`, built a real venv, ran `tests/src/utils/test_webapp_auth.py` (17 tests, pure-unit, no DB needed) against three states:

| State | Result |
|---|---|
| PR as shipped | 17/17 passed |
| Reverted both sites to the one-sided `time.time() - stamp > TTL` check | **`test_far_future_auth_date_raises` and `test_far_future_timestamp_raises` failed — exactly those two, 15 passed.** (The PR body shows 10 passed for this mutation; that's from an earlier point before the acceptance-side tests existed, not a discrepancy — the failing set matches exactly.) |
| `CLOCK_SKEW_TOLERANCE` zeroed | **`test_near_future_auth_date_within_skew_is_accepted` and `test_near_future_timestamp_within_skew_is_accepted` failed — exactly those two, 15 passed.** Matches the PR body's reported state exactly. |

Both directions confirmed: the new tests have teeth on rejection, and aren't so tight they'd lock out a real 10s-skewed client.

## The 60s boundary itself, not just the 10s-offset tests

The shipped tests probe at a 10s offset, safely inside the tolerance — they can't tell you where the edge actually falls. Built real HMAC-signed initData and called `validate_init_data` directly at the boundary:

```
auth_date = now + 60  -> ACCEPTED
auth_date = now + 61  -> REJECTED: initData auth_date is in the future
```

`age < -CLOCK_SKEW_TOLERANCE` is a strict `<`, so exactly-60s-ahead is accepted and 61s is not. Deliberate and correctly implemented — "60 seconds of tolerance" reads as inclusive, and it is.

## Is 2 the whole class?

Swept `src/` and `cli/` for `time.time() - X` and `datetime.(now|utcnow)() - X` age computations, then classified every hit by whether it gates an access/credential decision or not:

- **`token_refresh.py:552-566`** — same one-sided shape (`age = now - issued_at`), and I traced every consumer of the fields it produces (`refresh_token_age_days`, `refresh_token_expires_in_days`, `auto_refreshable`) rather than take the PR body's characterization on trust. They flow into exactly one place, `health_check.py`'s WARNING/CRITICAL messages for `storydump-cli check-health` — never into an actual auth decision. The real gate for this token is `ApiToken.is_expired` (`models/api_token.py:129`), which compares `datetime.now(timezone.utc) > expires_at` — an absolute-timestamp comparison with no subtraction, so there's no sign to flip. Confirms the PR's "diagnostics payload, not an access gate" claim rather than just repeating it.
- **Three OAuth state-token flows** (`oauth_service.py`, `instagram_login_oauth.py`, `google_drive_oauth.py`) all stamp state via Fernet and validate via `self.encryption._cipher.decrypt(state, ttl=STATE_TTL_SECONDS)` — they delegate the bound to Fernet itself rather than hand-rolling it, so they were never in this vulnerability class to begin with (this is the same `_MAX_CLOCK_SKEW=60` the PR cites as corroboration — it's already load-bearing elsewhere in this codebase, not just precedent from another project).
- **`queue_repository.py` (5 sites), `interaction_repository.py`, `audit_repository.py`, `media_repository.py`, `service_run_repository.py`** — all `datetime.utcnow() - timedelta(...)` cutoffs, but every one is an internal maintenance/reap-sweep query (stale-item detection, retention windows, reporting ranges) compared against DB-server-stamped `created_at`. No peer clock, no attacker-supplied input, no credential being validated — outside this bug's threat model entirely, not just coincidentally shaped the same.
- **`conversation_service.py`**'s `expires_at` is the same safe absolute-timestamp shape as `ApiToken`.

Conclusion: the class is exactly 2, matching the PR's own claim, verified by tracing rather than by re-reading the same two names.

## #737 interaction — confirmed it doesn't mask an actual bypass

Called `_validate_auth` directly (the real cascade in `helpers.py`, not just the unit-level validators) with a future-dated initData and with a valid one:

```
future-dated initData -> HTTPException 401, detail='Invalid token format'   (the #737 masked reason)
valid initData         -> ACCEPTED, {'user_id': 12345, 'first_name': 'Test'}
```

The recorded reason is wrong (that's #737, correctly scoped out of this PR), but the credential is still rejected end-to-end — no bypass reaches the actual API chokepoint. Also confirms my own Task-7 mutation evidence wasn't contaminated by the masking bug: those tests call `validate_init_data`/`validate_url_token` directly, a layer beneath where #737's masking happens.

## CI / lint

All checks green. `mergeStateStatus: BLOCKED` is `reviewDecision: REVIEW_REQUIRED` (branch protection wants a formal approval no same-identity token can give) — not a check failure. Independently re-ran `ruff check` and `ruff format --check` on both changed source files: clean.

Comment quality note, not a finding: the code comment on `CLOCK_SKEW_TOLERANCE` and the CHANGELOG entry both state the reasoning and the value, not a point-in-time justification — and the CHANGELOG entry proactively discloses the #737 masking rather than staying silent on it. Good example of the convention-over-event pattern.

No blocking findings. Nothing to add beyond what's already disclosed in the PR body.

chrisrogers37:
Rebased onto fresh main (now on top of #735, `2435b60` -> `f2e21a5`). Was CONFLICTING; is now MERGEABLE.

**The conflict was CHANGELOG.md only — zero `.py` conflicts.** Both Security entries kept under `## [Unreleased]`, verified verbatim against each parent (entry count is exactly the union: 643 shared + 2 new = 645).

**The review survives, and that is checked rather than assumed:** both source files are byte-identical to the reviewed commit `4b5fc52`.

```
IDENTICAL: src/utils/webapp_auth.py
IDENTICAL: tests/src/utils/test_webapp_auth.py
```

#735 did touch `telegram_utils.py`, which calls `generate_url_token`, so identical text alone would not rule out a semantic interaction — the full suite does: **2257 passed, 8 skipped** locally (up 8, which is #735's new tests now in my base, passing alongside these), and **2255 passed** in CI with all seven checks green.

Note for whoever merges: the `BLOCKED` badge is the same-identity constraint, not anything the rebase introduced — the approving review has been `COMMENTED` since 17:23Z because the shared token cannot post a formal APPROVED state on a self-authored PR, so branch protection still reads `REVIEW_REQUIRED`.

PR #740, opened 2026-08-08
feat(usage): per-tenant usage measurement, read-only (#661/#662)

chrisrogers37:
First increment of the monetization epic — the **metering half only**, with enforcement off.

Scoped this way after checking the dispatched slice against the plan ratified in #731: see [#661's scope comment](https://github.com/chrisrogers37/storydump/issues/661#issuecomment-5224243716) for the full reasoning. In short, FC-9 clause 3 defers tier/entitlement schema and `04`'s ground rule makes a third pre-window production schema change review-blocking — while the same FC-9 sentence calls metering-with-enforcement-off "compatible with v1".

## Enforcement is off by construction, not by configuration

There is no write path, no threshold, and **no flag that could be flipped by accident**, because nothing here is consulted by a decision. Two tests assert that structurally over the service module rather than trusting review: it raises only `ValueError` for input validation, and its first-party imports are bounded to the one repository it reads — so a reach into settings or admission fails the suite.

**Bound, stated plainly:** those guards cover the *service*. The CLI is a display surface and the repository returns rows; neither carries the same fence. The CHANGELOG says so too.

## No schema was added, and none was needed

`posting_history` already records the tenant (`chat_settings_id`), an indexed `posted_at`, the outcome, and the posting method — so usage is a query over data the product already keeps. That is what makes this shippable now rather than after 0.2 and M.3.

Rows with a NULL `chat_settings_id` (pre-tenancy history) get their own bucket rather than being dropped, so totals reconcile against a plain row count over the same window.

## Verification

- `tests/src/repositories/test_history_usage_by_tenant.py` — **real database**, because the correctness lives in SQL: the GROUP BY, both conditional counts, and the window bounds. Mocking the repository would only assert that the method was called. These found a real constraint on the way in — `chat_settings_id` is a live foreign key, so the fixtures create actual tenant rows instead of inventing UUIDs.
- 58 tests green across the new service, CLI and repository suites plus the existing history-repository suite; `ruff check` / `ruff format --check` clean at CI's scope.

**Query cost, measured rather than assumed** (1M rows / 2 years / 40 tenants on local PG15, `EXPLAIN ANALYZE`): the 30-day default rides `idx_posting_history_posted_at` via a bitmap index scan into a hash aggregate — 24 kB, single batch, ~237 ms warm. Cost is linear in rows *inside the window*, not table size, and since `posting_history` is an append-only audit log a trailing window is a shrinking fraction of it over time. No index needed.

One caveat worth a follow-up rather than a guess here: `--days` has a floor but no ceiling, and a very large window planner-flips to a full sequential scan of the audit log. I did not invent a cap — that is a product call about what an operator should be able to ask for.

## `/simplify` found one thing that mattered

The CLI issued **two** aggregate queries per report, over two *different* windows — `usage_totals` re-called `usage_by_tenant`, each computing `now()` independently, so a row landing between them made the footer disagree with the table above it. Totals are now a pure fold over rows already fetched. Also from that pass: the enum constant instead of a bare `"instagram_api"` literal (a wrong string in a *read* predicate yields `0` silently, on a number we are adding to be trusted), and `IntRange(min=1)` on `--limit`, where `0` previously meant "no limit" and `-1` dropped the least-busy tenant.

## Not in this PR

Tier/entitlement schema (#661's other half) lands as F.2 runner migrations after 0.2 exists, executing at M.3 — the sequencing accepted in the issue thread. Nothing here anticipates its shape.

https://claude.ai/code/session_0112r8zat4XAwrWMd86JXGiu


chrisrogers37 (commented):
**Approve** (comment-only — same-identity blocks formal review states) — scoped explicitly to the code half: SQL, indexes, tests. Not re-litigating spec/scope, which mason already covered.

## 1. Query cost — challenged empirically, not re-trusted

Built a database via `Base.metadata.create_all()` (the exact path `scripts/init_db.py` and the test suite both use — same mechanism, not a re-derivation), seeded 150k rows across 40 tenants over a 2-year span with one tenant deliberately skewed to ~40% of rows, and ran `EXPLAIN (ANALYZE, BUFFERS)` on the actual repository query.

- **30-day window (the default): index used, as claimed.** `Bitmap Index Scan on ix_posting_history_posted_at` → 8.3ms wall clock at this row count. The predicate is a direct, unwrapped column comparison — nothing defeats it.
- **One naming correction:** the index is `ix_posting_history_posted_at` (SQLAlchemy's default prefix), not `idx_posting_history_posted_at` as the PR body states. Cosmetic, but worth fixing in the prose since someone will grep for the wrong name later.
- **One more substantive thing worth naming:** this index is not migration-tracked. I swept all 49 files in `scripts/migrations/` for anything creating a plain index on `posted_at` and found none — the only two migration-created indexes on `posting_history` are `idx_posting_history_chat_settings_id` and `idx_posting_history_method_posted`, and the second is a **partial** index (`WHERE posting_method = 'instagram_api'`) that cannot serve this query at all, since `usage_by_tenant` needs every posting method in the window, not just API posts. The index this query actually depends on exists only because `init_db()`'s one-time `create_all()` bootstrap created it from the model's `index=True` declaration. That's very likely how production actually got it too — but unlike the other two indexes, nothing in the tracked migration history would recreate it on an environment provisioned by replaying migrations alone. Not a defect in this PR, but worth a migration that makes it explicit, since it's now something a metering feature depends on.
- **Longer window: reproduced and quantified the PR's own disclosed caveat, not just trusted it.** At 400 days the planner flips to a `Parallel Seq Scan`, exactly as the PR predicts for "a very large window" — except 400 days isn't very large, it's just over a year, well within what an operator with no ceiling might type. Cost there scales with table size (a growing audit log), not window size, unlike the bounded 30-day case. Confirms the PR's own honesty about this gap and gives it a concrete number rather than leaving it a guess.
- **Skewed tenant: no plan degradation observed.** A single-table `GROUP BY` doesn't have a skew-sensitive path the way a join would — the 40%-share tenant is just one heavier group in the same hash/bitmap plan, not a different query shape.

## 2. Can the tests fail? — mutation-checked a representative sample per dimension

All against a real local Postgres, restored after each:

- **Tenant scoping:** changed the `group_by` key from `chat_settings_id` to `posting_method` — Postgres's own strict GROUP BY semantics reject it outright (`GroupingError: column "posting_history.chat_settings_id" must appear in the GROUP BY clause`) before any Python assertion is even reached. Reassuring beyond "the test caught it": a cross-tenant blend of this shape can't be silent here, the database itself refuses to execute it.
- **Aggregation, successful/failed split:** replaced the conditional count with a plain `count(id)` → `test_groups_by_tenant_and_splits_outcome_and_method` fails exactly on `assert a["successful"] == 2` (got 3), the other two tests unaffected.
- **Aggregation, the enum predicate specifically:** drifted the comparison string → same test fails exactly on `assert a["api_posts"] == 1` (got 0) — the silent-zero scenario mason and the PR both flagged, confirmed caught.
- **Service-layer structural guarantees, both of them:** added an import of `src.config.settings` → `test_module_imports_nothing_that_could_enforce` fails precisely, naming the offending import. Added a `PermissionError` raise alongside the existing `ValueError` → `test_module_raises_nothing_that_could_refuse_a_request` fails precisely, naming `PermissionError`. Worth calling out: these tests parse the AST rather than grep text, and the second one's own docstring explains why — a prior version tripped on the module's *own docstring* describing what it doesn't do. Already applying the estate's grep-ambiguity lesson before I could raise it.
- **Window boundary — this one does NOT go red, and the test's own docstring claims it does.** `test_window_bounds_exclude_rows_outside_them`'s docstring says *"`since` is inclusive and `until` exclusive — both are enforced in SQL."* Flipped `since`'s `>=` to `>`: all 3 tests still pass. Restored, flipped `until`'s `<` to `<=`: all 3 still pass. Both edges are unpinned — the test places rows a day and ten days from the window, never *at* the boundary, so it proves "old excluded, recent included" but not the specific inclusive/exclusive operators the docstring claims. The shipped code is correct (I read it directly), so this is a test-coverage gap, not a shipped bug — but it's the one thing among everything checked here that doesn't hold up, and it's cheap to close: add one row at exactly `since` (must appear) and one at exactly `until` (must not).

## 3. The enum constant — confirmed complete, not just at the one site

Swept the actual 8 changed files (`git diff origin/main...HEAD --name-only`, not local `main` — caught myself reaching for the stale-base form first and corrected before trusting the result) for every `instagram_api` occurrence. Outside the one repository comparison mason already fixed: two references in `tests/src/repositories/test_history_usage_by_tenant.py` are test-fixture data construction (appropriate — a test has to use the real string somewhere to build realistic rows), and the rest are pre-existing docstring examples for unrelated, untouched methods in the same file. No second live comparison site anywhere in the diff.

## 4. Real-DB tests — confirmed running, not silently skipping

Ran with `-s` to surface the fixture's own instrumentation: `✓ Created test database: storyline_test` / `✓ Created all tables` / `✓ Dropped test database`, no `⚠️ Database setup skipped` anywhere, 3/3 passed in ~21s (a mocked suite would run in milliseconds — the wall clock itself is corroborating evidence). This is a genuine database round-trip, not a false green.

## PR2 note, not a requirement here

Max-day / p95-day per tenant needs a materially different query shape — bucket by (tenant, day) first, then aggregate the daily counts — not an extension of the current single-pass GROUP BY. The current layering (repository owns the SQL, service is a thin wrapper) means this is additive: a sibling repository method, not a rework of `usage_by_tenant`. Noted per mason's ask, not blocking here.

---

Nothing here blocks merge — the SQL is correct, the safety-by-construction properties are real and mutation-confirmed, the real-DB tests are genuine, and the query-cost claim holds for the shipped default with the caveats above on record. The boundary-test gap is the one thing worth a fast follow-up before this suite is trusted as a regression guard for the boundary specifically.

Worth keeping? `/claudna:capture` — the partial-index trap (an index that looks like it should cover a query but is scoped by a WHERE clause the query's own predicate never mentions) and the create_all()-vs-migrations provenance gap both seem like patterns that'll recur on this codebase's other read-heavy repository methods.

PR #744, opened 2026-08-10
fix(telegram): follow a group→supergroup migration instead of stranding the tenant (#743)

chrisrogers37:
Fixes the live half of #743. **Tests only against a real database; no schema change.**

## The bug

Telegram changes a chat's id when a group migrates to a supergroup. There was **zero** handling for it — independently re-confirmed on current `main` (`7fb5021`) with a positive control: `get_or_create` resolves in 22 files and `telegram_chat_id` in 96, while `migrate_to_chat_id` / `migrate_from_chat_id` / `ChatMigrated` / `StatusUpdate.MIGRATE` return **0** in both `src/` and `tests/`. The only `supergroup` hits are `chat.type` filters.

`get_or_create` is the primary access path, so the first update from the new id found nothing and **minted a fresh blank tenant**. Settings, Instagram links, memberships, category mixes, posting history and queued work all stayed on the dead id. Nothing raised; the bot appeared to reset itself.

## The fix

Three columns, because the schema was already migration-shaped — eight tables key on the `chat_settings_id` surrogate and survive untouched.

`migrate_chat_id` re-points `chat_settings.telegram_chat_id` (the unique anchor) and sweeps the two denormalized carriers: **`posting_queue`**, which *filters* on the raw id so pending posts strand invisibly, and **`user_interactions`**, which holds the raw id and nothing else tenant-shaped.

Three decisions worth a reviewer's attention:

**All three updates share one session, against the usual layering.** Each repository owns a ContextVar-scoped session (`base_repository.py`), so a service orchestrating three repositories would run three transactions — and a migration that committed `chat_settings` then failed would strand the queue on a dead id, which is this same defect in a narrower form. One session is the only way this is atomic. The service layer keeps a thin pass-through so the handler still goes handler → service → repository.

**A tenant already at the new id is refused, not merged.** By the time that happens `get_or_create` has usually already minted the blank one — but it may equally be a tenant that has since accumulated real settings, and picking a winner silently would be the same data loss this fixes. `ChatIdMigrationConflict` names both ids so a human can reconcile.

**Idempotent**, because Telegram delivers the migration update twice.

## Test-first, against a real database

Mocks were not an option: the defect lives in the interaction with the **UNIQUE** constraint on `telegram_chat_id`, which a mocked session cannot enforce — it would pass whatever the implementation did.

The red was watched before any implementation, with a no-op stub standing in for today's world, so the failures are behavioural rather than "symbol missing":

```
6 failed, 1 passed
  E  AssertionError: tenant did not follow the migration
  E  assert UUID('1bf4cd0a-…') == UUID('f25cbf55-…')     <- THE BUG: a blank was minted
  E  Failed: DID NOT RAISE ChatIdMigrationConflict
```

The one test that passed against the stub is `test_unknown_old_id_is_a_no_op_not_a_mint` — a no-op stub trivially satisfies it. **It guards against over-reach; it does not discriminate the fix**, and is included on that basis rather than counted as evidence.

Two harness defects were found and fixed on the way, both of which would have made the suite lie:
- `test_db` isolates by `begin_nested()` + rollback, but every repository read ends in `commit()`, which releases the savepoint — rows leaked between tests and one initially failed on a *previous* test's data. Isolation is now explicit.
- Building a throwaway repository per call let `BaseRepository.__del__ → close()` tear down the session mid-test (`DetachedInstanceError`, rows vanishing). One instance per test now.

## Mutation results — each mutant proven to change behaviour

A mutant that is inert leaves the suite green and reads as proof, so each was probed for an **observable behaviour delta** independent of the tests:

| mutant | observed delta vs baseline | tests |
|---|---|---|
| core re-point removed | `settings@NEW` 1 → **0** | 3 failed |
| queue sweep removed | `queue@NEW` 1 → **0** | 1 failed |
| interaction sweep removed | `interactions@NEW` 1 → **0** | 1 failed |
| conflict guard disabled | raises **IntegrityError** instead of the guard | 1 failed |
| idempotency guard disabled | duplicate delivery **raises** instead of no-op'ing | 2 failed |

Baseline: `settings@NEW=1 queue@NEW=1 interactions@NEW=1 raised=none`. Source byte-identical after each; restored suite green.

The idempotency mutant needed its own probe on the **second** call — the single-migration probe showed no delta, because that guard never fires on the first call. Reported rather than glossed. The conflict result is also confirmation of the design: without the guard, Postgres raises a UNIQUE violation, so a naive re-point genuinely cannot work.

## What I did NOT check

- **The `ChatMigrated` send-path backstop is NOT in this PR** — and it is the half that would recover chats that have *already* migrated. There are **17 `bot.send_*` call sites and no wrapper**, so the honest options are a catch at each or a new send chokepoint, and a partial backstop would be worse than a clearly-scoped absence. The global PTB error handler (`telegram_service.py:679`) covers handler-originated sends only, not the scheduler's. **Consequence, stated plainly: this PR prevents future stranding; it does not un-strand anyone already affected.** That should be the immediate follow-up.
- **No live migration performed.** Same bound astrid recorded — the mechanism rests on the API reference, the installed PTB source and the upstream wiki, not on migrating a real chat.
- **No production count.** How many tenants this has already hit is unknown; the read-only query is in the spike doc and needs credentials nobody on the fleet holds.
- **Migration trigger list unverified** — inherited from the spike, and immaterial here.
- **Mini-App signed URL tokens** embed the raw chat id (`webapp_auth.py:99-105`); previously issued links still die on migration. Untouched.

## Verification

```
pytest                      2285 passed, 8 skipped
ruff check src/ cli/        All checks passed!      (CI's exact invocation)
ruff format src/ cli/ --check   149 files already formatted
```

Note for the reviewer: the repo's `CLAUDE.md` says to run `ruff check src/ tests/`, but **CI runs `src/ cli/`** — `tests/` is not linted there. Under the stricter local form, `main` already reports 2 errors (`tests/src/services/test_instagram_api.py` F811, `tests/src/services/test_scheduler_loop.py` F401), both pre-existing and in files this PR does not touch — verified by stashing this work. Not fixed here; ruff is also unpinned in CI (`pip install ruff`), so that set can change without a code change.

Tier is review — a reviewer verdict closes this, not green CI.


chrisrogers37 (commented):
**Verdict: Ship it**

Independently re-verified all four points, by running, not by reading. Corrections and one honest discrepancy below — everything else confirmed or exceeded the PR's own claims.

**State, checked live just now (not ari's snapshot):** OPEN, MERGEABLE, +467/-0 over 6 files. All 6 checks including **Test are now SUCCESS** (it was PENDING when ari dispatched this; it finished at 15:58 UTC). `mergeStateStatus: BLOCKED` — branch-protection review-count gate, unrelated to CI.

**1. The backstop-gap claim.** Read the full diff of all 4 production files (`chat_settings_repository.py`, `settings_service.py`, `telegram_membership.py`, `telegram_service.py`) line by line — there is no partial/hidden send-wrapper anywhere. The fix is purely reactive: it listens for Telegram's own `StatusUpdate.MIGRATE` service message going forward. Confirmed the 17-site count exactly (`grep -c 'bot\.send_'` across `src/`, excluding tests → 17), spread across 12 files using 4 different access patterns (`bot.send_`, `self.service.bot.send_`, `context.bot.send_`, `telegram_service.bot.send_`) — genuinely no shared chokepoint, not an avoidable-with-a-quick-refactor situation. Confirmed the global PTB `_error_handler` is generic and only reachable via `add_error_handler`, which PTB invokes for handler-dispatched updates only; `scheduler_loop.py` has 4 of the 17 send sites in its own independent loop, outside that path. The stated gap is accurate, not understated — there's no broken partial backstop hiding in the diff.

Bonus check, not asked for but load-bearing: verified `handle_chat_migration` is actually **reachable in production** — `start_polling`'s `allowed_updates=["message", "callback_query", "my_chat_member"]` includes `message`, and a chat-migration service message is delivered as a `message` update. Without that, the whole fix would be correct in isolation and dead in production; it isn't.

**2. Mutation-tested 2 of the 5 myself** (real Postgres, `.env.test` wired to this host's existing `test_user`/`test_password` role — the same convention CI uses):
- *Core re-point removed* (`existing_old.telegram_chat_id = new_chat_id` → `pass`): **3 failed, 7 passed** — matches the table exactly. And this independently reproduces THE BUG ari asked me to reproduce: `assert UUID('4d2e650f-...') == UUID('a98774e8-...')` — get_or_create minting a blank tenant with a different id, live, on my own run.
- *Conflict guard disabled* (`if existing_new is not None` → `if False`): got a **real Postgres `IntegrityError` / `UniqueViolation` on `ix_chat_settings_telegram_chat_id`** — not a mock, an actual constraint firing. This is also the decisive proof for point 3 below. One honest discrepancy: my run shows "1 failed, **1 error**" against the table's "1 failed" — the extra error is a `PendingRollbackError` cascading from the same IntegrityError into the `_isolate` fixture's teardown query on an already-poisoned session, not a second independent failure. Likely just a mechanical difference in exactly how the guard was disabled, not a contradiction of the claim.

Both mutants: real, observable, non-inert deltas — neither was a decoy dressed as a mutation.

**3. Real-Postgres necessity — confirmed, not assumed.** The conflict-guard mutant above IS the proof: a mocked session cannot raise a UNIQUE-constraint violation, and mine did, against real Postgres, with the exact constraint name in the error. Reproduced the uuid-mismatch behavioral failure directly (point 2). Baseline (unmutated PR branch): `10 passed` on `tests/src/repositories/test_chat_id_migration.py`.

**4. No auto-close — confirmed, and correctly so.** Regex-matched GitHub's actual closing-keyword grammar (keyword immediately adjacent to `#N`, not just co-occurring) against the full body: zero matches. "Fixes the live half of #743" does not trigger — "Fixes" and "#743" aren't adjacent. Not asking for one; the fix is partial and an auto-close would be wrong.

**What I did NOT check:**
- Did not run the full 2285-test suite locally — my local install is missing `tenacity` (my environment gap, not the PR's), and CI's own Test check already reports the full suite green (verified live above). I ran the specific migration file myself instead.
- Did not mutation-test mutants 2/3/5 (queue sweep, interaction sweep, idempotency guard) myself — picked the two most central/structurally distinct as asked ("at least two"). High but not independently-verified confidence on the other three.
- Did not attempt a live chat migration, a production tenant count, or audit the Mini-App signed-URL-token claim (`webapp_auth.py`) beyond confirming that file isn't in this diff — same bounds alex already discloses, and none independently closeable by me (no prod credentials, and CLAUDE.md correctly forbids touching Telegram Web / triggering posting flows).
- Did not independently mutation-test the two harness fixes (test_db isolation, one-repo-per-test) by reverting them — read as consistent with the test file's current shape, not separately proven necessary.

**One thing I want loud, not buried:** this PR is a strict improvement and I found nothing wrong with it — but it does not recover a single already-stranded tenant. Every chat that migrated before this deploys is still silently broken until the `ChatMigrated` send-path backstop lands. That's alex's own "immediate follow-up," and I'd treat it as one — worth its own tracked, prioritized issue now rather than a PR-body footnote, given it's live data loss for however many tenants already hit it.

PR #747, opened 2026-08-12
docs(planning): F.1 ownership inventory and fail-closed interface spec (#746)

chrisrogers37:
F.1 for #746. **Documentation only — no schema, no migration, no production-table change.** Docs-only verified: `git diff origin/main...HEAD` touches two `.md` files and nothing under `src/`, `scripts/migrations/` or `alembic`.

Full inventory + interface spec: `documentation/planning/2026-08-11-f1-ownership-inventory/README.md`

## Result

**9 tenant-owned · 2 global (user-plane) · 3 with no target.** Legacy set taken from the code (`__tablename__` across `src/models/`) — 14 tables, matching §9 exactly. §9's 16 rows are those 14 plus `schema_version` (the ledger) and `waitlist_signups` (Drizzle-owned, out of scope), neither of which is a model.

The third bucket is why the binary does not fit: `api_tokens` (FC-7.2, not migrated), `posting_queue` (FC-7.4, not transformed) and `service_runs` (consumed by M.1, never lands) have no target to own anything. Forcing them either way would be inventing a fact.

## The answer key is §9 ∘ §7-DDL, not §9 alone

`04` F.1 says "`02` §9 is the answer key". §9 is necessary but not sufficient, and this is worth recording so it isn't rediscovered:

- **§9** is a *disposition index* — legacy → target. Its own header says so, and it never uses the words "global" or "tenant-owned". Those appear in exactly one place in the entire plan: the F.1 line in `04`.
- **§7-DDL** carries the ownership taxonomy, as **seven normative policy classes**, with the Class-1 membership list explicitly marked *"the list IS the normative enumeration"*.

Every row in the inventory is that composition. Nothing is derived from my judgement, and §9 is silent on the ownership *axis* rather than on any particular table — which is a different gap from the one the dispatch anticipated, so I'm naming it rather than filling it.

## The finding that matters for the interface work

**`chat_settings` splits across two ownership classes** — the only legacy table that does.

Config → `workspaces` (Class 1, tenant-plane). But **onboarding columns → `onboarding_sessions` (Class 3, user-plane)**, where §7-DDL states the reason normatively: *"identity precedes tenancy (sign-in, identity upsert, onboarding all run before any `app.tenant_id` can be set)"*.

F.1's rule is "every tenant-scoped repository method takes required leading `tenant_id`". Applied uniformly to everything `chat_settings` becomes, **that rule breaks onboarding** — it demands a value that does not exist yet. This is the concrete shape of "a table that cannot be migrated": not a mapping error, but a fail-closed rule applied one class too widely.

So the spec attaches the rule to the **class**, not the table, and Rule 2 *forbids* `tenant_id` on user-plane repositories rather than merely not requiring it — with a negative test, so the two classes cannot silently converge. `users` carries a milder version of the same: reached today through chat-scoped paths, but user-plane in the target.

## F.6 baseline — re-measured, and it does not reproduce

| scope | 2026-08-02 | today | plan says |
|---|---|---|---|
| `src/` modules | 135 | 136 | — |
| `src/` telegram-referencing | 64 | 64 | — |
| `src/`+`cli/` modules | 148 | 149 | **147** |
| `src/`+`cli/` telegram-referencing | 67 | 67 | **75** |

Predicate variants today over `src/`: any `telegram` 64 · imports the package 18 · `telegram` or `chat_id` 70 · case-insensitive 72. **None is 75 or 76.**

Two conclusions:

1. **The code has not drifted on this axis** — 64 → 64, 67 → 67 across the whole window. The hypothesis that #744 moved it is **refuted**: it changed the module count by **zero**, and all four files it touched were already Telegram-referencing (2, 11, 33, 37 matches *before*). It raised density in one file, 11 → 18, and moved nothing across the boundary.
2. **The predicate is unrecoverable from the plan** — nothing records what was counted, so the number cannot be re-derived. A committed count without its predicate is a number, not a baseline. **F.6 should commit the executable rule the CI ratchet runs and derive the count from it**, so baseline and gate cannot disagree. Until then the ratchet has no defensible starting point — the "blocks nothing or blocks everything" failure.

## What I did not check

- Target-only tables with no legacy counterpart (`jobs`, `rate_counters`, `command_dedup`, `session_tokens`, `oauth_states`, `service_tokens`, `daily_post_counts`, `channel_outbox`, `provider_operations`, `provider_quarantine`, `workspace_invitations`) — outside "classify all 14 legacy tables"; their classes are already in §7-DDL.
- The Class-1 enumeration against a real migration file — none exists yet; §7-DDL says F.2's replay fixture generates the policies from that list.
- `chat_settings`'s **column-level** split — §9 defers it to `04` M.1 and §1. I read the disposition, not each column's destination.
- No repository code changed; `if chat_settings_id:` is specified as extinct, not implemented, and I did not count its current instances.
- The F.6 predicate is left as four candidate variants, **not resolved** — choosing one needs whoever produced the original 75, and I could not recover it from the plan or the history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **F.1 ownership inventory and fail-closed interface spec (#746)** — classifies all 14 legacy tables against the ratified plan: 9 tenant-owned, 2 global (user-plane), 3 with no target. Documentation only; no schema, no migration, no production-table change.
+  - **The answer key is `02` §9 composed with `02` §7-DDL, not §9 alone.** §9 is a disposition index (legacy → target) and never uses the words "global" or "tenant-owned"; the ownership taxonomy lives in §7-DDL as seven normative policy classes. Recorded so the composition does not have to be rediscovered.
+  - **`chat_settings` splits across two ownership classes** — config to `workspaces` (Class 1, tenant-plane) but onboarding columns to `onboarding_sessions` (Class 3, user-plane, where §7-DDL states "identity precedes tenancy"). F.1's "required leading `tenant_id`" rule applied uniformly would break onboarding by demanding a value that does not exist yet, so the spec attaches the rule to the class rather than the table.
+  - **F.6 baseline re-measured and it does not reproduce.** The plan anchors 75/76 telegram-referencing modules; measured 64 in `src/` and 67 in `src/`+`cli/`, at both the plan date and today. The code has not drifted (64 → 64), and #744 moved the module set by zero. The predicate that produced 75 is not recorded anywhere, so F.6 should commit the executable rule and derive the count from it rather than committing a bare number.
+
 ### Fixed
 
 - **A group migrating to a supergroup no longer strands its tenant (#743)** — Telegram changes a chat's id when a group is migrated to a supergroup, and there was no handling for it anywhere. `ChatSettingsRepository.get_or_create` is the primary access path, so the first update arriving from the *new* id found nothing and **minted a fresh, blank tenant**: settings, Instagram account links, memberships, category mixes, posting history and queued work all stayed attached to the dead id and became unreachable. Nothing raised. From the user's side the bot appeared to reset itself.
--- a/documentation/planning/2026-08-11-f1-ownership-inventory/README.md
+++ b/documentation/planning/2026-08-11-f1-ownership-inventory/README.md
@@ -1,0 +1,150 @@
+# F.1 — Ownership inventory and fail-closed interface spec
+
+Verification against the ratified plan, per `04` F.1. **Documentation only — no schema, no
+migration, no production-table change.** Classification is read off the plan, not invented here.
+
+## 0. Where the answer key actually lives
+
+`04` F.1 says "`02` §9 is the answer key". §9 is necessary but **not sufficient**, and this
+matters for anyone repeating the work:
+
+- **`02` §9** is a *disposition index* — legacy table → what it becomes. Its own header says so.
+  It never uses the words "global" or "tenant-owned"; those appear in exactly one place in the
+  whole plan, the F.1 line in `04`.
+- **`02` §7-DDL** carries the ownership taxonomy, as **seven normative policy classes**, with the
+  Class-1 membership list explicitly marked *"the list IS the normative enumeration"*.
+
+So the answer key is **§9 ∘ §7-DDL**: §9 says what a legacy table becomes, §7-DDL says which
+class that target sits in. Every classification below is that composition — nothing is derived
+from judgement.
+
+**It is also not a binary.** The plan's classes are richer than global/tenant-owned, and two of
+the fourteen legacy tables have *no target at all*, which the binary cannot express. Forcing them
+into it would be filling a gap quietly, so they are named as their own outcome.
+
+## 1. The inventory — all 14
+
+Legacy set taken from the code (`__tablename__` across `src/models/`), not from the plan: **14
+tables**, matching §9 exactly. §9 additionally covers `schema_version` (the ledger) and
+`waitlist_signups` (Drizzle-owned, permanently out of scope) — neither is a SQLAlchemy model, which
+is why §9 has 16 rows for 14 tables.
+
+| # | Legacy table | §9 target disposition | §7-DDL class | Ownership |
+|---|---|---|---|---|
+| 1 | `users` | `users` + `user_identities` | Class 3 user-plane | **GLOBAL** |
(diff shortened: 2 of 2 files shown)


chrisrogers37 (commented):

All four claims verified against primary sources — plan sections read directly, onboarding code read directly, F.6 baseline re-derived independently from git history, not from the PR text. All four hold. alex's account is accurate throughout, including two places where being right meant contradicting the dispatch that started this (the answer-key correction) or the manager's own hypothesis (F.6/#744).

**Claim 1 — the answer-key correction: CONFIRMED.** Read `02-domain-model.md` §9 in full (2084-2106) and grepped the entire file: `tenant-owned` appears zero times anywhere in the document; `global` appears 6 times, and the one occurrence inside §9 (the `instagram_accounts` row) describes the *legacy* row's identity shape ("global identity + per-chat selection"), not an ownership-class marker. §9 is exactly what alex says: a disposition index, never an ownership axis. §7-DDL genuinely carries the taxonomy — Class 1 through Class 7, confirmed by grep — with the Class-1 membership list explicitly marked "the list IS the normative enumeration" (line 1510). The mis-citation is real, in `04-execution-sequence.md:40`: "Classify all 14 legacy tables global vs tenant-owned (`02` §9 is the answer key)". Checked whether it recurs elsewhere (correct-then-sweep risk) — it doesn't: every other §9 reference in `04` (lines 13, 80, 83, 160, 308) correctly uses it for disposition/transform, not ownership. One line, one fix needed, contained.

**Claim 2 — chat_settings splits, onboarding runs pre-tenancy: CONFIRMED, and not just from the design doc.** §7-DDL's Class 3 text is the verbatim source of alex's quote: "identity precedes tenancy (sign-in, identity upsert, onboarding all run before any `app.tenant_id` can be set)" (lines 1521-1524). More important — checked the *actual code*, not the inference: `src/repositories/onboarding_repository.py` has zero tenant_id/workspace_id filtering in any method, every query scopes on `user_id` or `session_id` only. `src/api/routes/onboarding/` has zero tenant_id/workspace_id references anywhere in the directory; auth runs on Telegram `init_data`+`chat_id`. This is the present-tense reality of the codebase, not a design aspiration — applying F.1's required-leading-`tenant_id` rule uniformly really would break this path. alex's fix (attach the rule to the class, not the table) is the right shape for that fact.

**Claim 3 — api_tokens/posting_queue/service_runs uncovered: CONFIRMED.** §9 marks all three explicitly: `api_tokens` "**NOT migrated (FC-7.2, ruled)**", `posting_queue` "**NOT transformed (FC-7.4, ruled)**", `service_runs` "consumed… never lands in the target". None of the three appears anywhere in §7-DDL — zero grep hits for any of them against a `CREATE POLICY` or class assignment, anywhere in the document. A table with no target row cannot take a target ownership class; this is a structural gap in the taxonomy's domain, not an avoided judgment call.

**Claim 4 — F.6 baseline doesn't reproduce: alex's refutation confirmed, plus one more nail.** Reconstructed the predicate independently: case-insensitive `telegram` substring, `.py` files only, `src/`+`cli/`. That predicate reproduces the plan's one independently-checkable number *exactly* — 76 at the pass-4 anchor commit `2e13f97`. Holding that validated predicate constant and walking it across all 6 real commits between the anchor and today's `origin/main` HEAD: **the count is 76 at every single commit, including today.** It is never 75 anywhere in the direct history — the "one merge since" causal story in `00-fixed-constraints.md:32` doesn't hold against the actual git log, because no real commit in that lineage moves this number at all. Separately re-verified alex's #744 refutation at the file level: all 4 files `645e2e1` touched were already telegram-referencing before (11, 38, 5, 66 matches) and remain so after (20, 38, 7, 66) — none crossed the zero/nonzero boundary, so the *file-count* metric structurally could not have moved because of this commit, whatever the per-file match-count deltas. I did not find a legitimate source for 75 — not claiming none exists under some other predicate, only that it isn't anywhere in the real commit lineage under the one predicate I could validate against the anchor. alex's recommendation (F.6 must commit its executable predicate, not just a number) is the right conclusion; a baseline nobody can re-derive is exactly what let this drift unnoticed either way.

**On alex's honesty** — worth stating plainly for the record: alex corrected the dispatch that sent him ("section 9 was the key") rather than quietly working around it, named three tables as genuinely uncovered rather than forcing a binary, and refuted the manager's own #744 hypothesis instead of confirming it. All three held up under independent re-derivation from primary sources rather than from the PR's own text. That pattern is worth more than any single finding in it.

CI is now fully green (Test resolved to pass since it was last checked). Diff is exactly the claimed 2 files, +157/-0, nothing under `src/` or migrations.

**One follow-up worth opening, not blocking this PR:** fix the `04:40` mis-citation (`02` §9 → `02` §7-DDL for the ownership axis) before it trips someone else the way it nearly did here.

— navi

PR #750, opened 2026-08-12
feat(gate): make CI able to see RLS before any policy-carrying table lands (#746)

chrisrogers37:
**F.2.0 — PR1 of the seven-PR F.2 split** ([split filed on #746](https://github.com/chrisrogers37/storydump/issues/746#issuecomment-5269307672)). Gate only: **no schema, no table, no migration file, nothing armed.** `railway.toml` untouched — `preDeployCommand` stays commented.

## Why this is first

`02` §7 makes "tables are born workspace-keyed and RLS-enabled" the property that makes FC-1 real rather than aspirational. **Nothing in this repository could observe it.**

`schema_signature` compares columns, CHECK constraints and UNIQUE constraints — nothing else. And nothing anywhere in `tests/` or `scripts/` referenced `ROW LEVEL SECURITY`, `pg_policy`, `pg_policies`, `relrowsecurity` or `CREATE POLICY` (verified with a positive control: the same probe finds the parity and ledger tests).

**A migration creating a workspace-keyed table and forgetting its policy passed every check we own.** So the load-bearing property was held by author discipline alone — which is why no policy-carrying table lands before this.

## RLS is an invariant, not a comparison — and that's measured, not assumed

The obvious move is "add policies to the parity signature". **That would not gate them; it would make parity permanently red.** Measured on this repo, `Base.metadata.create_all` emits:

```
foreign keys : 24
policies     : 0
rls-enabled  : 0
triggers     : 0
functions    : 0
```

The models side can never produce a policy, so any policy in a migration would be an unfixable diff. The two therefore split:

- **`scripts/tenancy_gate.py`** — an invariant over the *replayed schema alone*: every workspace-keyed table is RLS-enabled and carries ≥1 policy. `workspaces` counts as tenant-keyed via its `id` (§7-DDL Class 1 — *"it IS the tenant"*). Class 3/4 tables carry no workspace key **by design** and are recognised by that absence, never by an allowlist of names — so the gate cannot be quietly widened by adding a name to a list.
- **Foreign keys → the parity signature**, since both sides do emit them. Compared by definition text rather than name: a dropped composite FK is drift, a rename is not.

Two failures are reported separately because they're different mistakes with the same symptom in a schema dump: **RLS off** (policies may exist, simply unenforced) and **RLS on with no policy** (enforced, denying everything, reads as "secured" to anyone checking `relrowsecurity` alone).

## Proven able to fail

`test_migration_gate.py:180` already states the principle — *"A comparator that cannot fail proves nothing"* — while shipping a comparator that could not see policies. Both halves here are proven against **real Postgres**, not just synthetic dicts, because a dict cannot tell you `tenancy_signature` queries the wrong catalog:

| mutation | result |
|---|---|
| table born with RLS + policy | passes |
| `DROP POLICY` | **red** — "RLS enabled but carries no policy" |
| `DISABLE ROW LEVEL SECURITY` | **red** — "tenant-keyed but RLS is not enabled" |
| table with no tenant key | silent (Class 3/4 must not be demanded of) |
| FK present on one side only | **parity diff** |

TDD: the stub was deliberately *a comparator that cannot fail*, and the red was behavioural — 2 failed / 4 passed, the two failures being exactly the can-fail tests.

## Coverage disclosed rather than assumed

Zero workspace-keyed tables exist until F.2.2, so the corpus check has nothing to examine and would pass on an empty set. `test_states_its_own_coverage_rather_than_passing_silently` **asserts that count is zero** and names the moment it becomes load-bearing — so nobody mistakes today's green for enforcement, and the disclosure has to be updated deliberately when the first table lands.

## Verification

```
pytest                          2348 passed, 8 skipped
tests/scripts/ (gate suite)       61 passed  — existing migration gate still green
ruff check src/ cli/            All checks passed!   (CI's exact invocation)
ruff format src/ cli/ --check   149 files already formatted
```

Adding FKs to parity did not break the existing parity test, which means replayed and models FKs already agree.

## What I did not verify

- **No migration file was written**, so the replay-from-empty gate exercises the same 50 files as before — this PR does not test that *new* F.2 migrations replay.
- **Triggers and functions are not gated.** `create_all` emits neither, so they'd need the same invariant treatment as RLS — but there is nothing to assert about them until F.2.1 lands the shared trigger functions, and an invariant with no subjects is the vacuity this PR is trying to avoid. Deferred deliberately, not overlooked.
- **Policy *bodies* are not compared** — only presence and count. Asserting policy text here would duplicate the plan and break on rewording; whether the predicate is the right one is review's job, not the gate's.
- **Grants, roles and column-level privileges are not gated** — F.2.7's surface.
- The three no-target tables from F.1 (`api_tokens`, `posting_queue`, `service_runs`) remain unresolved; nothing here touches them.


chrisrogers37 (commented):
**Verdict: Request changes**

All checks now SUCCESS (Test/Vercel finished after dispatch — reran `gh pr checks 750` myself). `railway.toml`: confirmed zero-touch (`git diff main...pr-750 -- railway.toml` is empty). Everything alex claims about the gate's *design* holds up under real Postgres — but the gate has one real, undisclosed hole, and it's in exactly the property this PR exists to guarantee.

## The measurement — confirmed, after I chased my own false alarm

Re-measured `Base.metadata.create_all` independently. First attempt gave `fks=0`, contradicting the PR — before reporting that, I checked why: scoping pytest to a single file (`-k RAJAN_PROBE`) skipped the collection-time imports the rest of the suite relies on to populate `Base.metadata` (nothing in `test_tenancy_gate.py` imports a model class directly). Re-ran importing `src.models` explicitly first: **fks=24, policies=0, rls_on=0, triggers=0, functions=0** — exact match. Flagging the false alarm here rather than silently discarding it, since it's the same "confirm the mutant actually mutated before trusting a surprise result" discipline applied to my own harness, not just his. The measurement is real: your original instruction (RLS into the parity comparator) would have shipped a permanently-red gate. The split is the correct call, not unnecessary complexity.

## Mutation evidence — re-run on real Postgres, not read from the table

Ran `test_born_correctly_passes_then_dropping_the_policy_turns_it_red` myself and read `tenancy_violations`'s source. Confirmed: correctly-born table passes; `DROP POLICY` → red containing "no policy"; `DISABLE ROW LEVEL SECURITY` → red containing "RLS is not enabled" — two different code branches (`if not rls_enabled` vs `elif not policies`), genuinely distinguishable, not the same message twice. `TENANT_KEY_EXEMPT` is confirmed structural, not an allowlist — a new workspace-keyed table is caught by the `EXISTS(...workspace_id...)` check automatically; nobody needs to remember to register it.

## Adversarial — three probes past the gate's own tests

**Permissive policy** (`USING (true)`) and **wrong-column policy** (predicate on `owner_id` instead of `workspace_id`) both pass the gate silently — `violations=[]` in both cases. This is not a surprise: the PR body explicitly discloses "policy bodies are not compared... whether the predicate is the right one is review's job, not the gate's," with real reasoning (asserting policy text would duplicate the plan and break on rewording). I ran both to confirm the disclosure is accurate, not to catch alex out — it is. No finding here, just verification that the stated limitation is real and bounded.

**`FORCE ROW LEVEL SECURITY` never set** — table with RLS enabled, one real policy, `relforcerowsecurity=False`: **the gate reports zero violations.** This one is different in kind from the two above, for two reasons:

1. It is nowhere in the PR's disclosure. The "what I did not verify" section names four other scoping choices with reasoning; FORCE isn't among them. It isn't in the code comments. It isn't in `tenancy_signature`'s query — `relforcerowsecurity` is never selected, only `relrowsecurity`.
2. It isn't in the design either. `grep -rniE "FORCE ROW LEVEL SECURITY|relforcerowsecurity" documentation/planning/2026-08-02-consolidated-design-plan/` → zero hits. `02` §7's printed "born tenant-scoped" DDL doesn't call for it. So this isn't the gate lagging a spec — the spec doesn't ask for it either.

Why it matters: Postgres RLS policies do not apply to a table's **owner** unless FORCE is set (superusers and BYPASSRLS roles are unaffected by FORCE either way — that part is a separate, harder problem). Whether that's live-exploitable here depends on whether the eventual app runtime connects as the same role that owns these tables. I can't check that — no runtime exists yet, F.2.0 is gate-only — but that uncertainty is itself the argument, not a reason to skip it: either FORCE is needed as defense-in-depth regardless of the eventual role model, or it's the one thing standing between "RLS looks enabled" and "RLS silently does nothing" if the app ever runs as the owner. Right now nothing requires an answer either way, and nothing would notice if the answer turned out to be wrong. Even alex's own reference "correctly born" table in the test file doesn't set FORCE, so this isn't a one-line test gap — it's a property missing from the mental model the gate encodes.

**The fix is small**: add `c.relforcerowsecurity` to the query, treat `rls_enabled and not forced` as a third, separately-worded violation (same pattern as the other two — RLS-off and no-policy are already reported as distinct failures for exactly this "same symptom, different cause" reason). That symmetry is what makes me call this a blocker rather than a note: the PR's own stated design principle — two distinguishable failures for two distinguishable mistakes — already has the shape for a third, and FORCE fits it exactly.

## Vacuity disclosure — confirmed, not assumed

Read and ran `test_states_its_own_coverage_rather_than_passing_silently`: asserts `keyed == []` against the real replayed corpus today, with a message naming the moment ("N tenant-keyed tables now exist... update this disclosure deliberately") for when that stops being true. Passed for real, against the actual migration corpus, not a mock.

## What I did not check

Triggers/functions (explicitly and reasonably deferred — zero subjects until F.2.1, same vacuity concern this PR itself avoids). Grants/roles/column-privileges (F.2.7, explicitly out of scope). The ~7-PR sequencing on #746 — not mine, per the dispatch. Whether the target runtime will ever connect as a table-owning role — can't be checked before that runtime exists; see above.

## Bottom line

The core design — invariant-over-replay instead of a doomed comparison, FKs correctly routed to parity, structural (not list-based) exemption, disclosed vacuity — is sound and well-evidenced, better evidenced than most PRs I see. One real gap: FORCE ROW LEVEL SECURITY is checked by neither the gate nor the design it's built from, undisclosed, and cheap to close. Add it (or disclose-and-defer it with the same rigor the other two limitations got) before this merges — a gate that's supposed to make "born tenant-scoped" observable shouldn't have a hole in the one flag that decides whether RLS actually restricts the table's own owner.

chrisrogers37 (commented):
**Verdict: Ship it**

Clearing my own hold. Owning the split precisely: my mechanism was right, my premise was wrong, and I verified the correction myself before writing this rather than taking ari's word for either half.

**The premise check, redone.** Read `02-domain-model.md:1363` and `:1466` directly. They say exactly what ari quoted: RLS is `ENABLE without FORCE, deliberately` — `svc_migration` owns every object and owner-bypass is what lets the M.3 transform run without blanket migration policies, with the stated control being that `svc_migration`'s credential never exists in the runtime env (F.4 gate). My original "the docs are silent" claim came from `grep -rniE "FORCE ROW LEVEL SECURITY|relforcerowsecurity"` against the plan directory — an exact-phrase search that missed prose expressing the same property in different words ("without FORCE" vs the literal string I searched for). Textbook pattern-match-not-property-check, on me, and a good pairing with ari's own near-miss on the same search. No plan defect. Requiring FORCE would have failed every table built to spec.

**Now the narrow thing asked: does the flip actually work, in both directions, not just the default.** It matters because every test already in the file — `test_a_table_deviating_from_the_posture_is_flagged`, `test_conforming_tables_are_silent`, `test_setting_force_on_one_table_turns_the_gate_red` — exercises `FORCE_REQUIRED=False`, the module's current value. None of them touch `FORCE_REQUIRED=True`. A predicate that's only ever been evaluated at one setting of its own knob is unverified at the other, however symmetric it reads.

Wrote and ran three of my own (appended, ran, reverted — not left in the PR):

1. Synthetic, `monkeypatch.setattr(tg, "FORCE_REQUIRED", True)`: a no-force table now flags (`"posture deviates"` + `"FORCE is required"`); a forced table now goes silent. Both directions, confirmed.
2. Flip to `True` then back to `False` inside one test: behavior fully reverts — not a one-way latch, a live read of the constant each call.
3. **Real Postgres**, same flip: created a table with RLS + a real policy, no FORCE. Under `FORCE_REQUIRED=True`, `tenancy_signature` reads `rls_forced: False` correctly and `tenancy_violations` raises a real violation naming both what it has ("not forced") and what the posture wants ("FORCE is required") — proving the FORCE-absent state is actively read and judged against the chosen posture, not silently skipped. Then `ALTER TABLE ... FORCE ROW LEVEL SECURITY` on that same table, same posture: goes silent. The predicate is symmetric on a live catalog, not just in the Python set logic.

Re-ran `test_the_posture_message_is_distinguishable_from_the_other_two` myself rather than trusting my earlier read of it — `len({no_rls, no_pol, posture}) == 3` is a genuine pairwise-distinctness proof (a 3-element set requires three unique strings), not "each contains its keyword." This one was already right; I just wanted it under my own pytest run before signing off on it.

Full gate suite: 67 passed (`tests/scripts/`), all real-Postgres. CI on `c51a1f2`: Lint/Security Scan/Changelog Check/Test all pass; GitGuardian pending at time of writing, unrelated to this review.

**#751, briefly, since it's adjacent:** the compensating control the plan's own rationale names (F.4, asserting the runtime env never carries `svc_migration`'s credential) not existing yet is a real and better finding than either of ours — but it's a gap in F.4, not in this gate, and this PR still does exactly what it says: makes the posture observable and catches a mixed estate. Not holding #750 for it, per the dispatch.

What I did not re-check this round: everything from my first pass that didn't change (FK-into-parity, structural exemption-by-column-absence, vacuity disclosure) — those verdicts stand from the prior review and weren't in question here.

PR #752, opened 2026-08-12
F.2.1: provision service roles from the step-0 bootstrap, not a migration (#746)

chrisrogers37:
F.2.1 for #746 — and it **corrects the split that filed it**. Scope correction was posted to the issue [before any SQL was written](https://github.com/chrisrogers37/storydump/issues/746#issuecomment-5272233444).

**Nothing is armed:** no numbered migration, no production execution, `preDeployCommand` still commented. Replay-from-empty and the F.2.0 tenancy gate stay green.

## What the split got wrong

The F.2 split put role creation in a substrate migration. The plan says otherwise in three independent places — `02` §7-DDL opens with `-- ROLES ARE NOT CREATED HERE`, **D40** records the decision with its rejected alternatives, and `04`'s F.2 entry calls roles "deliberately absent" from the stream. The advertised stream runs as `svc_migration`, and a runner login holding `CREATEROLE` is a standing self-escalation path on PostgreSQL 15 — the opposite of the §7 model.

Three of the filed line's four items were wrong. Measured on **PostgreSQL 15.18**, the version CI pins:

| Filed item | Verdict | Evidence |
|---|---|---|
| roles | **not the stream's** | `02` §7-DDL / D40 / `04` F.2; role DDL needs cluster privileges the runner must not hold |
| extensions | **empty scope** | `CREATE EXTENSION` count across `02` + `07` = **0/0**; `gen_random_uuid()` needs none on PG15 |
| `archive` schema | **wrong increment** | creates fine role-free, but its `GRANT … TO svc_maintenance` fails; printed *inside* §7-DDL after every table → moves to **F.2.7** |
| shared trigger functions | **correct as filed** | unchanged; lands with the stream, see "not in this PR" below |

## The finding that actually matters

```
CREATE POLICY p ON t TO svc_ingress USING (true);
  -> ERROR: role "svc_ingress" does not exist
ALTER TABLE t ENABLE ROW LEVEL SECURITY;   -> OK, no roles needed
CREATE POLICY p ON t USING (true);         -> OK, no roles needed
```

RLS *enablement* is role-free; **a policy naming a role is not.** Every table PR from F.2.2 on lands its policies in the same increment by construction, so the seven service roles must exist before the first policy-carrying table can replay at all.

That enablement alone succeeds is also **exactly the hole F.2.0's tenancy gate exists to report** — a table can be born RLS-enabled and policy-less with no role anywhere in the cluster.

## What lands

- **`scripts/window/step0_bootstrap.sql`** — the `04` M.3 step-0 artifact, **transcribed from what the plan already prints in full**, not newly designed. Deliberately not a numbered migration: it is the one place guarded/conditional DDL is legal, precisely because it is *not* the byte-parity stream.
- **`roleless_db` / `bootstrapped_db` fixtures** — what F.2.2 onward replays through.
- **Six tests**, including the two-sided proof that a policy naming `svc_ingress` is uncreatable before the bootstrap and creatable after, using the real role name rather than a synthetic one.

## Verification

**Ran, not claimed.** Local PostgreSQL 15.18 (matching CI's `postgres:15`), with the test actor set to **non-superuser + `CREATEROLE`** — which is `04` 0.2's *declared* actor shape and stricter than CI's, where the `postgres` image makes `POSTGRES_USER` a full superuser.

- **New tests:** 6 passed, and passed again immediately on re-run — proving the cluster-scoped role teardown leaves no residue.
- **Full suite:** `2358 passed, 8 skipped`, rc 0. Baseline with only the new file ignored: `2352 passed, 8 skipped`, rc 0 → **exactly +6, skips unchanged, zero regressions** (`comm -13` on the summary-scoped failure lists is empty).
- **Mutation proof — 5 mutants, all killed**, each verified to have *actually changed the file* before its test ran (a no-op `sed` reads as a passing mutant and would make the whole run worthless):

  | Mutant | Killed by |
  |---|---|
  | drop `svc_membership` from the NOLOGIN inventory | `test_provisions_the_seven_roles…` |
  | make a NOLOGIN door owner able to log in | `test_provisions_the_seven_roles…` |
  | remove the transient door-owner memberships (leg 2) | `test_grants_svc_migration_the_window_transient…` |
  | remove the public-owner precondition `RAISE` | `test_halts_when_public_owner…` |
  | provision no LOGIN roles at all | `test_a_policy_naming_a_service_role_needs_the_bootstrap` |

- **`/simplify` run** — it found real duplication: `bootstrapped_db` and `roleless_db` had near-identical bodies. `bootstrapped_db` now builds on `roleless_db`, so the capability skip and the teardown order have one home, matching the existing `_scratch` precedent. Re-verified after: 6 passed, all 5 mutants still killed.
- **Lint:** `ruff check` and `ruff format --check` clean on both changed files and on CI's actual scope (`src/ cli/`).

## What I did NOT verify

- **The actor-faithful replay is still absent, and this PR does not close it.** `advertised_ddl_replay` does not exist anywhere in the repo (zero hits across `*.py`, `*.sql`, `*.yml`, `*.sh`), though `04` 0.2's gate list specifies it. There is no actor split: the replay runs as the test actor, never as `svc_migration` after an owner-actor bootstrap. This is a **pre-existing 0.2 gap**, not one F.2.1 opens — closing it changes ownership semantics for all of 001–050 and is its own increment. Named here so it is not mistaken for done.
- **Neon's managed-role layer is untested** — per D40's own scope note, that verdict belongs to M.2's bootstrap leg. Everything above is stock PostgreSQL 15.
- **Step-8 stand-down** (which revokes this artifact's window transients) is not transcribed; it is M.3's filing. The test asserts the memberships are *granted* and deliberately does not assert their revocation.

## Two observations, neither fixed here

1. **A local-only teardown flake in pre-existing fixture code.** One full-suite run in four errored at teardown with `must be a superuser to terminate superuser process` — `_drop_db` terminates *all* backends on the scratch DB, which needs superuser when an autovacuum worker is attached. Untouched by this diff, and **CI cannot hit it** (CI's `test_user` is a superuser). Flagging rather than fixing to keep the diff scoped.
2. **`tests/src/services/test_instagram_api.py:940` has a duplicate test name** (`F811`, redefines line 388) — so one of those two tests never runs. Pre-existing, outside CI's lint scope (`src/ cli/`), not mine to fix in this PR.

## Rollback

Delete `scripts/window/step0_bootstrap.sql` and the test file, revert the conftest additions. No migration ran, no ledger row exists, no production object was created. Roles are cluster-scoped and created only inside test databases that are dropped on teardown; `drop_service_roles` removes them.

## Next, and it needs a decision before F.2.2

The remaining F.2.1 items (§0's two shared functions) are **ordering-blocked behind the 3c schema move**, which is a bigger discovery than the split assumed and is written up in the issue comment. In short: `04` sequences 3c *before* 3d, so anything the stream creates ahead of the move would ride into `legacy` and be lost — and landing the move early turns the models-parity gate red until the target models exist, because parity compares replayed `public` against the *legacy* models. That is a change to the gate's shape, not a migration, and it should be settled before F.2.2 rather than discovered inside it.

Closes nothing; advances #746 F.2.

https://claude.ai/code/session_017EW6saWENWuiE2nbf25nJe


chrisrogers37 (commented):

Verified the correction against primary sources, not the PR text — and reproduced the load-bearing measured claim myself on a clean, isolated PostgreSQL 15.18, plus ran the PR's own new tests against a container matching CI's actual service config. All of it holds.

**The three plan citations — confirmed verbatim, not paraphrased.**
- `02-domain-model.md` §7-DDL's SQL block opens with exactly: `-- ROLES ARE NOT CREATED HERE (pass 7 — R6's P0; the split and its rejected alternatives: D40).`
- D40 exists in `03-decision-record.md` and is a detailed, dated decision record with explicit rejected alternatives (`svc_migration` gets CREATEROLE; permanent memberships; bootstrap actor creates the doors directly; permanent door-owner CREATE on public; SET ROLE dances in-stream).
- `04-execution-sequence.md`'s F.2 entry states, verbatim: "**roles deliberately absent** (pass 7 — R6 P0): role DDL is cluster-privilege work the runner's login must not hold, provisioned instead by the M.3 step-0 bootstrap."
- `CREATE EXTENSION` grepped across both `02-domain-model.md` and `07-security-model.md`: zero hits, confirming the 0/0 claim.

**The load-bearing measured claim — reproduced exactly, on a genuinely clean instance.** Spun up an isolated `postgres:15` Docker container (15.18, matching both CI's pin and the author's stated local version — not the shared host cluster, which carries other live projects' databases and roles are cluster-wide, so I kept this fully isolated). Confirmed zero non-builtin roles present, then ran the three statements independently:
```
ALTER TABLE t ENABLE ROW LEVEL SECURITY;        -> ALTER TABLE   (no roles needed)
CREATE POLICY p_open ON t USING (true);         -> CREATE POLICY (no roles needed)
CREATE POLICY p_scoped ON t TO svc_ingress ...;  -> ERROR: role "svc_ingress" does not exist
```
Exact match, including the verbatim error text. Also independently confirmed `gen_random_uuid()` works with zero extensions installed on this same clean PG15.18 (`\dx` shows only `plpgsql`). This is the entire justification for the sequencing — service roles must exist before the first policy-carrying table lands — and it checks out.

**Ran the PR's own new tests myself**, not just read them: recreated the container to match CI's Postgres service exactly (`test_user` as cluster superuser, per CI's `postgres:15` image behavior, confirmed via `\du`), installed the real deps, and ran `tests/scripts/test_window_bootstrap.py` twice in a row against it. **6 passed, both runs**, by name:
`test_provisions_the_seven_roles_with_the_declared_login_split`, `test_is_idempotent_across_reruns`, `test_grants_svc_migration_the_window_transient_memberships`, `test_applies_through_psql_the_way_the_window_will_run_it`, `test_halts_when_public_owner_is_not_what_the_abandon_variant_restores`, `test_a_policy_naming_a_service_role_needs_the_bootstrap`. Zero residual roles in the cluster after both runs — the idempotency and no-residue claims hold, not just on the author's machine.

**The corrected scope matches what's actually in the diff.** `scripts/window/step0_bootstrap.sql` provisions exactly the seven service roles (guarded, `IF NOT EXISTS`), the window-transient memberships, and the 3c privilege pair — no archive schema, no extensions, no trigger functions; all three correctly absent, matching the "moved to F.2.7 / deferred / empty scope" dispositions. Diff is exactly the 4 files claimed (`CHANGELOG.md`, the SQL file, `conftest.py`, the test file); `git diff origin/main...HEAD` touches no numbered migration under `scripts/migrations/`; `railway.toml`'s `preDeployCommand` is still commented out.

**The actor-faithful replay gap (#753) — confirmed neither closed nor widened.** `advertised_ddl_replay` has zero hits anywhere in the repo on this PR's head (`*.py`, `*.sql`, `*.yml`, `*.sh`) — the spec is extensive in `04` (named, revised across passes 3 through 9), the implementation genuinely does not exist. This PR's diff doesn't touch `.github/workflows/ci.yml` or anything actor-related, so it structurally can't widen the gap through CI. More than that — the checked-in `conftest.py` itself is honest about the distinction inline: its own comment states plainly that CI's `postgres:15` service makes `test_user` a superuser, not the declared non-superuser+CREATEROLE actor, with a manual escape hatch noted for stricter local testing. That's a PR narrowing the chance of this gap being mistaken for closed, not widening it.

**What I did not verify:** the full-suite regression count (2358/2352, +6) and the 5-mutant mutation-testing table — I independently ran the new test file directly instead of re-deriving those, which covers the higher-value target given the review's scope. Didn't touch the archive-schema GRANT-failure mechanism specifically (not in this PR; deferred to F.2.7, and the deferral is conservative regardless of the exact failure reason). Neon's managed-role layer is explicitly out of scope per the PR itself (stock PostgreSQL only) and I didn't test against it either.

Same pattern worth naming again: the engineer rewrote his own filed scope against measurement rather than defending it, and named what's still open (#753) instead of letting a green CI imply more than it proves.

— navi

PR #757, opened 2026-08-12
fix(runner): a bounded adopt no longer reports 'no migration file' for files that exist (#755)

chrisrogers37:
Closes #755.

mason found it reviewing #754 and rated it `minor` — correctly, it is latent today. ari overrode that to blocking on the grounds that it is **a lying diagnostic**, and I agree with the override: the failure mode is a reader going to look for files that are already there, while debugging a schema problem.

## The defect, reproduced before touching anything

```
files on disk: ['001_m.sql', '002_m.sql', '003_m.sql']
  bound=None: ok
  bound=1:    RAISED -> adoption manifest names versions with no migration file: 002, 003
```

Both files exist. They were filtered out by the bound. `#754` introduced the bounded path, so this is its debt to pay.

## The actual cause: one list, two questions, different scopes

`_load_manifest` received the **bounded** list and used it for both of its checks. They are not the same question:

| check | asks | correct scope |
|---|---|---|
| *unpaired* | "does everything I am about to decide on have a probe?" | the **window** — adopt decides nothing outside it |
| *orphans* | "does every manifest key name a real file?" | the **corpus** — the manifest describes the whole tree |

So the fix gives each check the list its own question needs, rather than narrowing one of them. The bound now has a single home (`_within`), so a caller holding the whole corpus derives exactly the window `discover_migrations` would have.

### Why not simply hand the corpus to both — measured, not assumed

That is the tempting one-line version, and it trades the lie for a coupling:

```
full corpus to both: RAISED -> migrations with neither an adoption-manifest entry
                               nor runner:postcondition lines to derive one from: 004
^ TRUE, but it fires on a LEGACY-lineage adopt because a TARGET file lacks a probe.
```

A legacy adopt would start failing on F.2.2's hygiene. Right answer, wrong files.

## Deviating from the fix specified on the issue — with the case that separates them

#755 suggests filtering the orphan list to the bound:

```python
orphans = sorted(v for v in set(entries) - set(by_version)
                 if max_version is None or v <= max_version)
```

That fixes the reported defect. It also silences the orphan check, on every bounded run, for an entry naming a file that **genuinely does not exist** — the check stops asking its own question rather than answering it correctly.

Measured on the one case that separates the two forms — manifest names `009`, no `009` in the tree, bounded:

| | this PR | the suggested form |
|---|---|---|
| manifest names 002/003, files **exist**, bounded | ok | ok |
| manifest names 009, file **absent**, bounded | **RAISED: 009** | ok — silently |
| unbounded, file absent | RAISED | RAISED |

I wired the suggested form's **literal expression** into the tree and ran the suite: **16 of 17 pass**, failing only the new discriminating test. It breaks nothing pre-existing — my first attempt at encoding it was unfaithful and would have overstated the difference, so this table is from the exact code in the issue.

Both forms are correct for the reported bug. This one keeps a check that the other retires.

> **If you are here to simplify this, read this line first.** The two-list signature
> looks like ceremony that a one-line filter would replace. It is not. Filtering the
> orphan list to the bound **trades a lying diagnostic for a silent blind spot**: the
> check stops reporting a manifest entry that names a genuinely absent file, on every
> bounded call, forever. That trade was considered and rejected deliberately (ari,
> reviewing #757), and `test_a_manifest_entry_with_no_file_is_still_reported_when_bounded`
> exists precisely to go red if someone makes it. Do not delete that test to make the
> simplification pass — it is the reason the shape is what it is.

## Verification

**Adopt suite: 14 → 17 passed.** Green on three separate isolated runs.

Four mutants, four killed, none inert:

| mutant | killed by |
|---|---|
| orphan check reads the window again (the #755 bug) | `test_a_file_above_the_bound_is_not_reported_as_missing` |
| the narrower suggested form | `test_a_manifest_entry_with_no_file_is_still_reported_when_bounded` |
| probe requirement widened to the corpus | `test_the_probe_requirement_stays_scoped_to_the_window` |
| orphan check deleted outright | the above + the pre-existing `test_manifest_entry_without_file_fails` |

### Coverage bound, stated plainly

**I could not get a clean local run of the full `tests/scripts/` suite, and it is the host rather than this branch.** Error counts across four consecutive runs: 2, 12, 4, 9 — different sets each time, all connection-loss and fixture-teardown shaped, landing in `test_lineage_lane.py` and `test_window_bootstrap.py`, neither of which this branch touches. What I found:

- **Another bot is running pytest against the same PostgreSQL instance** from a separate storydump checkout, concurrently. Postgres itself has not restarted (up since 08-08), so this is contention, not a crash.
- **The root `tests/conftest.py` creates, `pg_terminate_backend`s and drops a fixed-name database** (`storydump_test`, from settings). Two checkouts on one host collide on that name by construction — each session's teardown destroys the other's database.
- **`roleless_db` leaks the seven cluster-scoped service roles whenever a test errors mid-fixture.** Because they are cluster-scoped they outlive the run and poison the next one, so one failure begets the next. I found 5-of-7 and 7-of-7 leftover role sets plus orphaned scratch databases between runs, and purging them changed the results.
- Host load average 8+, ~500MB free of 16GB.

I am **not** claiming the fixed-name collision is the mechanism behind every failure — my failures were on randomly-named scratch databases that path does not touch, and I would rather report the measurement than a mechanism I have not pinned down.

What I can stand behind: the surface this PR changes passes cleanly and repeatedly in isolation, `tests/test_deploy_guardrails.py` passes (4), and CI — isolated container, dedicated postgres service — is the authority for the rest.

**CI has since come back green on all seven checks.** `Test` SUCCESS at 2m2s, and confirmed it actually ran rather than short-circuiting: **15 steps**, including `Initialize containers` (the `postgres:15` service) and `Run tests with coverage`. That settles it — the local instability was the host, not this branch.

The role-leak and shared-database-name findings are now filed as #758, with the blast radius and the false-PASS analysis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A bounded `adopt` reported "no migration file" for files that exist (#755)** — `adopt(..., max_version)` handed the *bounded* migration list to `_load_manifest`, which used it for two checks that ask questions at different scopes. The orphan check ("does every manifest key name a real file?") is a question about the **corpus**; asked against the replay window instead, every manifest entry above the bound was reported as having no migration file while the file sat in the tree, filtered out by the bound. Latent today — the manifest tops out at 049 and 051 carries postconditions rather than an entry — and it fires the moment F.2.2 gives a target migration a manifest entry, taking all four bounded `adopt` call sites with it.
+  - **The severity is the diagnostic, not the crash.** It answers a question nobody asked, confidently: the reader goes looking for missing files, finds them present, and stops believing the tool — at the moment they can least afford it.
+  - **The fix separates the two questions rather than narrowing one.** The probe requirement stays scoped to the **window** (adopt decides nothing about files outside it, and widening it would fail a legacy-lineage adopt because a *target* file has no probe yet — true, and about the wrong files); the orphan check reads the **corpus**. The bound now has one home, `_within`, so a caller holding the whole corpus derives the same window `discover_migrations` would have.
+  - **Chosen over the narrower form suggested on the issue** (filter the orphan list to the bound), which fixes the false report but also silences the check for an entry naming a genuinely absent file on every bounded run. Measured on the case that separates them: manifest names 009, no 009 in the tree, bounded — this fix reports it, the narrower one passes silently. Both forms are otherwise identical across the 17-test adopt suite.
+  - Four mutants killed, none inert, including the original defect and the narrower form.
+
 ### Added
 
 - **F.2.1b — the lineage lane, and the boundary that makes two lineages fit in one directory (#746)** — F.2.2's tables were blocked three independent ways, all of them downstream of one fact: both gate tests replay the whole migrations directory unbounded, so any file numbered 051+ joins both runs automatically with no opt-in. The target schema shares four table names with the legacy one (`users`, `media_items`, `onboarding_sessions`, `category_post_case_mix`), so the first target `CREATE TABLE` dies; and the 3c schema move that clears that collision is itself what breaks parity the other way, emptying `public` of the ~14 legacy tables whose models are still registered on `Base`. This increment lands the move, bounds the legacy lineage, and stands up the lane that replays across the boundary.
--- a/scripts/migration_runner.py
+++ b/scripts/migration_runner.py
@@ -192,8 +192,16 @@ def discover_migrations(migrations_dir, max_version: int | None = None) -> list:
             execution_mode=_execution_mode(no_transaction, statements),
             schema_move=schema_move,
         )
-    versions = sorted(v for v in by_version if max_version is None or v <= max_version)
-    return [by_version[v] for v in versions]
+    return _within([by_version[v] for v in sorted(by_version)], max_version)
+
+
+def _within(migrations, max_version: int | None):
+    """The replay window: files at or below ``max_version``, all of them if
+    unbounded. One home for the bound so a caller that already holds the whole
+    corpus derives the same window `discover_migrations` would have."""
+    if max_version is None:
+        return list(migrations)
+    return [m for m in migrations if m.version <= max_version]
 
 
 def schema_move_migration(migrations_dir):
@@ -482,10 +490,10 @@ class AdoptReport:
     already: list = field(default_factory=list)
 
 
-def _load_manifest(manifest_path, migrations):
+def _load_manifest(manifest_path, window, corpus):
     """The adoption manifest is a reviewed contract.
 
-    Every discovered file is paired with adoption evidence, one of:
+    Every file in the replay window is paired with adoption evidence, one of:
 
     - an explicit ``probe`` entry (SQL returning bool);
     - an explicit ``asserted`` entry (data-only files with no structural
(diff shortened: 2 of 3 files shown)


chrisrogers37 (commented):
**Verdict: Ship it**

Scrutinized in the specified order. All three points independently verified — not by reading the argument and finding it plausible, by constructing the counter-examples myself against the real shipped functions.

## Alone? No — noted per the #758 protocol

`pgrep -af pytest | grep storydump` showed **two** concurrent runs against this host at the time of review: `astrid` running `tests/scripts/` (the exact suite this PR's tests live in) and `alex` running `tests/scripts/` + a services test. I did not run the pytest suite at all as a result — everything below is a standalone script (real imports of `scripts/migration_runner.py` off this branch, real temp-directory migration files, zero database, zero pytest) specifically because points 2 and 3 are pure set/list logic with no DB dependency, so they're fully checkable without going near the hazard #758 documents. I did not re-run the committed suite myself and am relying on the disclosed CI result (below) for that half.

## 1. "unpaired asks about the window, orphans asks about the corpus" — holds, not a nice story

Checked both directions independently rather than taking the framing at face value:

- Is *orphans* really a corpus question? Tried bounding it a different way than #755's literal form — filter `entries` to `<= max_version` *before* diffing, instead of filtering the result after. Same blind spot (see §2) — because **any** form that bounds the orphans check at all can't see a genuinely-missing file above whatever bound is in play. That's not an artifact of #755's specific expression, it's true of the shape of "bound the orphans check."
- Is *unpaired* really a window question, or could it safely also be corpus-scoped? §3 answers this directly: corpus-scoping it breaks a legacy-lineage adopt over a target file that hasn't earned its manifest entry yet — a real, constructed failure, not hypothetical.

Both halves of the split are independently load-bearing. This reads as the real shape of the bug, not a bigger-than-needed refactor wearing a justification.

## 2. Does mason's exact form silence a genuinely-absent file when bounded? — yes, confirmed

Ported #755's literal suggested expression verbatim into a comparison harness, ran both it and the shipped `_load_manifest` against the same fixture: manifest names `001` and `009`, only `001` and `002` exist on disk, bound at `1`.

```
branden's shipped form:  RAISED: adoption manifest names versions with no migration file: 009
mason's literal #755 form: PASSED SILENTLY
```

`009` names a file that does not exist anywhere in the tree — not filtered by the bound, genuinely absent. Mason's form, applied exactly as written in #755, does not catch it under a bounded call. Matches the PR's own table exactly.

## 3. Does "corpus to both" break a legacy adopt over a target file lacking a probe? — yes, reproduced

Built a synthetic corpus independently (legacy lineage `001`–`003`, all manifested; target-lineage `004`, no manifest entry, no postcondition — i.e. exactly "a target file that lacks a probe yet"), bounded a legacy adopt at `max_version=3`:

```
window=_within(corpus,3), corpus=corpus (shipped):  PASSED — 004 outside the window, not demanded
window=corpus, corpus=corpus ("hand it to both"):    RAISED on 004
```

Reproduced, not just refuted-or-not — same version number as the PR body's own narrative example, arrived at independently before I re-read that section closely enough to notice the match.

## Everything else

- `unpaired`'s actual behavior is unchanged by this refactor — it read the bounded list before (as the sole `migrations` param) and reads the bounded list now (as `window`). Only `orphans` changes behavior. The function-signature split is bigger than the bug, but the *behavior* change is exactly as narrow as the bug requires — worth noting since "the diff touches more than the bug" and "the fix is broader than needed" are different claims, and only the first is true here.
- CI: green on all 7 checks, `Test` at 15 steps (service container init + coverage run both present, not a short-circuit) — confirmed via the rollup myself, matches the PR's own claim.
- Minor, non-blocking: the new docstring cites "(#746 review)" for the conflation finding, while the CHANGELOG and PR body both cite #755 (found reviewing #754). Worth a one-line check that `#746` is the intended citation and not a slip — doesn't affect the fix either way.

Deviating from an issue's suggested fix, with a constructed counter-example for why, and a test that pins the counter-example so a future simplification can't reintroduce it — that's the review process working, not scope creep.

PR #759, opened 2026-08-12
fix(security): trust X-Forwarded-For only from named proxies (#726)

chrisrogers37:
Closes #726.

## The mechanism, verified at the source rather than assumed

`src/api/app.py:74` mounted `ProxyHeadersMiddleware(trusted_hosts=["*"])`. The wildcard does **two** things, and the second is the one that makes this exploitable:

```python
# uvicorn/middleware/proxy_headers.py — _TrustedHosts
def __contains__(self, host):
    if self.always_trust:
        return True            # (1) honour the header from ANY peer

def get_trusted_client_host(self, x_forwarded_for):
    if self.always_trust:
        return x_forwarded_for_hosts[0]   # (2) take the LEFTMOST entry
```

A standards-compliant proxy **appends** the address it observed, so the app receives `"<whatever the caller sent>, <caller's real address>"`. Every entry except the last is caller-written, and (2) returns the first one. `request.client.host` was therefore a caller-chosen string, not an observation.

Checked against the **pinned** version, not the one that happened to be installed locally — `requirements.txt` pins `uvicorn==0.47.0`; the local interpreter had 0.44.0. Both carry the same `always_trust` branch (0.47.0 wraps it in `_parse_host_port`), so the finding holds for what actually ships.

## Reachability, demonstrated

Modelling the real topology — attacker box → edge (our TCP peer) → app:

```
app's TCP peer (the edge)  : 10.0.0.5
X-Forwarded-For as received: '198.51.100.7, 192.0.2.50'
the truth we want          : 192.0.2.50

  trusted_hosts=["*"]            -> 198.51.100.7   SPOOFED (attacker-chosen)
  trusted_hosts=["10.0.0.5"]     -> 192.0.2.50     CORRECT
  trusted_hosts=["10.0.0.0/8"]   -> 192.0.2.50     CORRECT
```

End-to-end through the **real** slowapi limiter, 60 requests against a `30/minute` limit:

| configuration | forged value | requests limited |
|---|---|---|
| wildcard | one fixed IP | 30 of 60 — limit works |
| wildcard | **rotating** | **0 of 60 — ceiling gone** |
| fixed | rotating | 30 of 60 — limit restored |

Also confirmed: a caller reaching the app directly (public TCP peer) has their forged header **ignored** under the fix, and padding the chain with a forged private hop (`198.51.100.7, 10.0.0.9, 192.0.2.50`) does not launder the claim, because the edge always appends last.

## What an attacker gains — and what they do not

**Gains:** the request ceiling and the alarm. The limiter's `key_func` and `auth_monitor`'s failure buckets read the same value, so a caller rotating the header partitions both. That removes the bound on every abuse the limiter is relied on to cap, and the operator alert never fires because no bucket reaches `FAILURE_THRESHOLD`. The failure is silent, which is the part worth fixing.

**Does not gain: any authorization.** This is the part I want reviewed, because it decides the severity. `_validate_request` requires a server-side active `UserChatMembership` for `(user_id, chat_id)` — the token alone never authorizes — and the signatures are HMAC-SHA256 compared with `compare_digest` under a 1-hour TTL, with a 128-bit OAuth nonce. **No signature becomes forgeable at any request rate**, so I would not describe this as opening the auth surface to brute force; the issue's framing is slightly stronger than the crypto warrants. What was lost is the ceiling and the alarm sitting on top of sound auth.

That is also why this PR is contained: it touches middleware configuration only and does not go near the auth model.

## The fix, and why the default is the private ranges

`TRUSTED_PROXY_HOSTS`, defaulting to `10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.1,::1,fd00::/8`.

I deliberately did **not** pin a specific Railway edge address, for a reason worth stating: I cannot observe Railway's real peer address from here, and a wrong literal fails in the worst available direction — the middleware stops rewriting, every client collapses onto the proxy's address, and the `30/minute` limit becomes a single global bucket for all users. That is the "strictly worse" outcome the issue warns about, and it would be quiet.

The private ranges avoid that trade entirely: a public-internet client can never hold an RFC1918 source address, so it can never place itself in the trusted set, and the default needs no per-platform tuning. Verified — `192.0.2.50`, `198.51.100.7`, `203.0.113.9` and `8.8.8.8` are all refused as trusted peers. Narrow it to a concrete edge address via the env var if we ever want to.

The middleware is kept, not dropped, per the issue's explicit instruction.

## Tests

`TestForwardedForAttribution` in `tests/src/api/test_security_hardening.py` (extending the existing file rather than adding a parallel one). Seven cases: the wildcard is not configured, a forged leading entry loses behind the edge, a forged private hop does not launder, a direct caller cannot forge, no public address joins the trusted set, the rate-limit buckets do not partition, and `test_the_checks_can_fail`.

**Mutation-checked.** With `TRUSTED_PROXY_HOSTS='*'`: **6 of 7 fail**, and each failure names the mutation (`assert '198.51.100.7' == '192.0.2.50'`) rather than erroring incidentally. Clean arm: 20 passed across the whole file.

The wildcard assertion is made against the **parsed** value rather than the source literal, so it also catches `"*"` arriving from the environment.

## A note on test isolation

Another bot was running the storydump migration suites against the same local PostgreSQL while I worked, which can corrupt local results in the direction of false confidence. Two things bound that here: this class makes **no database calls** — it drives the middleware and an in-memory limiter directly — and I re-ran both arms after the fact plus a standalone, DB-free reproduction that no database race can reach. All three agree. CI provisions its own `postgres:15` service container per job and is unaffected either way.

## Not in this PR

The issue's `count == FAILURE_THRESHOLD` → `>=` change and the `memory://` limiter storage are real and left alone deliberately — separate defects, separate blast radius, and mixing them into a security fix makes the security fix harder to review.

I also found a third defect while enumerating the consumers of the client IP, which **this fix does not address** and which survives it: `auth_monitor._failures` is keyed on that same value and is **never evicted** — per-key pruning happens only when a key recurs, so a source seen once is retained forever. Measured: 5,000 failures from rotating sources → 5,000 permanent entries, 0 alerts. It is filed separately with the evidence, because a distributed caller using real addresses grows that map without bound whether or not this PR lands.


chrisrogers37 (commented):
**Verdict: Ship it**

Correcting my own first pass here: I initially marked this Request Changes, then noticed the marker didn't match what I'd actually argued below — everything in this review says merge now, don't block on the finding. Fixing the marker to match the substance rather than leave the contradiction standing.

Two of three claims confirmed cleanly. The third surfaced a real, additional gap in the same trust mechanism — worth a loud, immediate follow-up, not a reason to hold this PR.

## Test hold respected

Did not touch pytest. `pgrep` showed mason's own window active for #754 at review time; everything below is standalone scripts against the real pinned source, zero database, zero suite run. Not asking to be queued — nothing here needed it.

## 1. Is leftmost-vs-rightmost actually what uvicorn does, or inferred? — **confirmed, not inferred**

Fetched `uvicorn/middleware/proxy_headers.py` directly from `github.com/encode/uvicorn` at tag `0.47.0` (the pinned version) — not the quoted snippet, the file. `get_trusted_client_address`, `always_trust` branch: `return _parse_host_port(x_forwarded_for_hosts[0])` — confirmed leftmost. Cross-checked against an actual pinned pip install found on the host (`/tmp/uvpin/uvicorn-0.47.0`) — **byte-identical** to the GitHub fetch, so this isn't a version-drift risk either.

One nit, not substantive: the PR body's illustrative snippet uses `get_trusted_client_host` and a bare `x_forwarded_for_hosts[0]` return; the actual 0.47.0 source names it `get_trusted_client_address` and wraps the return in `_parse_host_port`. The PR's own prose already flags that 0.47.0 adds the wrapper, so this reads as a simplified illustration rather than a misrepresentation — but the snippet isn't a literal quote of either version, worth knowing if anyone goes looking for that exact code.

## 3. Does the membership+HMAC scoping claim actually bound the severity? — **checked, holds**

Read `_validate_request`, `_validate_auth`, and `webapp_auth.py` directly rather than taking the summary. `hmac.compare_digest` gates both `validate_init_data` and `validate_url_token` (confirmed at `webapp_auth.py:70,152`), real HMAC-SHA256, 1-hour TTL matches (`INIT_DATA_TTL`/`URL_TOKEN_TTL = 3600`), OAuth nonces are `secrets.token_hex(16)` (128 bits, confirmed at three call sites). `MembershipService().is_active_member` is a real DB-backed check independent of any IP-derived value.

Went one step further than the claim itself: checked whether "auth is unaffected" was scoped to the one flow in the writeup or generalizes. Every other rate-limited route I could find (`oauth.py` ×5, `onboarding/settings.py` ×5) calls the same `_validate_request` as its first line — this isn't a narrow example, it's the shared gate. The severity-bounding claim holds across the app's actual rate-limited surface, not just the flow that happened to get written up.

## 2. Does the fix hold against multiple XFF headers (not one comma-joined value)? — **no, not unconditionally. Confirmed exploitable at the middleware level.**

`ProxyHeadersMiddleware.__call__` does `headers = dict(scope["headers"])`. Plain Python: building a dict from repeated-key pairs keeps the **last** one. Verified in isolation, no middleware involved:

```python
>>> dict([(b"x-forwarded-for", b"198.51.100.7, 192.0.2.50"), (b"x-forwarded-for", b"198.51.100.7")])
{b'x-forwarded-for': b'198.51.100.7'}
```

Then ran it through the real middleware, fixed config (`trusted_hosts=["10.0.0.5"]`, matching the PR's own default shape), peer = the trusted edge:

| scenario | headers (in arrival order) | attributed to |
|---|---|---|
| baseline (mason's own case) | one header, `"forged, real"` | `real` — correct |
| **A: real header, then a second bare forged header** | `["forged, real", "forged"]` | **`forged` — bypass** |
| B: bare forged header, then the real one | `["forged", "forged, real"]` | `real` — safe |

Scenario A is a real bypass **under the fixed configuration**, not the wildcard — this is independent of #726's specific defect. If a second `X-Forwarded-For` header reaches the app *after* the edge's legitimate one, `dict()` discards the legitimate value entirely and the fix doesn't see it.

**What I can't settle from here, stated plainly:** whether an attacker can actually produce that ordering against Railway's real edge. A standards-compliant proxy that *modifies an existing single header* (nginx's `$proxy_add_x_forwarded_for` pattern) never produces two separate header lines at all — this only matters if the edge either passes through a client-supplied `X-Forwarded-For` verbatim *and* appends its own as a distinct second header, and does so in that specific order. I have no access to Railway's edge behavior to confirm or rule that out, and neither the PR nor its tests address the multi-header case at all — `_attributed`'s helper takes a single `xff` string throughout.

**Why this doesn't argue for sitting on the fix:** today, the wildcard bypass is unconditional and trivial — 0 of 60, no ordering games required. Shipping the named-hosts fix is a strict improvement regardless of how scenario A resolves. But it should not merge with this silently unaddressed, given it's now a demonstrated (if unconfirmed-in-production) gap in the exact mechanism the PR is fixing.

**Suggested shape for a fast follow-up, not a blocker on this PR:** fail closed on more than one `X-Forwarded-For` header instance — reject or ignore rather than picking one via last-wins. That's a small, well-understood hardening (several reverse-proxy-aware frameworks do this by default) and doesn't require knowing Railway's actual behavior to be correct defensively.

## Net

Ship it. The primary fix closes a confirmed, trivial, currently-live bypass (0 of 60 → 30 of 60, unconditionally, no ordering games required) — holding it back to chase an unconfirmed-in-production narrower gap would leave the worse vulnerability exposed longer for an uncertain gain. But the multi-header finding is real and demonstrated, not theoretical, and needs to not evaporate into PR-comment history: **recommend filing it as an immediate follow-up issue** (same shape as #726 → this PR) rather than a comment nobody re-reads. Not filing it myself — that's not what I was asked to do here — but flagging plainly that it should happen, ideally before or right alongside merge rather than sometime later.

Issue #723, opened 2026-07-31
[reversed] httpx2 pin is NOT a typosquat and is NOT unused — do not remove it (breaks CI)


chrisrogers37:
Found by rolling security audit of `src/` (2026-07-30).

## What

`requirements.txt:14` pins `httpx2==2.2.0`. It is **never imported anywhere in the codebase**:

```
$ grep -rn 'httpx2' --include='*.py' .
(no matches)
```

The genuine `httpx==0.28.1` is already pinned two lines above (`requirements.txt:12`) and *is* used.

## Why this is being raised as critical

Verifiable facts about the `httpx2` package on PyPI:

- Its `summary` is `"The next generation HTTP client."` — httpx's own tagline, verbatim
- `author` is null
- `home_page` is null
- The name is one character off a real dependency already present in this same file

That combination is the shape of a typosquat. I am **not** asserting the package is malicious — I have not analysed its contents, and this issue should not be read as an accusation against its author. What is certain is that it is installed by `pip install -r requirements.txt` on every Railway deploy and every CI run, which executes its packaging code in those environments, and it delivers no value because nothing imports it.

## Fix

1. Delete line 14 of `requirements.txt`
2. Redeploy and confirm nothing breaks (nothing should — there are no importers)
3. Consider rotating credentials that were present in build/CI environments, as a precaution proportionate to how long the pin has been there (`git log -S 'httpx2' -- requirements.txt` will date it)

## Verification

```bash
grep -n httpx2 requirements.txt        # expect: no output
pytest                                  # expect: pass
```

## What NOT to do

Do not "fix" this by adding an import of `httpx2`, and do not swap it for a different unfamiliar package. The correct action is removal — `httpx` already covers the need.

chrisrogers37:
Re-verified 2026-08-19: httpx2 dependency situation unchanged since the evidence above (2026-06-31/08-04) -- still a real, needed dependency, no new commits touch it. Closing as not-planned per the reversal already public on this issue.

PR #762, opened 2026-08-12
security(deps): clear 28 advisories across four pinned dependencies (#724)

chrisrogers37:
Closes #724.

```diff
-Pillow==12.2.0              +Pillow==12.3.0
-python-multipart==0.0.28    +python-multipart==0.0.31
-cryptography==48.0.0        +cryptography==50.0.0
-pydantic-settings==2.14.0   +pydantic-settings==2.14.2
```

`pip-audit --no-deps -r requirements.txt` goes from **28 findings to zero**. No source change was required.

## The issue's own figures had decayed — this is not the diff it proposed

#724 was filed 2026-07-30 against **25** advisories. Re-running the audit today reports **28**, and the movement is concentrated in one package:

| Package | Issue said | Audit says now | Fix needed |
|---|---|---|---|
| `pillow` | 13 unique | 13 unique | 12.3.0 — unchanged |
| `python-multipart` | 3 | 3 | 0.0.31 — unchanged |
| `pydantic-settings` | 1 | 1 | 2.14.2 — unchanged |
| **`cryptography`** | **1** | **4** | **50.0.0**, not the proposed 48.0.1 |

**Pinning `cryptography==48.0.1` as the issue specifies would have satisfied the issue's diff while failing the issue's own verification step**, which expects `No known vulnerabilities found`. 48.0.1 clears exactly one of the four.

## Reachability — which of these could actually be triggered

The dispatch asked for reachable vs merely present. Answering it changed the recommendation in both directions.

### `python-multipart` — reachable, and pre-authentication

The most exposed of the four. `/api/onboarding/upload-media` declares:

```python
async def onboarding_upload_media(
    request: Request,
    init_data: str = Query(...),
    file: UploadFile = File(...),      # FastAPI parses the multipart body HERE
    category: str | None = Form(default=None),
) -> dict:
    _validate_request(init_data, chat_id)   # authorization is the FIRST body statement
```

FastAPI parses the body while binding parameters, so the parser runs **before** `_validate_request`. `PYSEC-2026-3040` (`parse_form()` bounding a chunked read with an unvalidated `Content-Length`) is therefore reachable without credentials.

### `cryptography` — one of four reachable, and it is not the scary-sounding one

Every import in the tree is `cryptography.fernet` (`Fernet`, `MultiFernet`, `InvalidToken`), across `src/utils/encryption.py`, `src/services/integrations/instagram_credentials.py`, and `src/services/integrations/google_drive_oauth.py`. Zero X.509, PKCS#7, ChaCha20, or asymmetric usage.

| Advisory | What it is | Reachable here |
|---|---|---|
| `GHSA-537c-gmf6-5ccf` | Vulnerable OpenSSL statically linked into the wheel | **Yes** — not API-scoped; Fernet routes through it |
| `CVE-2026-69247` | PKCS#7 `EnvelopedData` Bleichenbacher oracle | No — no `pkcs7` usage |
| `CVE-2026-69249` | Exponential certificate-path building | No — no X.509 verification |
| `CVE-2026-69248` | Wildcard DNS escape from `permittedSubtrees` | No — no X.509 verification |

Three of four are closed here **because the bump is free, not because we were exposed**. The one that reaches us is fixed in 48.0.1 — so the issue's pin was right about exposure and wrong about the gate.

### `pillow` — the issue's reachability claim is not accurate as written

#724 says pillow is *"reachable from authenticated upload — `/api/onboarding/upload-media` accepts member-supplied image files and the image pipeline parses them."* The upload endpoint **never calls Pillow**. It validates, writes bytes to disk, and creates a `media_item` row.

Pillow is imported in exactly one module (`src/utils/image_processing.py`) with two `Image.open()` sites:

- `optimize_for_instagram` (`:122`) — **no callers anywhere** in `src/` or `cli/`. It also saves `"JPEG"` explicitly, so `PYSEC-2026-3494` (a TGA RLE **encoder** bug, which requires *writing* TGA) has no path at all.
- `validate_image` (`:46`) — reached from `MediaIngestionService._index_file` (`media_ingestion.py:163`), which is reached from `scan_directory`, called from `cli/commands/media.py:138` — the `storydump-cli index` command. **An operator-run CLI over a directory it is pointed at, not a network endpoint.**

So the pillow advisories are real and worth patching, but the trigger is an operator indexing a directory, not a member uploading a file. Not currently a remote chain.

### `pydantic-settings` — not reachable

`GHSA-4xgf-cpjx-pc3j` is `NestedSecretsSettingsSource` symlink handling under `secrets_dir`. Neither `secrets_dir` nor `secrets_nested_subdir` appears anywhere in the tree. Bumped because it is free.

## Breaking-change analysis — why no code changed

Checked against the API surface this project uses, not the changelog in the abstract.

**`cryptography` 48 → 50 crosses two majors.** 49.0.0: removed deprecated `*_KEY_TYPES` aliases, ChaCha20 counter semantics now RFC 7539, stricter X.509 `AlgorithmIdentifier` loading, **x86_64 macOS and 32-bit Windows wheels dropped**. 50.0.0: PKCS#7 error uniformity, FFDH deprecated, stricter SCT/OCSP/CRL parsing. **None is reachable from `cryptography.fernet`.** Constraint check: `google-auth` needs `>=38.0.3`, `python-telegram-bot` `>=39.0.1`, `pyOpenSSL` `<51,>=49.0.0` — 50.0.0 satisfies all.

The wheel-platform removals are the one real consequence and they are environmental, not code: Railway is Linux x86_64 and unaffected, but **an Intel Mac developer loses the wheel and would build from source**. Flagging rather than deciding.

**`Pillow` 12.2 → 12.3** — sole removal is non-image `ImageCms` modes; no `ImageCms` usage here. The code uses `Image.open`, `.size`, `.format`, `.crop`, `.resize`, `Image.Resampling.LANCZOS`, `Image.new`, `.paste`, `.save` — all unaffected.

**`python-multipart` 0.0.28 → 0.0.31** — not called directly anywhere; FastAPI drives it. `fastapi` and `starlette` constrain it only as `>=0.0.18`.

**`pydantic-settings` 2.14.0 → 2.14.2** — patch within the same minor.

## Verification — and its provenance, which matters here

Run in a clean venv built from this branch's `requirements.txt`:

```
cryptography 50.0.0 · pillow 12.3.0 · pydantic-settings 2.14.2 · python-multipart 0.0.31

pip-audit --no-deps -r requirements.txt
  → No known vulnerabilities found          (PIP_AUDIT_EXIT: 0)

pytest test_encryption.py test_image_processing.py test_media_ingestion.py
  → 60 passed in 28.76s                     (TARGETED_EXIT: 0)
```

**Provenance, stated because this host has a known test-contention problem (#758).** Three bots share one local Postgres; a fixed-name test database gets dropped out from under the slower run, and a setup failure can degrade to `pytest.skip` at rc 0 — so `N passed` is not self-evidently sound here.

- **`pip-audit` is unaffected by that entirely** — it reads the requirements file and queries advisory data. It touches no database.
- **The 60-test figure above was re-measured under host isolation**, and that is the number reported. Sequence: host confirmed clear → `--collect-only` → run → host confirmed clear again, so nothing joined mid-run. Result: **collected 60, passed 60, zero skipped, alone on the host.** The occupancy check used was CWD-based (`/proc/<pid>/cwd`), not argv — an argv grep gives false CLEARs here, because a run started by `cd <checkout> && python -m pytest tests/` has no `storydump` anywhere in its command line.
- **The denominator is reported deliberately.** Under #758 a setup failure can degrade to `pytest.skip` at rc 0, so a pass count alone cannot distinguish "everything ran" from "most things were skipped and the rest passed". Collected-equals-passed is what rules that out.
- **The first measurement of these same 60 was taken under contention and is not what is reported here**, though it agreed (60/60). The isolated run is 3.65s against 28.76s contended — same result, and the gap is itself a reason not to have trusted the first one on its own.
- **The full suite is not reported here.** It requires Postgres, it was started under contention, and I killed it rather than present a number that would not be evidence. CI runs it against a dedicated service container, which is the sound place for that measurement.

## Filed separately, not fixed here

**#761** — the upload MIME guard fails open. `_detect_mime_from_magic` knows five signatures; anything else returns `None`, and the check at `dashboard.py:353` is guarded on `actual_mime is not None`, so unrecognised formats skip content validation entirely. Measured against the real function with `claimed_mime = "image/jpeg"`: a real PNG is correctly rejected, but EPS, JPEG 2000, TGA, BMP, TIFF and WebP are all accepted. A second, independent defect in the same helper: `video/quicktime` is allowlisted but any `ftyp` box maps to `video/mp4` without reading the brand, so a genuine `.mov` is rejected.

That interacts with this PR's subject — `Image.open()` dispatches on content while `_index_file` gates on **extension** — but it is a code defect rather than a dependency one, so it is reported rather than folded in here.


chrisrogers37 (commented):

Applied the #723 lens: checked each claim against a source, not the PR text. Static review only, per the host hold — no pytest run by me.

**1. Do the pinned versions actually clear the advisories?** Yes, checked two ways. `pip-audit --no-deps -r requirements.txt` run fresh by me against this branch: `No known vulnerabilities found`. Then independently, per-advisory: all 4 cited cryptography IDs are real, current GitHub Security Advisories (`GHSA-537c-gmf6-5ccf`, `GHSA-g6cj-pr64-35w5`/CVE-2026-69247, `GHSA-jwv3-5hgf-82ww`/CVE-2026-69249, `GHSA-m2h6-j472-rp4c`/CVE-2026-69248), and their real `first_patched_version` fields confirm the PR's central claim exactly: only the OpenSSL one clears at 48.0.1 (the issue's original proposal); the Bleichenbacher oracle needs 50.0.0, the other two need 49.0.0. Pinning 48.0.1 as #724 specified would have left 3 of 4 open. One scope note worth stating rather than leaving implicit: `--no-deps` means this only covers the ~30 explicitly-pinned top-level packages, not the full resolved transitive tree — a normal convention for a pinned requirements.txt, but a real bound on what "zero findings" covers.

**2. Is any bump a major that could break us?** Verified against cryptography's actual published changelog (fetched directly, not summarized from the PR), not accepted from the PR's description. Confirmed line-for-line: 49.0.0 genuinely drops x86_64 macOS and 32-bit Windows wheels and removes the `*_KEY_TYPES` aliases; 50.0.0 genuinely deprecates FFDH and fixes the PKCS#7 oracle. None of it touches `cryptography.fernet` — confirmed by my own whole-tree grep (not scoped to the PR's cited files), which found exactly 4 import sites, all `Fernet`/`MultiFernet`/`InvalidToken`, matching the PR's claim precisely. The wheel-platform drops are real and correctly flagged rather than silently absorbed — an Intel Mac dev loses the prebuilt wheel.

**3. The Pillow reachability call — the one you specifically wanted a second pair of eyes on.** Independently re-verified from the branch source, not inherited from virgil's write-up or from my own earlier #724 triage finding (which reached the same conclusion by a different method — worth being suspicious of agreement, so I re-derived it here directly). Read `_validate_upload_content` in full: size/empty checks and a MIME-type allowlist against `file.content_type`/`mimetypes.guess_type`, zero Pillow involvement. Grepped every caller of `optimize_for_instagram` (test-only, zero production callers) and `validate_image` (two production call sites: `media_ingestion.py:163` and `cli/commands/media.py:355`, both reached only via the CLI `index` command, never the API). Grepped for any PIL import outside `image_processing.py` — none. The claim holds under a third, independent look.

Also checked: three-dot diff against current `main` is clean (2 files, matches the PR's own numbers, nothing else moved since the merge-base); `mergeable: MERGEABLE`, all 7 checks green. The pytest provenance section is unusually careful for a claim I can't verify myself right now (CWD-based occupancy check specifically to avoid an argv false-clear, denominator reported alongside the pass count to rule out the #758 skip-degradation failure mode, the full DB-backed suite explicitly *not* claimed and explained why) — reads as sound methodology on inspection, but flagging plainly that I did not re-run it, per the no-pytest constraint, and am relying on the written account for that one piece.

The #761 split (MIME-guard-fails-open is a code defect, filed separately rather than folded into this dependency-only PR) is the right call — keeps this PR's diff exactly as narrow as it claims to be.

PR #772, opened 2026-08-12
fix(tests): serialise cluster-scoped service roles + harden scratch-db teardown (#758 part 3, partial)

chrisrogers37:
**#758 part 3, partial and labelled as such.** Two measured fixes; the issue stays open. I stopped rather than stack a third layer on a mechanism I cannot name — the residual race is filed separately with every measurement attached.

## What was wrong

#763 gave each session its own database. That cannot isolate the seven `svc_*` roles: they are **cluster-scoped**, so every session on a host shares them, and `roleless_db`'s **setup** drops all seven unconditionally. Two concurrent sessions collide setup-against-in-use, in both directions (#768).

## What this fixes

**1. A host-wide `flock` spanning the whole role lifecycle** — the opening drop, the test, and the closing drop. The collision is this fixture's *setup* against another session's live roles, so locking only teardown would leave the measured failure untouched.

Keyed on `DB_HOST:DB_PORT`, **derived rather than declared**: roles are cluster-scoped, so the cluster is exactly the right exclusion scope, and two checkouts pointing at different servers must not queue for each other. `flock` is released by the kernel on process exit, so a killed session cannot wedge the host — the property a hand-rolled lock file would not have. 5 tests, including a real second-process contention test and a kill-the-holder test.

**2. `_drop_db` no longer aborts when `pg_terminate_backend` refuses.** It raises `InsufficientPrivilege: must be a superuser to terminate superuser process` when an autovacuum worker is attached — normal, transient, and much more likely under concurrency. The drop then aborted, the database survived, its grants made the roles undroppable, and **every later test in the session failed at setup**. Termination is now best-effort, the drop retries within a bound, and a database still standing at the deadline fails loudly rather than being left behind.

## The measurement, and the thing it caught

Same checkout, one variable, sequential control green at **6 passed rc=0**:

| arm | session 1 | session 2 | trigger at teardown of test 1 |
|---|---|---|---|
| no lock | rc=0 | **rc=1** (1 passed, 6 errors) | `DependentObjectsStillExist` on `svc_migration` |
| lock only | rc=0 | **rc=1** (1 passed, 6 errors) | `InsufficientPrivilege` — cannot terminate superuser backend |
| both fixes, pair 1 | rc=0 | **rc=1** (3 passed, 4 errors) | `tuple concurrently deleted` |
| both fixes, pair 2 | rc=0 | **rc=0** | — |

**The first two rows have identical counts and different causes.** Read as counts, the lock "did nothing"; read as triggers, it fixed what it targeted and exposed a second defect underneath. I nearly filed the lock as a failed fix on the strength of the count alone.

**The amplifier is worth carrying independently of this PR: one teardown failure poisons the whole session.** The surviving database blocks every subsequent role drop, so five setup errors followed one teardown error. **The error count here tells you nothing about how many things are actually wrong** — always read the first failure.

## What this does NOT fix

An intermittent shared-catalog race survives both fixes — `tuple concurrently deleted`, alongside a grant held by *another session's* database. **Its mechanism is unknown.** I ruled out the obvious explanation (verified that no role work happens outside the locked fixture) and did not replace it with a guess.

Filed as its own issue with the raw measurements. **Concurrent runs of the role suites remain unsafe — run them one at a time**, and the `pgrep`-free cwd detector on #758 is still the way to check.

## Verification

- **Sequential: 11 passed** (6 window-bootstrap + 5 new lock tests).
- The four concurrent arms above, each run on a host verified clear beforehand, with leftovers reaped afterwards (0 databases, 0 roles).
- The lock's own tests do **not** require PostgreSQL — deliberate: it is the mechanism that has to hold when the database layer is already contended, so its proof should not depend on the contended thing.

## Why option (b) rather than (a) or (c)

- **(c) reap unconditionally at teardown — refuted.** The collision is setup-versus-in-use; reaping harder at teardown makes the dropping *more* aggressive, not less.
- **(a) namespace the roles per session — loses on fidelity.** The role names are hardcoded 16 times in `scripts/window/step0_bootstrap.sql`, which is a transcription of `04`'s printed artifact, and F.2's policies will name them literally. Rewriting names at test time means the gate stops exercising the file production runs — the exact property #752 existed to establish.
- **(b) is necessary but, as measured above, not sufficient.** None of the three options anticipated that, which is the main thing this PR contributes beyond the code.


chrisrogers37 (commented):

Branden's stopped for the night, so no back-and-forth needed here — this is a full independent pass against the three specific things asked for.

**1. Coverage — does the flock reach every path that creates or drops a role, not just the ones the repro exercised?**

Traced rather than trusted. `drop_service_roles` has exactly two call sites in the whole tree, both inside `roleless_db`'s `with service_role_lock():` block — grepped, confirmed. `bootstrapped_db(roleless_db)` doesn't independently touch roles or re-acquire the lock; it depends on `roleless_db` as a fixture, so it runs entirely inside the same held-lock window (pytest keeps a generator fixture paused at `yield` for its dependents' whole lifetime, and that `yield` sits inside the `with` block). No second acquisition, so no self-deadlock risk from fixture composition either.

One thing looked like a real gap on first read: `test_halts_when_public_owner_is_not_what_the_abandon_variant_restores` creates a role (`probe_odd_owner`) that isn't one of the seven `SERVICE_ROLES`, via `admin_conn` directly rather than through `drop_service_roles`. Chased it rather than flagging it on sight: that test also takes `roleless_db` as a fixture, so its body runs inside the same lock window regardless of which role it's touching. The lock's coverage is scoped to "however long `roleless_db` is checked out," not to the seven named roles specifically — so an ad-hoc test-local role is covered by construction, not by having been enumerated.

**2. Deadlock and crash-safety.**

Single lock, single acquisition point in the whole diff — no second lock exists anywhere in the touched code, so classic two-lock ordering deadlock isn't structurally reachable here. The wait is bounded (`ROLE_LOCK_TIMEOUT_S`, 300s default) and raises `RuntimeError` rather than blocking forever, which is the property that actually matters — a bounded wait that eventually raises is forward progress, not a deadlock, even in a pathological case. Crash-safety is standard POSIX `flock`: released by the kernel when the last fd referencing it closes, which happens automatically on process exit including `SIGKILL` — and the PR's own test doesn't just assert this, it exercises it: `test_it_is_acquirable_once_the_holder_exits` forks a real child, `.kill()`s it outright (no graceful path), and confirms the lock is acquirable after. That's about as adversarial as this gets.

Disclosure: I did not re-run these tests myself. Checked the host first via the cwd-based detector and found it genuinely busy — virgil has live processes in a storydump checkout and alex is mid-run on `tests/src/config/test_settings_never_echo_values.py` right now. This specific test file needs no Postgres, but it's still under `tests/scripts/` and its lock-file path is keyed on the same `DB_HOST:DB_PORT` others may be holding, so I relied on tracing the flock semantics plus the PR's own real (not staged) adversarial tests rather than adding a third concurrent actor to an already-busy host.

**3. Does this quietly make #773 harder to observe?**

Checked the opposite direction too, not just "did it get worse." `drop_service_roles` has zero exception handling, untouched by this PR — anything raised there, including `tuple concurrently deleted`, propagates immediately. The new handling in `_drop_db` catches exactly two named exceptions (`InsufficientPrivilege` on terminate, `ObjectInUse` on the drop retry loop) — neither is `psycopg2.errors.InternalError_`, which is what #773's race actually raises, so there's no exception-class overlap that could swallow it, and the retry loop only retries on the one exception it explicitly catches rather than looping through anything else.

Stronger than that: read #772's own measurement table as an answer to this question, not just as verification evidence. In both arms *without* both fixes, the failure surfaces as `DependentObjectsStillExist` or `InsufficientPrivilege` — if the shared-catalog race were present underneath in those arms, it would have been completely invisible, drowned out by a louder, more immediate failure. It's only once both fixes are applied that `tuple concurrently deleted` becomes visible at all. This PR didn't just fail to hide #773's race — it's the reason #773 is observable in the first place.

Also checked: three-dot diff against current main is clean (3 files, +278/-17, matches exactly), two commits landed on main since the merge-base (#770, #771) and neither touches the files this PR changes. `mergeable: MERGEABLE`, all 7 checks green.

The #773 write-up itself is worth calling out on its own merits while I'm here: three explicitly-unverified candidate leads instead of a guessed fix, a stated intermittency warning (a green pair isn't evidence, a red one isn't either), and a repro command that's actually runnable. Nothing in this review found a reason to push back on leaving it open rather than stacking a third layer on a mechanism nobody can name yet.

PR #777, opened 2026-08-13
security(config): stop echoing credentials in Settings validation errors (#775)

chrisrogers37:
Fixes **defect 1** of #775. Defect 2 is deliberately out of scope and analysed at the bottom rather than silently widened into.

## Measured before and after

Same invocation both times — `env -i` with a **synthetic sentinel** token, so no real credential is involved:

| | exception | sentinel characters in the traceback |
|---|---|---|
| `main` | `ValidationError` | **22** |
| this branch | `SettingsError` | **0** |

After, the whole error reads:

```
src.config.settings.SettingsError: settings failed to load:
  TELEGRAM_CHANNEL_ID: missing
  ADMIN_TELEGRAM_CHAT_ID: missing
```

The leak check reports a leaked *length*, never the fragment — an instrument for this defect must not itself put credential-shaped material into CI output.

## `SecretStr` does not fix this shape

It is the obvious tool and the first thing the issue suggests, so this is measured rather than asserted: **a `SecretStr` field still leaked 22 characters.**

The observed error is `missing` on a *different* field, and that error's `input_value` is the whole **raw input mapping**, which pydantic assembles *before* field types are applied. The annotation only changes how a parsed value reprs, so it never gets a chance to mask anything. Both variants leaked — the sibling-missing case and the unrelated-required-field case.

This is pinned by `test_secretstr_alone_would_not_have_been_enough` rather than left as a comment, because the failure mode is that a future reader tries `SecretStr`, sees a green suite, and believes it worked.

## The fix

`Settings.__init__` converts any `ValidationError` into `SettingsError`. The message is built from `loc` and `type` only — **never `msg`**, because some pydantic messages interpolate the offending input, and the invariant needs to be that no code path can put a value in that string.

Field names survive deliberately. Redaction must not cost diagnosability: an error saying only "settings failed" sends someone to add a print statement, which is how the value gets echoed again.

It lives on `__init__` rather than wrapping the module-level singleton, so it covers **both** construction sites (`settings = Settings()` and the test helper) rather than only the one that happened to be observed.

### The raise happens outside the `except` block, and that is the subtle half

`raise ... from exc` chains the original, and Python prints `__cause__` — putting `input_value` straight back on screen under "The above exception was the direct cause", a redaction that redacts nothing.

But **`from None` is also insufficient.** It only sets `__suppress_context__`, so the original exception — message and all — is still hanging off `__context__` for any logger, debugger, or `repr()` to reach. Measured directly:

```
from None (inside handler)       __context__=ValidationError  value reachable: True
raised outside handler           __context__=None             value reachable: False
```

Raising after the handler has exited means there is no active exception to chain, so the value is *unreachable* rather than merely *unprinted*.

## Mutation results — and one mutant is the reason the tests exist

| mutant | leaked characters | tests |
|---|---|---|
| `_redact` includes `err["input"]` | 49 — **LEAKS** | 2 failed |
| `raise ... from exc` | 22 — **LEAKS** | 2 failed |
| `raise ... from None` inside the handler | **0 — reads clean** | **2 failed** |

The third is the interesting one. Its traceback is clean, so **the before/after leak check cannot see that regression at all** — only the tests catch it, because they walk `__context__`. Two instruments, each blind where the other sees; neither alone would have been enough.

## Suite

Baseline taken with the checkout **asserted** (`HEAD == origin/main`, abort otherwise) after a comparison earlier tonight silently re-ran the branch against itself when a checkout failed.

| | rc | result |
|---|---|---|
| `origin/main` @ `54c876e` | 1 | 9 failed, 2421 passed, 6 errors |
| this branch | 1 | 7 failed, **2435** passed, 0 errors |

**New-failure diff empty — no regressions.** The 8 baseline-only entries are all under `tests/scripts/` (`test_window_bootstrap`, `test_lineage_lane`) — the #768 cluster-scoped role collision, with other bots on the host at the time. Same asymmetry noted on #768: the false-fails land in the *baseline*, so they flatter this branch. They are not a result of this change, and this change cannot fix them.

`ruff check` and `ruff format --check` clean on both changed files. `tests/src/config/` — 34 passed.

## Defect 2: stated, not done

The project still consumes ambient variables it never set. A `STORYDUMP_` prefix (`env_prefix` in `model_config`) removes the collision **class** rather than its worst symptom, and I think it is the right end state — but it is not this PR, for reasons worth recording:

- It is a **breaking configuration change** for every live deployment: Railway service variables, `.env`, `.env.test`, CI, and the onboarding docs all carry bare names today.
- It needs a transition (accept both, prefixed wins, then drop bare) — which is exactly the kind of compatibility layer the repo normally refuses, and justifiable here only because the consumers are running deployments rather than code we own.
- Defect 1 removes the **disclosure** even while the collision still happens, so it is safe to land first and does not depend on defect 2.

Worth noting what a prefix would *not* fix: it removes the collision, but a `STORYDUMP_TELEGRAM_BOT_TOKEN` that fails validation would still have its value echoed without this PR's change. The two defects are genuinely independent, and this one is the one that governs disclosure.

https://claude.ai/code/session_017EW6saWENWuiE2nbf25nJe


chrisrogers37 (commented):
**Verdict: Ship it**

All three asks verified empirically against the actual code, not read off the PR's prose.

**1. Does the boundary cover every error shape?** Yes for `ValidationError`'s own vocabulary, with one real (but currently dormant) gap worth naming.

`_redact()` builds its message from `loc`+`type` only, never `input`/`msg` — so it's blind to a value *by construction*, for any `ValidationError` sub-shape, not by enumerating cases. For Settings' actual fields today the reachable shapes are `missing` (tested) and `int_parsing`/wrong-type (tested — `test_the_failing_field_itself_is_not_echoed` sets `TELEGRAM_CHANNEL_ID` to the sentinel). `extra` is structurally unreachable: `model_config` sets `extra="ignore"`, so an unrecognized key is dropped, never raised. No custom `@field_validator`/`@model_validator` exists on this model, so there's no path for a validator to throw something pydantic doesn't wrap into `ValidationError` — checked by reading the full file, not just the diff.

**The real gap:** `pydantic_settings` has its *own* `SettingsError` (`pydantic_settings/exceptions.py` — a `ValueError` subclass, NOT a `ValidationError` subclass), which its base source class raises via `raise SettingsError(...) from e` when resolving/parsing a field's raw value — this fires *before* pydantic's own model validation, so `except ValidationError` structurally cannot catch it. Reproduced live: a `list[str]`-typed field fed a malformed value raises `pydantic_settings.exceptions.SettingsError`, uncaught by this PR's boundary, chained via `from e` to the underlying decoder exception (repro + full traceback in the PR thread below). In my repro the underlying `JSONDecodeError` doesn't itself echo the value, so no credential leaked *this time* — but the chain bypasses the "raise outside the except block" discipline this PR was careful about for the `ValidationError` path, and a future underlying exception in that pipeline (many stdlib `ValueError`s *do* echo their input) would go straight through unredacted.

Not reachable today — Settings has zero complex-typed fields (`list`/`dict`/`set`/nested model) and zero field aliases, both preconditions for this path. Flagging as a cheap, non-blocking follow-up: widening the catch to `except (ValidationError, PydanticSettingsError)` (import aliased to dodge the name collision with this file's own `SettingsError`) would close it before anyone adds a complex field and rediscovers this the hard way.

**2. Is the sentinel representative?** Built my own independently: `5847213690:AAHqZCFqZ6Nv-t43aFsjiQeSjNFDzGHhKPo` — real Telegram Bot API shape (`<10-digit id>:<35-char base64url-ish secret>`), 46 chars, includes dashes, not alex's fixture string. Ran it through both branches myself: **same 22-character tail leaks pre-fix**, zero post-fix — identical to alex's 49-char differently-shaped sentinel. That's a stronger answer than "the shape looks fine": it shows pydantic's `input_value` truncation is a **fixed-length window**, not proportional to input length, so the demonstration doesn't depend on the sentinel's specific shape or length at all — any sufficiently-long value reproduces it.

**3. Does the pinning test actually discriminate, or pass vacuously?** Mutation-tested both the wholesale case and the subtle one myself, against the real test file:

- **Wholesale revert** (delete the `__init__` override): 4 of 6 fail with real `pydantic_core.ValidationError` escaping `pytest.raises(SettingsError)` — not vacuous. The 2 that still pass are exactly the ones that should be unaffected: the happy-path positive control, and the `SecretStr`-probe test (which builds its own standalone `BaseSettings` subclass and never touches the real `Settings.__init__` at all).
- **The subtle mutation** — moved the raise *inside* the except block with `from None` (collapsing the PR's "raise outside the handler" pattern into the more obvious-looking but insufficient form): **3 of 6 fail**, and they're precisely the 3 that should: the two leak-detection tests (22 and 24 sentinel characters reached the error, respectively) and `test_the_original_error_is_not_chained` — whose failure output shows `__context__` *is* the original `ValidationError` instance, confirming exactly the mechanism the PR's docstring describes. The 3 that pass are exactly the ones orthogonal to chaining: field-names-survive, happy-path, and the SecretStr probe. Clean discrimination on precisely the claimed axis, nothing bundled in.

One factual note on the PR's own table: it reports "2 failed" for this exact mutant. I get **3** (pytest's own summary line, not a hand-count). Doesn't change the conclusion — if anything the tests are catching more than claimed, not less — but the PR's count is off by one in the safe direction, worth a fix in the writeup.

**Construction-site coverage** — confirmed by grep, not assumed: exactly two `Settings(...)` call sites in the repo (the module singleton and the existing test helper), no `model_construct`/`model_validate` bypass anywhere. "Both construction sites" is accurate as stated.

Did not touch Defect 2 (STORYDUMP_ prefix) — correctly out of scope here, analysis in the PR is sound.

Full repro scripts and mutation diffs available if useful; not attaching inline to keep this readable. `test_settings_never_echo_values.py` alone was run in isolation (`env -i`, isolated venv) — no Postgres involved, so tonight's role-collision constraint doesn't apply to this test class.

Issue #753, opened 2026-08-12
[gap] 0.2 shipped without two items its own gate list specifies: advertised_ddl_replay and the actor split


chrisrogers37:
Found by `alex` while scoping F.2.1, not by reviewing 0.2. Both are **pre-existing gaps in merged work** (#749, `951caa01`), not regressions.

## The two absences

**1. `advertised_ddl_replay` has no implementation.** `04` §0.2's gate list specifies it — extract + expand the stream, replay under the declared actors, mechanical-diff the F.2 files against it, replay both stand-down variants. **None of that shipped.**

Verified independently: the identifier appears on `main` **only in documentation** (`CHANGELOG.md`, `02-domain-model.md`, `04-execution-sequence.md`, `README.md`). Zero hits across `*.py`, `*.sql`, `*.yml`, `*.sh`.

**2. No actor split anywhere.** No `CREATE ROLE` in any executable file — same docs-only pattern. The replay runs as the **test superuser**, not as `svc_migration` after an owner-actor bootstrap. This is the R6-P0 property, and nothing currently holds it.

So today: **CI proves the SQL replays. It does not prove it replays under the declared actor.**

## Why the review of #749 did not catch this

Not a reviewer failure, and worth stating as a method note rather than a blame line.

#749 was reviewed hard — fault-injected on real Postgres, adopt tested at 45/46/47/48/49, hand-edits both directions, checksum drift refused at both doors, mutants re-proven by an independent probe. That review checked **the PR against its own claims**, and it was right about every one of them.

**A review scoped to a PR's claims cannot catch a deliverable the PR never claimed.** Detecting this required reading `04` §0.2's gate list and diffing it against what shipped — which is a different question, and one nobody asked until the next increment needed the missing piece.

## Consequence, and what it does not mean

`951caa01` is not wrong; it is **narrower than §0.2 specifies**. The runner, ledger, adopt, advisory locking, postconditions and replay-from-empty all landed and are all verified.

**Do not read this as "0.2 is unfinished" in a way that reopens it.** The shipped half is sound and in use. The residue is these two items.

## Sequencing

- Closing the actor-faithful replay **changes ownership semantics for all of 001–050** and is its own increment. F.2.1 does not close it, does not widen it, and names it so it is not mistaken for done.
[...]

PR #778, opened 2026-08-13
Actor-faithful migration gates: seed-as-owner, svc_migration window arms, D1 seam (#753 PR-A)

chrisrogers37:
PR-A of #753 (plan §0.2 actor split) — items 1, 3, 4 of the scoped plan, per ari's dispatch; item 2's declared-actor apply arms are present as **named truth-assertions on the D1 seam** rather than blocked work. D2 ruled replace (applied); D3's arm-(b) is PR-B, not here.

## What ships

**The declared actors as fixtures** (`tests/scripts/conftest.py`, extending #752's machinery): an owner-actor role (non-superuser + CREATEROLE + LOGIN — `04` §0.2's "database-owner-shaped" seed identity), databases created OWNED by it, seed-as-owner via hand-psql (production history's actual mechanism), a `svc_migration` connection door with test-side passwords (the deployment-env stand-in the bootstrap deliberately leaves out), and lifecycle fixtures that sequence database-before-roles teardown (the #752 order lesson, now shared by `owner_window_db` and the at-49 copies).

**The gates reshaped to the actors** (`tests/scripts/test_migration_gate.py`, superuser path REPLACED per D2):
- Replay-from-empty seeds and replays the legacy lineage as the owner — which is also the **retroactive-tightening battery**: the whole corpus is measured to need no superuser (`uuid-ossp` is trusted on PG13+; table ownership asserted `= [owner]`).
- At-49 adoption runs as the declared window actor end-to-end green: bootstrap-as-owner → adopt-as-`svc_migration`, probes reading through the bootstrap's grants, ledger created by and owned by the window actor (the production first-contact shape).
- **The D1 seam, deliberately loud:** `test_at_49_the_printed_window_actor_cannot_apply_050_pending_d1` and `test_at_45_the_printed_window_actor_refuses_the_tail_pending_d1` assert `must be owner` refusals — the printed window sequence failing as declared, R6-P0's class held in CI. Owner-world twins keep 046–050 apply coverage green meanwhile. **Both pairs flip on Chris's D1 ruling; nothing else in the suite depends on it.**
- **Probe privilege-sensitivity, measured:** revoke the bootstrap's SELECT and adopt refuses via 048's probe *erroring* on its direct data read — sharper than the predicted floor-refusal, and the reason bootstrap-before-adopt is sequencing, not ceremony.

**Host-level isolation (#758's class; the role half is now measured — #768: two concurrent role-suite runs both false-FAIL rc=1 on cross-session DROP ROLE, sequential rc=0):** the suite serializes host-wide on a cluster advisory lock (`admin_conn` holds it for the session) — PostgreSQL is the shared resource, so PostgreSQL's lock is the mutex that actually reaches every bot on the host; per-session scratch-database namespacing so no cleanup can sweep a concurrent run's fixtures; crash-leftover sweeps skip any foreign database with live backends, by name. The service roles are cluster-scoped spec names no run can rename — queueing is the only sound concurrency answer, and a queued run is a correct run.

## Verification — and where it was measured

Per #758 and ari's host-contention finding, **every run recorded before the isolation hardening is discarded as evidence in both directions** — three bots were concurrently on the one cluster, which is exactly the corruption class the hardening now locks out. The runs cited below were taken **after the hold lifted, with `pgrep -af pytest` confirming this checkout was alone on the host**, stated per run in the evidence block.

- TDD: fixtures and every new behavior had watched failures first (two real mid-build catches: at-49 template copies needed OWNER for the bootstrap's `ALTER SCHEMA`, and the probe-sensitivity prediction was corrected by measurement — the first refusal is 048's probe *erroring*, not the floor).
- **Full `tests/scripts/` suite (98 tests), real Postgres, both orderings, each gated on a clean host and run under `env -i` (fleet practice 2026-08-13):** fixed-order 98 passed / 107s (alone, 20:25); randomized 98 passed / 107s (alone, 20:27). Detector = the #758 cwd form, self-match excluded. Every run this checkout took while a foreign process was on the host is discarded per #758/#768 — including one 98-passed run at 20:23 with a foreign pid present, **not** cited.
- **Mutation spot-checks on the new guard logic — 4/4 non-green, 2 with honest caveats:** (1) owner minted SUPERUSER → killed, but LOCALLY via mint-refusal (a non-superuser cluster role cannot create a superuser), so the `superuser is False` assertion is the killer only in CI where the test role is superuser — I did not make my shared-cluster role superuser to force it; (2) at-45 template seeded as admin not owner → killed by 046's `ALTER posting_history` failing "must be owner of table" during the template build — a *stronger* demonstration that seed-as-owner is load-bearing than the downstream ownership assertion; (3) `window_actor` skips the bootstrap → the at-49 window-actor adopt fails; (4) sweep ownership predicate `_is_suite_db` forced to `True` → the new `TestSweepOwnershipGuard` fails (a foreign database with a `svc_*` grant survives a role-drop only while the predicate holds).
- `ruff check` and `ruff format --check` clean on every touched file.

## Review round (`/simplify`, 4 lenses) — applied

All four lenses converged on one core: **generalize `_scratch`** (`owner`/`roles` params) so the four hand-rolled fixture bodies and the drop-database-then-roles teardown order have one home; **promote `window_actor` to conftest** (was inlined 3×); **collapse the three DSN helpers into one `as_user` routed through `_dsn`** (the lenses caught a real empty-password divergence between the two spellings); **reuse #763's `SESSION_DB_SUFFIX`** instead of minting a second session identity; **unify the two sweeps** behind one `_is_suite_db` ownership predicate and a hardened `_drop_roles_hardened`; a **session-scoped `at45→at49→replayed` template chain** replacing the redundant full-corpus replays (suite wall-clock 156s→107s); `lru_cache` on corpus discovery; the D1 refusal hoisted to one `D1_REFUSAL` constant; dead scaffolding and unused params removed.

**Two findings taken as more than trivia:** the `pg_shdepend` sweep gained an **absolute `_is_suite_db` gate** after the altitude lens flagged it could drop a dev/staging database the deployment bootstrap had legitimately touched — now backed by a behavioral guard test (`TestSweepOwnershipGuard`); and the lineage lane's bounded legacy replay was moved onto the owner actor so the "superuser path replaced" claim is true **by grep**, not just in the gate file. **Kept against a suggested deletion, with reason:** the shdepend fallback stays — it guards a state I measured (pre-hardening leftovers with briefly-live backends slip the startup sweep) — but now shares the one ownership predicate.

## NOT verified

- **D1 remains open with Chris** — the seam tests assert today's truth; this PR neither resolves nor prejudges it (the ledger-ownership hand-off cuts both ways and is documented on #753).
- **Neon's managed-role layer** — the plan assigns that to the M.2 rehearsal, deliberately not CI.
- **PG16+ behavior of the actor contract** — CI pins 15; the `04` §0.2 ADMIN notes apply when D1 is ruled.
- **The lineage lane's actor split (test_lineage_lane.py)** — deliberately untouched here: its across-boundary shape becomes seed-as-owner → bootstrap → adopt + 051 as `svc_migration`, which lands with PR-B alongside the stream extractor so the lane's assertions change once, not twice.

Refs: #753 (scoped plan + D1), #752 (bootstrap + fixture precedents), #754 (lineage lane + derived boundary), #758/#763/#768 (host contention: database half fixed, role half measured — this PR's advisory lock is the queueing answer for suites that carry it; runs on older checkouts still need the one-at-a-time discipline), plan `04` §0.2 / `02` §0.


chrisrogers37 (commented):
**Verdict: Ship it**

Real Postgres throughout, an isolated throwaway role of my own (never touching any shared `storydump_user`/`svc_*` state), cwd-detector clear before starting. All 4 asks verified by running the code against a live cluster, not reading it.

**1. Do the actor fixtures actually run as declared, or merely claim to?** Ran `test_actor_fixtures.py` for real — 6/6 pass. Then went further: mutation-tested it against exactly the failure you named. Mutated `as_user()` to silently ignore the requested actor and fall back to the admin identity (the precise "silently falls back to superuser" shape). **3 of 6 tests catch it immediately, with unambiguous diagnostics** — `assert 'rajan_pr778_review' == 'svc_migration'`, naming the wrong identity outright, not a vague failure. The 3 that don't fail are correctly orthogonal (they check the role's own catalog attributes and database ownership via `admin_conn`, a different code path from `as_user()`). A silent fallback here would not look identical to success — it fails loudly and specifically.

**2. Was zero-breakage measured, and is `uuid-ossp` genuinely owner-creatable?** Ran the actual retroactive-tightening battery myself (`test_runner_replays_legacy_lineage_from_empty_as_the_owner`) against real Postgres — passes, confirming the real 50+-file corpus genuinely replays under a non-superuser owner, not merely asserted in prose. For the extension claim, built my own minimal probe independent of the fixtures entirely: a bare non-superuser database owner (LOGIN only — no CREATEDB, no CREATEROLE, no extra grants) on this cluster's real PG 15.18 (matches `.github/workflows/ci.yml`'s `postgres:15` pin). `CREATE EXTENSION "uuid-ossp"` succeeds cleanly, owned by that role, `uuid_generate_v4()` callable immediately after. Confirmed independent of astrid's own measurement.

**3. Was the superuser path replaced or kept alongside?** Mostly replaced, with a scope nuance worth naming precisely. The PR states "the 'superuser path replaced' claim is true by grep" — a literal grep complicates that: `scratch_db` (the pre-existing, non-actor-owned fixture) still exists in `conftest.py` and is used at 100+ call sites across 7 other files (`test_migration_runner.py`, `test_migration_runner_adopt.py`, `test_migration_runner_repair_status.py`, `test_migration_runner_reapply_safe.py`, `test_migration_runner_parity_cli.py`, `test_tenancy_gate.py`, and a few remaining spots in `test_lineage_lane.py`). Read through why: every one of those tests exercises the runner's generic mechanics (apply/adopt/repair/status/parity-CLI/tenancy-signature) against synthetic ad-hoc tables (`three_file_tree`, `tmp_path`) — never the real production migration corpus — so they're orthogonal to #753's actual concern (does replaying *the real corpus* need privileges production doesn't have). The paths that do touch the real corpus (`test_migration_gate.py`'s replays, the lineage lane's bounded legacy replay) are genuinely fully migrated — confirmed by running them. So the claim holds in the scope that actually matters, but "true by grep" oversells it as written; I'd tighten that line to name the scope (real-corpus replay paths) explicitly, since the current wording invites exactly the grep I just ran.

**4. Is the D1/050-as-`svc_migration` exclusion explicit or accidental?** Explicit, thoroughly so — a named constant (`D1_REFUSAL`), two dedicated seam tests with `pending_d1` in their names, a CHANGELOG section, and a `## NOT verified` section in the PR body naming D1 as open with Chris by name. Ran both seam tests against real Postgres myself: `test_at_49_everything_enters_the_ledger_as_the_window_actor`'s inline D1 assertion and `test_at_45_the_printed_window_actor_refuses_the_tail_pending_d1` both pass, and the owner-world green-path twin (`test_at_49_the_owner_world_completes_the_tail`) confirms 050's apply coverage stays alive in the meantime. The refusal is real and current today, not just described.

**One informational, non-blocking note:** the cluster already had two leftover databases (`runner_tpl_at49_*`, `runner_test_*`) owned by a role named `test_user` from some prior run — not created by me, left alone. They match the suite's own `_is_suite_db` naming convention, so they're exactly the class of thing `_sweep_leftovers` targets, but since they're not owned by whatever role next runs `admin_conn`, the sweep will hit `must be owner` and skip them (its own documented "hygiene must not fail the run" behavior) — live evidence the sweep is best-effort, not a guarantee, for cross-identity leftovers. Not this PR's problem to solve, just worth knowing the cluster has some accumulated cruft only a superuser can clear.

Did not run the full 98-test suite (your own report already has fixed-order + randomized runs on a clean host) — scoped my own execution to the 4 specific claims plus a mutation probe, which is where independent verification adds the most over re-confirming numbers already measured carefully.

PR #784, opened 2026-08-13
Advertised-DDL extractor: manifest ratchet + F.2 prefix-diff (#753 PR-B)

chrisrogers37:
PR-B of #753 (plan §0.2 `advertised_ddl_replay`) — the extractor, the block-manifest, the F.6-ratchet gate, arm (b)'s F.2 prefix-diff, **and arm (a): the stream replayed under the declared actors**. Arm (a) landed after #778 merged (it needs that PR's actor fixtures); the extractor/manifest/arm(b) half is pure text and independent of both the #772 cluster-lock question and #787 (the D1 ALTER-rights ruling). Only arm (c) (stand-down variants) is deferred, to the M-phase (#790).

## What ships

**`scripts/advertised_ddl.py`** — standalone (stdlib only, zero `src` imports, the runner's rule; a read-only doc tool needs none of the app's env). Extracts the advertised stream `02` §0 defines: every ```sql fenced block from `02` top-to-bottom then `07` in file order, **minus** illustrative examples, **plus** the mechanical expansion of the two §7-DDL policy lists (13 tenant-plane + 2 user-plane = the 15 policies literal-only replay would leave absent, `02` §0 line 35).

**`scripts/advertised_ddl_manifest.json`** — the committed classification. Every one of the 22 blocks (18 normative, 4 illustrative) is keyed by content hash with a class and, for illustrative, a stated reason. "Illustrative" is a reviewed fact on the record, never a regex over the SQL — the failure mode pass 6/R5 named ("verbatim" and "complete" were contradictory under literal-only replay) is closed by *classifying*, not by *guessing*.

**The F.6-ratchet gate** (`tests/scripts/test_advertised_ddl.py`) asserts the manifest and the docs are an exact **bijection, both directions**: a block the manifest does not classify is a defect (a new or edited plan block); a manifest hash no block matches is a defect (a deleted or edited block whose classification is now stale). Any plan-doc edit that touches a block fails the gate until a human re-classifies it. The stream build **fails closed** over an unsatisfied ratchet — it can never silently drop or admit a block.

**Arm (b): the F.2 prefix-diff.** The concatenated target-lineage migration files (above the 051 move — F.2, alex's active lane) are held equal to a **prefix** of the advertised stream at the normalized-statement level (reusing the runner's quote/dollar-aware splitter). F.2 is empty today, so the prefix is **vacuous — and disclosed as such** (the #750 non-vacuity shape), with the mechanism wired and a tripwire test that flips it load-bearing the moment a target-lineage file appears.

**Arm (a): the stream replayed from empty under the declared actors** (`tests/scripts/test_advertised_ddl_replay.py`, DB-backed and actor-coupled, so it sits in its own host-serialized suite). The step-0 bootstrap runs as the owner, the extracted stream replays as `svc_migration` into the empty `public` — the M.3 step-3 shape — and the four §0.2 behavioral assertions hold on the replayed database:
- a seeded intent transition (`scheduled → prompt_pending`) succeeds;
- an unseeded one (`scheduled → posted`) raises through the §4 guard trigger (with a `check_violation` errcode, its deliberate choice);
- `pg_policies` carries the 15 pattern-expanded policies — `p_tenant` on all 13 tenant-plane tables, `p_user_plane` on both user-plane tables;
- the door-owner transient bracket is proven **exercised then closed**: each door-owner system role (`svc_claim`/`svc_clock`/`svc_maintenance`/`svc_membership`) *owns* the SECURITY DEFINER doors the stream created for it (which required the transient `CREATE` on `public`, D40) *and* holds no surviving `CREATE` on `public`.

This is the first end-to-end execution of the extracted stream, so it also proves the plan's target DDL is valid, correctly ordered, and installable by the declared actor — an ordering or privilege defect surfaces here in CI rather than in the cutover window. The intent chain was built against the *live replayed schema* (the deferred owner-member constraint, the §4 actor-GUC guards, the media_source→media_item→intent FK chain), not guessed from the markdown.

## Deferred, named (not silently dropped)

- **Arm (c)** — the stand-down-variant replays — stays with the M-phase (**#790**), per the scoped plan (the variants are untranscribed and replay after window files that do not exist).

## Verification

- **TDD throughout** — every behavior had a watched failing test first (module-missing REDs, then per-feature REDs; the arm-a foundational replay was the moment-of-truth that the extractor produces valid ordered SQL).
- **25 tests**: 20 pure-text (extractor/manifest/ratchet/arm-b) + 5 arm-a (foundational replay + the four assertions). The pure-text suite needs no database; the arm-a suite is real Postgres, host-serialized (#758) and env-isolated (#777).
- **Mutation battery — all bite:** ratchet-ignores-unclassified / -orphaned, illustrative-reason-not-enforced, f2-prefix-accepts-divergence, policy-expansion-drops-tables, and `build_stream` fail-closed (a survivor found in the first pass and fixed with a direct test) all fail a test when mutated; for arm (a), stubbing the stream fails 5/5 assertions (the door-owner test included, after the #784-review fix), dropping a tenant table fails the policy assertion, and swapping the illegal edge for a legal one fails the guard assertion.
- **`/simplify`** applied (shared `_ratchet` core so the docs are parsed once; a stray duplicate import removed). `ruff check` / `ruff format --check` clean.

## Review-round fixes (#784)

- **The door-owner assertion now proves exercised-then-closed, not just absent** (rajan): a pure negative (`has_schema_privilege=False`) passed vacuously when the stream was stubbed out — it could not distinguish "granted then revoked" from "never ran." It now also asserts each door-owner *owns* its SECURITY DEFINER doors, which requires a real replay; the stubbed-stream mutation now fails it, as it should.
- **This body was swept** for the pre-arm(a) state (it had led "all pure text" and listed arm (a) as deferred — both true when written, neither true now).

## NOT verified

- **The M.3 window itself** is unbuilt — arm (a) proves the stream replays and installs under the declared actors; it does not exercise the transform (M.1), the rehearsal (M.2), or the runbook (M.3). Those are the M-phase (**#790**).
- **Arm (b)'s equality is a vacuous prefix today** (F.2 empty); the diff is wired and tripwired but asserts over an empty prefix until a target-lineage file lands.
- **Block classification rests on the doc-map inventory** pinned by the count assertions; a *misclassification* (a normative block wrongly marked illustrative) would surface only downstream — arm (a)'s replay finding an object missing, or arm (b) diverging from an F.2 file — not in this PR's bijection+counts guarantee.

Refs: #753 (scoped plan), #778 (PR-A — arm (a)'s fixtures), #790 (M-phase — arm c), #787 (the 3b/050 ALTER-rights ruling, does not gate this PR), #754 (lineage lane / target `Base`), #750 (non-vacuity shape), plan `02` §0 / §7-DDL / D40.


chrisrogers37 (commented):
**Verdict: Request changes**

**Scope: reviewed the complete head (`95756279`), including arm(a)** — the stream replaying from empty under the declared actors. I extended rather than restarted after arm(a) landed mid-dispatch (ari's call, either was fine): my clone picked up `9575627` via a live `git fetch` before I'd drafted any conclusions, so nothing here reflects a pre-arm(a) partial pass. Arm(a) got the deepest scrutiny of the four points and is finding #1 below — 5 files, +1073/-0, matches the PR's own numbers exactly.

## Two things to fix

**1. `test_no_door_owner_carries_create_on_public` passes without the stream ever executing.**

Mutation-tested it directly: swapped `cur.execute(stream)` for `cur.execute("SELECT 1")` inside `_replay_as_window_actor` — the stream never runs, `public` stays empty — then re-ran all 5 replay tests.

```
4 failed, 1 passed in 16.52s
```

The 4 that failed, failed for the right reason (`UndefinedTable: relation "users" does not exist`, `p_tenant missing on workspace_invitations`) — they genuinely require real execution, exactly as advertised. The one that passed is the door-owner check. It's a negative assertion (`has_schema_privilege(role, 'public', 'CREATE') is False`), and "never granted" and "granted then correctly revoked" are indistinguishable to it — so as written it can't tell "the transient bracket closed behind itself" from "the transient bracket was never opened because nothing ran." In the suite as composed today this isn't exploitable (the other 4 would already be screaming), but in isolation it doesn't prove what its own docstring claims, and that's the one of your four behavioral assertions this review was specifically asked to lean hardest on.

Suggested fix: have this test also assert something that requires the schema to exist first (e.g. `post_intents` present, same as the foundational test), or assert the roles held the grant at some point mid-replay before checking it's gone.

**2. PR body is stale since arm(a) landed.**

Still reads "All pure text (no database, no conftest touch)" as the lede, and lists arm (a) under "Deferred... needs #778's fixtures, which are not on main yet." Both were true when written; neither is true now (#778 merged 01:22Z, arm(a) is in *this* PR and is DB-backed, `@pytest.mark.integration`). Worth a sweep before merge — a future reader trusting the body over the diff gets a materially wrong picture of what shipped.

## Everything else checks out — verified empirically, not read-and-trusted

- **PR-A ancestry**: `git merge-base --is-ancestor 90a084a HEAD` — clean, properly rebased onto post-#778 main. Three-dot diff (`git diff origin/main...HEAD`) matches the PR's own +1073/-0 exactly.
- **Manifest ratchet (scrutiny 1)**: mutated the real plan docs three ways, each reverted via `git checkout --`.
  - ADD a block → `ok=False`, exactly 1 unclassified. ✓
  - EDIT a block's content → old hash orphaned *and* new content unclassified — both directions of the bijection break, as claimed. ✓
  - Pure REORDER (full reversal of all 18 blocks in `02`, content unchanged) → ratchet stays `ok=True`. Not a gap — it's hash-set bijection by design, and neither the PR body nor the module docstring claims to catch reordering (only "added, removed, or edited"). Confirmed `build_stream`'s output *does* change to match the new order (not stale/cached), so a reorder that breaks DDL dependency ordering is arm(a)'s job, not F.6's — and arm(a) would catch it live (a real Postgres error executing out-of-order DDL).
- **Arm(b) tripwire (scrutiny 2)**: confirmed `target_lineage_files()` is genuinely empty today, not just asserted so. Dropped a synthetic `052_*.sql` above the move boundary — the tripwire fires immediately. A matching prefix reports `ok=True, vacuous=False`; a diverging one reports `ok=False` naming the exact statement index. Both directions genuinely evaluated, not skipped.
- **Mutation survivor (scrutiny 3)**: reverted the exact guard `build_stream` fails closed with. Found something worth knowing along the way — the guard's protective value is asymmetric between the two ratchet-violation shapes. Without it, an *unclassified* block still crashes (uncaught `KeyError` — accidental and unhelpful, but it does stop you). An *orphaned* (stale) manifest entry does **not** crash without the guard — `build_stream` silently ignores it and returns a normal-looking 121KB stream. That's the failure mode the guard actually exists to prevent, and it's real. With the guard restored, both shapes raise `AdvertisedDDLError` cleanly. The survivor is genuinely dead.
- **Real execution, not static analysis**: ran both suites for real, alone, under `env -i` (your own methodology). `test_advertised_ddl.py`: 20 passed. `test_advertised_ddl_replay.py`: 5 passed against a live local Postgres (`test_user`/`test_password`, the same convention CI uses). 25/25, matching your report exactly.
- **Host discipline**: checked `tests/scripts/` for concurrent use via a cwd-based `/proc` probe (not pgrep argv) before every run, and waited out a window where alex was actively running `test_rls_harness.py` on the same cluster before touching Postgres myself. `env -i` throughout — hit `SettingsError: TELEGRAM_CHANNEL_ID: missing, ADMIN_TELEGRAM_CHAT_ID: missing` once while probing credentials, field names only, no values leaked — #775's redaction held under a live trigger.
- CI green (Test, Lint, Security Scan, GitGuardian, Changelog Check). No scope creep, no dead code, naming is clear throughout, manifest-load error handling is solid at the boundary (illustrative blocks require a reason at load time, class values validated, duplicate hashes rejected).

Strong work overall — both findings are narrow and fast to close. Re-review once addressed.

chrisrogers37 (commented):
**Verdict: Ship it**

Re-reviewed at `1971d75`. Both fixes verified by re-running my own original checks, not by reading the diff and taking the description's word for it.

**1. Door-owner test — genuinely fixed, confirmed in one command.** Re-applied my exact original mutation (`cur.execute(stream)` → `cur.execute("SELECT 1")`, stream never runs) against the current `test_advertised_ddl_replay.py` and re-ran the full file:

```
5 failed in 20.65s
FAILED …TestTheFourBehavioralAssertions::test_the_door_owner_create_bracket_was_exercised_then_closed
E   AssertionError: svc_claim owns no SECURITY DEFINER door — the stream did not run
    (so there was no bracket to close), or the door-owner model changed
E   assert 0 > 0
```

Previously this was 4 failed / 1 passed — the door-owner test was the one that passed vacuously. Now it's **5 failed / 0 passed**: every behavioral assertion in the file requires real execution. The fix is a stronger shape than what I suggested (schema existence via `post_intents`) — it asserts each door-owner role *owns* the SECURITY DEFINER functions the stream creates for it, which is a positive claim scoped to what this test is actually about, not a borrowed check. Reverted the mutation; confirmed the real (unmutated) suite is back to clean — ran both files together fresh: `26 passed`.

**2. PR body — swept, not just stripped.** Confirmed the false lede is gone (you'd already caught that), and checked what replaced it, not just that it was gone. The "Deferred, named" section now lists only arm (c) — arm (a) correctly removed, not left dangling. The "What ships" section states plainly that arm (a) is DB-backed and host-serialized (directly contradicting the old claim, in the right direction). Grepped the full current body for "deferred" / "pure text" / "no database" / "not on main" looking for a second stale instance the way a partial sweep leaves one — the only surviving hits are correctly *scoped* ones ("the extractor/manifest/arm(b) **half** is pure text," "**20** pure-text… needs no database") that are true statements about a named subset, not the old whole-PR claim. Body now states directly: "This body was swept for the pre-arm(a) state… both true when written, neither true now."

**Bonus, unprompted:** `1971d75` also closes the orphan/unclassified asymmetry I noted as an observation rather than a formal finding — `build_stream` now has a test for the orphaned-manifest-entry direction, not just the unclassified one. Read it against my own original standalone verification of that exact asymmetry; it's the right shape (asserts `AdvertisedDDLError` with "orphaned" in the message, using the existing `extra=[ghost]` fixture pattern).

Nothing half-done. No new issues from the fixes themselves. Both original findings closed on their merits, not on trust.

PR #786, opened 2026-08-13
security(api): key the global rate limit on the proxy-corrected client (#776)

chrisrogers37:
Closes #776.

## What

`SlowAPIMiddleware` was assembled **above** `ProxyHeadersMiddleware`, so the global `30/minute` default was evaluated before the scope's client had been corrected. It keyed on the raw connecting peer.

`SlowAPIMiddleware` runs that default itself, in its own dispatch, for every route carrying no `@limiter.limit` decorator. So behind a single edge, **every tenant on every undecorated route shared one bucket** — a capacity ceiling on the whole service, not a per-abuser control.

The fix is one move: add the limiter *before* `ProxyHeadersMiddleware`, which puts it *below* on the request path. No limiter configuration changed.

## Measured

Real `src.api.app:app`, pinned uvicorn 0.47.0 (this box had 0.44.0 — the XFF trust walk moved across that range, so the pin is load-bearing). 60 requests per arm against the 30/minute default.

| arm | before | after | after + #774 |
|---|---|---|---|
| M1 distinct tenants behind one edge | **30** | **0** | **0** |
| M2 issue #776 repro, verbatim | 30 | 0 | 0 |
| M3 forged *leading* entry, edge appends | 30 | 30 | 30 |
| M4 two XFF headers (#765 shape) | 30 | **0** | **30** |
| M5 control: direct untrusted caller | 30 | 30 | 30 |

Read M1 as the defect: 60 clients making one request each cannot exceed any per-minute limit, so 30 blocked — a contiguous tail — means one shared bucket.

Note every arm reads 30 in the `before` column. That uniformity **is** the signature: the limiter never saw `X-Forwarded-For` at all.

## The trade the issue named was one shape wide, and #774 has closed it

> **Update — rebased onto `main` with #774 merged. The ordering constraint below is satisfied, not outstanding.** Stack is now `CORS → DropAmbiguous → ProxyHeaders → SlowAPI → SecurityHeaders`.

#776 declined to propose a fix because moving the limiter "would also make it forgeable by exactly the routes #765 describes." Measured, that was true of **one** shape only:

- **Forging a leading entry does not open it** (M3, 30 → 30). uvicorn's right-to-left trust walk still attributes to the address the edge appended. The reorder inherits that resistance rather than losing it.
- **The two-header #765 shape did open it** (M4, 30 → 0). `ProxyHeadersMiddleware` reads via `dict(scope["headers"])`, which keeps the last instance — the attacker's.
- **#774 closes exactly that** (M4 back to 30), leaving every arm strictly better than `main`.

### The `xfail` fired, which is how the marker got retired

While #774 was unmerged, the gap was pinned in code rather than prose: `test_a_caller_cannot_mint_buckets_with_repeated_headers` carried `xfail(strict=True)`. On the rebase it came back **`[XPASS(strict)]` — a build failure demanding its own removal**. Marker removed deliberately, on a signal, rather than left behind asserting a gap that had closed.

### That test does *not* cover this change, and now says so

Re-derived on the post-#774 stack by reverting one thing at a time:

| mutation | that test | `test_distinct_clients…` |
|---|---|---|
| revert **only** the #776 ordering | **passes** | **fails** — `30 of 60 blocked, first at request 30, contiguous to the end` |
| revert **only** #774's middleware | **fails** | passes |

So the two-header test is an **inert mutant** for this PR: with the limiter back above `ProxyHeadersMiddleware` the ceiling still bites, just for the wrong reason. It guards #774, not #776. That is now stated in its docstring, because a green test sitting inside a class named for #776 otherwise reads as coverage of #776 — the same mistake that let this bug survive in the first place.

Evidence for *this* change is carried by `test_distinct_clients_behind_the_edge_do_not_share_one_bucket` and `test_the_limiter_is_assembled_below_proxy_headers`, both of which fail on that mutation.

## The property already had a test, and it could never have caught this

`test_rate_limit_buckets_do_not_partition_on_a_forged_value` asserted this against a FastAPI app built inline for the purpose, wired with `ProxyHeadersMiddleware` **outside** the limiter — the correct order, and never the deployed one. No change to `src/api/app.py` could make it fail.

Its `blocked == 30` could not separate the two worlds either: one shared bucket and one correctly-attributed caller both produce exactly 30, for opposite reasons. Even pointed at the real app it would have passed and told you nothing.

Removed rather than repaired. The replacement drives the assembled app and varies the two candidate keys **independently**, which is the only thing that separates them.

## Verification

- **Watched it fail first.** On unpatched `app.py`: `30 of 60 blocked, first at request 30, contiguous to the end`, plus the structural assertion at `assert 2 < 1`. After: 4 passed, 1 xfailed.
- **No regressions.** Re-baselined after the rebase on a clean `git worktree` of the current `origin/main` (with #774): **182 passed**. This branch: **186 passed**, zero xfail remaining. Reconciles exactly — 182 − 1 removed + 5 added. (Pre-rebase figures were 173 → 176 against the older `main`.)
- **Rebase kept #774 byte-identical.** Both sides touched the same block and the same test file. `TestForwardedForAmbiguity` is verbatim from `main` (diffed line-for-line, 181/181), and `DropAmbiguousForwardedForMiddleware` is untouched — the `app.py` diff is only the limiter move.
- **0/10 on a post-rebase stress of `tests/src/api/`**, on top of the 0/22 below.
- `ruff check src/ cli/` and `ruff format src/ cli/ --check` both clean (rc 0, read directly, not through a pipe).

### Two things I got wrong on the way, since both are the more general trap

**The new tests were vacuous at first.** `tests/src/api/conftest.py` has an autouse fixture setting `limiter.enabled = False` for every API test. Under it, "distinct clients do not share a bucket" passes against an app with **no limiter running at all**. Only watching the tests fail first caught it — the failure pattern came out inverted. The helper now asserts the limiter is live before measuring, so it cannot go quiet again.

**A flake I could not explain, and what I did about it.** `test_distinct_clients` failed roughly one full-suite run in twelve, never when the file ran alone (0/25, 0/14, and passing in isolation every time). I could not establish the mechanism. Rather than diagnose it by assertion, I removed its precondition: every drive now allocates a **fresh key space** from a counter instead of a per-test constant, so no key is ever reused within a session. The storage reset stays as well — one of those is a fix for a cause I could not name, and it should not also be the only thing standing between a shared singleton and a false security result. The assertion now reports the blocked **index pattern**, so a contiguous tail (real defect) and a scattered few (pollution) name themselves.

**How much that is worth, stated rather than implied:** 0 failures in 22 consecutive full-suite runs afterwards. That is *consistent with* a fix and does not establish one — at the prior rate, 22 clean runs come up by chance roughly 15% of the time. If it recurs, the assertion message now says which of the two modes it is, which is the thing I actually lacked while chasing it.

I also miscounted that flake as 12/12 at one point: my filter matched the word "error" in the suite's pre-existing `ResourceWarning` noise, so the number measured nothing. And one later "failure" was my own doing — I stress-ran the file while editing it, catching `_drive` mid-change (`assert [] == 0`).

## Not verified

- **No real edge.** Every arm is `TestClient` with a synthesised peer. Whether Railway's edge appends to or overwrites a caller-supplied `X-Forwarded-For` is untested here and is the open question #774 already flags. It does not change M1 (the defect) but it does bound M3.
- **CI checks may not appear.** `pull_request` events are not creating runs right now. The branch is named `feature/*` so the existing `push` trigger (`[main, develop, 'feature/*']`) covers it — verified in `.github/workflows/ci.yml`, not assumed. I have not re-pushed to chase a run.
- **Integration tests did not run** — no PostgreSQL on this host, so `tests/integration/` and the DB-backed suites were skipped, honestly, by the #758 probe. This change touches no database path.
- The pre-existing `rc=1`-with-all-passing on `tests/src/api/` (unraisable `ResourceWarning`s) is **on `main` too** and is not addressed here.

https://claude.ai/code/session_01JJpXAhbmfDQXChyqfUuao3


chrisrogers37 (commented):
**Verdict: Ship it — behind #774, not before it.** The code is correct and thoroughly verified; the merge order is a hard precondition, not a nice-to-have. Do not merge either PR until both are sequenced (ari has recorded this on both PRs already).

All 3 asks verified by running the code and mutating it, not by reading virgil's (already very thorough) write-up and nodding along.

**1. Is "#774 closes it" actually true, or does #786 open something #774 doesn't cover?** Independently enumerated every consumer of client-IP attribution in the codebase (`grep -rn "request\.client\|get_remote_address\|record_failure"` across `src/`) rather than trusting the claim. There are exactly three: `SlowAPIMiddleware`'s own default-limit dispatch (the one thing #786 moves), the 10 `@limiter.limit()` decorator sites (`oauth.py` ×5, `onboarding/settings.py` ×5), and `auth_monitor.record_failure`'s 3 call sites via `onboarding/helpers.py`. The latter two read `request.client.host` from inside route-handler code, which is architecturally always downstream of every middleware in the ASGI stack regardless of middleware-to-middleware ordering — so they were never exposed by #776 and aren't touched by this reorder, independent of virgil's or the issue's say-so. That leaves exactly the one shape virgil measured (M4, two-header XFF) as the actual opened surface, and #774's own 8-scenario suite plus its end-to-end 30-blocked test close precisely that. I didn't find a fourth shape neither PR's tests cover. **Current state, rechecked just now: #774 is `MERGEABLE` (alex's rebase landed, ~01:27Z) and #786 is `MERGEABLE` — the conflict ari flagged is resolved, both are individually clean, only the sequencing remains.**

**2. Does the reorder break anything that depended on the old order?** Enumerated rather than assumed. #786's diff only swaps the relative order of `SlowAPIMiddleware` and `ProxyHeadersMiddleware` — `SecurityHeadersMiddleware` (innermost either way; read its `dispatch()` directly, it never touches `request.client`, only sets response headers off `request.url.path`) and `CORSMiddleware` (outermost either way, reads `Origin` not client IP) keep the same relative position to that pair before and after. So the full 4-middleware stack has exactly one order-sensitive relationship, and it's the one this PR changes on purpose. Nothing else in `src/` reads `request.client`/`get_remote_address` outside the three sites in point 1.

**3. Would a do-nothing change also show 30-to-0?** No — mutation-tested it directly rather than trusting the number. Took #786's branch, surgically reverted *only* the middleware ordering back to pre-#786 (swapped the two blocks back, touching nothing else — isolates ordering as the sole mutated variable) while keeping virgil's actual test file, and reran `TestGlobalRateLimitKeysOnTheCorrectedClient`. A no-op does **not** produce 0-blocked: `test_distinct_clients_behind_the_edge_do_not_share_one_bucket` fails with the exact contiguous-tail signature (`[30, 31, ..., 59]` blocked) virgil's own write-up describes as the defect, and `test_the_limiter_is_assembled_below_proxy_headers` fails naming the actual wrong order directly. 3 of 5 tests flip red under the revert; the other 2 stay green for a legitimate, non-decoy reason — their assertions are "caller cannot mint fresh buckets," which holds trivially whenever the limiter isn't reading `X-Forwarded-For` at all (the *other*, already-known failure mode), including the `xfail(strict=True)` test, which XPASSes under the revert for that same reason rather than contradicting anything. That's consistent with virgil's own "every arm reads 30 in the before column — the limiter never saw X-Forwarded-For at all" framing, not a gap in it. Restored the file clean afterward; full `test_security_hardening.py` back to 23 passed, 1 xfailed, matching virgil's claim exactly.

**On the write-up itself:** virgil's own PR body already did most of the hard part correctly — the M1–M5 table, the xfail-that-retires-itself mechanism, catching their own vacuous-test and flake issues before shipping, the honest "22 clean runs is ~85% confidence, not proof" framing. Nothing here contradicts it; this is independent confirmation of a genuinely well-verified PR, not a correction of one.

Not merging. Reporting per the dispatch.

PR #788, opened 2026-08-13
fix(scheduler): log auto-approve notification failures instead of swallowing them (#782)

chrisrogers37:
Closes #782.

alex found this by reading `scheduler_loop.py` against `ChatMigratedError.parse_pair`'s docstring, and said plainly it was **not verified by running**. So the first work was reproducing it — the fix itself is one line.

## Proven before, gone after

Same four tests, same command, against unmodified `main` and then against the fix:

```
BEFORE (main):   2 failed, 2 passed
  test_a_failed_notification_is_logged_rather_than_swallowed
    AssertionError: a failed auto-approve notification left no warning: []
  test_a_migrated_tenant_leaves_a_RECOVERABLE_pair
    AssertionError: the new chat id is not recoverable from the log line: []
    assert (-100123, -100999) in []

AFTER  (fix):    4 passed
```

`[]` is the whole bug: **zero log lines of any kind**. The before-run is the mutation proof — I did not need to gut anything, because the unfixed code *is* the mutant.

## Why the silence is the defect, not the failure

The post genuinely succeeded; only the courtesy confirmation was lost, and this send already sits inside `_scheduler_tick`'s per-chat `try`, so it ends nothing. Nothing downstream looks wrong. **A tenant simply stops hearing from the bot and no signal exists anywhere.**

That is #758's shape one surface over: a failure that degrades into silence becomes indistinguishable from the thing never having happened.

## Both sides of the distinction are pinned

This is the part #764 established, and it is why two of the four tests assert *absence*:

| case | expected |
|---|---|
| notify raised | **logged**, naming the tenant |
| not auto-approved → no notify attempted | **silent** |
| no Telegram service → no notify attempted | **silent** |
| notify raised `ChatMigrated` | logged **and machine-recoverable** |

A fix that logged unconditionally would satisfy the first row and destroy the signal. The two silent rows pass before *and* after — they are the controls that make the first row mean something.

## The #743 recovery-corpus hole, closed and asserted through the real parser

`ChatMigratedError.parse_pair`'s docstring names this exact site as one of three shapes that produce no row — the reason "not in this corpus" has to be read as *unknown* rather than *did not migrate*. A migrated tenant raising here lost the new chat id outright, which is the single fact a recovery pass needs.

The test round-trips the emitted line back through the real `ChatMigratedError.parse_pair` rather than matching prose:

```python
pairs = [ChatMigratedError.parse_pair(ln) for ln in alert_lines if "Auto-approve" in ln]
assert (CHAT, new_id) in pairs
```

So a future reword that stopped being machine-recoverable fails this test, which a substring assertion would not catch.

## Two things worth knowing beyond the fix

**`caplog` cannot see these logs.** `src/utils/logger.py` sets `propagate = False`, so a test built on `caplog` would observe nothing and read as *still swallowed* even against a correct fix. The repo uses `patch("<module>.logger")` everywhere and `caplog` in **zero** places; these tests follow that. Worth stating because the failure mode is a silently vacuous test.

**`_log_alert_failure` had no test coverage at all.** It shipped in #781 with two callers (`Pool depletion`, `Drive token health`) and no test references it — `grep -rn '_log_alert_failure' tests/` returned nothing before this PR. These tests give the helper its first coverage, via this third site and specifically through its `ChatMigrated` rendering path. The other two call sites remain uncovered; that is not this PR's scope and I have not bundled it.

## Scope note answered

The issue asked whether a failed notify should also stop counting the post as delivered. **It does not, and that is deliberate** — `session_state.posts_sent += 1` runs before the notify block and the post really was sent, so the failure is logged rather than re-classified. Verified by position in the source, not assumed. Agreeing with alex's read, and flagging it as the product call he correctly declined to make inside the fix.

## Verification

- 4 new tests: 2 red before / green after, 2 controls green throughout.
- **68 passed** across `test_scheduler_loop.py`, `test_scheduler_queue_reliability.py`, `test_reapers_honor_states.py`, `test_main_scheduler_loop.py`.
- Zero bare `except Exception: pass` remain in `scheduler_loop.py`.
- Run under `env -i` per the fleet practice; no `tests/scripts` involvement, so no role provisioning and no host serialisation needed.


chrisrogers37 (commented):
**Verdict: Ship it**

Reproduced independently, not just re-read. Swapped `scheduler_loop.py` back to the pre-fix (`be77376`) version underneath the PR's new test file and reran `TestAutoApproveNotificationIsObservable`: **2 failed / 2 passed**, identical `[]` symptom, identical assertion text to what's quoted in the PR body. Restored the fix, reran — 4/4 green. Real mutation test, not a re-statement of the PR's own numbers.

Checked ari's three scrutiny points directly against source:

1. **Never-happened vs happened-and-lost** — confirmed via the mutant run itself: `test_a_notification_never_attempted_is_not_logged` and `test_no_telegram_service_is_not_logged_either` PASSED under *both* pre-fix and post-fix source. The guard (`auto_approved and telegram_service`) short-circuits before the `try` even opens, so no path was made to over-log. The fix adds a signal only where one was lost.
2. **Reaches somewhere real** — `_log_alert_failure` isn't new; it already has two live callers (Pool depletion, Drive token health — `scheduler_loop.py:293,338`, both pre-existing, untouched by this diff). This PR brings the third named caller up to the same already-trusted `logger.warning` channel, not a new/weaker one.
3. **Test count 2→4, not weaker versions** — the two "new" tests are exactly the two that flipped in my mutant run; the two that didn't flip are true negative controls, not padding.

Checked independently of the PR's own claims:
- Issue #782's body (pre-dates this PR) already names `_log_alert_failure` "(added in #781)" and this exact site as its third caller — corroboration from a source that isn't the PR itself.
- `except Exception:$` still has one hit at scheduler_loop.py:93, but it's `rollback(); raise` — not a swallow, unrelated, correctly out of scope. Zero bare `except Exception: pass` remain, as claimed.
- Full claimed regression set (`test_scheduler_loop.py` + `test_scheduler_queue_reliability.py` + `test_reapers_honor_states.py` + `test_main_scheduler_loop.py`): **68 passed**, rc=0. Exact match.
- Two-dot diff vs `origin/main`: only the 3 claimed files changed, branch forks cleanly off main's current tip — no stale-base cargo, no scope creep.
- All 6 CI checks green.

Nothing to add beyond what's here. Reproduced before touching anything, correct distinction pinned on both sides, closes a corpus hole named in someone else's docstring rather than invented after the fact.

PR #792, opened 2026-08-13
feat(m1): the FC-8 zero-row gate, executable (#790 M.1)

chrisrogers37:
Toward #790 **M.1**. Lands the piece of M.1 that is buildable today, and reports precisely why the rest is not.

## What ships

`scripts/fc8_gate.py` — the **FC-8 zero-row gate** (`04` L87), which M.1 names and which had no implementation. It is a window-prep **halt**, not a report: live local/upload-origin `media_items` = 0, and `posting_history` rows resolving to such media = 0. Exit `0` CLEAR / `1` HALT / `2` ERROR.

**Why this piece and not the transform files.** M.1's per-column mapping and transform `INSERT`s write into F.2's target tables. Those do not exist: `src/models/target/` is an empty `Base` whose own docstring says *"zero target tables exist today; F.2.2 onward registers them here"*, and no shipped migration creates a target table. This gate reads **legacy tables only**, so it is buildable and testable now rather than behind F.2. See "What M.1 still needs" below — I have not narrowed the scope silently.

## Three properties held deliberately

**A gate that cannot answer does not report CLEAR.** A missing table or unreachable DSN exits `ERROR`, not zero — #758's false-PASS class. Mutation-proven: swallowing the error and returning the safe answer reddens exactly the test that names it, and nothing else.

```
MUTATED: missing table -> CLEAR instead of ERROR
FAILED TestCannotAnswerIsNotClear::test_a_missing_table_is_error_not_clear
1 failed, 13 passed
```

<sub>Verbatim capture from when the mutation was run, at the 14-test state — left unedited on purpose. The suite is 20 tests now; the `13 passed` here is history, not a current count.</sub>

**The two counts are different predicates on purpose.** Count 2 does not filter on `is_active` — a posted row whose media was later deactivated still has no target destination for that media. There is a test that goes red if the two queries are ever "tidied" into one.

**It discloses what it has no authority to rule.** Covered next, because it is the finding rather than a feature.

## The finding: a fourth source_type nobody has classified

FC-8 names `local` and `upload`. The legacy corpus also carries **`instagram_backfill`** (in use in `src/`), and the plan classifies it nowhere — I grepped all eight plan documents.

Meanwhile the target's media-source CHECK is closed at exactly one provider (`02` §2):

```sql
provider TEXT NOT NULL CONSTRAINT ck_sources_provider CHECK (provider IN ('gdrive'))
-- v1 ships exactly 'gdrive'
```

So on the printed target schema, an `instagram_backfill` row has **no destination either** — the same condition FC-8 exists to catch — while falling outside the two counts that halt. Window prep would proceed.

**I have not resolved this, and the gate does not.** It halts on exactly FC-8's two counts and reports any other origin as `UNCLASSIFIED`, so the question is visible rather than silently folded into the safe side by a `NOT IN ('local','upload')` that would read as deliberate. Which way it should go is an owner ruling.

Same shape for liveness: `media_items.is_active` is **nullable**, so "live" has three states. `IS NOT FALSE` would quietly call NULL live; `IS TRUE` alone would quietly call it dead. NULL rows are excluded from the live count **and** counted in their own bucket.

## Verification

**20 tests, all green** (was 14 before the disclosure-bucket fix; the CHANGELOG was swept to 20 and this line was not — corrected before merge), run against `replayed_db` — the *real* replayed legacy lineage rather than a hand-built fixture, which is what caught a `NOT NULL` column that reading the model had missed (`queue_deleted_at`). `env -i`, host taken one at a time via a cwd-based probe.

Includes a positive control (`google_drive` does **not** halt), without which a gate that halted on everything would pass every other count test.

## What M.1 still needs — stated, not implied

Not built here, and none of it silently dropped:

- **Per-column mapping + transform files** for the 14 legacy tables — blocked on F.2 registering target models. `02` §9 is the disposition index and says the column-level tables live in M.1; writing them against a schema with no machine-readable form would create the second hand-maintained enumeration the F.2 fork ruling (a) explicitly refused.
- **Owner derivation + quarantine list**, membership role mapping, FC-7.2 accounts/sources config-only, the `fn_safe_tz` timezone rule, history→terminal-intents with the `service_runs` ±6h attribution timeline, and the FC-7.4 queue clause. The *derivations* in these are legacy-side and testable now even though their `INSERT`s are not — that is the natural next increment and the split is clean.
- **The quarantine procedure document** (part of M.1's gate).

M.1's gate — "every transform file carries postconditions; the reconciliation queries are review-approved; the quarantine procedure is written" — is therefore **not** met by this PR, and #790 stays open. This is one named bullet of M.1's contents, landed whole.


chrisrogers37 (commented):
**Verdict: Request changes**

Reviewed `a660b3a` against fresh `origin/main` (three-dot diff matches the PR's own +394/-0 exactly — `CHANGELOG.md`+9, `fc8_gate.py`+205, `test_fc8_gate.py`+180). All four things asked of this review, empirically, not read-and-trusted.

## The one thing to fix

**The two human-facing disclosure messages in the default (non-`--json`) text output are untested — correct today, unguarded against regression.**

Mutated `render()` twice, independently: disabled the `UNCLASSIFIED source types are present…` block, then separately the `is_active IS NULL` block. **Both times, all 14 tests still passed.** The JSON path is covered (`test_json_mode_carries_the_disclosure_not_only_the_verdict`), but nothing captures plain-text stdout and asserts on it — so a future edit to `render()` (the kind of thing that happens when this file gets touched again for M.1's next piece) could silently drop either explanation and the suite would stay green throughout.

This is the direct answer to "does UNCLASSIFIED genuinely surface or is it cosmetic": **today it's genuine** — I confirmed with a real subprocess run against a freshly replayed DB, unmutated:

```
media_items by source_type (disclosure, not a verdict):
  instagram_backfill     live=1      total=1  <-- UNCLASSIFIED by the plan
  local                  live=0      total=1

UNCLASSIFIED source types are present. FC-8 names only local/upload,
and the target's media-source CHECK is closed at 'gdrive' (`02` §2), so
...
1 media_items have is_active IS NULL — neither live nor dead.
```

But "genuine today" and "guarded" are different claims, and only the second one survives someone else editing this file next month. The fix is small: two tests capturing `capsys` on a plain (non-JSON) `main()` call, asserting `"UNCLASSIFIED"` and `"is_active IS NULL"` appear in stdout for the two respective cases. Everything else below is sound — this is narrow and fast to close.

## Ari's three scrutiny points — verified, not read

**1. Vacuous-pass risk (the #758 shape) — sound.** Reproduced the mutation you quoted in the PR body myself: swapped `census(conn)`'s error handling in `main()` to swallow `psycopg2.Error` and fall back to a zero `Census` (fake CLEAR) instead of returning `ERROR`.

```
1 failed, 13 passed
FAILED tests/scripts/test_fc8_gate.py::TestCannotAnswerIsNotClear::test_a_missing_table_is_error_not_clear
E   AssertionError: a gate that could not answer must not exit CLEAR
E   assert 0 == 2
```

Exactly the shape claimed. Traced why it's structurally sound rather than accidentally sound: `_rows()` has no internal try/except, so *any* `psycopg2.Error` from *any* of the four queries in `census()` — missing table, missing column, connection loss mid-query — propagates unmodified to `main()`'s single broad `except psycopg2.Error`, which is the only place that decides CLEAR vs ERROR. There's no per-query seam where a future "helpful" error handler could reintroduce a silent zero.

**2. Two counts, genuinely independent — sound.** Mutated count 2 to also filter `is_active IS TRUE` (the exact "tidied into one" the code comment names as the risk):

```
1 failed, 13 passed
FAILED …TestTheTwoNamedCounts::test_an_inactive_local_row_is_not_live_but_its_history_still_counts
E   assert 0 == 1
```

Caught cleanly, by exactly the test built for it.

**3. Nonzero actually halts, constructed independently.** Didn't rely on `test_halt_exits_one` — built a fresh scratch DB via the real legacy-lineage replay (`psql_apply` + `migration_files`, same as `replayed_template`), ran `scripts/fc8_gate.py` as an actual subprocess (not `main()` in-process) before and after inserting one real `local`-origin row:

```
=== empty corpus: exit=0 ===
VERDICT: CLEAR
=== after real nonzero insert: exit=1 ===
VERDICT: HALT — window prep stops
  Routes to the owner with the counts. Resolutions per `04` L87: …
```

It stops. Not a log line — an exit code a caller would actually observe.

## Policy grounding — nothing invented, checked against source

- The two-count / zero-threshold / halt-and-route-to-owner language is **verbatim** from `04-execution-sequence.md` L87 (read it directly, not via the epic's paraphrase) — down to the exact resolution wording ("migrate the files to Drive first, or an explicit accept-loss list… weakening the `02` NOT NULL chain is recorded as rejected") that `render()` reproduces on HALT.
- `is_active IS TRUE` for "live" isn't invented — it's the **dominant existing convention**: `media_repository.py` uses `.is_(True)` nine separate times for "active" media, `== True`/`== False` elsewhere, and *never* an `IS NOT FALSE`/NULL-inclusive form anywhere in the app. A NULL row is already invisible to every real query path today; FC-8's separate bucket matches that rather than inventing new semantics.
- `google_drive`'s exemption from `UNCLASSIFIED` is grounded in the target's own `CHECK (provider IN ('gdrive'))` — the only provider with a real destination.
- The 0/1/2 exit-code shape matches `migration_runner.py`'s own `return 2` convention — a sibling pattern, not a new one.
- `instagram_backfill` and null-`is_active`: genuinely unruled by the plan (confirmed — grepped `04` L87 myself, it names only `local`/`upload`), and the PR does not resolve them either. It reports and refuses to guess, which is what "no invented policy" should look like — not that the gap goes away, but that it doesn't get silently decided as code. #793 is filed for it, so the ruling has a home.

Strong PR — the primary risk you named was the right one to name, and it holds up under direct mutation. The one gap is small and specific.

chrisrogers37 (commented):
**Approve**

Verified at HEAD, not from the reply. Fresh fetch, checked out detached at `1957633` after confirming it matches `gh pr view`'s `headRefOid` — no stale-checkout risk from the crash.

20 tests, real `replayed_db`, zero skips, genuine green baseline.

Then four independent mutations, one at a time, `git checkout --` between each:

| Mutation | Kills | Result |
|---|---|---|
| Neuter the UNCLASSIFIED reason block only | `test_the_unclassified_block_keeps_the_reason_not_just_the_label` | exactly that one, 19 passed |
| Neuter the inline row marker only | `test_an_unclassified_origin_is_named_and_marked_in_the_text` | exactly that one, 19 passed |
| Neuter the null_active block | both claimed tests, incl. the DB-backed wiring test (`test_the_default_path_prints_the_text_disclosure_to_stdout` — the one no pure test can reach) | exactly those two, 18 passed |
| Force both blocks to print unconditionally (the strawman the negative control's own docstring names) | `test_a_clean_census_prints_neither_disclosure` | exactly that one, 19 passed |

Four mutations, four exact kills — no over-kill, no under-kill, negative control included. Not a decoy; the gap from the original review is closed.

CI reconfirmed independently via GraphQL on the same SHA: Lint/Security Scan/Test/Changelog Check all `SUCCESS` with real step counts (9/10/15/5). The two `steps=0` entries are GitGuardian and Vercel's App-based checks — confirmed absent from `ci.yml`, so they're not workflow jobs and never report steps; expected shape, not a silent no-run.

Two non-blocking asides:
- CHANGELOG.md was swept correctly (14→20, no stale copies anywhere in-repo), but the PR body's own Verification section still reads "14 tests, all green" — a stale copy in a sibling document, not shipped code.
- The halt-routing text (render(), the block under `if c.halts:`) still has no content assertion. Different category from what this fix closed — losing it wouldn't hide anything the exit code and VERDICT line don't already say — so not blocking, just naming it.

PR #796, opened 2026-08-13
security(gdrive): a named tenant never resolves to the service account (#627)

chrisrogers37:
Closes #627.

## The defect

A broad `except Exception` around the per-tenant Google Drive OAuth lookup fell through to `get_provider()` — the deployment-wide service account. The operator and a tenant are different parties in this product, so that is a cross-tenant access path, and it was **silent**: the auth error that caused it was downgraded to a warning on the way past.

**Two distinct boundaries crossed, not one.**

1. The service account's **credentials** stood in for the tenant's.
2. Because `get_provider()` reads `root_folder_id` from the service account's *own stored metadata* when passed none, a tenant with no configured root was handed a provider **rooted in the operator's Drive folder** — whose files would then be indexed under that tenant.

The second is the sharper one, and the credential assertions do not catch it, so it is asserted separately.

It also **masked the cause**: the tenant's error was swallowed and the service account's surfaced in its place, telling a tenant to run an operator-only CLI command for a credential they do not own. `get_provider_for_media_item`'s existing regression test already described this exact chain in its docstring — the fallback that produced it was still live.

## Was the broad `except` load-bearing?

Checked rather than narrowed blind. **It does not have a legitimate case.**

The only candidate was "this tenant has not connected their own Drive". But onboarding refuses to store a media root until that tenant's own OAuth has proven access to the folder — `/api/onboarding/media-folder` validates through `get_provider_for_chat` and returns 400 on failure. So "named tenant, no working OAuth" is a **broken state to surface**, not a configuration to substitute around. Nothing anywhere asserted the fallback.

The untenanted path — no `telegram_chat_id` claimed — still uses the service account, and crosses nothing. That is preserved and directly tested.

## `media_sync.py` was the same defect wearing a different hat

`telegram_chat_id or settings.TELEGRAM_CHANNEL_ID` named a tenant the sync is not, so an untenanted sync resolved — and later refreshed — whichever tenant that env var happens to point at's OAuth credentials, then used them against *this* sync's folder. The tenant is now passed through as-is, including `None`.

## Two existing tests were asserting the crossing as correct

`test_create_provider_google_drive` and `test_sync_uses_overridden_source_type` both pinned `telegram_chat_id=-100123456789` on a caller that named no tenant. Both were rewritten to assert the substitution does *not* happen, against that same distinctive value so a reintroduction fails by name rather than by a bare `None` check.

## Verification

The defect is a silent substitution, so a test that cannot detect silence is worthless here. Every new assertion is on the negative — `assert_not_called` on the service-account door — and each was mutation-proven to fail when the crossing is restored.

**Mutation 1 — restore the broad `except`** (`+25/-3` in `factory.py`, verified applied):

```
FAILED ...::test_a_tenant_oauth_failure_does_not_resolve_to_the_service_account
FAILED ...::test_the_tenants_auth_error_is_not_masked_by_the_service_accounts
FAILED ...::test_a_non_auth_failure_also_does_not_resolve_to_the_service_account
FAILED ...::test_a_tenant_without_a_configured_root_is_not_given_the_operators
4 failed, 12 passed
```

Exactly the four crossing assertions red, **both controls green**. The mutant's own captured log shows the crossing happening: `User OAuth failed for chat -100123: No root_folder_id configured..., falling back`.

**Mutation 2 — restore the `TELEGRAM_CHANNEL_ID` substitution** (`+9/-2` in `media_sync.py`):

```
FAILED ...::test_an_untenanted_sync_does_not_borrow_the_channel_tenants_identity
FAILED ...::test_sync_uses_overridden_source_type
2 failed, 36 passed
```

The controls are load-bearing, not decoration: an untenanted caller *still* reaching the service account is what proves the negative assertions are green because the door is shut to tenants, rather than because the fixture never opens it.

**Suite:** `2086 passed, 8 skipped` (`tests/src -m unit`), rc 0, zero failures. Affected files ruff `check` + `format --check` clean. No integration test touches this path (`tests/integration/` has no reference to `MediaSourceFactory`, `_create_provider`, or `get_provider_for_chat`).

## Observed, not applied

- **Pre-existing on `main`, unrelated to this PR:** `ruff check src/ tests/` fails with `F811 Redefinition of unused test_check_response_errors_revocation_subcode_458` at `tests/src/services/test_instagram_api.py:940`. Confirmed present on `main` (`git show main:...` + isolated ruff run), untouched here. Flagging rather than fixing — different lane.
- **Reuse (not applied):** the `gdrive_registered` fixture writes `MediaSourceFactory._providers` directly rather than going through the public `register_provider()` classmethod used by `test_register_provider` in the same file.
- **Simplification (not applied, deliberately):** the fixture was flagged as dead on the grounds that `create()` self-registers `google_drive`. That registration is wrapped in `except ImportError`, so it only holds where `google-api-python-client` is installed. Removing the fixture would make the suite pass *without testing anything* on a host lacking that dependency — failing open, which is the failure shape this PR exists to close. The fixture's docstring was corrected to state the condition precisely; the fixture stays.


chrisrogers37 (commented):
**Approve**

Reproduced both mutation proofs myself in an isolated worktree at the PR's exact head SHA (d9bfe83) — not just read them.

**Item 1 — both sites, not one.** `factory.py:86-99` (broad `except Exception` → service-account fallback) and `media_sync.py`'s `telegram_chat_id or settings.TELEGRAM_CHANNEL_ID` are both removed. I expected the second to be the miss going in; it isn't — confirmed by diff and by mutation.

**Item 2 — mutation proof, independently reproduced:**
- Reverse-applied just the `factory.py` hunk in a worktree (confirmed byte-identical to `main` after), ran the new test class: exactly the 4 named tests fail (`test_a_tenant_oauth_failure_does_not_resolve_to_the_service_account`, `test_the_tenants_auth_error_is_not_masked_by_the_service_accounts`, `test_a_non_auth_failure_also_does_not_resolve_to_the_service_account`, `test_a_tenant_without_a_configured_root_is_not_given_the_operators`), both controls pass. The captured log line matches the PR description verbatim: `User OAuth failed for chat -100123: No root_folder_id configured..., falling back to service account`.
- Same for `media_sync.py`: reverse-applied that hunk alone (other file back at PR head), ran the full `test_media_sync.py` (38 tests): exactly the 2 named tests fail (`test_an_untenanted_sync_does_not_borrow_the_channel_tenants_identity`, `test_sync_uses_overridden_source_type`), 36 pass — matches the claimed `2 failed, 36 passed` exactly.
- Both mutations run in isolation, one file mutated at a time, so neither result is contaminated by the other fix.

**Item 3 — fail-closed.** `factory.py`'s tenant branch no longer catches at all — the exception is the tenant's own and propagates. Checked one representative real caller, `scheduler.py:822` (the unattended auto-post path): it's `try/finally`, not `try/except`, and the only nested `except Exception: pass` scopes to Cloudinary cleanup, not the provider call — the auth error still propagates past it. Did **not** walk the other two callers (`telegram_autopost.py:333`, `telegram_notification.py:147`) individually — stated bound, not checked.

**CI, per-SHA:** all 6 checks (Test, Lint, Security Scan, GitGuardian, Changelog Check, Vercel Preview Comments) `success` on `d9bfe83db0e357cab0973100f3284c5160f66e0d`, the current head — queried directly against that SHA via the check-runs API, not the checks tab.

**Scope:** clean — 5 files (2 source, 2 test, CHANGELOG), nothing in migration/#630/Phase M.

**What I did not check** (time-boxed): did not re-run the full suite myself, only the tests exercised by the two mutations (~43 total across both runs); did not independently verify the ruff/format or no-integration-test-touches-this-path claims; only 1 of 3 non-factory callers checked for a re-introduced catch.

Comment-only — same-identity blocks native Approve on this account (confirmed: `Can not approve your own pull request`).

PR #738, opened 2026-08-07
docs: describe storydump as the hosted product it is (repo-wide sweep)

chrisrogers37:
The product owner ruled 2026-08-07: **"Product I host and charge for!"**

`README.md` and `CLAUDE.md` both opened by calling storydump "self-hosted". That is now wrong, and it is the kind of wrong that costs real time: it made the monetization epic (#661–665) look like it contradicted the architecture.

**Credit to `mason`**, who found this during monetization pre-scope and was right to treat the contradiction as load-bearing rather than cosmetic. It took one question to the product owner to resolve, and the answer was that the *docs* were stale, not the plan.

## Why the plan was never the problem

The consolidated design plan (#731) has assumed a hosted, multi-tenant product all along — by mechanism, not by statement, which is why grepping for the word would not have found it:

- `06` §5: **one operator**, and *every operator action on tenant data is audit-visible to that tenant's admins*. That mechanism is pointless when the operator and the tenant are the same person.
- The hosting substrate (Neon, Railway, Vercel, one app-level Cloudinary) is named **as given**, while the plan deliberately builds pluggable seams for three other things — channels, media sources, auth/email providers. It declined a seam here on purpose.
- **0.1 App Review is a single submission and the program's long pole.** A self-hosted product structurally cannot share one Meta app.
- **T3: "workspaces are rows, not deploys"**, with FC-0 sizing one deployment at ~5,000 workspaces.

So this PR corrects the project's self-description to match what it was already being built as.

## The third edit, and why it is here

Fixing only the two sentences would have left the same falsified claim alive on the same page, expressed structurally: README's **"Quick Start → Installation → clone the repo, make a venv, add your credentials"** reads as instructions for standing up your own instance. It is retitled **"Local Development Setup"** and now says plainly that it is not a deployment guide.

That one is a judgement call rather than a strict consequence of the ruling — **drop it if you disagree**; the two sentence fixes stand on their own.

## Scope

Deliberately **not** here: **FC-9**, which records the ruling normatively, lands on #731 where the plan lives. A plan PR that also edits the project's identity docs would be two changes wearing one number.

Also noted, not fixed: `README.md:277` and `documentation/README.md:102` still advertise a "Quick Start Guide — get running in 10 minutes", and `documentation/guides/quickstart.md` behind them is written for a self-installer. That is a larger docs pass than this PR should carry, and it deserves its own review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #798, opened 2026-08-13
docs(0.4): verify Meta platform facts against primary docs; fix #734's actual defect (#745)

chrisrogers37:
Implements plan **§0.4** (#745) with **#734** folded in — one unit, since both edit the same operator-facing docs and #734's number is exactly what §0.4 re-confirms.

Documentation only: no code, no schema, no production access.

## How this was verified

Every claim below was checked against Meta's **live published documentation on 2026-08-13**, fetched raw and de-tagged, with the wording quoted verbatim. No summarizer sits between Meta's text and the quotes — for a figure that has already gone stale once, a paraphrase is the failure mode being guarded against. Sources and full evidence: [`0.4-meta-primary-doc-verification.md`](documentation/planning/2026-08-02-consolidated-design-plan/0.4-meta-primary-doc-verification.md).

| # | Item | Verdict |
|---|---|---|
| 1 | Publish cap 100 / rolling 24 h | **Confirmed** — and Meta contradicts itself elsewhere |
| 2 | 200 calls/user/hr | **Corrected** — not a per-user ceiling |
| 3 | IG-Login scopes + refresh semantics | **Confirmed**, min-age 24 h |
| 4 | Usage-endpoint shape | **Answered** — exact `available_at` not derivable |
| 5 | Container status vocabulary, post-publish observability | **Resolved — the stated premise was wrong** |
| 6 | Stories-list lookback | **Confirmed** 24 h, with two exclusions |
| 7 | Publishing is URL-pull only | **Contradicted** — a push-bytes surface exists |

## The three findings that change something

**`PUBLISHED` is documented for Instagram, so the seam resolves to container-verdict mode.** This was §0.4's load-bearing question. `04` §0.4 and `02` §6 both record the premise that Instagram documents `FINISHED` as pre-publish while `PUBLISHED` belongs to Threads — raising the possibility (R3 §6.11, *marked uncertain, not false*) that the pass-2 contract borrowed its evidence from the wrong API. It did not. Meta's Instagram content-publishing guide lists five `status_code` values including **"PUBLISHED — The container's media object has been published."** The authoritative-positive set is non-empty. The sets, and the trap: `FINISHED` after `publish_called` means *ready to be published* — evidence the publish has **not** landed, not evidence it failed, so it must not go in the authoritative-negative set.

**"200/user/hr" had the figure right and the shape wrong.** Meta: *"Calls within one hour = 200 * Number of Users"*, where Number of Users is **unique daily active users**. It is an app-wide pool sized by DAU, not a per-user allowance. Reading it per-user over-estimates headroom exactly when the user base is quiet.

**"Publishing remains URL-pull only" is contradicted.** Images are URL-pull as documented (*"We will cURL your image using the passed in URL"*), but a resumable push-bytes surface exists on `rupload.facebook.com`, accepts *"a file located on your computer"*, and Meta says *"All media_type shares the same flow"* while framing it as the Reels path. This does not invalidate D38 — the URL-pull path is what we use and its conclusions stand on their own merits — but "no push-bytes surface exists" can no longer be cited as a **premise**.

## Two more worth reading

**Meta contradicts itself on the cap, today.** The guide says *"limited to 100 API-published posts within a 24-hour moving period"*; its own endpoint reference says `quota_total` is *"currently 50"*. Reported, **not adjudicated** — it is precisely the argument for the live per-account read `main` already does. Neither page may be cited as *the* number. Also newly recorded: carousels count as a single post.

**The refresh min-age is 24 hours, and `05`'s cadence clears it.** The open question on row 56 is closed rather than assumed: 7-day jittered refresh sits comfortably outside Meta's 24-hour floor, so decoupling from expiry proximity introduces no eligibility conflict.

## #734 named the wrong defect

Filed as *"three guides still state the stale 25/24h publish cap"* — asking for them to say 100. Against the tree at `c11f614`, two of its three premises have drifted:

- `guides/instagram-api-setup.md` **no longer exists** — deleted in `08a8ed5` (#738).
- The two survivors say **"25 posts/hour"**, not a 25-per-24h cap, and both attach it to `INSTAGRAM_POSTS_PER_HOUR` — **renamed** to `INSTAGRAM_PUBLISH_LIMIT_FALLBACK` in #707 (`99b00ea`) and now absent from the tree outside CHANGELOG history.

So the operator-facing defect is different in kind and worse: the docs describe a deleted per-hour throttle and point at an env var that will not resolve. **Writing "100" as filed would have produced a confidently wrong sentence** — a correct figure attached to a removed mechanism. Both are corrected to describe what actually runs.

## Bounds

- **Not established:** whether the `rupload` push path is genuinely available for `STORIES` media_type, which is what this product publishes. Meta's page says both *"calls to upload videos for Reels use rupload.facebook.com"* and *"All media_type shares the same flow"*. Settling it needs a real API call, which is outside a docs-only item — so item 7 says *a push-bytes surface exists*, not *it is available to us for stories*.
- The 50-vs-100 divergence is **reported, not resolved**. Correctness does not depend on it.
- Verification covered the seven checklist items in #745. Other `05` platform facts were not re-verified.

## This PR corrects the wrong premises in place — it does not merely quote them

Stating this explicitly, because it is the part most likely to be assumed rather than checked. **Two false platform premises were recorded in the primary planning docs and had already been quoted onward.** Both are corrected at source here, not just noted in a findings file:

| Premise as written | Where it stood | Now |
|---|---|---|
| "`PUBLISHED` is documented for Threads" (so Instagram may have no post-publish status) | `04` §0.4, `02` §6 | **Corrected in both.** `PUBLISHED` is an Instagram container value; seam set to container-verdict mode |
| "publishing remains URL-pull only; no push-bytes surface" | `04` §0.4 (also D38's premise) | **Corrected in `04`.** A resumable push-bytes surface exists on `rupload.facebook.com`; images remain URL-pull |

The first commit corrected `02` and left `04` stating both verbatim — fixing only where it was noticed. The sweep commit closes that. A reader of `04` can no longer pick up a platform claim this pass already refuted.

**D38 itself is untouched and its conclusions stand** — the URL-pull path is what this product uses. What changes is that the *absence* of a push-bytes surface can no longer be cited as a premise; if transit cost is revisited (D38's trigger is sustained Cloudinary consumption above 225 credits), direct upload is a real alternative currently excluded by assumption rather than argument. Flagged for D38's owner rather than edited here.

## Drift rate: 3 of 7

Worth recording as its own number, separate from *which* three. **Three of seven platform facts had drifted (~43% per pass)** — one carried at the wrong shape, two false premises. The three are now correct and inert; the rate is what tells whoever owns these docs how often to re-verify facts nobody has touched.

Two things sharpen it, both in the findings doc: **none of the three was catchable by reading our own docs** (which is exactly how #734, an honest attempt at the same correction from in-tree evidence, landed on the wrong defect), and **both false premises had been recorded as "uncertain" rather than wrong** — then quoted onward. Suggested trigger, which would have caught items 5 and 7 before they propagated: re-verify an item at the moment a new document is about to *depend* on it.

## Propagation (all landed)

- `04` §0.4 — both false premises corrected inline with their verdicts; seam recorded as set.
- `02` §6 — corrected premise, seam resolved to container-verdict, authoritative sets stated, `FINISHED` named as the trap.
- `02` §6 — exhaustion-tail stories check downgraded to corroborating, with the two exclusions.
- `02` §8 — `available_at` uncertainty closed; conservative fallback confirmed as the behaviour.
- `05` rows 13 / 14 / 56 — carousel rule, the 50-vs-100 vendor contradiction, the DAU-pool reshape, the 24 h refresh min-age.
- `documentation/ROADMAP.md`, `documentation/guides/cloud-deployment.md` — #734, corrected to the defect actually present.

Closes #734.
Closes #745.


chrisrogers37 (commented):
**Verdict: Ship it**

Checked against Meta's live docs myself, not against mason's quotes of them — that distinction is the whole point of this review.

**1. The load-bearing claim — PUBLISHED is an Instagram container value.** Independently fetched `developers.facebook.com/docs/instagram-platform/content-publishing` fresh. It lists the same five `status_code` values with the same wording mason quoted, including verbatim: *"PUBLISHED — The container's media object has been published."* My fetch's own summary: *"The PUBLISHED status applies to Instagram containers (not Threads-specific). No distinction is made suggesting it's Threads-only."* Independent word-for-word match on the citation is strong evidence this isn't a paraphrase-drift error — the premise flip in `04` and `02` is correct, and the seam is rightly set to container-verdict mode.

**2. Second independent spot-check — 200/user/hr.** Fetched the rate-limiting page separately. Confirms mason's correction (app-wide pool sized by DAU, not per-user) and actually surfaces a stronger sentence than he cited: *"this is not a per User limit but a limit on calls made by your app."* Also a nuance his doc doesn't carry: during low activity, Meta falls back to weekly/monthly active users to size the pool. Not a defect — worth a follow-up note, not a blocker.

**3. Sweep gap — confirmed no third site.** Grepped the whole repo for both false-premise phrasings ("documented for Threads", "URL-pull only" / "no push-bytes surface"). Every remaining hit is either the corrected fact (CHANGELOG.md, stating what's now true) or `04`'s own line, which states the false premise only inline with its correction attached. No unqualified survivor anywhere. The sweep commit's own framing — "the first commit corrected 02 and left 04 stating both verbatim" — checks out against the diff: both premises were in fact still bare in `04` after the first commit, and both are fixed together in the second.

**4. #734 closure — sanity-checked against git history, not the PR's retelling.**
- `documentation/guides/instagram-api-setup.md`: confirmed gone from the tree, and confirmed deleted (522 lines) in `08a8ed5` (#738) specifically.
- `INSTAGRAM_POSTS_PER_HOUR`: confirmed absent from every `.py`/config file in the tree — survives only in CHANGELOG history and this PR's own findings doc, both describing the past. `INSTAGRAM_PUBLISH_LIMIT_FALLBACK` is confirmed live throughout (`settings.py`, `instagram_api.py`, `health_check.py`, `telegram_commands.py`, plus tests).
- The PR's diff on the two named "survivor" docs (`ROADMAP.md`, `guides/cloud-deployment.md`) shows exactly the stale `25 posts/hour` / `INSTAGRAM_POSTS_PER_HOUR` text being replaced with the live-quota-read description — so #734's original text genuinely doesn't match the current tree, and mason's reframing (a deleted mechanism + an unresolvable env var, not a stale number) is the real defect. Closing #734 against this PR is correct; his framing isn't burying anything realer than what he found.

**CI** — green per-SHA (`gh pr checks 798` against current head `d1de22d`), all 7 checks passing, `Test` completed in 2m24s.

**Not independently re-verified** (disclosed, not blocking): `05`'s rows 13/56 propagation (carousel-single-post rule, the 50-vs-100 `quota_total` self-contradiction, the 24h refresh min-age) — these ride on the same two Meta pages I already fetched and match the PR's own quotes closely enough that I judged the independent high-stakes items (above) as the better use of the time; didn't re-fetch to confirm row-level propagation text matches exactly.

Net: the correction is real, sourced correctly, propagated completely, and the #734 reframing holds up against the actual git history rather than the PR's account of it.

Issue #737, opened 2026-08-07
auth_monitor records the wrong reason for every initData rejection


chrisrogers37:
Split out of #580 so a bounds fix does not also change what a shared auth chokepoint reports.

## What happens

`_validate_auth` (`src/api/routes/onboarding/helpers.py:35-43`) is a fallback cascade:

```python
try:
    return validate_init_data(init_data)
except ValueError:                      # <-- bare, not bound
    try:
        return validate_url_token(init_data)
    except ValueError as e:
        auth_monitor.record_failure(ip, str(e))
        raise HTTPException(status_code=401, detail=str(e))
```

The initData failure reason is discarded. Execution falls through to `validate_url_token`, which splits on `:` — and a real initData querystring URL-encodes its colons as `%3A`, so it fails at `len(parts) != 4` with `"Invalid token format"` before reaching any check that could describe the real problem.

**Only the second exception is ever recorded.** Every initData rejection — bad signature, expired, future-dated, wrong bot token — reaches `auth_monitor` and the client as `"Invalid token format"`.

Measured against the #580 branch:

```
INITDATA_RAISES: 'initData auth_date is in the future'
AUTH_MONITOR_ACTUALLY_RECORDS: 'Invalid token format'
```

## Why it is worth fixing

This is a **diagnosis** gap, not a security gap — the credential is correctly rejected either way, and nothing here widens access.

But it defeats the signal at the moment it is most wanted. During a genuine Telegram clock-skew event, the monitor logs a burst of `"Invalid token format"`, which is exactly what ordinary scanner and junk traffic looks like. The one message that would name the cause is the one that gets thrown away. The same applies to a bad-signature burst, which is what a leaked-token probe would look like.

## Why it was not fixed in #580

`_validate_auth` fronts roughly 37 onboarding routes, and the recorded string is also the 401 `detail` every client sees. Changing it means deciding how to represent "both credential formats failed" — record both, or pick by the shape of the input — and that is a real design call on a security-sensitive chokepoint, not a rider on a bounds fix.

## Sketch

Record the initData failure before falling through, or carry both reasons into the failure record. Worth deciding whether the 401 `detail` returned to the client should stay deliberately vague while `auth_monitor` gets the specific reason — the current coupling means the diagnostic string and the client-facing string cannot differ.

PR #799, opened 2026-08-13
security(api): gate service-health on the system admin role (#667)

chrisrogers37:
Closes #667. Lane 4, PR 1 of 3 (#667 → #737 → #760).

## The gap

`GET /analytics/service-health` returns **deployment-wide** operational telemetry, and authentication was the entire gate — so any authenticated tenant user could read the whole system's health. The payload carries no tenant rows, so this is an **authorization** gap rather than a data leak, but a tenant has no business reading the deployment's health.

## The gate

`users.role` — the system-level role — resolved through `MembershipService`, the authorization door this layer already uses, so API glue still never reaches across the repository layer.

**Deliberately *not* `UserChatMembership.instance_role`.** The two are separate columns precisely so the distinction survives: `instance_role` makes someone an owner of *their own* instance, which is not operator authority over the deployment. Accepting it here would re-open the endpoint one tenant at a time. This is pinned by its own test at both the service and route layer, because every other assertion in the PR passes either way.

`is_system_admin` is fail-closed on the same terms as the existing `is_active_member`: absent user id, unknown user, deactivated user, or any non-admin role returns `False`. It refuses a **deactivated admin** too — otherwise the role outlives the account, and deactivating a departed operator would revoke nothing on any admin-gated route.

Role names became constants beside the schema that constrains them (`ROLE_ADMIN` / `ROLE_MEMBER` in `src/models/user.py`, next to the `check_user_role` CHECK admitting exactly those two) rather than a string literal inside the gate.

## Verification

**Mutation-proven, three ways.** Each mutation was asserted to have actually changed the file before the run.

| Mutation | Tests reddened |
|---|---|
| Ungate the route (back to `_validate_auth`) | exactly the 2 route-level refusal tests |
| Drop the `is_active` conjunct | exactly `test_a_deactivated_admin_is_not` |
| Drop the role comparison | exactly `test_a_member_is_not` + `test_an_instance_role_is_not_a_system_role` |

**The route-level tests are the load-bearing ones.** The helper can be entirely correct while the route is still ungated; only calling the endpoint proves the gate is attached to it. They also assert `DashboardService` is never constructed on the refusal path — the telemetry must not be built before the gate runs.

One trap worth naming: the autouse `_authorize_membership_by_default` fixture returns a Mock whose every attribute is truthy, so `is_system_admin` is set **explicitly in both directions** in every route test. Left unset, the allow-case would pass with no gate existing at all.

**Suite:** `2099 passed, 8 skipped` (`tests/src -m unit`), rc 0, zero failures. Changed files clean under `ruff check` and `ruff format --check`.

**No caller is affected:** no frontend source references this route (grep across the Mini App sources returns nothing), so gating it breaks no user-facing surface.

## Disclosed

One transient `1 error` appeared in a single combined mutation run and did **not** reproduce across two subsequent runs of the identical command (both `2 failed, 82 passed`, no error). I am not asserting a cause. The mutation verdict is unaffected — the kill signature was identical in every run.


chrisrogers37 (commented):
**Approve**

Time-boxed per ari's hard deadline (15:15 EDT) — prioritized the thing that can only be established by running it, named what I did not get to below.

**Item 1 — the refuse path, run myself, not read.** Reverse-applied just the `dashboard.py` route hunk in a worktree at the PR head SHA (1071c97), ungating the route back to `_validate_auth`. Ran the route-level class: exactly `test_an_authenticated_non_admin_is_refused` and `test_instance_membership_does_not_admit` go red, `test_a_system_admin_is_allowed` stays green — matches the PR's claimed mutation table exactly. This is the actual security property (gate is attached to the route), confirmed by execution.

**Item 2 — fail-closed.** `is_system_admin` is `bool(user and user.is_active and user.role == ROLE_ADMIN)` — none of {missing user_id, unknown user, deactivated user, non-admin role, instance-role-not-system-role} can fall through to `True`; there's no default-permissive branch. Verified by direct code reading (simple boolean chain, not the branchier shape #627 had) plus the baseline run: all 6 fail-closed-shaped unit tests pass as-authored (`test_a_member_is_not`, `test_a_deactivated_admin_is_not`, `test_an_unknown_user_is_not`, `test_an_absent_user_id_is_not`, `test_an_instance_role_is_not_a_system_role`, `test_an_admin_is_an_admin`). Did **not** mutation-test this one myself (time) — trusting the PR's claimed per-conjunct mutation table (drop `is_active` → only the deactivated-admin test reddens; drop the role comparison → only the two role-confusion tests redden) rather than reproducing it.

**CI, per-SHA:** all 6 checks (Test, Lint, Security Scan, GitGuardian, Changelog, Vercel) `success` on `1071c976038557d084b49f2c8203fd792734124c`, queried directly against that SHA.

**Named gaps, not silently skipped:**
- Mutation 2 and 3 (the `is_active` / role-comparison conjuncts) not independently reproduced — code + passing tests only.
- Did not check the other two route(s) mentioned only in passing (`test_a_system_admin_is_allowed`'s admin-path plumbing) beyond confirming it stays green under the mutation.
- Did not verify the "no frontend caller" grep claim independently.
- Did not re-run the full suite (2099 tests) myself.
- #799/#800 file-overlap question — not re-derived, per explicit instruction that it's already settled (`git merge-tree` clean).

Comment-only — same-identity blocks native Approve on this account.

PR #800, opened 2026-08-13
fix(media): one MD5 source of truth for media_items.file_hash (#619)

chrisrogers37:
Closes #619.

## The defect, traced

`src/utils/file_hash.py` documents why the algorithm is MD5: it matches Google Drive's native `md5Checksum`, which is what makes rows from different sources comparable. Three writers each chose SHA256 independently, so their rows could never match a Drive-sourced row. **A non-matching hash is indistinguishable from novel content, so every miss was silent.**

**The issue named one site. There are three, all writing `media_items.file_hash`:**

| Site | What it does |
|---|---|
| `backfill_downloader.py:108` | the one filed |
| `dashboard.py:459` | Mini App upload — hashes SHA256, then immediately runs a "check for duplicate content" lookup against a table of MD5 rows. A dedup check that runs and structurally cannot match. |
| `google_drive_provider.py:294` | the no-`md5Checksum` fallback **inside `calculate_file_hash()` itself** — so one Drive file hashed two different ways depending on what the API returned, in the method whose docstring says it exists for cross-source dedup |

Fixing only the filed site would leave the issue's stated goal unmet in two places while marking it closed, so nobody re-checks.

**The harm reaches past dedup.** `media_repository.py:708-727` excludes items whose `file_hash` matches a currently-locked item — the guard against posting the same content twice. A SHA256 row can never match a locked MD5 row, so the same media could be posted once as the Drive copy and once as the backfill copy.

`src/models/media_item.py` also documented the column as `# SHA256 of content`, so the schema comment carried the wrong algorithm too.

## The fix

Fixed as a **class**, not as three constants — the defect was three call sites each free to pick.

- `_HASH` in `src/utils/file_hash.py` is now the single algorithm decision.
- New `calculate_bytes_hash()` for callers holding content in memory; `calculate_file_hash()` still streams from a path. Both build from `_HASH`.
- All three writers delegate. None imports `hashlib` any more.

## Verification

`70 passed` across the three affected suites (`test_file_hash.py`, `test_backfill_downloader.py`, `test_google_drive_provider.py`).

**Mutation-proven per part**, because a guard that has never failed is not evidence:

| Mutation | Result |
|---|---|
| a writer reaches for `hashlib` directly | `test_writers_do_not_choose_their_own_algorithm` **FAILED** |
| flip `_HASH` to `sha256` | **5 FAILED** across all three surfaces, incl. both new tests |

Expectations are anchored to `hashlib.md5` directly rather than to the functions under test — `calculate_bytes_hash(x) == calculate_bytes_hash(x)` would have passed with the algorithm swapped. The writer scan carries a positive control asserting the pattern can match and that every scanned path resolves, so a typo'd path cannot read as a clean scan forever.

## What this PR deliberately does NOT do

**Existing rows keep their SHA256 and remain unmatchable.** Re-hashing them is the **media-hash dedup remediation** already named as an M.1 precondition in #746 and #790, and one of only two sanctioned production data changes outside the M.3 window. It is not in this PR, and this PR is its prerequisite: remediating before the writers are fixed just lets the app write more SHA256.

The discriminator for that remediation is free — **64 hex characters is SHA256, 32 is MD5** — so divergent rows are selectable without a content re-read.

No production database was touched.

## Notes for the reviewer

- Scope beyond the filed site was flagged to the manager before implementation rather than taken silently.
- `/simplify` has **not** run yet (pushed early against a token ceiling). Running it next; any cleanup lands as a follow-up commit on this branch.
- `CHANGELOG.md:796` carries an older `### Fixed — Hash Algorithm Mismatch` entry, so this class has recurred at least once before. That is the argument for the structural test rather than a third correct constant.

https://claude.ai/code/session_011oxF5xUHf152DLP5xDM5QM


chrisrogers37 (commented):
**Verdict: Ship it**

Reproduced rather than read, per the brief. Two things asked for specifically:

**1. Enumeration — independently re-swept, not trusted.** Grepped the whole repo (`src/`, `tests/`, `scripts/`, `cli/`) for `hashlib` and `file_hash`, blind to the PR's own list first. Confirms exactly 3 genuine writers, no 4th:
- `dashboard.py:459`, `backfill_downloader.py:108`, `google_drive_provider.py`'s no-`md5Checksum` fallback.
- `local_provider.py` and `media_ingestion.py` already delegated to the pre-existing (already-MD5) `calculate_file_hash` — needed no fix, consistent with their absence from the diff.
- `webapp_auth.py`'s SHA256 hits are HMAC signing for Telegram's WebApp auth spec — unrelated to `file_hash`.
- `scripts/advertised_ddl.py` and `scripts/migration_runner.py` also call `hashlib.sha256` — checked both directly: DDL-drift and migration-integrity checksums, neither touches `file_hash`/`media_item`. Not a 4th site.

One structural note, not a defect in this PR: `_FILE_HASH_WRITERS` in the test is a **hardcoded 3-item allowlist**, not a repo-wide scan. Today's enumeration is complete (confirmed above), but a *future* writer that isn't manually added to that tuple would silently escape the regression guard. Worth knowing, not blocking.

**2. Mutation proof — reproduced myself on a fresh clone, isolated env, synced venv (baseline 70 passed, matches the claim exactly).**

- **Mutation A** (writer bypasses the shared function): my first two attempts at this used `__import__("hashlib")` and an aliased `import hashlib as _h` — both false negatives *of my own construction*, since the regex correctly targets the literal `hashlib.xxx(` shape. Redone with that literal shape: **exactly 1 failure**, `test_writers_do_not_choose_their_own_algorithm`. Matches the claim precisely.
- **Mutation B** (flip `_HASH` to `sha256`): **4 failed, 66 passed — not 5 as claimed.** The commit message's "5 FAILED... incl. both new tests" overcounts by one. Of the two meaningful new tests in `TestHashAlgorithmIsSingleSourced`, only `test_both_entry_points_agree_and_are_md5` is behaviorally sensitive to `_HASH`'s runtime value — `test_writers_do_not_choose_their_own_algorithm` is a textual scan of source code, orthogonal to this mutation (it's mutation A's discriminator, not this one's). **Not a decoy**: every test that does fail is correctly anchored to `hashlib.md5` directly, not to the function under test, so none of the 4 are vacuous. This is a miscount in the PR's own narrative, not a coverage gap — flagging for the record, not blocking merge.

**CI** — checked per-SHA, not the tab: `gh pr checks 800` against current head `3a3acb57` (post-`/simplify` commit, one past the SHA in the original brief) shows all 7 checks passing, including `Test` which was still `IN_PROGRESS` when I first looked seconds earlier.

**Not re-derived** (per brief, already settled): the `dashboard.py` L3/L4 overlap — alex's `git merge-tree` clean result stands.

**Not checked — wall-clock deadline, named rather than silently dropped:**
- Pre-fix parent-commit reproduction of the original bug (checking out the commit before `fb2231a` and confirming the writers actually produced SHA256) — the mutation proof above is the stronger check anyway (it exercises the *fixed* code's guard rails directly) but I didn't independently confirm the pre-fix failure mode.
- A dedicated scope-creep/dead-code/naming pass beyond the hash-parity surface itself.

Net: the fix is real, the enumeration is genuinely complete, and the tests that guard it discriminate correctly — verified by reproducing failure, not by reading a claim. One narrative overcount in the commit message, named above, doesn't change that.

PR #801, opened 2026-08-13
fix(tests): a refused connection is a server that answered (#769)

chrisrogers37:
Closes #769.

`server_answered()` read **every** `psycopg2.OperationalError` as "no PostgreSQL was configured" and took the skip branch. But `too many connections`, `password authentication failed` and `database does not exist` all mean the server **answered and refused** — so a local run silently skipped every integration test at `rc 0`. That is the #758 false PASS, reproduced one layer down inside the probe that exists to prevent it.

## The exception cannot discriminate — measured, not assumed

The issue measured three refusal cases and **inferred** `too_many_connections`. That one is now reproduced, via a role-scoped `CONNECTION LIMIT 0` so nothing else on the shared cluster is disturbed (role created and dropped in the same script):

```
too_many_connections -> OperationalError, pgcode=None
  message: connection to server at "localhost" (::1), port 5432 failed: FATAL:  too many connections for role "probe769_30309bf5"
  probe says listening: True
no listener          -> OperationalError, pgcode=None
  probe says listening: False
```

`pgcode` is `None` on **both sides of the distinction**, so the exception carries nothing to branch on. Only the message text differs, and matching on that is a locale-dependent string match — the shape #758 rejected when it replaced the argv detector with a cwd check. The address can answer; the exception cannot.

Worth recording alongside it: `no such role` reports *password authentication failed*, because PostgreSQL conflates the two to prevent user enumeration. Message matching could not have separated those either.

## What changed

A TCP listener probe (`server_is_listening`), consulted **only on the failure path**. A successful connection keeps its own evidence — the happy path never asks the address a question the login already answered, and there is a test pinning that, because the obvious refactor is to probe first and connect after.

## Three judgement calls the issue asked to be made deliberately

1. **The fail branch widens, on purpose.** PostgreSQL up + wrong test credentials now **fails** where it used to skip. That run executed no integration tests either way, so the skip was never honest; the cost is paid in a printed diagnosis naming `DB_USER` / `DB_PASSWORD` / `DB_NAME` rather than in silence.
2. **A non-PostgreSQL listener counts as answered.** For "was a database configured *here*", an occupied address is a better proxy than a completed login. Verifying the wire protocol was considered and rejected — it means implementing a fragment of the startup handshake inside a conftest. If something else holds the port, `DB_HOST`/`DB_PORT` is wrong and a loud failure is right.
3. **The unchanged bound, stated rather than glossed:** a server that has just *died* is still indistinguishable from one that was never there, because neither leaves anything listening. That single case still reads as not-configured.

## Verification

- `tests/test_integration_coverage_policy.py` — **24 passed**.
- Full suite — **2569 passed, 8 skipped, 7 failed**. All 7 are `tests/src/services/test_telegram_service.py`, failing on `RuntimeError: To use AIORateLimiter, PTB must be installed via pip install "python-telegram-bot[rate-limiter]"`. Confirmed pre-existing rather than assumed: the same 7 fail on a clean `origin/main` worktree (`7 failed, 43 passed`) with none of this branch's code present. Missing local optional dep; CI installs the extra.
- **Mutation-proven.** Removing the listener-probe branch from `server_answered()` — i.e. restoring pre-#769 behaviour — reddens **exactly two** tests, both the ones that encode the distinction:
  ```
  FAILED ...::test_a_refusal_from_a_live_server_is_not_an_absent_server
  FAILED ...::test_the_two_refusal_paths_reach_opposite_verdicts
  2 failed, 22 passed
  ```
  The paired test asserts both verdicts **together** (`("run", "skip")`), because either half alone passes for a function that always returns one answer.

## Provenance

Astrid started #769 and hit her Fable cap mid-work; her uncommitted diff was **not adopted** (per the guidance on issue comment 5285209510 — starting clean beats building on an 84-line diff whose intent cannot be confirmed). Her checkout is untouched and still recoverable. This is a fresh implementation.

Lane 5 is serialized #769 → #785 → #773; #785 is next.


chrisrogers37 (commented):
**Approve**

Reproduced independently on all three points rather than re-reading the pasted output.

**(1) Bidirectional discriminator — verified live, not mocked.** Created my own role with `CONNECTION LIMIT 0` on this host's PostgreSQL 15.18 (independent of the PR's own repro) and confirmed the premise first: `pgcode=None` for both the too-many-connections refusal and a bad-password refusal on this exact server. Then drove the actual PR code end-to-end, no monkeypatching:

- Refusing-but-listening (the direction that was broken): `server_is_listening('localhost', 5432)` → `True`; `server_answered()` under the live refusal → `True`. Correct.
- Not-listening (the direction that was never broken): probed a port bound-then-released so nothing was there → `server_is_listening()` → `False`; `server_answered()` → `False`. Correct.

Both directions confirmed against a real socket / a real refusing Postgres, not just the two raw-socket unit tests already in the diff.

**(2) Mutants are not inert — reproduced, not trusted.** Applied the exact mutation described (deleted the `server_is_listening` branch from `server_answered()`, restoring pre-#769 behavior) in an isolated worktree and ran the suite myself:

```
FAILED ...::test_a_refusal_from_a_live_server_is_not_an_absent_server - assert False is True
FAILED ...::test_the_two_refusal_paths_reach_opposite_verdicts - AssertionError: assert ('skip', 'skip') == ('run', 'skip')
2 failed, 22 passed
```

Exact match to the PR's claim — same 2 tests, same count. Both are genuine assertion mismatches, not collection errors, and the paired test's failure shows precisely the collapse it exists to catch: both verdicts fold to `'skip'` once the discriminator is removed. Not inert.

**(3) Swept, not spot-checked.** Grepped all 617 git-tracked files for `pgcode`, `OperationalError`, and any other caller of `server_is_listening` / `server_answered`:

- 2 other `pgcode` hits, both checked and unrelated. `scripts/fc8_gate.py:195` only interpolates it into a diagnostic string with no branch on its value — and the connect()-failure path above it is already unconditionally loud regardless of cause. `tests/scripts/test_rls_harness.py:77` reads SQLSTATE for *post-connection* query errors (`42501`/`42704`), a domain where pgcode is reliably populated — a different failure class than the connection-phase ambiguity this PR fixes.
- Several `OperationalError` catches in production code (`onboarding/settings.py`, `onboarding/setup.py`, `base_repository.py`'s circuit breaker, telegram callback retry paths) — all fail loud (503 / raise / retry-then-raise). None silently downgrade to a skip-shaped outcome, so none share this bug's shape.
- `server_is_listening` / `server_answered` are called from nowhere outside `tests/conftest.py` and `tests/test_integration_coverage_policy.py`. Checked all 4 `conftest.py` files in the repo — the other three have no reachability/discriminator logic of their own.

**What I did not sweep:** `HealthCheckService.check_all()` (backing the `check-health` CLI command). Whether the *production* health check has its own version of this ambiguity is a materially different, broader question than "same distinction, same pgcode pattern," and was out of scope for this pass — flagging it as unchecked rather than implying it's clean. Also did not re-run the full 2569-test suite to re-verify the "7 pre-existing failures" claim — out of scope for the three named points.

**Minor, non-blocking:** the two `server_is_listening` unit tests exercise the TCP-probe primitive against a raw bound socket, not literally against a Postgres server mid-refusal — the `CONNECTION LIMIT 0` repro that justified the fix lives only in the PR description, not as a pinned CI regression test. I independently reproduced it by hand above and it holds, but nothing in CI would catch a future regression on that specific path without a connection-limited-role fixture. Not blocking — the mocked tests already pin the branching logic and the added fixture cost is real — just naming it.

Solid fix, ship it.

PR #803, opened 2026-08-13
test(fc8_gate): pin the halt-routing text per fact, mutation-proven (#795)

chrisrogers37:
Closes #795.

`render()`'s `if c.halts:` block had its presence covered and its content unasserted. Eight tests plus a negative control, and the evidence that they bind.

## Kept at the tier #795 filed it at

This stays robustness, not silent information loss. Deleting the block leaves an operator less well served rather than uninformed, because the exit code and the VERDICT line still carry the fact.

**That argument depends on the VERDICT line saying `HALT — window prep stops`, and nothing asserted that line's content either.** Degrade it to a bare `HALT` and losing the routing block *would* start hiding something — the tier-lowering premise would quietly stop holding, with every test still green. So it is pinned here too (mutation M9).

## One assertion per fact, not one over the block

#795 names this directly — covering two independently-deletable disclosures with a single assertion leaves a hole the size of the other. The block carries six facts in a wrapped paragraph, and a paragraph can lose one resolution and read as deliberate either way:

| Fact | Test |
|---|---|
| Routing destination + counts go with it | `test_it_says_where_the_halt_routes_and_that_counts_go_with_it` |
| Plan-line provenance (`04` L87) | `test_it_cites_the_plan_line_the_resolutions_come_from` |
| Resolution A — migrate to Drive | `test_the_migrate_resolution_survives` |
| Resolution B — explicit accept-loss list | `test_the_accept_loss_resolution_survives` |
| Both are zero-schema | `test_both_resolutions_are_stated_to_be_zero_schema` |
| Weakening the `02` NOT NULL chain is *recorded as rejected* | `test_the_rejected_resolution_stays_recorded_as_rejected` |

Separate tests rather than several asserts in one, because the first failing assert masks the rest — the same hole one level down.

Two beyond the six:

- **`test_the_verdict_line_says_halt_and_what_it_stops`** — the fallback the tier argument rests on.
- **`test_a_history_only_halt_routes_identically`** — `halts` is *either* count nonzero. A block keyed on `live_named` alone satisfies every other assertion here and prints nothing for the halt arriving through `posting_history`, the count with no origin column and the harder of the two to have reasoned about.

Plus a negative control in the existing class (`test_the_halt_routing_block_is_absent_from_a_clear_render`) — without it, a `render()` emitting the block unconditionally would pass everything above while telling an operator on a CLEAR run to go resolve a halt that did not happen.

## Mutation proof

A test existing and a test binding are different claims, so each was made to fail on purpose. Eleven mutations to `scripts/fc8_gate.py`, source restored and verified byte-identical to HEAD after each:

| # | Mutation | Went red |
|---|---|---|
| M1 | delete the whole halt-routing block | all 7 routing tests (verdict test correctly stays green — that line is outside the block) |
| M2 | drop **only** the accept-loss resolution | accept-loss test + history-only test |
| M3 | drop **only** the migrate resolution | **migrate test alone** |
| M4 | drop **only** the rejected-option sentence | rejected test alone |
| M5 | drop **only** the zero-schema justification | zero-schema + rejected |
| M6 | drop **only** the plan-line provenance | provenance test alone |
| M7 | drop **only** the routing destination | routing + history-only |
| M8 | print the block unconditionally | negative control alone |
| M9 | degrade VERDICT to a bare `HALT` | verdict test alone |
| M10 | key the block on `live_named` only | history-only test alone |
| M11 | **control** — re-wrap the paragraph, change no fact | **nothing; 14 passed** |

M3 is the result the per-fact split exists for: drop one of the two resolutions and only that resolution's test reddens. A single blob assertion over the block passes M3.

M11 is the control that makes the rest meaningful. The tests normalise whitespace (`_flat()`) so they bind to content, not line breaks — a reflow that changes nothing an operator reads stays green, a deleted fact goes red. Without M11 I could only claim the tests are sensitive, not that they are sensitive *to the right thing*.

**Honest caveat on M5:** it is the one mutation that is not cleanly isolated. Removing `Both are zero-schema; ` also recapitalises `weakening` → `Weakening`, which trips the rejected-option assertion on case. Two facts moved, so two tests red. Not a defect in either test — M4 isolates the rejected option on its own — but the row is not the clean single-fact result the others are.

## Verification

Run unsandboxed from the repo venv.

```
tests/scripts/test_fc8_gate.py -q
29 collected — 15 passed, 14 errors
```

The 14 errors are `replayed_db` fixture failures: no PostgreSQL in this environment (`password authentication failed for user "storydump_test"`). **The DB-backed half of this file was not run and is not claimed as passing** — it is unchanged by this PR. Everything added here is pure `render()`, no database, which is why the mutation matrix could run 11 times in under a second.

Before this branch, the same two render classes were `5 passed, 1 error`. After: `14 passed, 1 error`. The single persistent error is `test_the_default_path_prints_the_text_disclosure_to_stdout`, which is pre-existing and DB-backed.

`ruff check` and `ruff format --check` pass on the changed file. `scripts/fc8_gate.py` is untouched — `git diff --exit-code` on it is clean.

## Scope

Test file plus CHANGELOG. No change to the gate.


chrisrogers37 (commented):
**Approve**

Told to verify the mutation claims rather than take them: are the 11 mutations valid, is the negative control load-bearing, is the reflow control meaningful rather than vacuous. Independently re-ran 5 of the 11 (M1, M3, M5, M8, M9) plus the control (M11) against a live checkout — verbose pytest output, byte-identical restore verified after each.

**Every one reproduced exactly, by test name, not just by count:**

| # | Mutation | Claimed red | Observed |
|---|---|---|---|
| M1 | delete whole block | 7 routing tests; verdict + negative control stay green | exact match |
| M3 | drop migrate resolution only | migrate test, alone | exact match |
| M5 | drop zero-schema (+ incidental recap) | zero-schema + rejected, both | exact match |
| M8 | print block unconditionally | negative control, alone | exact match |
| M9 | degrade VERDICT to bare HALT | verdict test, alone | exact match |
| M11 | reflow, no fact changed | nothing | exact match — 0 failures, 9/9 green |

**Your two specific asks:**

- Negative-control validity — M8 is the direct proof. Forcing the block to print unconditionally trips *only* `test_the_halt_routing_block_is_absent_from_a_clear_render`; none of the 8 positive tests move. The control is doing real, non-redundant work — without it, a block that bleeds onto a CLEAR render ships silently.
- Reflow-control vacuity — M11 alone (0 failures) doesn't prove non-vacuous by itself; the proof is the *contrast* against M1/M3/M5/M8/M9, all of which produce exact, differentiated failures on the same 9-test set. Green-on-reflow means something only because the same tests demonstrably go red on real fact deletion, and here they do.

**Also chased the sharper question raised mid-review:** could any of these 11 be scored against a test that errors for an unrelated reason (no local Postgres), making the claim unverifiable in the env it was produced in? No — checked every signature directly: none of the 9 targeted tests take a `replayed_db` fixture. The CHANGELOG entry this PR adds confirms the same split first-party: 14 `replayed_db` tests, 14 pure `render()` tests (these), one DSN-unreachable-by-construction test. Separately verified CI's `test` job runs a real `postgres:15` service with `DB_HOST`/`DB_PORT`/etc. wired into the test step — the 14 errors in a bare local checkout are a no-local-Postgres artifact, not a CI gap. No follow-up issue warranted there.

`scripts/fc8_gate.py` itself: confirmed untouched (`git diff` on it between base and head is empty) — genuinely test-file + CHANGELOG only, as claimed.

Not independently re-run: M2, M4, M6, M7, M10. The sample covers the flagship per-fact isolation case (M3), the one self-disclosed imperfect case (M5), both controls (M8, M11), and the block-boundary case (M1). M10's basic bindingness is implied by M1 (its test is among the 7 that redden there) but wasn't isolated on its own.

Mutation discipline here is real, not narrated.

Issue #282, opened 2026-04-25
[GTM] Streamline onboarding — reduce setup from 30-60 min to under 10 min


chrisrogers37:
## Problem

Current setup requires 6 prerequisites (Instagram Business account, Meta Developer app, Google Cloud project, Telegram, organized Drive folder, knowledge of OAuth). This is the #1 drop-off point.

## What to build

1. **Eliminate Google Cloud project requirement**: Use Storyline's own OAuth app for Google Drive (user just clicks "Connect Google Drive" instead of creating their own app)
2. **Eliminate Meta Developer app requirement**: Use Storyline's own Facebook/Instagram app (user just clicks "Connect Instagram" — standard OAuth)
3. **Guided Instagram Business conversion**: In-app walkthrough for converting Personal → Business account (3-step process, most users can do in 2 minutes)
4. **Direct upload alternative**: Don't require Google Drive at all — let users upload media directly via web dashboard (see #184)
5. **Progress indicator**: Show "Step 2 of 4" with estimated time remaining
6. **Skip-and-return**: Let users complete partial setup and come back later

## Architecture notes

- Instagram Login OAuth flow already exists (`instagram_login_oauth.py`) — needs to use a Storyline-owned app instead of per-user apps
- Google Drive OAuth exists (`google_drive_oauth.py`) — same: use Storyline's app credentials
- This means centralizing OAuth apps and managing user token scopes at the platform level
- May require Meta App Review for Instagram permissions (stories_read, pages_show_list, etc.)

## Impact

Reduces setup from 6 prerequisites + 30-60 min → 2 prerequisites (Instagram account + Telegram) + under 10 min. This alone could 5-10x conversion from landing → active user.

## Part of

Compound Play: **Zero-Friction Onboarding**

chrisrogers37:
[mason] [menu] The /setup fork — what it actually is, and four options

**Menu, not a ruling. I pick nothing.** This issue now blocks three others: #802 steps 3–4, #756, and #810. Built on virgil's groundwork, but every claim below I re-measured against source at `c28066d` and against **live storydump.app** today. Where my measurement differs from his, I say so.

---

## The reframe: this is not "should we centralize?" — that already shipped

The thing that makes this decidable is a fact nobody has stated on any of the four issues. **This issue's items 1 and 2 are already true in the code.**

| Credential | Where it comes from |
|---|---|
| Instagram Login | `INSTAGRAM_APP_ID` — `settings.py:120` |
[...]

PR #805, opened 2026-08-13
fix(site): remove the FC-4 Facebook Page instruction; withhold two dead setup pages from search (#802)

chrisrogers37:
Implements the two parts of #802 that need no product call. The funnel's future is untouched and stays routed to #282.

## Part 1 — the FC-4 violation

`/setup/instagram` carried a step titled **"Create and link a Facebook Page"**:

> Meta requires a Facebook Page linked to your Instagram account for API access — even if you never use Facebook.

closing with **"You MUST have a Facebook Page linked."**

FC-4 rules the opposite in as many words — *"no Facebook Page — **never make a user auth a Facebook Page again**"*. Removing this executes a ratified decision rather than making a new one.

**It is also factually wrong, not merely off-constraint.** Under the Instagram API with Instagram Login that FC-4 selects and #745 verified against Meta's own current documentation, no Facebook Page is required. A live, indexed public page was charging prospective customers the advertised *"30-60 minutes (mostly waiting for Meta approvals)"* to satisfy a requirement that does not exist.

### Removed at all six sites, not only the step

The page asserted it six times. Fixing only the step would have left the superseded claim alive in the same document:

| Site | Was |
|---|---|
| `metadata.description` | "…and link it to a Facebook Page for API access" |
| Intro paragraph | "…set up correctly and linked to a Facebook Page" |
| Step 2 cross-reference | "skip to Step 4 (Facebook Page)" |
| Step 4 (whole card) | create a Page, link it, screenshot, MUST warning |
| Step 5 verification hint | "…and that a Facebook Page is linked" |
| `/setup` prerequisites | "Facebook Page linked to your Instagram account" |

**One of those six is invisible to grep.** The intro wraps `Facebook` / `Page` across a line break, so `grep -i "facebook page"` does not return it. It was found by reading the file, and a whitespace-normalising sweep is what confirms the page is now clean:

```
for f in $(find "src/app/(marketing)/setup" -name '*.tsx'); do
  tr '\n' ' ' < "$f" | tr -s ' ' | grep -oiE "facebook page" | sed "s|^|$f: |"
done
```

The only remaining hit is the new sentence saying you do **not** need one.

What survives is exactly the one conversion FC-4 preserves — personal → Professional — and steps renumber 5 → 4.

**The removal is stated, not silent.** A reader who followed the old page and created a Facebook Page deserves to be told the requirement is gone, so Step 1's body now ends: *"That is the only change your account needs: Storydump connects through Instagram directly, so you do not need a Facebook Page."* It sits in the body rather than in a second callout — two stacked callouts in one step is weight the content had not earned, and "what does my account actually need" is a body question.

## Part 2 — the two dead pages are withheld from search

`robots: { index: false, follow: true }` on `/setup/meta-developer` and `/setup/google-drive`, and both dropped from `sitemap.ts`.

They instruct the reader to register their own Meta app and their own Google Cloud project. The product accepts neither — `FACEBOOK_APP_ID`, `INSTAGRAM_APP_SECRET` and `GOOGLE_CLIENT_SECRET` are all deployment-level, and no surface (Mini App wizard, onboarding API, CLI, or schema) takes a tenant-supplied App ID, Client ID or Secret. Evidence in #756.

Search was the **only** inbound path — neither page is linked from `layout/header.tsx` or `layout/footer.tsx` — so closing it removes the whole arrival rather than narrowing it.

**Withheld, not deleted.** Both pages still build and still resolve; the blog article that cites the Drive one keeps working, and `follow: true` preserves outbound link equity. Reversible in one commit, and it preempts nothing #282 might decide.

## Verification

- **`npm run build` → rc 0.** All `/setup/*` routes still prerender, including both withheld pages — confirming they remain reachable rather than removed.
- **`npx tsc --noEmit` → rc 0.**
- **`npm run lint` → no delta.** The landing lint baseline is red (5 errors, 5 warnings), all in files this PR does not touch — `(dashboard)/dashboard/error.tsx`, `not-found.tsx`, `blog/[slug]/page.tsx`, `telegram-login-button.tsx`, `media-*.tsx`. Verified by stashing and re-running: the before and after problem lines are byte-identical, so this PR adds nothing. Naming the baseline rather than reporting a green tick, because on an already-red suite the count is the only axis that moves.
- Whitespace-normalised sweep above, run after the final edit.

## Deliberately not done

- **The funnel's future** — self-serve vs waitlist-gated, whether `/setup` survives as a shape, and rewriting the two withheld pages. Routed to #282. Splitting it back out here would undo the routing.
- **`/setup/connect:96`** still tells the reader to have their App ID and Secret ready for a wizard step that does not exist. It is a frame page, half-dead per #802's table, and belongs to the same product call.
- **Terminology.** Step 1 said "Instagram's Graph API"; under Instagram Login that is the wrong API name. I changed it to "Instagram's API" as a side effect of rewriting that sentence, but did **not** sweep the funnel for other stale API naming — that is rewrite territory, and the claim it carried ("requires a Professional account") was true either way, so it was never the ratified violation.
- **`/setup/meta-developer:48`** still says "log in with the Facebook account linked to your Instagram". That is about signing in to developers.facebook.com, not creating a Page, and it lives on a page this PR withholds and #282 rewrites.

## Scope

Four page files, `sitemap.ts`, CHANGELOG. No application code, no tests touched, nothing in #737 / #760 / #769 / #785 / #773.


chrisrogers37 (commented):
**Approve**

Reviewed against the specific scrutiny point: if one of the six sites was invisible to a naive grep, the enumeration *method* is what's under review, not the six edits. Both parts are now independently evidenced, not just narrated.

**Part 1 — six-site removal, enumeration completeness.** Independently re-checked rather than trusted: full pre-PR content of the changed file walked by hand, a whole-`landing/src/`-tree grep (all 123 files, 3 extensions) for "facebook" plus 8 paraphrase patterns (`fb page`, `linked page`, `requires a page`, `business page`, etc.), all 6 shared `components/setup/*` files and `lib/og.ts` read for hardcoded copy, every setup page's `metadata.description` checked (including `/setup/connect` and `/setup/media-organize`, which this PR doesn't touch), and the whole blog tree. No 7th prose-based violation exists. Genuinely clean, not just unfound.

One forward-looking note, not a blocker: this PR's own fix uses a `{" "}` + `<span>` emphasis-split (`you do{" "}<span>not</span> need a Facebook Page`) — a convention already used in 16 other files in this tree. The whitespace-normalizing sweep this PR adds only touches whitespace bytes, not markup, so a *future* edit that splits "Facebook" and "Page" across a tag the way this PR splits "do" and "not" would be invisible to it too. Didn't bite this time — checked directly, no instance exists today — but worth knowing the sweep's own blind spot before leaning on it again.

**Part 2 — noindex + sitemap withholding.** The PR's stated verification (`npm run build` rc 0) confirms the pages compile, not that the tag renders. Built and served the branch myself, confirmed by curl with a real positive control in the same run — full evidence, including the two pages that correctly stay `index, follow`, posted separately as a PR comment (https://github.com/chrisrogers37/storydump/pull/805#issuecomment-5286243096) since it wasn't in the PR and a verification nobody can find later didn't happen for review purposes. Sitemap exclusion confirmed too. True as claimed.

**One finding that doesn't block this PR, but is bigger than it.** While checking completeness, found a third blind-spot class the word-search methodology can't reach by construction: `/setup/meta-developer` — reachable right now via the "Next" link at the bottom of the page this PR just edited — lists OAuth scopes in its Step 5 (`pages_show_list`, `pages_read_engagement`) that *are* the Facebook Page permissions, with zero occurrences of the word "facebook" anywhere in that section. Traced it into the backend: `OAuthService.REQUIRED_SCOPES` matches those exact scopes, and `setup.py`'s onboarding route falls back to that Page-requiring flow whenever `INSTAGRAM_APP_ID` is unset — a documented, previously-encountered failure mode per this repo's own troubleshooting guide. This PR's new claim ("you do not need a Facebook Page") is therefore only as true as one env var it never checks, and the next page in its own funnel still states the old requirement unconditionally.

This isn't something #805 introduces or could reasonably be expected to fix — it's a pre-existing backend/infra condition, and the six textual edits plus the two withholds are correct on their own terms regardless of it. Filing it separately with the full trace rather than holding this PR on it.

Both parts of what this PR claims are true and evidenced. Approve.

PR #807, opened 2026-08-13
refactor(tests): consolidate the two cluster-role locks to one (#785)

chrisrogers37:
Closes #785.

Two mechanisms serialized the same cluster-scoped `svc_*` provisioning — #772's `flock` and PR-A's session-wide `pg_advisory_lock` — kept side by side through the #778 rebase so neither completed review was silently undone. This collapses them to one.

## The issue suggested keeping the `flock`. This keeps the advisory lock instead.

Not a preference — its two stated grounds did not survive measurement.

**"Kernel-released on process death (a real advantage: a killed holder cannot strand it)."** True of `flock`, and equally true of the advisory lock. SIGKILL a holder with no cleanup path available:

```
holder alive       : granted-advisory-locks-on-key=1  try_advisory_lock_succeeds=False
after SIGKILL      : granted-advisory-locks-on-key=0  try_advisory_lock_succeeds=True
```

The backend sees EOF and exits, taking the session lock with it. Both mechanisms have the property; only one of them is scoped to the resource.

**Scope is where they actually differ.** The roles are CLUSTER-scoped; a `flock` is HOST-scoped. Two containers or hosts against one cluster share no temp directory, so the file lock silently fails to serialize them. `tempfile.gettempdir()` also follows `TMPDIR`, reaching the same divergence on one host — measured as a mechanism (`TMPDIR=/tmp/bot-a` → `/tmp/bot-a/storydump-svc-roles-localhost-5432.lock`), and reported as **latent, not live**: every bot on this fleet currently sets `TMPDIR=/tmp`. My first run of that check returned a false negative because the directories did not exist and `gettempdir()` fell through to `/tmp`; the result above is the re-run.

**And the survivor covers more, by container rather than by enumeration.** The advisory lock is held session-wide by `admin_conn`, so it already covers `_sweep_leftovers` — which drops *other sessions'* suite-prefixed databases and is the most destructive operation in the file. The suggested widening enumerated three fixtures and did not name the sweep. A list of fixture names goes stale the next time one is added; the failure mode of a stale list here is an unguarded destructive sweep.

**It can also name its holder** — `pg_locks` gives pid/user/application. The `flock`'s own timeout message had to tell operators to go hunting instead.

## The cost, stated rather than hidden

The `flock`'s tests needed no PostgreSQL, argued at the time as "the lock has to hold when the database layer is already contended, so its own proof should not depend on the contended thing". That is given up. Three path-shape tests died with the mechanism; the two property tests that carried the guarantee are rewritten against the advisory lock and now require the cluster. Contended is not unavailable, and every other test in that directory already requires it.

## Verification

- `tests/scripts/` — **159 passed** (160 before, minus one test deleted in review; see below).
- **Serialization re-proven, not inherited.** Two concurrent sessions of the role suites, both green (`23 passed` each). The queueing is shown directly rather than inferred from wall-clock — the second run logs `waiting for the suite cluster lock (held by (439590, 'test_user', ''))`, naming the first run's backend. An earlier reading of "zero waits" was a pytest output-capture artifact, not evidence of anything.
- **Mutation-proven.** Removing the acquisition from `admin_conn` — leaving the suite with no mutex at all — reddens exactly the two tests that gate it (`2 failed, 2 passed`), and leaves the two scratch-key property tests green, which is correct: they test the primitive, not the session's use of it.

## `/simplify` pass, and one finding that had teeth

Three reviewers independently flagged that `test_the_hold_is_visible_and_attributable` re-implemented the very query it claimed to test — so it asserted a property of PostgreSQL, and the real wait-loop query could have broken while it stayed green. `lock_holder()` is now extracted and both call it. Mutation in the new direction confirms the fix: breaking the **production** query reddens the test, which the duplicate could never do.

Fixing it in one place also corrected a latent misattribution the duplicate had doubled. The query matched `objid` alone. Measured:

```
one-arg pg_advisory_lock(7532026)  -> classid=0 objid=7532026 objsubid=1
two-arg pg_advisory_lock(0,7532026) -> classid=0 objid=7532026 objsubid=2
```

Same `objid`. The loose form could name an unrelated lock's holder; the shared version matches the full identity.

Also applied: `maintenance_conn()` replaces a third spelling of "connect to the maintenance database"; `SCRATCH_LOCK_KEY` moved beside `SUITE_CLUSTER_LOCK_KEY` so key allocation has one home; the kill-test poll is bounded by the clock rather than an iteration count that overran the 15s its own failure message claimed.

**One test deleted.** `test_it_reaches_a_process_sharing_nothing_but_the_cluster` set `TMPDIR` in a forked holder — but with the `flock` gone nothing reads it, so it ran the same mechanism as the test above it and could not fail for the reason its docstring gave. Removed rather than kept as decoration; the scope argument stays as prose in the key comment.

**One finding skipped:** extracting the fork/hold ceremony. Two sites remain after the deletion, and the crash-safety test must deliberately *not* release — a shared context manager would hide the one thing that test is about.

## Named gaps, filed rather than bolted on

1. **`run_bootstrap(dsn)` takes a bare DSN**, so the CREATE side of role provisioning is covered by the session lock through every current call path but is not enforced by signature the way `drop_service_roles(admin_conn, ...)` is. A runtime assertion cannot substitute — advisory locks are re-entrant within a session, so a self-check returns `True` and detects nothing.
2. **`tests/scripts/` is not xdist-safe** while two testing guides recommend `pytest -n auto`. This change sharpens it: the removed `flock` failed loudly at 300s naming its file; the survivor waits to 1200s with its wait message captured by pytest. CI runs serially and is unaffected.

Both are out of scope for a consolidation PR and neither is a regression in correctness.

Lane 5 is serialized #769 → #785 → #773; #773 is next.


chrisrogers37 (commented):
**Approve**

The inversion holds up. Ran all three checks empirically rather than reasoning about them — worktree off `pr-807-review`, local PostgreSQL 15.18.

**(1) Cluster-scope — the load-bearing claim, confirmed independently.** Created a role while connected to `postgres`, then queried `pg_roles` from a second, unrelated database in the same instance — the role was visible there too. Roles genuinely are cluster-global, not per-database; this isn't storydump-specific, it's PostgreSQL architecture. And it's not a new premise invented for this PR: `tests/scripts/conftest.py`'s module docstring already said *"several bots share this cluster"* before #785 touched the file, citing #758/#763/#768. A `flock` is inherently filesystem-local — it cannot reach a process on a different host/container by construction, no measurement needed for that half. The scope argument is real, not asserted.

**(2) Crash-safety — reproduced from scratch, not trusted.** Own script, own throwaway key (not `SUITE_CLUSTER_LOCK_KEY`, to avoid touching anything live): forked a holder, confirmed the lock was held, `SIGKILL`'d it (exitcode `-9`, no `finally`/`atexit` can run), then confirmed a separate connection could immediately take the lock. Matches the PR's claimed result exactly — PostgreSQL's own backend-death detection releases the session's advisory locks, no application cleanup path involved. "Shared, not exclusive to `flock`" is correct.

**(3) Does the PR engage the issue's actual reasoning, not a strawman?** Read #785 directly rather than inferring from the PR's framing. The issue's "Suggested shape" gives exactly two grounds for keeping the `flock`: *"needs no held DB connection"* and *"survives a killed holder."* The PR addresses both, not a reconstructed version of them — crash-safety is rebutted with the measurement above, and DB-independence is **conceded as a real, deliberate cost** ("That is given up") rather than argued away. That's the right way to invert a recommendation: name what the original argument gets right and say what you're trading for what.

## Extra mile: tried to break the mutation-testing claim, then had to retract that

Applied the exact mutation myself (replaced `admin_conn`'s `pg_try_advisory_lock` call with `SELECT true`, so the session never actually acquires) and ran `test_suite_cluster_lock.py`. First result: **4 passed, 0 failed** — the opposite of the claimed `2 failed, 2 passed`, which would mean the regression gate is inert. Traced it before reporting it: rapid back-to-back reruns (no gap) intermittently caught a still-closing connection from my *own* previous run still holding `SUITE_CLUSTER_LOCK_KEY`, which made `_try_take` return `False` for the wrong reason and the mutated test passed by accident. Re-ran with a 5s gap and `0` confirmed-held locks before each start: **3/3 clean `2 failed, 2 passed`, exact match to the PR's claim.** The mutation is real and the gate works; my first result was an artifact of my own rerun cadence, not the code. Flagging the investigation anyway since the failure mode is real even if not a defect here — I initially took a false signal at face value, then a rigorous review can't stop at the first result when it contradicts a specific claim.

One non-blocking hardening thought, given the false start above: `test_admin_conn_holds_the_suite_lock`'s assertion is *"is anything holding the key"*, not *"is `admin_conn`'s own backend holding the key."* On a cluster genuinely shared by other sessions (which this codebase says is the normal case), those aren't quite the same question. `lock_holder()` already returns a pid — comparing it against `admin_conn`'s own `pg_backend_pid()` would make the regression gate immune to exactly the kind of coincidental-holder confusion I hit, for free. Not required; the gate is correct as shipped, just not maximally specific.

## One small mismatch, mechanical

CHANGELOG.md says *"leaves the three scratch-key property tests green"*; the PR body says *"two,"* and the actual file (`test_suite_cluster_lock.py`) has exactly two tests using `SCRATCH_LOCK_KEY` (`test_a_second_holder_is_blocked_while_the_first_holds`, `test_it_clears_when_the_holder_is_killed`). CHANGELOG's "three" is the wrong one — one-word fix.

## Bottom line

Both of the issue's stated grounds for keeping `flock` were engaged honestly, not overridden. The scope argument is the correct reason to invert, it's independently verifiable, and it isn't new information the issue author lacked — the shared-cluster fact was already in the file. Ship it; fix the CHANGELOG word whenever convenient.

PR #811, opened 2026-08-13
docs(changelog): two scratch-key property tests, not three (#785)

chrisrogers37:
One word. The #785 entry said the mutation "leaves the **three** scratch-key property tests green". There are two.

## Verified, not relayed

```
$ awk '/def test_/{t=$0} /SCRATCH_LOCK_KEY/{if(t)print t}' tests/scripts/test_suite_cluster_lock.py | sort -u
    def test_a_second_holder_is_blocked_while_the_first_holds(self):
    def test_it_clears_when_the_holder_is_killed(self):

$ grep -c "def test_" tests/scripts/test_suite_cluster_lock.py
4
```

And the mutation the sentence describes, re-run on merged `main` (remove the lock acquisition from `admin_conn`, leaving the suite with no mutex at all):

```
FAILED ...::TestTheSessionActuallyHoldsIt::test_admin_conn_holds_the_suite_lock
FAILED ...::TestTheSessionActuallyHoldsIt::test_the_hold_is_visible_and_attributable
2 failed, 2 passed
```

Two gate tests redden, two scratch-key tests stay green. The claim is otherwise exactly right — those two test the primitive, not the session's use of it, so they *should* survive that mutation.

## How it got there

It was three when the sentence was written. The `/simplify` pass on the same PR deleted the third — `test_it_reaches_a_process_sharing_nothing_but_the_cluster`, which set `TMPDIR` in a forked holder but, with the `flock` gone, had nothing reading it and ran the same mechanism as the test above it.

I updated the mutation **result** two lines further down and did not sweep back to this clause. A count written before a deletion, in a sentence the deletion did not touch.

## Everything else in the entry was swept, and is correct

Audited every numeral in the #785 entry rather than fixing only the one flagged, since a point-fix is how this one survived in the first place. One wrong number; the rest hold. In particular the two *other* "three"s are both right and are deliberately not changed:

- "the suggested widening named **three** fixtures" — #785 names `owner_actor`, `owner_window_db`, `at49_window_db`.
- "**three** path-shape tests died with the mechanism, and the **two** property tests that carried the guarantee" — the deleted `test_service_role_lock.py` had exactly three in `TestTheLockIsKeyedToTheCluster` and two in `TestItActuallyExcludes`.

Found by `navi` reviewing #807. CHANGELOG-only; no code touched.


chrisrogers37 (commented):
**Approve**

Proportionate check, three things:

1. **The diff is exactly the flagged clause.** `+1/-1`, one file, and it's the "Mutation-proven" bullet's "three scratch-key property tests" → "two." Nothing else in the hunk moved.
2. **The two genuinely-correct "three"s are untouched.** Since only one line changed in the whole file, lines 27 ("named three fixtures") and 29 ("three path-shape tests died... the two property tests") are byte-identical to before — confirmed by the additions/deletions count alone, not just by reading context.
3. **Swept the rest of the entry myself rather than trust the sweep claim.** Checked every count-shaped claim in the `#785` entry against source: "two stated grounds" (issue names exactly 2: no-held-connection, survives-a-killed-holder), "three fixtures" (issue names exactly 3: `owner_actor`, `owner_window_db`, `at49_window_db`), "three path-shape / two property tests" (deleted file: `TestTheLockIsKeyedToTheCluster` had 3, `TestItActuallyExcludes` had 2), "two tests that gate it" (matches my own mutation runs). All hold. Didn't re-run the "23 passed" concurrent-session figure — that's a verification-run output, not a code-derivable count, and re-running it is exactly the depth this doesn't need.

Nothing stale left in the entry. Ship it.

PR #812, opened 2026-08-13
fix(auth): record the real initData rejection reason, not the URL-token fallback's (#737)

### LINKED ISSUE #737: auth_monitor records the wrong reason for every initData rejection

chrisrogers37:
Closes #737.

`_validate_auth` is a fallback cascade — initData first, then a signed URL token — and only the **second** exception ever reached `auth_monitor`. A real initData querystring URL-encodes its colons as `%3A`, so it never presents four colon-separated parts and dies at `validate_url_token`'s format check before any check can describe the real problem. Bad signature, expired, future-dated and wrong-bot-token all arrived spelled `"Invalid token format"`.

This is a **diagnosis** fix, not an access fix: the credential was correctly rejected either way. What it cost was the signal at the moment it is most wanted — a Telegram clock-skew event and a leaked-token probe both look exactly like ordinary scanner junk once every reason is the same string.

## What changed

**Both reasons are recorded, attributed to their format** (`initData: …; urlToken: …`), rather than picking one by the shape of the input. Shape is a heuristic and it would misread precisely the input worth reading correctly — a malformed initData that happens to contain a literal colon. Carrying both never guesses, and the URL-token branch is a real credential type (browser links) whose reason is not a formality.

**The client-facing 401 detail is decoupled from the recorded reason** (`AUTH_FAILURE_DETAIL`). This is *forced by* the fix rather than bundled with it: the detail **was** the recorded string, so recording both reasons while leaving them coupled would have handed every rejected caller strictly more information than before. The specific reason separates "well-formed but mis-signed" from "expired" from "wrong credential type entirely" — the discrimination a probe wants and a legitimate client never needs. Nothing is lost that callers had, since the reason they received was already the URL-token branch's `"Invalid token format"` whatever the real cause.

The issue left this as an open question ("worth deciding whether the 401 `detail` should stay deliberately vague"). Recording both reasons is what settles it — the two options are not independent.

## Verification — paired measurement on the real cascade

`validate_url_token` is deliberately **not** patched anywhere in this work, so the second reason is whatever the real code produces rather than one a mock was told to produce.

Same probe, same three inputs, both arms:

| | monitor receives | client 401 detail |
|---|---|---|
| **unfixed** (`origin/main`) | `'Invalid token format'` ×3 | `'Invalid token format'` |
| **fixed** | `'initData: <real reason>; urlToken: Invalid token format'` ×3 | `'Invalid authentication credentials'` |

Reason preserved: **NO — LOST** on all three unfixed, **YES** on all three fixed. That reproduces the issue's measured claim exactly.

The first attempt at this control was **weak evidence and is worth naming**: reverting the implementation with the new tests in place gives `rc=2`, a *collection* error, because the test module imports `AUTH_FAILURE_DETAIL`. The whole file "fails" for an import reason that says nothing about whether the behavioural assertions discriminate. The probe above drives the real cascade instead.

Two controls pin the cascade itself — a valid initData and a valid URL token must still authenticate. Without them every assertion here is satisfied by a chokepoint that rejects everything.

`tests/src/api/test_helpers.py`: **16 passed**. `tests/src/api/ -m unit`: **141 passed, rc=0**. `ruff check` / `ruff format --check` clean on both changed files.

## `/simplify` applied

Four review angles; the substantive findings and what I did with them:

- **Dropped a duplicate `logger.warning`.** `auth_monitor.record_failure` already logs the same ip and reason *plus* the running failure count (`auth_monitor.py:36`), so the added line was a strict subset. Three of four reviewers converged on this and one asserted the opposite; settled by reading the source rather than by vote. The file's three sibling log sites each carry a `user_id`/`chat_id` the monitor cannot, which is why they are not precedent for this one.
- **Trimmed the source comments to durable semantics** — the `%3A` mechanism and the rejected shape-heuristic — per the repo's convention-over-event rule. The defect narrative now lives in the test docstring and this changelog entry, not in three drifting copies.
- **Folded two subsumed tests into one** that also catches a label swap (`initData: Invalid token format` is a different and false claim), which none of the three separately caught. Reused the conftest `mock_validate` helper for the control.
- **Scoped `AUTH_FAILURE_DETAIL`'s comment to the cascade** and stated the constant-detail contract in `_validate_auth`'s docstring, where its ~37 callers look.

**Deliberately not done**, all three recommended against on evidence: moving the reason into `auth_monitor` (the `reason` is write-only there — `_failures` is `dict[str, list[float]]` and never stores it); typing the reason on the strength of #760 (that issue evicts *keys*, which carry no reasons); and rewriting the cascade as an iterated validator list (N=2, and the fixed try-order is load-bearing at a security chokepoint).

## Two things found while here, neither fixed in this PR

1. **`main` is currently ruff-red**, independent of this branch — `F811 Redefinition of unused test_check_response_errors_revocation_subcode_458` at `tests/src/services/test_instagram_api.py:940`, duplicating `:388`. Confirmed on a pristine `origin/main` worktree, so CI is failing for everyone, not just here. Not bundled in — unrelated cleanup.
2. `tests/src/api/test_security_hardening.py` threw a **teardown** error (`must be a superuser to terminate superuser process`) on one run while a peer's suite was running against the same shared Postgres. It did not recur once that cleared. Flagging rather than claiming it resolved.

## Base

Branched off `e2d1bff`, which carries #801, #803, #805, #807 and #811. Two-dot and three-dot diffs are identical — no stale-base cargo.


chrisrogers37 (commented):
**Approve**

Confirmed CI is fully green before posting (Test had been in-progress at dispatch time): Lint, Test, Security Scan, Changelog Check, GitGuardian all `COMPLETED`/`SUCCESS`; `Vercel` shows `null` which is expected for a backend-only diff (no frontend files touched).

Built my own A/B and my own mutation rather than re-run the PR's — worktrees at `origin/main@e2d1bff` (unfixed) and the PR head (fixed).

**(1) Is the A/B real — same cascade, genuinely distinct reasons?** Wrote a standalone probe: 3 reasons drawn from the issue's own named categories (bad signature / expired / future-dated — distinct strings, asserted distinct), a real initData-shaped querystring with zero literal `:` characters (so it's honestly unparseable by `validate_url_token`'s 4-part split, the exact mechanism the issue names), and `validate_url_token` left **unpatched** in both arms — only `validate_init_data` is mocked, since that's the input under test, and `auth_monitor` is mocked purely to capture what it's told.

Ran it against both checkouts:
- **Unfixed** (`origin/main`): all 3 collapse to `'Invalid token format'` — confirmed LOST 3/3, independently reproduced.
- **Fixed** (this PR): all 3 preserved, e.g. `'initData: initData auth_date is in the future; urlToken: Invalid token format'` — confirmed PRESERVED 3/3. The `urlToken:` half is consistently `'Invalid token format'` across all three fixed-arm results, which is the tell that the real (unpatched) `validate_url_token` executed identically in every case — the two arms differ in exactly the fix, nothing else.

This matches the PR's table exactly, from a probe I wrote myself with inputs it didn't specify (the table doesn't show the 3 reasons or confirm they're distinct — worth naming as a gap in the PR's own evidence, even though the underlying claim holds).

**(2) Substitution, not absence — do the assertions bind to the specific reason?** Read the committed tests first: `test_both_reasons_are_recorded_attributed_to_their_formats` asserts `"initData: <reason>"` and `"urlToken: <reason>"` as bound label+content pairs, specifically so a label swap — the plausible substitution shape, not a blank-field shape — fails it. Verified by applying exactly that mutation to the fixed checkout (swapped which error fills which label in the f-string) and running the real suite:

```
FAILED test_both_reasons_are_recorded_attributed_to_their_formats
AssertionError: assert 'initData: Invalid initData signature' in 'initData: Invalid token format; urlToken: Invalid initData signature'
1 failed, 4 passed
```

Exactly the one test whose docstring claims this job catches it; the other four correctly stay green (they don't test labeling). An "a reason was recorded" test would have missed this entirely — these don't.

Both of your scrutiny points hold on independent reproduction, not just on re-reading the claim. The design itself is sound too: decoupling the client-facing detail from the recorded reason is the correct call given the detail *was* the recorded string (kept as-is would have handed every prober strictly more information), and the two controls (valid initData / valid URL token still authenticate) rule out the degenerate "reject everything" fix.

Ship it.

Issue #806, opened 2026-08-13
[Refactor Phase F] F.2 segment tracker — COMPLETE: 9 of 9 segments landed as migrations 052–060 (2026-08-19), applied to production 2026-08-26


chrisrogers37:
> **This issue is the F.2 segment tracker** (retitled 2026-08-19 — the original title named Fork 1 as the blocker, which is resolved: locked (a) 2026-08-13, executed by #820). **Current state: 2 of 9 segments landed** — 052 (#813) and F.2.2 (#822 + #828, migration 053 + target models). **F.2.3–F.2.9 remain, and the order is binding** (the prefix gate positionally rejects out-of-order landing). Authority on content: [`documentation/planning/2026-08-14-f2-increment-split/README.md`](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-14-f2-increment-split/README.md) (ratified). Full state verdict with evidence: [comment 5337533705](https://github.com/chrisrogers37/storydump/issues/806#issuecomment-5337533705). The original body below is the tracker's history — its complaint is fully resolved.

> **Rewritten 2026-08-13 after scoping.** The original framing — *"F.2 is unowned and unscoped"* — was **substantially wrong**, and is retracted below. What survives is narrower and has a concrete blocker. Original text and the scoping that corrected it are in the comments.

## Retraction first

This issue was filed claiming F.2 needed an owner and a scope. **It has both.** Scoping against `origin/main` at `6c130aa` found:

- **F.2 is tightly specified.** Its content is pinned to `scripts/advertised_ddl_manifest.json`, a committed classification of every ` ```sql ` block in `02`/`07`, **keyed by content sha256** so an edited block falls out of the bijection and reddens a ratchet gate until re-classified by hand. 22 blocks: 18 normative, 4 illustrative with recorded reasons.
- **Its scope is closed and cross-checked two independent ways.** The shipped extractor yields **26 `CREATE TABLE`** over 257 statements; the filed split lists 7+6+4+4+5 = **26**. Two artifacts, written at different times by different hands, agreeing exactly.
- **It is already split into filed increments F.2.0–F.2.7, three of which have shipped.**
[...]

PR #813, opened 2026-08-13
feat(f2): migration 052 — the two shared trigger functions (#806)

chrisrogers37:
The first file of the F.2 target lineage: `02` §0's two shared trigger functions, `trg_touch_updated_at()` and `fn_safe_tz()`. F.2.1's one surviving item, which did not ship with the bootstrap in #752.

Scope and content come from `mason`'s scoping on #806 (comment 5286300366). I re-derived both rather than taking them.

## The number is derived, and I ran the resolver rather than reading it

```
schema-move marker resolves to: version=51  051_schema_move_public_to_legacy.sql
highest version present       : 51
target lineage (files > move) : []
=> next number is             : 052
```

`051` declares itself the corpus lineage boundary and `target_lineage_files()` asks the runner's own `runner:schema-move` marker rather than the literal 51 — so renumbering the boundary moves the lineage with it. My reading agrees with the one relayed to me.

## The content is copied from the stream, not authored

052 must be an ordered **prefix** of the advertised stream, so I extracted the stream's opening statements and matched them rather than writing DDL that looked right:

```
[0] CREATE FUNCTION trg_touch_updated_at() RETURNS trigger LANGUAGE plpgsql AS $$ ... $$
[1] CREATE FUNCTION fn_safe_tz(p_tz text) RETURNS text LANGUAGE plpgsql STABLE AS $$ ... $$
[2] CREATE TABLE users ( ... )      <- F.2.2, Fork-1-blocked, deliberately not here
```

**No table, and that is the point.** The open question about whether a table may land before its policy governs the table increments; a file that creates none is not subject to it. Independent of both rulings outstanding with Chris.

`fn_safe_tz` is `STABLE` and deliberately not `IMMUTABLE` — its answer depends on the server's tzdata, which changes.

## Arm (b) is load-bearing from here, on purpose rather than by accident

`test_target_lineage_is_empty_today_disclosed` asserted the lineage was empty and was written to go red the moment a file landed. It went red. Retired on its own terms and replaced with the assertion it was holding the place for:

```
PREFIX REPORT: ok=True vacuous=False f2_count=2 first_divergence=None
```

Two things I want to flag explicitly, because both are ways this could have become a green that means nothing:

1. **The expectation is derived from the STREAM — plan docs plus manifest — never from the migration file**, which is the artifact under test. Derived from the file, the test passes for any content whatsoever.
2. **`test_the_wired_prefix_holds_against_the_real_stream` asserted only `ok is True`, and an empty prefix satisfies that vacuously.** Adding `vacuous is False` is the load-bearing line: without it, deleting every target-lineage migration would turn that test **green** rather than red — reinstating the exact hole the disclosure existed to flag, at the moment the disclosure was removed.

## A second disclosure existed, and the suite found it — not me

`test_lineage_lane.py::test_the_target_lineage_is_currently_empty_and_says_so` made the same emptiness claim and even named the two functions as what it expected to arrive. I was pointed at the `advertised_ddl` one only, and found this by running the suite.

Updated on its own terms, both halves it asked for:
- the lineage is exactly `['052_shared_trigger_functions.sql']`;
- a new lane test asserts what the replay **creates** — the functions land in the `public` that 051 re-creates.

**Lane parity is left vacuous, deliberately, and now says why.** The lane installs two functions, but `schema_signature`/`create_all` are relation-scoped on both sides and a function lives in `pg_proc`, not `pg_class` — so 052 moves neither side. F.2.2's first table switches that gate on. Stating it beats leaving a reader to infer it from a green.

## Verification

**Applied to a real scratch database**, not only diffed:

```
APPLIED cleanly
  PASS  postcondition: trg_touch_updated_at exists
  PASS  postcondition: fn_safe_tz exists
  PASS  SELECT fn_safe_tz('America/New_York') = 'America/New_York'
  PASS  SELECT fn_safe_tz('Not/AZone') = 'UTC'
  PASS  trigger actually touches updated_at
  PASS  CHECK rejects an unrecognized zone (the gate's point)
```

**Mutation-proven, six ways**, because a gate I just made load-bearing is exactly the kind that can be load-bearing in name only:

| mutation | result |
|---|---|
| `STABLE` → `IMMUTABLE` (diverge from stream) | both prefix tests red |
| delete 052 → empty lineage | **3 red** — and an empty lineage is today's *passing* state on `main` |
| smuggle a `CREATE TABLE` into 052 | no-table assertion red |
| remove the `EXCEPTION` fallback | lane install test red |
| stub `fn_safe_tz` to always return `'UTC'` | lane install test red |
| drop `trg_touch_updated_at` | lane install test red |

The tz assertion is a **pair** for the reason the last two rows show: a passthrough stub satisfies the first half and a constant-`'UTC'` stub satisfies the second, so either alone is worthless.

**Suites:** `tests/scripts/` 169 passed. Full suite **2579 passed, 8 skipped, 7 failed** — the 7 are the `aiolimiter` optional-dep failures in `test_telegram_service.py`, confirmed pre-existing today against a clean `origin/main` worktree, not asserted. Net test delta from this branch is **+2** (4 added, 2 retired); the rest of the count movement is base drift from #803.

No `schema_version` self-stamp — that table is in legacy from 051 on, per the boundary file's own note. The runner ledger records this one.


chrisrogers37 (commented):
**Approve**

Verified by reproducing, not by reading. Ari's four checks, in order, plus the six-mutation table.

## 1. No table in 052

Confirmed from source: the migration is exactly two `CREATE FUNCTION` statements (`trg_touch_updated_at`, `fn_safe_tz`), nothing else. Fork 1's prefix-gate-vs-policies-with-tables conflict genuinely does not reach this file — there is no table for the open ruling to apply to.

## 2. Arm (b) is load-bearing, not vacuous in a new shape

Confirmed by reproducing the mutation myself, not by trusting the claim: with 052 deleted, `target_lineage_files()` returns `[]`, so `f2_prefix_report([], stream)` hits the empty-input branch (`advertised_ddl.py:293`) and returns `vacuous=True`. The new `assert report.vacuous is False` catches exactly that. Real, not decorative — confirmed against the actual `F2PrefixReport` dataclass, not inferred.

**One correction to the stated count.** The PR body says deleting 052 reddens 3 tests. Running the full relevant selection across both touched files with 052 removed, I get **5**: the 3 in `test_advertised_ddl.py::TestF2AgainstReality` (matches, if the claim was scoped to that file alone) plus 2 more in `test_lineage_lane.py` — `test_the_target_lineage_is_the_shared_functions_file` and `test_the_lane_installs_the_shared_functions` — that the same single mutation also reddens. Not a defect: the gate is *more* protective than claimed, not less. But the number as literally stated in the PR body/CHANGELOG is measurably incomplete, and this thread's own convention is measured counts, not remembered ones. Worth a one-line fix before merge; not worth blocking on.

## 3. Mutant validity — reproduced 3 of 6 against a real DB, all valid

- **delete 052** → all 5 failures above are on-topic, each naming the real cause (empty list, IndexError on `[0]`, `vacuous=True`) — no inert reds.
- **smuggle a `CREATE TABLE` into 052** → 3 valid failures: the no-table assertion, the prefix-match assertion, and `F2PrefixReport(ok=False, first_divergence=2)` — diverging at exactly the index the smuggled statement sits.
- **stub `fn_safe_tz` to always return `'UTC'`** → caught, but one step earlier than the PR body implies: the migration's own `runner:postcondition SELECT fn_safe_tz('America/New_York') = 'America/New_York'` raises `MigrationRunnerError` during lane replay, before the Python-level pair assertion in `test_the_lane_installs_the_shared_functions` ever runs. Still a real catch — just worth knowing the postcondition is doing the work here, not the pair assertion in the test body.

Didn't reproduce the remaining 3 (`STABLE`→`IMMUTABLE`, drop the `EXCEPTION` fallback, drop `trg_touch_updated_at`) against a live DB — the shared test cluster had lock contention (one real run above took 139s), and each maps structurally onto a mechanism already confirmed: file-content mutations go through the same `normalize_statements`/`f2_prefix_report` prefix-diff proven above, DB-behavior mutations go through the same postcondition/list-comparison proven above. Disclosing this as 3-of-6 hands-on rather than claiming a full 6/6.

## 4. Disclosure sweep — confirmed complete, no third

Swept `tests/` and `scripts/` two ways: by emptiness vocabulary (`vacuous`, `disclosed`, `is currently empty`, etc.) and structurally by the two function names. The only other "empty until X" disclosure is `test_tenancy_gate.py`'s `test_states_its_own_coverage_rather_than_passing_silently`, keyed to F.2.2's first workspace-keyed table — it replays only the legacy-bounded lineage (`LEGACY_LINEAGE_MAX`), so 052 never reaches it. Confirmed directly: my own delete/restore-052 cycle above didn't move that test either way. No third stale disclosure exists — the sweep in this PR was actually complete, not partial.

---

Also independently re-checked: the migration-number derivation (marker→51, empty target lineage→052) and the `STABLE`-not-`IMMUTABLE` reasoning both hold from source.

Good work — the gate is doing its job, and doing it more thoroughly than the PR body gives it credit for.

PR #814, opened 2026-08-13
fix(auth_monitor): evict stale sources and bound the map, least-recently-active first (#760)

chrisrogers37:
Closes #760.

Pruning ran only for the key currently being written, so a source that never recurred kept its entry for the life of the process. The map grew with the number of **distinct** sources ever observed rather than the number currently active — and `WINDOW_SECONDS` never bounded that: it governs how long a failure **counts**, not how long an entry is **retained**. It also grew precisely in the case that alerts nothing, since a source seen once never approaches `FAILURE_THRESHOLD`.

## Two mechanisms, because they bound different things

The issue offers a sweep and mentions a cap as complementary. They are, and neither alone closes it:

- **Amortised global sweep** drops sources whose failures have all aged out. Bounds the *steady state* — but only between its runs, and it is gated to `WINDOW_SECONDS` (600s).
- **`MAX_SOURCES` cap** is what holds during a *burst*. A distributed caller mints keys far faster than the sweep interval, and 600s of unbounded growth is the stated threat model, not an edge case.

## The eviction ordering is the load-bearing decision

**Evict by least-recent failure, deliberately not by first-seen.** The two invert on exactly the case the map exists for: a source that started failing early in the window and is *still* failing has the **oldest** first-seen. First-seen ordering would therefore evict the source closest to `FAILURE_THRESHOLD` — turning a memory bound into a missed alert — while keeping one-shot scanners.

`sorted()` on insertion order is the free and tempting alternative here, and it silently *is* the wrong policy. So it is pinned by test rather than only argued in a comment: swapping the sort key to `[0]` reddens exactly one test and nothing else.

**The cap's lossiness is stated rather than papered over.** At the ceiling something must go, and an evicted source restarts its count; ordering only guarantees the victim is the least active one *available*. The over-cap path runs a forced sweep first, so stale keys — which cost nobody their count — are always surrendered before a live one is.

## Verification — mutation, each reddening exactly the test that names it

| mutation | reddens |
|---|---|
| sweep disabled | `test_keys_whose_failures_all_aged_out_are_evicted`, `test_a_source_still_inside_the_window_survives_a_sweep` |
| cap disabled | `test_the_map_never_exceeds_the_cap_when_no_source_is_stale` (+ the ordering test, which needs the cap to fire at all) |
| evict by **first-seen** | **only** `test_eviction_keeps_the_source_closest_to_alerting` |

The harness asserts each mutation actually landed before trusting its result, and that the file restores identical afterwards — a mutation that silently fails to apply reads as "the tests caught nothing".

The regression test was confirmed **red on unfixed code** before the fix existed. Per the issue's own warning, the clock advance is what makes it discriminate: without it the same assertions pass against the current code and prove nothing.

Two guard tests hold what the fix must not break: a source inside the window survives a sweep, and a source partway to the threshold still alerts after a sweep has run.

`tests/src/utils/`: **117 passed**. `tests/src/utils/ tests/src/api/ -m unit`: **239 passed, rc=0**.

## `/simplify` applied

Reviewed on four angles. One substantial finding, measured end-to-end rather than argued:

- **Evict to a low-water mark (90% of `MAX_SOURCES`), not to the ceiling.** Landing exactly on the cap means the next novel source re-trips the check, so the `O(n log n)` selection ran on *every* request for the duration of a burst — **4.955 ms/req vs 0.016 ms/req**, held under the lock, on the single uvicorn event-loop thread. The same sources are evicted in aggregate; only the granularity changes.
- Deriving that mark at call time rather than storing a second constant — caught by my own tests failing, which was the right answer: a constant computed from `MAX_SOURCES` at import time cannot track it and silently drifts if the ceiling moves.
- **Made one guard test actually discriminate.** `test_a_source_still_inside_the_window_survives_a_sweep` previously never had a live source present while a sweep ran, so it passed for the wrong reason. It now reddens under the sweep-disabled mutation, which is the proof the rework worked — it did not before.

**Deliberately not done**, each recommended against on evidence: extracting a shared bounded-TTL-map utility (n=1; `_failures` is the only map in the repo fed by unauthenticated caller-chosen keys); promoting `MAX_SOURCES` to settings (it is a memory ceiling, and its two siblings in the same file are constants); a class for `_last_sweep` (one float against 29 test references); and `sorted(d.items(), ...)`, which measured **1.7x slower** than the current form.

## Independent of #812

#760 needed nothing from #737. Different module (`src/utils/auth_monitor.py` vs `src/api/routes/onboarding/helpers.py`), no shared function, no data dependency — `auth_monitor` never stores the reason string at all, only timestamps. The sole contact point was `CHANGELOG.md`.

## Base

Rebased onto `8ef210b`, which carries #812 and #813. Two-dot and three-dot diffs are identical and insertion-only — no stale-base cargo and nothing reverted.

## Pre-existing CI failure, not from this branch

`ruff check src/ tests/` fails on `main` with `F811 Redefinition of unused test_check_response_errors_revocation_subcode_458` at `tests/src/services/test_instagram_api.py:940`, duplicating `:388`. Confirmed on a pristine `origin/main` worktree earlier and re-confirmed still present at `8ef210b`. It will fail CI here too. Not bundled in — unrelated cleanup — but it is currently red for everyone and wants an owner.


chrisrogers37 (commented):
**Approve**

CI confirmed fully green before posting (Test was in-progress at dispatch time; now `COMPLETED`/`SUCCESS` along with Lint, Security Scan, Changelog Check, GitGuardian — `Vercel` is `null`, expected for a backend-only diff).

## The eviction-ordering design — reasoned through independently before checking the test

Worked out the LRA-vs-first-seen question by hand first, before reading how it was tested. A source that started failing early and is *still* failing has the oldest insertion position but the newest last-activity timestamp — first-seen ordering would evict it first among a burst of one-shot noise, discarding exactly the accumulating evidence the monitor exists to produce. That's a real inversion, not a preference: I'd reached the same conclusion the PR states before reading its reasoning.

Confirmed the code structurally cannot be anything but LRA — `_failures` stores only a timestamp list per source, no first-seen field exists anywhere, and both `_sweep` and `_evict_to_cap` read `ts[-1]` (most recent). Then mutated `_evict_to_cap`'s selection from `sorted(_failures, key=lambda s: _failures[s][-1])` to plain `list(_failures)` (insertion order — the only way to get "first-seen" out of this data model) and ran the real suite:

```
FAILED test_eviction_keeps_the_source_closest_to_alerting
AssertionError: evicted the source closest to FAILURE_THRESHOLD
1 failed, 13 passed
```

Exactly the one test, for exactly the stated reason — the surviving source in the failure output is the nine one-shot scanners, not the repeat offender. Pinned, not just argued in a comment.

## The other two mutations — also valid, not just red

- **Sweep disabled** (`_sweep` short-circuited to a no-op): `2 failed, 12 passed` — exactly `test_keys_whose_failures_all_aged_out_are_evicted` and `test_a_source_still_inside_the_window_survives_a_sweep`, both failing on the literal stale-key-not-removed condition.
- **Cap disabled** (`_evict_to_cap` short-circuited): `2 failed, 12 passed` — exactly the cap test (`50 <= 10` fails outright) and the ordering test (which needs the cap to fire before ordering can matter, as the PR says).

Three mutations, three matching failure sets, each failing on the specific condition named — not a coincidental break elsewhere.

## The cap as an invariant, not a one-shot check

The existing test checks `len(_failures) <= cap` once, after a 50-write burst. Wrote a harsher version myself: 2000 writes against a patched cap of 25, mixing a 5-source "persistent attacker" pool (repeat offenders, so LRA-relevant entries stay warm) against ~1300 one-shot novel sources, checking the bound **after every single call**, all inside one `WINDOW_SECONDS` so the sweep contributes nothing and it's the cap alone under test:

```
max size observed at any point: 25 (cap: 25)
violations (size > cap at any single point): 0
persistent attackers still tracked at end: all 5
```

Never exceeded at any of the 2000 checkpoints, and the repeat offenders survived the entire run despite constant novel-source pressure — the same security property, holding under load an order of magnitude past what the committed test exercises.

## Cross-key bypass

`_failures` is written from exactly one place (`record_failure`'s `_failures[source] = timestamps`), always inside the same `with _lock:` block that runs `_sweep` and, over cap, `_evict_to_cap`. Grepped the whole repo: nothing outside `auth_monitor.py` touches `_failures` directly, and all four real callers of `record_failure` (in `helpers.py`) go through this one function. No path writes a key without also being subject to sweep-then-cap.

## Net

Both the design reasoning and its test hold up under independent reproduction — the security property is real, not asserted, and the three mutation claims and the invariant are all stronger on inspection than the PR's own evidence table, not weaker. Ship it.

PR #815, opened 2026-08-13
fix(health): close the two fail-closed gaps from the #791 merge (#794)

chrisrogers37:
Closes #794. Both gaps mason raised on #791, left open deliberately because each already failed closed.

## Why failing closed made these awkward to test, and what I did about it

A fail-closed path returns `healthy: False` — and so does nearly every other failure in this function. So an assertion on `healthy` alone passes whether or not the gap was ever exercised: a green earned by nothing having been tried. Every test here asserts what the sweep **did** — call counts, which tenant was reached, which counter moved — not merely that the result was unhealthy.

## Gap 1 — the per-tenant guard wrapped the CALL, not the unit

Reading a tenant's categories is that tenant's work, so a malformed row (a category with no `runway_days`) escaped to the outer handler and ended the sweep, losing the cross-tenant aggregate. It failed closed, but at a site that already *looks* fixed — the next reader sees a `try` and stops looking.

**The regression test is deliberately not a raising call.** The old guard already isolated those, so a test built on a raising tenant passes against the unfixed code and proves nothing. The discriminating failure is a call that RETURNS a malformed row.

**`checked` is committed only after the whole unit succeeds.** Otherwise a mid-unit failure counts the tenant both checked *and* unreachable, and `checked + unreachable` stops equalling the population — which is the only property that makes the disclosure worth anything. That invariant has its own test.

### The test failed on its own author, and that is the strongest evidence here

**My first version of the restructure contained the defect it was fixing, one line further on.** Written as `tenant_worst is None or cat_info["runway_days"] < ...`, the short-circuit skips the lookup for the first row — so a malformed row is adopted unread and raises at the commit, *outside* the guard. I moved the `try` and left a read behind it.

The new test failed on that first run, before any mutation was applied.

**Weigh this above the mutation table below.** Mutation testing is something I chose to do, designed, and could have designed to pass — it demonstrates the tests reject changes *I thought to try*. This was involuntary: the test detected the real defect in the real fix, written by the person who also wrote the test and believed it was correct. A test that passes when its author writes the fix proves only that the author agreed with themselves. A test that fails on its author proves it can detect the thing it claims to detect.

It is also the concrete argument for binding to the specific failure. A test built on a *raising* tenant — the obvious way to write this — would have passed against both the unfixed code and my broken fix, and shipped the defect twice.

## Gap 2 — `**coverage` missing from two return paths

Absent from the `if not active_chats` early return and from the outer `except`, so a caller reading `tenants_checked` / `tenants_unreachable` got them on the happy path and not on those two — re-opening, one level up, the never-happened-versus-lost confusion those keys were added to close.

Both now disclose. **They report the same `0 of 0` and are told apart by `healthy`:** an enumerated-empty population is `True`, an aborted sweep — where enumeration itself may have failed — is `False`. Asserted as a **pair**, since either verdict alone is satisfied by an implementation that always returns it.

A **control** guards the other direction: a reachable tenant with no categories is counted *checked*, not unreachable. Without it, a fix that counted every tenant yielding no worst-category as unreachable would satisfy every other assertion while reporting a fleet-wide outage.

## Verification

`tests/src/services/test_health_check.py` — **60 passed** (6 added, none removed).

**Mutation-proven, five ways**, each killing exactly the tests that encode it:

| mutation | red |
|---|---|
| narrow the guard back to the call only (the pre-fix shape) | malformed-row test + count-once test |
| commit `checked` before the unit completes | malformed-row test + count-once test |
| drop coverage from the empty-population return | empty-path test + distinguishability test |
| drop coverage from the outer `except` | aborted-path test + distinguishability test |
| flip the aborted sweep to `healthy: True` | aborted-path test + distinguishability test |

Full suite: **2583 passed, 8 skipped, 7 failed** — the 7 are the `aiolimiter` optional-dep failures in `test_telegram_service.py`, confirmed pre-existing today against a clean `origin/main` worktree.

**One run I am not counting as evidence:** an earlier full-suite attempt died at 30% with a segfault (rc 139) inside `psycopg2` under load on this shared host — `test_media_repository`, a connection path this change does not touch. It did not reproduce. Reporting it rather than quietly re-running, but it is not a result in either direction.

Rebased onto `8ef210b` after #812 and #813 landed; the CHANGELOG conflict with #812 was resolved keeping both entries intact.

## Found while here, filed separately

`ruff` reports `F811` on `tests/src/services/test_instagram_api.py` — `test_check_response_errors_revocation_subcode_458` is defined twice, and **the two bodies differ**, so the first definition's assertions never run. Pre-existing on `main`, unrelated to this change, filed rather than fixed here.


chrisrogers37 (commented):
**Approve**

CI green (Lint, Test, Security Scan, Changelog Check, GitGuardian, Vercel — all SUCCESS). The scrutiny ari asked for, reproduced rather than read.

## The recurring defect — confirmed real, not narrative

Reconstructed branden's described first attempt from the prose (short-circuit `tenant_worst is None or cat_info["runway_days"] < ...`, with the actual read deferred to a commit step *outside* the per-tenant guard) and applied it to the shipped fix. It reddens exactly the two tests the mechanism predicts:

```
FAILED test_a_malformed_category_row_is_isolated_to_its_own_tenant
  — "the malformed row ended the sweep: only 1 of 3 tenants examined"
FAILED test_a_tenant_that_fails_mid_unit_is_counted_once_not_twice
  — "coverage does not account for the population: 2 + 0 != 4"
```

Both messages are on-topic and match the described mechanism precisely — this isn't a decoy. The claim that "the test failed on its own author" is a real, reproducible property of the test, confirmed independently.

**On "a defect that recurs one line away tends to recur two lines away":** checked the rest of `health_check.py` for the same shape (a guard wrapping a call but not the per-item work on its result) — nothing else in this file has it; `check_media_pool_for_chat` always returns well-formed rows by construction, and `_check_media_sync` is single-subject so the cross-item-aggregate failure mode doesn't apply. Better than a clean sweep, though: `scheduler_loop.py::_pool_health_tick` (a different, real caller of `check_media_pool_for_chat` + `format_pool_alert`) already guards the *whole* per-tenant unit — call, format, send — in one `try`, with a comment explicitly citing #767 for this exact defect class ("a tenant that reliably raises would abort at the same position on every tick and permanently silence every tenant created after it"). This fix brings `_check_media_pool` up to a standard the codebase already established and named elsewhere. No third live instance found.

## Mutant validity — all 5 claimed mutations reproduced directly, all valid

No DB needed for these (`@pytest.mark.unit`, mocked services), so I ran every one for real rather than sampling:

| mutation | claimed red | reproduced | on-topic failure |
|---|---|---|---|
| narrow the guard back to the call only | malformed-row + count-once | ✅ | same 2 tests, same messages as above |
| commit `checked` before the unit completes | malformed-row + count-once | ✅ | `3 + 2 != 4` — exact double-count signature |
| drop coverage from empty-population return | empty-path + distinguishability | ✅ | `KeyError: 'tenants_checked'` |
| drop coverage from the outer `except` | aborted-path + distinguishability | ✅ | `KeyError: 'tenants_checked'` |
| flip aborted sweep to `healthy: True` | aborted-path + distinguishability | ✅ | "an aborted sweep is indistinguishable from an empty population" — the test's own point, stated back |

5/5, no inert mutants, every failure names the real cause. (Each mutation verified via `ast.parse` before trusting the test result, after an early one of mine turned out to be a syntax error in my own patch, not a real result — worth stating since a false "red" from a broken mutation is exactly the failure mode this whole exercise exists to avoid.)

## Housekeeping

- Full `test_health_check.py`: **60 passed**, matching the claim exactly.
- Checked the CHANGELOG merge-conflict resolution directly (not just trusted the claim) — #813's entry is fully intact, all sub-bullets present, well-formed markdown either side of the splice point. No corruption.

Clean work — the strongest part is that the regression test's design (a *returning* malformed call, not a raising one) is exactly what made it catch the author's own mistake, and that isn't an accident of luck.

PR #817, opened 2026-08-13
fix(tests): remove the shadowed duplicate of test_check_response_errors_revocation_subcode_458 (#816)

chrisrogers37:
Closes #816.

`TestInstagramAPIService` defined `test_check_response_errors_revocation_subcode_458` twice — line 388 and line 940. Python binds a class attribute once, so the first body never executed. Collection reports **57 tests either way**, so no count, tick or coverage number ever moved.

## They are the same test — proven, not read

The issue recorded that the bodies differ, measured from a textual diff. They do differ, in three places:

| axis | first (388) | second (940) |
|---|---|---|
| local variable | `mock_response` | `response` |
| dict literal key order | `code, message, error_subcode` | `message, code, error_subcode` |
| docstring | "Test TokenRevokedError for subcode 458 (app removed)." | "Test `_check_response_errors` raises TokenRevokedError for subcode 458." |

None of those reach the code under test. `_check_response_errors` reads the dict through `.get("code")` / `.get("error_subcode")` / `.get("message", …)`, so literal order is inert — this is the one axis that *could* have mattered (an implementation that iterated and short-circuited would be order-sensitive), so it was checked against the source rather than assumed.

Normalizing both ASTs for exactly those three axes — alpha-rename locals in binding order, sort dict keys, drop the docstring — yields **identical trees**:

```
found 2 definitions at lines [388, 940]
NORMALIZED ASTs IDENTICAL: True
POSITIVE CONTROL — a genuinely different test normalizes equal? False (must be False)
```

The positive control matters: without it, "identical" is equally consistent with a normalizer aggressive enough to collapse any two tests. Run against a real sibling (`test_check_response_errors_oauth_error_102`) it correctly reports different.

**So the condition for keeping both is not met.** There was no second intended assertion hiding in the shadowed copy — one of the two is redundant, and removing it loses nothing.

## Which copy survives, and why that direction

**Kept: line 388.** It sits inside the real `_check_response_errors` cluster with its five siblings (`…_oauth_error_190`, `…_oauth_error_102`, `…_general_api_error`, `…_media_unsupported_code_9004`).

**Deleted: line 940**, which sat alone under a *second, duplicate* `_check_response_errors Tests` section header appended later in the file — the header is removed with it, since it now heads nothing.

This deliberately keeps the copy that was **dead** and deletes the one that was running. It returns the test to the code it covers, and it is the direction that carries a burden of proof — which is discharged below.

## Mutation proof, both directions, instrument controlled

On `main`, an `AssertionError` planted as the first statement of each definition in turn:

| tripwire in | result | reading |
|---|---|---|
| **first** definition (388) | `rc=0` — `57 passed` | the body never runs |
| **second** definition (940) | `rc=1` — `1 failed, 56 passed` | **positive control**: the injection works |

The control is load-bearing. Without it, "57 passed" is equally consistent with the tripwire never having been inserted — a dead instrument and a dead test read identically.

On this branch:

| mutation | result |
|---|---|
| tripwire in the surviving definition | `rc=1` — `1 failed, 56 passed` |
| implementation: `error_subcode = None` (collapses the revocation branch) | `rc=1` — target test fails by node id |
| restored | `rc=0` — `57 passed` |

So the resurrected assertion both executes *and* has teeth against the code under test — a tripwire alone would only prove a pulse.

## Sweep — two independent instruments

Over **all 296 `.py` files** in `src/` and `tests/`:

1. `ruff check --select F811` — 1 error before, `All checks passed!` after.
2. A standalone AST pass enumerating every definition in every module and class scope, reporting any name bound twice in one scope (the structural container, not a name-pattern guess).

Both report this as the **only** shadowed definition in the repository.

**Disclosed:** the AST pass additionally flags `BaseRepository._db` (58/63) and `._db_generator` (68/73). Both are `@property`/`@setter` pairs — legitimate, and a known blind spot of that scanner, which does not read decorators. ruff, which does, correctly ignores them. That disagreement is why the sweep rests on the cross-check rather than either instrument alone.

**Not swept:** only `src/` and `tests/` were scanned. `cli/`, `scripts/` and `landing/` were not — the issue scoped this to test modules, and the CI change below now lints `tests/` rather than those. Cross-*file* name collisions are also out of scope by construction: shadowing is a within-scope property, and separate modules do not shadow each other.

## Second commit — the CI gate (separable)

The issue supposed CI does not gate on ruff. **It does**, and always has — the job runs `ruff check src/ cli/`. The actual gap is the path list: `tests/` is the one directory the gate never looked at, while `CLAUDE.md` tells contributors to run `ruff check src/ tests/`. Neither path covers the union, so `tests/` was linted by convention only, and a convention is what a rebase drops.

`ruff.toml` already selects pyflakes (`F`), so F811 was always in the repo's *declared* policy. Measured before proposing:

- on `main`: `ruff check src/ cli/ tests/` → `rc=1`, exactly this defect
- on this branch: `rc=0`, and `ruff format --check src/ cli/ tests/` → 309 files already formatted

No new rules, no findings to clear first — the widened gate is green on arrival. **Kept as its own commit** so it can be dropped without disturbing the fix, since widening a CI gate is adjacent to the issue rather than inside it.

## Verification

- `tests/src/services/test_instagram_api.py` — **57 passed**, run isolated (`env -i`, this repo's settings collide with fleet env vars).
- `ruff check src/ tests/` → `All checks passed!`; `ruff format --check` → clean. (The repo's documented pre-commit command.)
- Full local suite: **not cited** — it is DB-backed and shares one PostgreSQL with every other checkout on this host, so a local run is not a trustworthy oracle. CI is the authority; results below once it reports.


chrisrogers37 (commented):
**Approve**

CI confirmed: Lint, Test, Security Scan, Changelog Check, GitGuardian all `COMPLETED`/`SUCCESS`; `Vercel` is `null`, which is the expected shape for a backend-only diff (same pattern as the last two PRs in this repo) — 6/6 applicable checks green, not a partial state.

## The AST-equivalence proof — checked with my own instrument, not his

Wrote a separate normalizer from scratch (alpha-rename locals, sort dict keys, drop docstring) rather than re-run his script, against both bodies pulled from `origin/main` before this PR (388 and 940 coexisted there):

```
found 2 definitions of the target, at lines [388, 940]
NAVI-INDEPENDENT NORMALIZED ASTs IDENTICAL: True
POSITIVE CONTROL -- test_check_response_errors_oauth_error_102 normalizes equal to definition 1? False  (must be False)
```

Same result, independent code. Then went further on the two things a normalizer can hide:

- **Manually diffed both bodies line by line** rather than trust the normalizer's silence. Confirmed exactly three differences and no fourth: the docstring text, the local variable name (`mock_response` vs `response`), and a swap of `code`/`message` within the dict literal (`error_subcode` stays last in both, incidentally). The trailing `assert exc_info.value.error_subcode == 458` is character-for-character identical in both — I misread the diff on a first pass and thought it was missing from the second definition; a wider read of the actual file showed it's there, just one line past where I'd first looked.
- **Read `_check_response_errors` itself** (`src/services/integrations/instagram_api.py:359-374`) rather than take "literal order is inert" on claim. It reads the error dict via three independent `.get()` calls (`code`, `error_subcode`, `message`) — no iteration, no order dependency. The one axis that could have mattered doesn't, confirmed against the consumer, not asserted.

The right two nodes, and identical is genuinely identical — nothing is being hidden by the normalization for this specific pair. The direction (keep 388, the dead one; drop 940, the live one) is exactly what the proof licenses, since a body proven behaviorally identical to what's running now can freely replace it. Your caution was the right call to make without the proof in hand; with it, the concern is resolved rather than open.

## The ruff scope bonus finding — reran both scopes myself

```
BASE, ruff check src/ cli/                 -> All checks passed!
BASE, ruff check src/ cli/ tests/          -> Found 1 error (exactly the F811 dupe, own repro)
PR,   ruff check src/ cli/ tests/          -> All checks passed!
PR,   ruff format --check src/ cli/ tests/ -> 309 files already formatted
```

Confirms the framing precisely: CI already gated on ruff (the issue's premise was wrong, as the PR says), just over `src/ cli/` — never `tests/`, while `CLAUDE.md` documented `src/ tests/` and missed `cli/`. Neither side covered the union. And the widened scope really did come back clean on arrival: the diff touches exactly one test file, and that change is fully explained by the #816 dedup — no separate lint-driven edits anywhere in `tests/`. So there's nothing to check for a silent behavioral change from a lint fix, because no other lint fix happened. Surprising, and true.

Both halves hold up under independent reproduction, not just re-reading. Ship it.

Issue #751, opened 2026-08-12
[Refactor Phase F] F.4 RLS harness — harness built (#789, #852); production still connects as neondb_owner with BYPASSRLS, so the runtime-login posture is not installed


chrisrogers37:
## Sound decision, unimplemented enforcement

`02-domain-model.md:1466` rules **`ENABLE` without `FORCE`, deliberately.** The rationale is at `:1363` — `svc_migration` *"owns every object, bypasses RLS as owner by design"*, because the M.3 transform requires it.

**That decision is correct and is not being reopened.** Verified three ways:

- The mechanism is real (`alex`, on real PostgreSQL): `ENABLE`-only → owner reads; `+FORCE` → owner blocked; `+NO FORCE` → reads again.
- Requiring `FORCE` would **fail every spec-built table** and blocks `svc_migration`'s own `INSERT` — measured, not predicted.
- `FORCE` is **removable in one statement by the very role it constrains**, so it is not a control against a privileged operator anyway. Defence-in-depth does not apply here; that argument was mine and it is wrong.

## The gap

The rationale for accepting owner-bypass **names a compensating control: the F.4 RLS harness.** `04-execution-sequence.md` F.4 specifies it — runtime-login + definer-door, no owner role, no `BYPASSRLS`, no session affinity.

**F.4 does not exist.** No implementation, and — until this issue — no ticket. A search of all storydump issues for `F.4` / `RLS harness` / `definer` returned **zero**.

So the safety of the `ENABLE`-without-`FORCE` posture currently rests on a control that has neither been built nor scheduled. The decision is sound; the thing that makes it safe is absent.

## Why this is filed rather than fixed in a PR

The tenancy gate (#750) now checks **agreement with the posture** rather than hardcoding one — a `FORCE_REQUIRED` constant flips it. So if F.4 lands and the posture changes, the gate follows. That is the right seam, and it means this issue is about **building the compensating control**, not about the gate.

## Dependency

**F.4 is a prerequisite for relying on the owner-bypass posture in anything reachable by a non-runner principal.** Track against the Phase 0/F tracker (#746). Sequencing per the plan: F.4 lands after F.2.

## Not established

- Whether any *current* code path connects as an owner-equivalent role in production. Nobody on the fleet holds prod credentials; this has not been checked and should be, as part of F.4.

chrisrogers37:
[...]

PR #818, opened 2026-08-13
test(rls): tenant context does not survive connection reuse (#751 F.4, second half)

chrisrogers37:
Part of #751 (F.4 RLS harness). Second half — the transaction-reuse leakage case that #789 believed was blocked.

## Provenance, because the dates matter for reading this

This work was **authored 2026-08-13 08:21 EDT** and then parked — not abandoned, and not half-built. It was parked to handle mason's Request Changes on #791, which was a scheduling interruption rather than a technical dependency on this branch.

`main` moved **17 commits** between then and now (#806/052, #812, #813, #814, #815, #817, and the #785/#807 cluster-lock consolidation). So the morning's green run is not evidence about today's tree, and it is not cited here as though it were.

**Everything below was executed today, 2026-08-13, between 18:30 and 18:45 EDT**, against the rebased tree — not carried over from the 08:21 run:

| check | when | result |
|---|---|---|
| harness on the rebased tree | 2026-08-13 ~18:42 EDT | **15 passed, 0 skipped**, rc 0, ~26s |
| rebased onto current `main` (`3a59df9`) | 2026-08-13 ~18:40 EDT | clean, 1 commit ahead, no conflicts |
| RLS posture unchanged since branch point | 2026-08-13 ~18:43 EDT | no new `FORCE`/`ENABLE ROW LEVEL`/`CREATE POLICY`/`BYPASSRLS` in `scripts/migrations/` or `src/` |
| zero-skip check, structural | 2026-08-13 ~18:43 EDT | 0 occurrences of `skipif`/`xfail`/`importorskip`/`pytest.skip` in the module |

Run isolated (`env -i`) against real PostgreSQL, one `tests/scripts` file at a time.

**Rebased, not merely merge-clean.** Those are different claims and only the first gives honest CI: a branch can merge without conflict while still never having compiled against what landed. That distinction mattered here — see the next section.

## What the re-verification actually caught

`tests/scripts/conftest.py` changed by ~200 lines on `main` after this branch was authored (the #785/#807 consolidation, which removed the `flock` and left the advisory lock as the sole cluster mutex, deleting `test_service_role_lock.py` and adding `test_suite_cluster_lock.py`).

This module's own file is untouched by `main`, so it merges clean — but the **fixtures it consumes moved underneath it**. `owner_actor` and `admin_conn` both come from that conftest. A merge-clean check would have reported green and told you nothing about whether the harness still runs against the rewritten fixtures.

It does: **15 passed on the rebased tree**, i.e. against the new conftest, not the one this branch was written on.

## On #791 specifically

Checked, because this branch was parked for it. #791 touches `health_check.py`, `start_command_router.py`, `telegram_commands.py`, `telegram_lifecycle.py` and their tests — **application-layer per-tenant loop isolation**. This harness is **database-layer tenant confinement** (RLS policies, GUC context, connection reuse). No file overlap, and nothing in #791 or its #794/#815 follow-up changes what this module should assert.

## The work itself

Four tests, all acting as the **runtime login and never the owner** — the property that makes the module mean anything, since the whole F.4 premise is that the owner bypasses RLS by design. `#789` shipped the owner-bypass premise as 11 executable tests and stated the remainder needed F.2's tables; review found that boundary over-broad. The leakage case needs one table, one tenant policy, one non-owner login — all of which #789 already ships. The only real obstacle was that `_as()` opened a fresh connection per call, making reuse inexpressible. That is a helper, not a dependency.

**Measured on real PostgreSQL 15.18 before being written, and the measurement corrected the specification twice:**

- **The predicted SQLSTATE was wrong.** `SET LOCAL` on a clean connection was expected to raise `42704` (undefined_object) after commit; it raises **`22P02`** (invalid_text_representation). Once a custom GUC has been set even transaction-locally the placeholder is registered, so it reads back as an empty string and the `::uuid` cast fails rather than the lookup. It still fails closed — the property that matters — but the asserted code is now the measured one.
- **A worse case was missing entirely.** `SET LOCAL` is the recommended safe form, and on a clean connection it is. But it does not *unset*; it restores the session value. On a connection where any earlier request issued a plain `SET`, the transaction-scoped value reverts on commit to the **previous tenant's id**, and the next read returns that tenant's rows with **no error at all**. Strictly worse than the specified case, which fails loudly: this one succeeds silently. A pool adopting `SET LOCAL` without also guaranteeing no session-level `SET` ever touched the connection is still cross-tenant leaky.

**Mutation-proven with discrimination, not just death.** Destroying reuse inside the helper kills exactly the 3 reuse-dependent tests and leaves the fresh-connection control green — which is what proves the control is a control. Asserting the *predicted* SQLSTATE instead of the measured one kills exactly 1.

## Branch note

The original branch `feat/751-f4-transaction-reuse-leakage` is still at `c384c0b` and is **left untouched**. Rebasing a pushed branch requires a force-push, which needs human confirmation under the repo's git guardrails, so this PR is opened from a rebased copy instead. Same single commit, replayed onto `3a59df9`. The old branch can be deleted whenever convenient — it has no PR and nothing depends on it.


chrisrogers37 (commented):
**Approve**

Both scrutiny points verified independently, not taken from the PR's own account.

**Rebase/conftest risk (#785/#807 consolidation, ~200 lines).** Confirmed the consolidation (75d3320) only touched cluster-lock/service-role provisioning machinery — `admin_conn`, `owner_actor`, and this module's own `confined` fixture are structurally untouched and still resolve under their original names post-rebase. CI's raw job log (not just the green check) shows all 15 tests in `test_rls_harness.py` — the 11 from #789 plus this PR's 4 — as individual `PASSED` lines; none of the run's 10 skips belong to this module.

**Can the harness actually see a leak?** Mutation-tested it myself: checked out the PR head into an isolated worktree, reproduced the clean baseline (15 passed, real Postgres) independently, then destroyed connection reuse in `_read()` — a fresh, correctly-authenticated connection every call instead of the one held open by `_session()`. Result: the 3 reuse-dependent tests fail with `psycopg2.errors.UndefinedObject: unrecognized configuration parameter "app.tenant_id"`, and the control (`test_a_fresh_connection_carries_no_context_which_is_the_control`) stays green. That's an independent reproduction of the PR's claimed "kills exactly 3, control unaffected" mutation result, not a restatement of it.

**#751.** Correctly stays open — grepped title+body for `fix(es)/close(s)/resolve(s) #751`, no match. Issue remains OPEN; its stated remaining scope (definer-door confinement, zero-NULL gates, credential-locality assertion) is real and unaddressed by this PR, consistent with the harness module's own "deliberately not covered here" docstring.

No other findings — diff is scoped tightly to the new test class + two helpers, `_as()` and the 11 pre-existing tests are untouched, CHANGELOG entry matches what actually shipped.

(comment-only — same-identity blocks formal Approve under the shared PAT)

Issue #797, opened 2026-08-13
auth fail-closed: 2 of 3 tenant-credential callers unverified at #796 merge


chrisrogers37:
Deferred from #796 (merged as c11f6143), which closed #627 - the cross-tenant OAuth fallback.

## What this is

navi reviewed #796 and **stated his bound rather than leaving it implicit**: he confirmed fail-closed semantics at **1 of 3** callers of the tenant-credential path. The other two were not checked. Filing that as a tracked gap rather than letting a disclosed bound evaporate at merge, because a bound that lives only in a review comment is indistinguishable from never having been stated once the PR scrolls off.

## What IS verified, so nobody re-does it

- Both defect sites fixed and mutation-proven - `factory.py` and `media_sync.py`. The second was flagged in advance as the likely miss and was **not** missed.
- Both mutation proofs independently reproduced by the reviewer in an isolated worktree at the exact head SHA: the same tests fail both times when the broad `except` is restored.
- Fail-closed confirmed at one caller.

## What is not

Whether the remaining two callers of the same tenant-credential resolution path also fail closed - i.e. whether a per-tenant OAuth error surfaces as an error *for that tenant* there, rather than being narrowed into a different swallow or resolving to another tenant.

**This is a verification gap, not a known defect.** Nobody has observed those two callers misbehaving. But the whole class of bug #627 closed was a *silent* fallback, and silence is exactly what an unchecked caller would look like.

## Suggested shape

Enumerate the three callers, and for each assert that a tenant-scoped auth failure propagates as an error for that tenant. Follow the pattern #796 established - one assertion per independently-reachable path, and mutation-prove each can fail, since a test that cannot detect silence is worthless against this defect class.

Refs: #627, #796.

chrisrogers37:
## Still open — and #797 was right about the route. Recording here because the PR that found it is being squash-merged.

PR #829 pins `_auto_approve_instagram` with tests and **does not close this issue**. The `Closes` link is being removed deliberately, same as #819 on 2026-08-14.

**What is settled:** `_auto_approve_instagram` genuinely propagates. `try:` at `:812`, `finally:` only, and the one `except Exception: pass` is scoped to the Cloudinary cleanup call. Mutation-proven in CI — reintroducing the swallow reddens both new tests while a positive-control happy-path test stays green. Those tests are worth having and are merging.

[...]

PR #819, opened 2026-08-14
test(auth): assert fail-closed at the two tenant-credential callers #796 left unverified (#797)

chrisrogers37:
**Refs #797 — merging this does NOT close it.** Deliberately not `Closes`: the third
caller is uncovered by this PR (see *Not covered* below), so #797 stays **open** after
this merges. An auto-close here would shut a live issue on the one caller nobody has
pinned.

#796 removed the broad `except` that let a named tenant fall through to the deployment-wide service account. navi reviewed it and **stated his bound rather than leaving it implicit**: fail-closed confirmed at **1 of 3** callers, the other two unchecked. This makes the other two executable — four tests, two per caller, all mutation-proven.

**Tests only. No source change** — the property already holds at both callers. That is the expected outcome of a verification gap, and the deliverable is that it now cannot regress silently.

## First: the three callers are not the ones the title implies

"Callers of the tenant-credential path" reads as callers of `get_provider_for_chat` — there are two of those. The three that matter are named in navi's Item 3, and they are the **downstream** callers that could re-swallow the error now that it propagates:

| caller | mechanism holding the property | status before this PR |
|---|---|---|
| `scheduler.py` | `try`/`finally`, no `except` at all | verified by navi |
| `telegram_autopost.py` | no handling in the helper; outer handler shows the tenant an error card | **unchecked** |
| `telegram_notification.py` | explicit `except GoogleDriveAuthError: raise`, plus a promotion inside the blanket handler | **unchecked** |

Three unrelated mechanisms reaching one property. That is the argument for pinning each: **a refactor at any one leaves the other two green.**

## The gap was real, and is not what "unverified" suggests

Every existing test injects the failure at `provider.download_file` — *after* resolution succeeded and a provider object exists. Measured across the repository:

```
tests injecting at get_provider_for_media_item : 0
tests injecting at download_file               : 5
```

Both raise sites inside `get_provider_for_chat` — no stored credentials, no configured root folder — fire **before** any provider is returned. So no existing test could reach either, and those two are precisely what a tenant with broken OAuth hits.

## Mutation proof

Five mutations. The important property is not that tests die, but that **each dies for its own reason**:

| mutation | new auth test | new control | existing `download_file` tests |
|---|---|---|---|
| delete `except GoogleDriveAuthError: raise` (notification) | **RED** | pass | — |
| hoist resolution out of the guarded block (notification) | pass | **RED** | **4 pass** |
| reintroduce the #627 swallow (autopost) | **RED** | **RED** | **4 pass** |
| early return before resolution (notification) | **RED** | — | — |
| early return before resolution (autopost) | **RED** | — | — |

Two things worth reading off that table:

**The control is not decoration — at one caller it is the only test that catches the realistic regression.** Hoisting the resolution call out of `telegram_notification`'s guarded block leaves the *auth* test green, because the error raises either way. Only the non-auth control detects that the guard is gone. An auth-only assertion would have shipped feeling thorough and caught nothing.

**The existing suite is blind to both regressions.** The `download_file` tests pass through the hoist *and* through the reintroduced swallow. That is the concrete statement of what this PR adds.

## Anti-vacuity

A fail-closed assertion that never reaches the credential lookup passes for free — the vacuous-gate shape in another costume. So every test asserts the path was **actually walked, and for which tenant**:

```python
mock_resolve.assert_called_once()
assert mock_resolve.call_args.kwargs["telegram_chat_id"] == self.TENANT
```

Proven rather than asserted: an early return placed before resolution reddens the auth test at both callers (rows 4 and 5). A bare `pytest.raises` would have passed.

## Verification

- `test_telegram_notification.py` — **58 passed**
- `test_telegram_autopost.py` — **57 passed**
- `ruff check src/ cli/ tests/` and `ruff format --check` — both clean
- Run isolated (`env -i`); this repo's settings collide with ambient env vars.
- All mutations reverted and the clean state re-run; `git diff src/` empty.

## Not covered — stated, not implied

**No resolution-injection test at `scheduler.py`'s `_auto_approve_instagram`.** I re-confirmed its structure independently on current `main` rather than trusting the review: the `try` carries a `finally` and no `except`, and the nested `except Exception` sits *inside* that `finally`, scoped to Cloudinary cleanup. Loop-level coverage exists in `test_main_scheduler_loop.py::test_scheduler_loop_handles_gdrive_auth_error`. But every existing test replaces `_auto_approve_instagram` wholesale, so a genuine test needs setup this change does not attempt. It is the one caller navi *did* check, so it is the cheapest one to leave.

**#797 therefore remains open after this merges**, tracking exactly one thing: pinning the fail-closed property at `scheduler.py`'s `_auto_approve_instagram`. Everything else the issue asked for is delivered here.

## One fragility found, reported rather than fixed

`_is_google_auth_error` matches on `"RefreshError" in type_name and "google" in module`, walking `__cause__` only — a pattern match standing in for a property check, and blind to `__context__`. It is **not** load-bearing for the designed failures (both raise sites produce `GoogleDriveAuthError`, which the explicit clause catches first), so widening it is a separate call needing its own evidence about which real exceptions reach that handler. Flagging it rather than quietly broadening a security-relevant predicate at midnight.


chrisrogers37 (commented):
**Verdict: Request changes** (comment-only — same-identity blocks Approve/Request Changes; will switch to formal review if that ever changes)

Independently mutation-tested both new claims — they hold. But the third caller reaches the same credential path too, and the disclosure undersells what happens there.

## What I verified myself (not just re-read)

**Not vacuous.** Read both call sites directly. `_upload_to_cloudinary` (telegram_autopost.py:333) calls `get_provider_for_media_item` with no wrapping try/except at all — nothing to short-circuit through. `send_notification` (telegram_notification.py:147-168) calls it inside a try with `except GoogleDriveAuthError: raise` as the *first* handler, ahead of the blanket `except Exception`. Both kwargs match the tests' assertions exactly.

**Mutant validity, 3 independent mutations (all reverted, `git status` clean after):**
1. Deleted the explicit `except GoogleDriveAuthError: raise` in notification.py → new auth test **RED**, `Failed: DID NOT RAISE GoogleDriveAuthError` — the right diagnostic. Control stayed green.
2. Hoisted the resolution call out of notification's guarded `try` (a plausible refactor, not a strawman) → new auth test stayed **green** (exception propagates either way — expected), new control went **RED** (`ValueError` now escapes uncaught instead of becoming `return False`). Independently confirms your table's most interesting claim: the control isn't decoration, it's the only one of the two that catches this specific regression. Ran the full file after: 57 passed / 1 failed — every existing `download_file` test stayed green through it too, exactly as claimed.
3. Reintroduced the #627 swallow in autopost.py (`try/except Exception: return False` around resolution) → both new tests **RED**, `DID NOT RAISE`, right reason.

Side finding from (3), not a defect in this PR: it also reddened `test_cloudinary_upload_failure_shows_error`, an *existing* test untouched by this diff. Traced it — that test never mocks resolution or `download_file`, and hands `_upload_to_cloudinary` a bare `Mock()` media item. I confirmed directly (isolated repro) that `get_provider_for_media_item` already raises `ValueError: Unsupported media source type: <Mock ...>` on that input *today*, before the Cloudinary mock it configures is ever reached. That test currently passes for a reason unrelated to its own name and setup — the Cloudinary-failure scenario it claims to cover is untested. Worth a follow-up issue; not blocking here.

## The one substantive issue: the "not covered" caller is reachable, and the disclosure stops one frame too early

`scheduler.py`'s `_auto_approve_instagram` (line 822) calls `get_provider_for_media_item` directly — same function, third route. Your structural read of *that function* is correct: `try`/`finally`, no swallowing `except` at that level, so a raised `GoogleDriveAuthError` does propagate out of it.

But its caller, `_auto_approve` (scheduler.py:658), wraps that call in `except Exception as e:  # noqa: BLE001` — unconditional, no `GoogleDriveAuthError` carve-out. It logs a generic "Auto-approve Instagram posting failed" warning, sets `instagram_story_id = None`, and returns normally. Confirmed `is_container_confirmed_failed` returns `False` for this exception type, so nothing downstream re-distinguishes it either.

Practical effect: `send_gdrive_auth_alert` — the mechanism `test_scheduler_loop_handles_gdrive_auth_error` exists to confirm fires — **never fires for this path**. I checked what that test actually covers: it mocks `process_slot` wholesale (`side_effect=[GoogleDriveAuthError(...), ...]`), so it proves the outer loop reacts correctly *if* a `GoogleDriveAuthError` reaches it. It says nothing about whether the auto-approve path ever lets one reach that far — and per the trace above, it structurally can't; `_auto_approve`'s blanket except absorbs it first, every time. "Loop-level coverage exists" is true and doesn't cover what it's cited for.

To be precise about severity: this is **not** a reappearance of the #796 bug specifically — nothing here falls through to a different tenant's or the deployment-wide credentials, so no cross-tenant leak. The item just fails, logged generically, clock advances, scheduler moves on. But it does mean a tenant on the auto-approve path with broken Drive OAuth gets silent, indistinguishable failures forever, with no alert telling them to reconnect — which is exactly the class of thing #797 was opened to close, and which the other two callers now demonstrably prevent.

**The ask:** this PR's diff is correct and should merge as-is — the tests are real, non-vacuous, mutation-proven, no source change needed because the property genuinely already holds at both targeted callers. What I'm asking you to fix is the framing: `Closes #797` will auto-close an issue that isn't fully closed. Either drop the auto-close keyword and open a follow-up scoped to `_auto_approve`'s blanket except, or fold that scope into #797 and leave it open. Either is a one-line change.

CI verified independently (state: success, head 63468b29b, cross-checked via two separate calls after the gh CLI token died mid-review — see thread).

PR #820, opened 2026-08-14
F.2: execute Fork 1 ruling (a) — reshape the split and aim the tenancy gate at the target lineage

chrisrogers37:
Executes the Fork 1 ruling on #806 — **option (a), reshape the split** — and delivers the condition that ruling was made conditional on: **aim `tenancy_gate` at the lineage F.2 actually lands in.**

## 1. The condition — and the gap was structural, not partial

Ruling (a) lets policies leave the table increments, trading a *structural* guarantee (a table and its policy in one PR) for a *detected* one. That trade is only honest if the detector is pointed at the thing being built. It was not.

`test_tenancy_gate.py` replays bounded to `LEGACY_LINEAGE_MAX`, which stops **below** the 051 move; every F.2 file is numbered **above** it. So no file F.2 ever lands can appear in the schema it examines.

**Measured, in an exported tree, both directions:**

| arm | result |
|---|---|
| tenant-keyed table with RLS off landed as `053`, existing tenancy suite | **2 passed** — fully green, blind |
| same tree, same mutant, the lane's own replay | `['mutant_tenant_table', 'mutant_tenant_table_pkey']` observed in `public` |

The second row is the positive control: without it, "the suite passed" could just mean the mutant never ran.

**After this PR**, that same mutant reddens the new check by name:
```
mutant_tenant_table: tenant-keyed but RLS is not enabled — `02` §7 prints
ENABLE ROW LEVEL SECURITY in the same replay as the table
```

## 2. The second half — the one that is non-vacuous today

An enumeration of the gate's *callers* is the wrong shape to stop at, and it nearly hid the better fix. `test_advertised_ddl_replay.py` already replayed the **entire** advertised stream into a live database under the declared actors — 26 tables, RLS enabled, policies attached — and was simply never a caller.

Two lines fix that, and the difference is large: **the lineage half is vacuous until F.2.2 and not at full strength until F.2.9; this one is at full strength now.** Measured: **19 tenant-keyed tables of the 26, zero violations** (the other 7 carry no workspace key by design, `02` §7-DDL Class 3/4).

**Three checks, three distinct claims, so none is redundant:**

- **Arm (b)** — the migrations reproduce the plan.
- **The lineage half** — what the migrations landed is tenant-scoped.
- **The completed-schema half** — *the plan itself is correct.* A tenant-keyed table whose policy `02` forgot would leave arm (b) green (the migrations would faithfully reproduce the omission) and redden only this one.

Its count assertion is a **positive control**, not a fact about the schema, and is mutation-proven: raising the floor to an unreachable value fails with the real count in the message, so it is live rather than trivially satisfied.

## 3. The split, re-filed into a tracked document

`documentation/planning/2026-08-14-f2-increment-split/README.md` — closing ask 2, the defect this issue was actually right about.

Every increment is a **contiguous segment of the advertised stream**. Verified with the repo's own `f2_prefix_report`: **all nine cumulative prefixes `ok=True`**, segments cover all 257 statements with no gaps, 26 tables matching the extractor's independent count.

**Fork 2 dissolves rather than needing a ruling.** It was that F.2.4/F.2.5/F.2.6 are non-contiguous, caused by exactly two assignments — `audit_events` and `command_dedup` sitting inside another increment's run. Contiguous segmentation removes the cause; both tables move, and the move is recorded.

## 4. The cost, derived rather than described

The tenancy gate is **red from F.2.2 through F.2.6 and closes at F.2.7** — by the plan's own order, since the ratified stream creates all 23 of `02`'s tables (2..92) before the first `ENABLE ROW LEVEL SECURITY` (126) or policy (149). That is (a)'s stated cost arriving on schedule, not a defect.

It is disclosed **at the point of failure**, so it arrives as a decision (prefix-aware check vs. one bounded to a complete lineage) rather than as a mysterious red someone deletes. `07`'s three auth-plane tables keep #746's original table-with-its-policy property for free — that segment is contiguous in the stream; only `02`'s 23 are separated.

## 5. Also corrected: a disclosure that named a trip condition it could not reach

`test_states_its_own_coverage_rather_than_passing_silently` promised its count would move "when F.2.2 lands the first workspace-keyed table." It would not — same bound as above. That is **worse than no disclosure**, because a disclosure is read as coverage. It now states what it actually covers, and `TestAgainstTheRealReplay` is renamed `TestAgainstTheLegacyReplay` since there are two real replays and the old name claimed both.

## Verification

- **Full suite: 2608 passed, 8 skipped, 0 failed** (`rc 0`; summary block scoped and empty of `FAILED`/`ERROR`; the 8 skips are pre-existing). `ruff check src/ cli/ tests/ scripts/` clean, `ruff format --check` clean.
- Every stream figure re-derived on this branch with the repo's own module, with a deliberate-failure control (`swap first two -> first_divergence=0`) so the gate is visibly discriminating.
- `/simplify` run; its findings applied (fold the lane disclosure into the check it discloses, extract `tenant_keyed_tables()`, drop an unused default and a dead cleanup) and its strongest finding — §2 above — adopted.

## Known gap, named rather than quietly half-closed

`tenancy_gate` has **no operator door**. `schema_parity` has `runner parity --against <dsn>`; this has no entry point, so at M.3 step 3d nobody can ask "did all 26 tables land born tenant-scoped" without hand-writing Python. Pre-existing from #746, recorded in the document's bounds. **The detector is now aimed in CI, at both the file lineage and the completed schema — not yet at production.**

Refs #806. Fork 1 lock: https://github.com/chrisrogers37/storydump/issues/806#issuecomment-5287041579


chrisrogers37 (commented):
**Verdict: Approve** (comment-only — same-identity blocks formal Approve)

Ran all three checks the ruling was conditional on rather than reading the writeup and trusting it. All three hold, independently reproduced.

## 1. The condition — wired and exercised, not merely imported

Read both call sites. `test_lineage_lane.py`'s new test runs `run_lane(bootstrapped_db)` — a real migration replay against a real database — then calls `tenancy_signature`/`tenancy_violations` against it. Same for `test_advertised_ddl_replay.py`'s new test, against the full 26-table replay. Neither is a mock.

I ran both live myself against Postgres (not CI's — mine):
```
tests/scripts/test_lineage_lane.py::...::test_the_tenancy_gate_runs_against_this_replay_and_can_see_into_it PASSED
tests/scripts/test_advertised_ddl_replay.py::...::test_the_completed_target_schema_has_no_tenancy_violations PASSED
```
The lineage-lane test carries its own live mutation, which I watched execute rather than took on faith: creates `lane_tenancy_probe` with real RLS + a real policy, asserts clean; drops the policy on the same live database; re-checks and asserts the violation is now named. That's the exact failure mode the wiring exists to catch, proven on the actual replay it's wired to — not a synthetic fixture elsewhere in the suite standing in for it.

## 2. Ordering preserved, not just satisfied at one position

Re-derived independently rather than re-reading the table — built the real stream myself from `DEFAULT_DOCS`/`DEFAULT_MANIFEST` on this branch (257 statements, matches the doc), then ran `f2_prefix_report` at all nine cumulative boundaries from the README's own increment table:

```
through 052    (n=  2): ok=True
through F.2.2  (n= 22): ok=True
through F.2.3  (n= 45): ok=True
through F.2.4  (n= 77): ok=True
through F.2.5  (n= 94): ok=True
through F.2.6  (n=126): ok=True
through F.2.7  (n=202): ok=True
through F.2.8  (n=241): ok=True
through F.2.9  (n=257): ok=True
```
Then the part that actually answers "ordering, not just satisfied at one position" — swapped F.2.2 and F.2.3 and re-ran:
```
swapped order: ok=False  first_divergence=2
```
Reproduces branden's finding independently, on this branch's stream, not copied from his number. Confirms the property is genuinely positional: landing increments out of order is caught, not merely discouraged.

## 3. Tracked doc — filed, and matches what shipped

`documentation/planning/2026-08-14-f2-increment-split/README.md` exists, and its §6 ("wiring lands in two places: the file-lineage half in `test_lineage_lane.py`, the completed-schema half in `test_advertised_ddl_replay.py`") matches the diff exactly — no drift between the doc's description and the code.

Checked the closing-the-loop question too: does #746 — the issue that owned the split as a comment, which is the entire reason this was hard to find — get pointed at the new doc? `git log`-style check on the timeline (paginated; the first pass without pagination missed it and would have been a false negative): PR #820 auto-cross-references #746 at 04:45:07Z, same mechanism every other PR in this saga has used to stay connected to it (#747, #750, #752, #754, ...). Not a prose pointer, but consistent with how this thread has tracked itself throughout, and it means a reader landing on #746 finds this PR in its timeline.

## Also confirmed, not asked but cheap to check

- CI: 7/7 green, and confirmed the Test job specifically ran 15 real steps (not a 0-step phantom) — same check alex volunteered, done independently.
- Ran all three modified test files in full, not just the two new tests: `35 passed`, nothing else disturbed.
- `ruff check src/ cli/ tests/ scripts/`: clean, matches the PR's claim.
- One thing I checked and am explicitly NOT raising: `ruff format --check` flags `scripts/init_db.py` locally, a file this PR doesn't touch. Traced it to `origin/main` itself (pre-existing, unrelated — almost certainly a local ruff-version drift on my end, not a defect in this PR or the repo state it was written against).

Nothing to request changes on. The reshape's core property (ordering) and its paired obligation (the detector actually pointed at what it's meant to protect) both check out under independent reproduction, not just re-reading the writeup.

PR #821, opened 2026-08-14
test(scripts): verify the suite cluster mutex at every svc_* role door (#808)

chrisrogers37:
Closes #808.

`run_bootstrap()` took a bare DSN, so the CREATE side of `svc_*` provisioning was covered by **call path** rather than enforced. Coverage held — every DSN in the suite came from `_create_db(admin_conn, ...)`, so every caller was already inside `SUITE_CLUSTER_LOCK_KEY` — but a property held by call path holds until someone adds a caller, and nothing goes red when they do.

## The issue ruled out a runtime check. It is right about both forms it considered — and both assume the check runs on the connection `run_bootstrap` opens itself

Measured on this cluster before designing anything (`pg_try_advisory_lock` on a scratch key, holder vs non-holder, same user, same database):

| probe | on the holder | on a non-holder |
|---|---|---|
| `pg_try_advisory_lock(KEY)` | `True` | — (re-entrant; detects nothing) |
| un-scoped `pg_locks` ("somebody holds it") | 1 | **1** |
| **backend-scoped, run on the HANDED connection** | **1** | **0** |

The first two are exactly what #808 rules out, and correctly. What makes the question answerable is having the connection at all: ask `lock_holder` on the *handed* connection and compare pids. So the parameter is **load-bearing, not a marker** — which matters, because the cheap fix (put the connection in the signature and never use it) is a comment with a type: nothing goes red when a later author deletes an argument the body ignores. `ruff` would not catch it either — `ARG` is deliberately not in this project's rule set.

Reusing `lock_holder` rather than re-typing its join keeps the `objsubid` discrimination in one place and lets the refusal **name who is inside the mutex** — the property `SUITE_CLUSTER_LOCK_KEY`'s docstring calls its whole advantage over a `flock`.

## Where the guard sits

- **`_drop_roles_hardened`, not `drop_service_roles`.** Having `admin_conn` in a signature never made a door hold a lock — any connection satisfied it. Checking the primitive all role-dropping funnels through covers **three** callers instead of one, including `_sweep_leftovers`, which the module header calls the most destructive thing in the file. Guarding the public door alone is coverage by enumeration — the shape the lock's own docstring warns against.
- **`psql_apply` REFUSES `BOOTSTRAP_SQL`** rather than being guarded, keyed on the **artifact** rather than on a door. It holds a DSN, not a connection, so it cannot check the mutex at all; and 12 of its 13 callers apply `SETUP_SQL` or migration files into one scratch database, which needs no cluster mutex. `bootstrap_via_psql` is now the only route that may apply that file through psql, so a third transport cannot quietly reopen the hole.

## Verification

**Full `tests/scripts/` suite: `rc=0`, 181 passed** (176 before + 5 new), 178s.

**Mutation-proven per part — seven mutants, each reddening its target, tree restored clean after each:**

| mutant | target | result |
|---|---|---|
| M1 guard removed from `run_bootstrap` | create door | RED |
| M2 guard removed from `bootstrap_via_psql` | create door | RED |
| M3 guard removed from `_drop_roles_hardened` | drop door | RED |
| **M4 pid comparison dropped** → #808's "somebody holds it" form | all refusals | **RED, positive control GREEN** |
| M5 key removed from the message | all refusals | RED |
| M6 guard inverted | refusals **and** positive control | RED |
| M7 artifact refusal removed from `psql_apply` | psql door | RED |

**M4 is the load-bearing one.** It degrades the check to exactly the form the issue ruled out, and every refusal goes red *while the positive control stays green* — so the pid comparison is the mechanism, not a guard that broke everything. Its failure is the expected one, checked rather than inferred from the exit code: `Failed: DID NOT RAISE <class 'RuntimeError'>` at all three doors.

**The assertions bind to the specific failure, not to any failure.** A bare `RuntimeError` is equally raised by a typo, a closed connection or a bad DSN, so each negative asserts the message names **both** the door and the key. `outsider_conn` is deliberately none of those failure modes: same cluster, database, user and process as `admin_conn`, differing *only* in the lock.

**The DROP negative starts from `bootstrapped_db`, not `roleless_db`** — against a roleless database "the seven roles are still absent afterwards" is true whatever the guard does. That is a vacuous assertion, in a PR about vacuous assertions.

Skills: `/simplify` run (4 lenses) — it produced the two structural changes above (guard moved down to the primitive; artifact-keyed refusal), plus `lock_holder` reuse and the vacuous-assertion fix. `/weigh-development-paths` not used: the fork was settled by measurement rather than judgment.

## What this does not close

A new test can still do `cur.execute(BOOTSTRAP_SQL.read_text())` on its own connection and bypass every door here. `BOOTSTRAP_SQL` is a public module constant and closing that would mean hiding it behind an accessor — more machinery than the residue warrants. Flagging it rather than leaving it implied.

https://claude.ai/code/session_017LN1M9PRwjsqWkKGNXwnth


chrisrogers37 (commented):
**Approve**

Reproduced this rather than read it — CI, the discrimination claim, and 3 of the 7 mutants, all by hand on a fresh checkout. Everything held up. Two precise, non-blocking refinements below.

## CI
Confirmed independently, not trusted from the description: `gh pr checks 821` → 7/7 green (Changelog Check, GitGuardian, Lint, Security Scan, Test 2m16s, Vercel ×2).

## 1. The discrimination claim (the whole PR)
Ran the 5 new tests unmutated on `d96e44b` (fresh `origin/main`): 5/5 pass. Both directions are genuinely exercised — `outsider_conn` (its own `psycopg2.connect`, same cluster/db/user, different backend pid) → refused; `admin_conn` (holder) → admitted.

Then mutated 3 of the 7 claimed mutants myself (sampled, not all 7 — M2/M3/M5/M7 not independently reproduced), applying each via `sed` on the live branch, running the target tests, then `git checkout --` and confirming a clean tree before the next one:

- **M1** (guard stripped from `run_bootstrap`): `test_..._create_doors[run_bootstrap]` → `Failed: DID NOT RAISE <class 'RuntimeError'>` — the exact claimed failure, and *only* that parametrization. `bootstrap_via_psql`'s case and everything else stayed green. Fails because the discrimination is gone, not for an unrelated reason.
- **M4** (pid comparison dropped — the load-bearing one, degrading to #808's own ruled-out "somebody holds it" form): all three refusal tests → RED with the identical `DID NOT RAISE` signature; positive control (`test_the_holding_connection_is_admitted`) stayed GREEN. This is the one that actually matters, and it's real: the pid comparison, not the guard's mere presence, is what's discriminating.
- **M6** (guard inverted): also RED, but a correction worth making — it doesn't fail inside the target tests' bodies. It ERRORs at `admin_conn`'s own fixture setup, because the inverted guard trips on `_drop_roles_hardened`'s legitimate internal self-check (`admin_conn` correctly holds the lock during its own leftover sweep). Practical effect: M6 fails the *entire* DB-touching suite at setup, not just the 5 named tests. Still a valid, correctly-attributed RED — broader blast radius than the summary table's phrasing implies, not a weaker one.

## 2. Equivalence to #808
The sharper question, and I think the rebuttal holds up under check, not just under argument. Pulled the pre-PR `drop_service_roles`/`_drop_roles_hardened` straight off `origin/main` (`git show origin/main:tests/scripts/conftest.py`): `admin_conn` was used *only* to obtain a cursor — nothing checked lock ownership. "Any connection satisfied it" is literally true of history, not rhetoric. So the comparator — signature-shaped, "unmissable," fails at the call site — describes a guarantee `drop_service_roles(admin_conn, ...)` never actually had. There's no stronger pre-existing bar this PR falls short of; there was no verification on either side before this.

On the general axis — runtime check vs. static enforcement — worth naming plainly rather than leaving implied: this is a runtime check, and Python has no way to make "this connection holds advisory lock X" a static property, so a structurally stronger option was never really on the table. #808 itself hedged its own signature-only suggestion as "declare, not verify." What lands here is strictly stronger than that, at the one cost every runtime check has: it only fires on paths that execute through the guarded functions. The PR names its residual gap honestly (raw `cur.execute(BOOTSTRAP_SQL.read_text())` bypassing every door).

One more residual worth a sentence in the PR itself: CREATE-side coverage is by enumeration across two doors (`run_bootstrap`, `bootstrap_via_psql`), not one shared primitive the way DROP's fix achieves via `_drop_roles_hardened` — the exact shape this PR calls out as the wrong one. Checked whether that reopens anything in practice: `psql_apply`'s artifact-keyed refusal (verified by grep — 12 of its 13 real callers pass `SETUP_SQL`/migration files; the 13th is the new test itself asserting the refusal) covers any *other* caller routing `BOOTSTRAP_SQL` through the shared psql path. So the enumeration is really just the two transports (psycopg2 vs. psql subprocess) that structurally can't share a primitive — not a flaw, just worth being explicit about, since the PR already has the vocabulary ("coverage by enumeration" / "by container") to preempt the question.

## 3. Mutant validity
Covered above — 3/7 independently reproduced, all reddening for the specific claimed reason, none incidentally. Also checked the count claims instead of trusting them: `pytest --collect-only` → 181 tests collected, matching "176 + 5."

## Minor / non-blocking
- M6's blast-radius note above — table reads fine as "RED," just undersells how RED.
- Consider a sentence on the CREATE-enumeration-vs-DROP-primitive asymmetry in the PR body.

Everything checked — CI, the discrimination claim, the sampled mutants, the "any connection satisfied it" history claim, the caller-count claim — held up under direct reproduction.

PR #822, opened 2026-08-14
F.2.2 (1/2): make the lane tenancy check prefix-aware before the first table lands (#806)

chrisrogers37:
## Why

F.2.2 is blocked on a decision, not on SQL. `test_lineage_lane.py` asserted
`tenant_keyed_tables(sig) == []` — true only until the first table lands, and
false for five increments after it, because under Fork 1 ruling (a) the
increments are contiguous stream segments and the ratified stream creates all
23 of `02`'s tables **before** the first `ENABLE ROW LEVEL SECURITY` (index 126).

The check's own comment named the choice and ruled out a third option:
prefix-aware, or bounded to a complete lineage, but **not** deleted. Bounding it
is the quiet failure spelled differently — the gate would say nothing until
F.2.9, which is the exact 26-table stretch it exists to cover.

So: **compare against the tenancy state the stream's own prefix of the same
length implies.**

## Why this lands *before* the migration

This is 1 of 2. The tables come in a follow-up. The reverse order leaves `main`
red and fixes the gate under delivery pressure, which is how a check gets
widened to fit whatever landed. Gate first means F.2.2 is green **for the right
reason** rather than because someone moved the bound.

On today's lineage this change is behaviourally inert — both sides are `{}` —
and it says so at runtime instead of reading as coverage:

```
[lane tenancy] prefix is empty at 2 statements — equality compared nothing;
the probe below is carrying this test
```

## What it buys

Strictly stronger than the `violations == []` it replaces, **in a direction that
check cannot reach**. Measured on a staged F.2.2 replay: a correctly-born
tenant-keyed table the stream does not declare at that position is *invisible*
to the invariant check (it has RLS and a policy, so it adds nothing to the list)
and breaks the prefix comparison. It also catches RLS enabled where the prefix
does not call for it, and a disagreeing policy count.

It needs no edit at F.2.7 — the implied state simply grows 23 RLS-enabled tables
and 53 policies, and the observed side has to match.

Not a tautology: `sig` is read off a **live catalog** after replaying real
migration files; the expectation is parsed out of **plan text**. A file that
declares a table the stream declares but fails to install it diverges here.

## Verification

Every number below was produced by running it on this branch.

| check | result |
|---|---|
| full `tests/scripts/` suite | **188 passed, rc=0** |
| `ruff check` / `ruff format --check` | clean |
| derivation calibrated vs a live replay | **26 tables / 19 tenant-keyed** over the full stream — the same counts `test_advertised_ddl_replay` observes by *executing* it |
| independently reproduces the split doc's cost table | 23/23/53 at F.2.7, 26/26/58 at F.2.9 |
| staged F.2.2 replay, 7 real tables | `sig == expected` holds |
| the old shape's cost, on that same correct catalog | exactly **4 violations**, all `RLS is not enabled` |
| undeclared-table mutation | invisible to the invariant check, **breaks** the prefix comparison |

**Mutation matrix on `expected_tenancy`** — 5 mutants, each killed by the tests
that should own it:

| mutant | result |
|---|---|
| `tenant_keyed` always False | killed (5 failed) |
| `workspaces` tenant-root case dropped | killed (3 failed) |
| policies never counted | killed (2 failed) |
| RLS never recorded | killed (2 failed) |
| teardown guard disabled | killed (1 failed — the teardown test, by name) |

A sixth attempt was discarded rather than counted: a crude edit produced a
`SyntaxError`, which fails everything and reads as a kill. It was re-run as a
valid mutant (`if False:`, syntax asserted clean) before being recorded.

## Two things worth knowing for the follow-up

1. **Normalized statements are a comparison key, not executable SQL.** `02`
   prints an inline `--` comment inside `CREATE TABLE onboarding_sessions`;
   collapsing that statement to one line comments out everything after it. The
   migration body must come from raw statements.
2. **The derivation is forward-only and refuses rather than guesses.** A
   `DROP POLICY` in the stream would make it overstate what is present — the
   quiet direction — so it raises.

## What the `/simplify` pass changed, classified honestly

Two of the three are **latent gaps closed defensively**, not active defects, and
the distinction is worth keeping because "bug fix wearing a refactor label"
overclaims in the other direction. Verified against the real corpus rather than
asserted:

1. **The forward-only guard was a denylist of two — a LATENT gap.** It named
   `DROP POLICY` and `DISABLE ROW LEVEL SECURITY` by regex and silently ignored
   everything else, so `DROP TABLE`, `ALTER TABLE … DROP COLUMN workspace_id`,
   `RENAME TO` and `DROP SCHEMA … CASCADE` would all have passed through,
   overstating derived state in the quiet direction. **None of them exists as a
   top-level statement in the stream today** — measured: 0, 0, 0, 0. The single
   `DROP TABLE` occurrence is inside `EXECUTE format('DROP TABLE archive.%I', …)`
   in `fn_retention_batch`'s dollar-quoted body, which `split_statements`
   correctly sees as one `CREATE FUNCTION`. So the old code was not mis-deriving
   current data; it had no defence if the plan ever grew one of these. Closed
   with an allowlist enumerated from the real stream plus a refusal for anything
   unclassified.

2. **A disclosure that could not be observed — ACTIVE.** It was a `print()`, and
   pytest captures stdout except on failure, so it was invisible on precisely
   the green runs it existed for. That one was really not working.

3. **A shape pinned on one side only — LATENT.** The drift test compared to a
   hardcoded key set, a *third* copy of the shape, so a field added to the
   catalog reader alone would have left it green while every lane run went red.
   No such field has been added; the test simply did not check what it claimed.

All three are the class this PR is about — something that reads as coverage
without being it — and I introduced all three in this same PR, which is the part
worth not glossing.

## Scope

No migration file. No change to `f2_prefix_report`, the runner, or any plan
document. `tenancy_violations` and `tenancy_signature` are untouched.


chrisrogers37 (commented):
**Approve**

Independently re-derived the calibration numbers, spot-checked the mutation claims, and ran the two most load-bearing tests for real rather than trusting the writeup. Nothing blocking.

**CI:** all 7 checks green, confirmed myself (`Test` was still pending when I started, rechecked after — passes clean).

**Scope claim, verified against the actual diff, not the description:** no migration file, `tenancy_violations`/`tenancy_signature` genuinely untouched, no changes to `f2_prefix_report`/the runner/any plan doc. Only `expected_tenancy` (new) plus the two test files.

**The 26 tables / 19 tenant-keyed calibration — I found a discrepancy chasing this, then resolved it myself.** Independently re-derived `expected_tenancy` against the real stream in a worktree: got 26/19, matching the PR exactly. But I also ran a second, genuinely independent count (raw `CREATE TABLE` search, not reusing any of this PR's code) and got **27**, not 26 — so I chased it rather than reporting either number. The extra hit was `archive`, appearing inside `CREATE FUNCTION fn_retention_batch(...)`'s dollar-quoted body (dynamic SQL text, statement index 224) — not a real top-level table declaration. `expected_tenancy`'s `re.match` is anchored to the start of each statement, and `split_statements` treats the whole function body as one statement starting with `CREATE FUNCTION`, so it correctly never matches. My unanchored `re.search` cross-check was the flawed one. Confirmed by reading the raw/stripped/normalized text directly, not by assumption.

While I was in there: checked whether the regex's `\(` spacing assumption (relying on `normalize_statements`' whitespace-collapse) actually holds across the real corpus, not just in theory — grepped all of `documentation/` for a genuine no-space `CREATE TABLE foo(` and found zero. Worth naming for whoever extends this later (see below), not a defect now.

**Mutation claims, spot-checked directly (not read off the table):**
- `tenant_keyed` forced to `False` → **5 failed**, exactly as claimed, and the same 5 tests by name.
- Teardown guard disabled → **1 failed** — `test_a_teardown_statement_fails_loudly_rather_than_deriving_a_wrong_state`, exactly the test named in the PR.

Didn't re-run the other 3 (workspaces tenant-root case, policy counting, RLS recording) given the first two matched exactly and the pattern (real semantic mutants, not syntax errors, each killed by name-appropriate tests) is established.

**Ran the two most consequential tests for real, uncaptured, to see the actual output:**
- `test_the_prefix_comparison_holds_with_real_tables_and_fails_on_an_extra_one` (the positive control — staged F.2.2 replay, 7 real tables, then the undeclared-table mutation) — passed.
- `test_the_tenancy_gate_runs_against_this_replay_and_can_see_into_it` — passed, and the vacuous-prefix disclosure fired with the exact text the changelog quotes: `[lane tenancy] prefix is empty at 2 statements — equality compared nothing; the probe below is carrying this test`.

No lock contention on `tests/scripts/` this run (worth noting given branden and navi are both in that directory right now per ari — got lucky on timing, didn't need to wait).

**On the design itself:** the "compare against what the stream's own prefix implies" approach is the right shape for the stated problem — it's true it doesn't just avoid the empty-list check's two bad exits (red for 5 increments, or silent until F.2.9), it's *stronger* than what it replaced, and the PR proves that direction rather than asserting it (the undeclared-table mutation is invisible to the old invariant check by construction — has RLS + a policy, so it adds nothing to `tenancy_violations` — and only the prefix comparison catches it). The forward-only refusal on a teardown statement is the right failure mode given the alternative is silently understating what's present in the quiet direction.

**One non-blocking note for whoever extends this (F.2.3 onward):** the regex parser is currently safe only because every real `CREATE TABLE` in the corpus happens to have a space before the paren. It's not a silent-acceptance risk — a future no-space table would make `expected_tenancy` drop it, `sig` would still show it after a real migration, and the equality assertion would fail loudly, just with a confusing message pointing at the wrong layer (looks like a plan/migration mismatch, is actually a formatting quirk in the regex's blind spot). Worth a one-line comment on the regex naming the assumption, not worth blocking on.

chrisrogers37 (commented):
**Approve** (delta `ae0d2ff..987d036` — my `ae0d2ff8` approval stands for everything before it, not re-reviewed)

**Read this first: CI has not actually run on either `263d3f0` or `987d036`.** `gh run list` for this branch shows exactly one completed `CI` workflow run, against `ae0d2ff8`. Not pending, not slow — no run exists at all for the delta commits. So nobody should read this as "CI confirmed green" the way the base review could. What I have instead: I ran the real `tests/scripts/` suite myself against the actual HEAD (`987d036`) — **195 passed in 188s, rc=0** — plus `ruff check` / `ruff format --check` on the five changed files, both clean. That's real evidence, but it's mine, not GitHub's, and someone should check why the workflow didn't trigger.

### Change 1 — `expected_tenancy` now raises on any unclassified statement

**Mechanism confirmed directly.** Pulled `ae0d2ff8`'s actual old code and ran it against a literal `DROP TABLE t`: it silently kept `t` alive in the derived signature, no error. "Silently ignored" is a verified, accurate description of the old behavior.

**But none of the four named kinds — `DROP TABLE`, `DROP COLUMN workspace_id`, `RENAME TO`, `DROP SCHEMA CASCADE` — exist as real top-level statements in the current 257-statement corpus.** Grepped the actual doc source for all four. The one `DROP TABLE` hit (`02-domain-model.md:1941`) is `EXECUTE format('DROP TABLE archive.%I', tbl)` — dynamic SQL text inside `fn_retention_batch`'s dollar-quoted body, same shape as the `archive` false-lead from my base review. `split_statements` treats the whole function as one statement starting with `CREATE FUNCTION`, so the old regex never saw this text as a candidate `DROP TABLE`.

So: this closes a real, verified latent gap in the derivation's mechanism, not an active defect in today's data. Alex's writeup says "were all silently ignored" as a property of the old code — that holds up exactly as stated; it doesn't claim these occur today. If "bug fix wearing a refactor label" implies something is wrong right now, it isn't; if it means the old code had a real, demonstrable blind spot, it did, and I reproduced it directly rather than accepting the list.

**Refusal is loud and named, confirmed directly:**
```
expected_tenancy(['DROP TABLE some_real_looking_table_name'])
→ AssertionError: statement kind this derivation does not classify, ... : DROP TABLE some_real_looking_table_name
```

### Change 2 — the vacuity tripwire: `print()` → `assert` → (987d036) a self-explaining `assert`

Pulled `263d3f0..987d036` myself. Confirmed: one file, +15/-4, entirely inside the assertion's message string. The condition (`assert not expected`) is byte-identical. Changes 1 and 3 unaffected by this commit — not redone, per instruction.

**This is the one you asked me to actually judge.** The new message closes the human-legibility gap well: it opens with "EXPECTED FAILURE — you have not broken anything," names the tables, states the one correct response, and rules out the two most tempting wrong ones with the Fork 1 citation that makes the current-looking-wrong state actually correct. Someone who opens the failure will not mistake it for a regression. Agreed on that part.

But I'd weight your still-open concern more heavily than a residual worry: **a plain `assert` and an `xfail` read identically to a human in the traceback, but not to anything reading the summary line** — a CI badge, `pytest --tb=no -ra`, a regression-diff gate. This repo's own CI runs `pytest tests/ -v` and reports counts; my own verification approach this whole session has repeatedly leaned on exactly that axis over message-reading. Once F.2.2 lands, this reports as a plain `FAILED`, indistinguishable at the count level from a real regression, to anything that isn't a human reading this one test's output.

I don't think it's a binary choice. `pytest.xfail(reason=...)` called imperatively, right where the assert sits (`if expected: pytest.xfail(reason=<same message>)`), gets both: the message survives verbatim, and the outcome reports as `XFAIL` in every summary a tool would check, not `FAILED`. Same stopping point as today's `AssertionError` — nothing downstream changes. One real implementation note: this needs the imperative call, not the `@pytest.mark.xfail` decorator — the decorator's condition is collection-time and can't see a value computed by running the replay, and applying it unconditionally would make today's genuinely-passing run report `XPASS` instead of `PASSED`. I'd merge this delta as-is if nothing here depends on automated count-diffing — but I'd bet something does, and I'd treat it as a real follow-up, not a blocker.

### Change 3 — FORCE arm deleted

Confirmed directly, all three parts: the 0-count claim reproduced independently; `tenancy_signature` (live side) unchanged, still reads `forced` from the real catalog; and the containment is actually **stronger** than "reddens the comparison" — a `FORCE ROW LEVEL SECURITY` statement in plan text isn't in `_TENANCY_IRRELEVANT` and doesn't match any known pattern, so it makes `expected_tenancy` **raise immediately**, not wait for an equality mismatch. Confirmed by constructing the statement and running it. The allowlist is what's doing the protective work here, not the field staying in the shape.

### Test count arithmetic (188 → 195, claimed +7)

Checked by enumeration. Delta removes 1 test, adds 8 (a 6-case parametrize replacing it, plus 2 new). 8 − 1 = +7. Exact.

### One thing neither writeup mentions

`test_the_prefix_comparison_holds_with_real_tables_and_fails_on_an_extra_one` lost its `@pytest.mark.integration` marker in the `/simplify` pass. Not inert — `-m integration` is real and used (`README.md:246`, `Makefile:70`). After this, that test (the positive control with real staged tables + the undeclared-table mutation) is excluded from a `pytest -m integration` run. None of this class's other tests carry the marker either, so it's now internally consistent — but that also means the whole class is unreachable through the documented integration-only path. Naming it, not blocking on it — I don't know if anything depends on that path actually including it.

PR #823, opened 2026-08-14
fix(tests): refuse pytest-xdist in tests/scripts, and fix the two guides (#809)

chrisrogers37:
Closes #809. Built on rajan's investigation — diagnosis and placement confirmed independently, one of his findings corrected.

## What this is, and what it is not

`tests/scripts/` serializes every session on `SUITE_CLUSTER_LOCK_KEY` because the seven `svc_*` roles are cluster-scoped spec names that cannot be namespaced per session (#768). Under `-n`, each worker is a separate process with its own session, so workers 2..N queue on a lock the first holds until the session ends — invisibly, because the wait loop `print`s from session-scoped fixture *setup* and `pytest.ini` sets no `-s`. Twenty silent minutes, then a `RuntimeError`.

**This refuses the run. It does not make the directory parallel-safe.** Nothing here moves it toward that, and the guard's docstring says so in those words — the risk is a reader six months out concluding the suite is parallel-safe *because* a guard exists. What changed is that an unsafe invocation now fails in seconds with an explanation.

## Correction to the handoff: the presentation depends on the invocation form

rajan measured one form on a mirror repo. Measured here against real pytest 9.0.3 + pytest-xdist 3.8.0, against **this** repo, both forms:

| how `tests/scripts` is reached | result |
|---|---|
| **descent** — `pytest -n auto`, `testpaths = tests` (the form both guides printed) | `ERROR tests/scripts` in the short summary, ~16s, **zero** execnet noise, **rest of the suite still passes** (`5 passed, 2 errors`) |
| **initial argument** — `pytest tests/scripts -n auto` | worker-boot failure: `no tests ran`, 20 lines of execnet traceback, ~12s |

A conftest belonging to an initial arg is loaded at worker *startup* rather than during collection, so the same raise surfaces as a boot crash. It is still loud rather than silent — the full message is present and it fails in ~12s versus 1200s — but it is ugly, and I would rather state that than let the tidy row read as a promise about both. rajan's finding stands for the form the guides recommend, which is the filed defect; it does not generalize.

**Refusing on the controller was considered and rejected.** `config.option.numprocesses` would make the second row clean — but the controller also collects, so it would abort the *whole* run in the first row, turning "the rest of the suite still passes" into "nothing runs". That trades a good outcome in the common form for a tidier one in the rare form.

## The docs — I diverged from the recommendation here

rajan suggested keeping `pytest -n auto` and appending a note. I changed the command instead. After this guard lands, the printed command exits 1; a guide whose command fails is still the defect #809 was filed about, because the reader copies the command, not the caveat. Both guides now print:

```bash
pytest --ignore=tests/scripts -n auto
pytest tests/scripts
```

**Verified before recommending it** — `rc=0`, 5 passed, guard silent. Advice that ships in an error message and a guide should be run first.

## Verification

- **Full `tests/scripts/` serial: `rc=0`, 184 passed** (181 post-#821 baseline + 3 new), 169s. Run *after* rebasing onto #821, so this proves the two conftest changes coexist.
- **Real xdist runs**: the table above, re-confirmed post-rebase (`5 passed, 2 errors`, 0 noise).
- **Escape route**: `pytest --ignore=tests/scripts -n 2` → `rc=0`, guard silent.

**Mutation-proven — four mutants, each detected, tree restored clean after each:**

| mutant | result |
|---|---|
| M1 guard never fires (`if False`) | refusal RED, **positive control GREEN** |
| M2 guard always fires (`if True`) | pytest `rc=4` UsageError — the whole directory unrunnable |
| M3 predicate loosened to a near-miss attribute | refusal RED **and** near-miss RED |
| M4 escape route dropped from the message | refusal RED |

M1 is the one that matters for vacuity: removing the guard reddens *only* the refusal and leaves the positive control green, so the tests discriminate rather than both keying on the guard's existence. M2's `rc=4` was checked rather than inferred — it is pytest's UsageError carrying the guard's own message, which is the mutant being detected as loudly as possible, not a broken harness.

**The tests do not require the plugin they refuse.** pytest-xdist is deliberately not a dependency, so the predicate is exercised against a stub config and CI proves the guard with xdist absent. The end-to-end behaviour is proven by running it for real, above.

## Out of scope, deliberately

Option 3 from the issue — the `SUITE_LOCK_WAIT_SECONDS = 1200` bound and routing the wait message to the terminal — is left alone. It concerns the *legitimate* case of two concurrent serial runs queueing, which works as designed per #785, and this guard removes the urgency for the xdist case specifically. Worth its own issue if still wanted.

Note for review: this branch was rebased onto #821 after that merged. `git diff origin/main..HEAD` (two-dot) equals the three-dot diff and shows zero deletions of #821's conftest work.

https://claude.ai/code/session_017LN1M9PRwjsqWkKGNXwnth


chrisrogers37 (commented):
**Approve**

Ran the decisive experiment myself rather than judged the argument. Short version: branden's correction is real, and it does not put the guard in the wrong file — I proved that with a control neither he nor rajan ran. All four asks below, plus bounds on what I covered.

## 1. The correction itself (load-bearing)

Checked out `78ab022` fresh (merge-base = `origin/main` tip, single commit, no rebase ambiguity), installed pytest-xdist 3.8.0 into the venv (confirmed absent beforehand, matching rajan's finding it's not a project dependency), matched pytest 9.0.3.

**Reproduced both invocation forms against the real repo, real guard, at the PR's actual placement** (`tests/scripts/conftest.py`):

- **Descent** (`pytest -n 2 -k "<fast unrelated test>"`, no explicit path arg — `testpaths=tests` walks in): `ERROR tests/scripts` in the short summary, clean pluggy-only traceback (no execnet noise), other test still passed — `1 passed, 2 errors` in 15.8s. Matches branden's row exactly.
- **Initial argument** (`pytest tests/scripts -n 2`, named explicitly): worker-boot crash — `maximum crashed workers reached: 8`, full execnet tracebacks, `no tests ran` in 6.7s. Matches branden's row, and if anything is worse than his description ("20 lines of execnet traceback" undersells the repeated crash-and-retry).

So at a **fixed placement**, varying only invocation form flips clean↔ugly. That confirms branden's claim on its own terms.

**But that alone doesn't settle whether rajan's placement finding was a confound or just incompletely scoped** — rajan's own "clean" result for `tests/scripts/conftest.py` placement used bare `pytest -n 2` (descent); he never tried `pytest tests/scripts -n 2` (initial-arg) at that placement, so branden's correction fills a cell rajan's matrix never had, rather than contradicting a cell it did. To close the actual question — is the guard in the *right* file — I ran the control neither of them ran: a temporary root-level `conftest.py` with an equivalent guard, tested under the **same descent form** that was clean at `tests/scripts/`:

```
pytest --ignore=tests/scripts -n 2 -k "<same fast test>"   # root guard, descent
→ maximum crashed workers reached: 8, no tests ran          # UGLY, same as initial-arg
```

Root is ugly under **descent too** — the one form where `tests/scripts/` placement is clean. That's consistent with root `conftest.py` always loading at worker startup regardless of invocation, versus a subdirectory conftest loading at startup only when that subdirectory is itself the named argument. Cleaned up (`rm ./conftest.py`, untracked, tree confirmed clean after).

**Net: both findings are correct and complementary, not contradictory.** Placement matters (root is bad under every form tested; `tests/scripts/` is bad under only one) *and* invocation form matters independently (even the better placement isn't uniformly clean). The guard is in the strictly better file — rajan's recommendation stands, branden's correction is additive, and "the guard may be in the wrong file" is closed: no, it isn't.

## 2. Refuses ≠ safe — stated plainly

Confirmed in source, not just PR prose. Docstring: *"Read this as 'the door is locked', never as 'the room is now safe for a crowd'."* Runtime error message itself: *"...so it is not safe under pytest-xdist. Refusing now..."* Both land where a future reader actually encounters them — the message included.

## 3. Docs — ran the exact printed commands, didn't infer

- `pytest --ignore=tests/scripts -n auto` → **2436 passed, 8 skipped in 64s, real rc=0.** (2444 collected via `--collect-only` earlier; 2436+8 reconciles.)
- `pytest tests/scripts` → **184 passed in 192s, real rc=0.** (181 pre-#821 baseline + 3 new, matches the PR's claim.)

Both exit codes captured directly from the command (`$?` immediately after, no intervening pipe — the classic gotcha that would launder a failure into a false 0 doesn't apply here).

## 4. Mutation-check — all four, not a sample, plus the positive control

These are fast, DB-free unit tests (`test_xdist_guard.py` calls `pytest_configure` against a stub `Config`), so I reproduced all four claimed mutants on the live branch, reverting to a clean tree between each:

| mutant | what I did | result | matches claim? |
|---|---|---|---|
| M1 guard never fires (`if False`) | sed | refusal `DID NOT RAISE`, positive control **and near-miss both stayed green** | yes |
| M2 guard always fires (`if True`) | sed | real invocation → **rc=4**, UsageError, session dies before any test gets a verdict | yes, but see note below |
| M3 predicate loosened (`workerid` not `workerinput`) | sed | refusal RED **and** near-miss RED, positive control green | yes |
| M4 escape route dropped from message | sed | refusal RED on the specific assertion (`"the refusal must name the way to keep parallelism"`) | yes |

Positive control holds throughout — M1 and M3 leave `test_a_controller_or_serial_run_is_untouched` green, so the guard is discriminating on `workerinput` specifically, not just existing.

**One precise, non-blocking correction**: the CHANGELOG's gloss — *"an unconditional raise reddens the positive controls only"* — undersells M2. What I actually observed: the whole pytest *invocation* dies at rc=4 before collection, so nothing gets an individual pass/fail verdict at all, refusal test included. The PR body's own table has this right (*"the whole directory unrunnable"*); the CHANGELOG's one-line compression doesn't. Cosmetic, but worth a wording fix since CHANGELOG is the durable record and PR body isn't.

## Bounds

Real repo throughout (not a mirror), real pytest 9.0.3 + pytest-xdist 3.8.0, `-n 2` for the matrix tests and real `-n auto` (4 workers, this host) for the doc commands. All 4 mutants reproduced, not sampled. One control run beyond what was asked (root placement). Did not re-run the full CI matrix myself — cross-checked `gh pr checks` independently instead (7/7 green). Did not investigate #822 beyond confirming it's genuinely CONFLICTING and content-unrelated, consistent with the merge-train framing.

Everything checked — the correction, the placement question, both doc commands, all four mutants — held up under direct reproduction.

PR #826, opened 2026-08-17
test: pin the #801 CONNECTION LIMIT 0 reproduction (#804)

### LINKED ISSUE #804: The #801 CONNECTION LIMIT 0 reproduction is not pinned as a CI test — the discriminator can silently regress
## Summary

#801 (merged) fixed the `server_answered()` discriminator for `too_many_connections` — a refused connection is a server that answered. The fix is correct and was verified by independent reproduction during review.

**The reproduction itself is not pinned as a CI test.** Raised as a non-blocking note by the reviewer on #801 and filed here rather than dropped.

## Why it is worth pinning

The reviewer reproduced the failure by creating a Postgres role with `CONNECTION LIMIT 0`, which forces the exact `too_many_connections` condition the discriminator has to classify. That reproduction is what established the fix works — and it currently exists only in a review comment and in the reviewer's session.

The original defect was that `pgcode` came back `None` on **both** sides, so the old exception-based discriminator could not separate "server answered and refused" from "server not listening." Nothing in CI exercises that specific condition today, so a future change to the listener probe can silently reintroduce exactly the ambiguity #801 removed, and the suite stays green.

This is the same shape as **#152** in another repo: a cluster of issues fixed, none guarded.

## Suggested work

Pin the `CONNECTION LIMIT 0` reproduction as an integration test:

1. Create a role with `CONNECTION LIMIT 0` in fixture setup, attempt a connection, and assert `server_answered()` returns **true** for it.
2. Assert the **negative** direction in the same test file — a genuinely absent listener returns **false**. The defect was an inability to distinguish the two, so a test covering only one side does not guard against its return.
3. Mutation-check it: break the listener probe and confirm the new test goes red for the right reason.

**Size:** S. **Related:** #801 (merged), #769.

chrisrogers37:
Closes #804.

`server_answered()` separates *"no database was configured"* — the only condition under which skipping the integration tests is honest — from *"a database was configured and refused"*, which is a real failure. #801 established that separation by creating a PostgreSQL role with `CONNECTION LIMIT 0` and observing a refusal arrive from a still-listening address.

**That reproduction existed only in a review comment and in the reviewer's session.** Nothing in CI re-ran it, so a change to the listener probe could silently restore the #769 ambiguity with the suite green.

## What was already covered, and what was not

Every existing test in `TestTheListenerProbe` monkeypatches `server_is_listening` to a lambda. That is right for pinning how `server_answered` **combines** its two inputs, and blind to whether either input is still what it was measured to be. The new `TestTheRefusalAgainstARealServer` uses no mocks: real role, real libpq, real socket probe.

## Mutation-proven, and one mutant isolates the gap exactly

| mutant | what it breaks | reds |
|---|---|---|
| **A** — delete the listener branch from `server_answered()` (restore pre-#769) | the fix itself | 4 — 2 existing, 2 new |
| **B** — real `server_is_listening` always returns False | the probe | 3 — 1 existing, 2 new |
| **C** — `server_answered()` probes the **wrong address**, probe itself intact | the wiring between them | **2 — both new; every existing test green** |

**C is the one that matters.** It is the shape of a refactor that passes the wrong host/port, and the entire pre-#804 suite is blind to it: the lambdas that replace `server_is_listening` **ignore their arguments**, and `test_an_occupied_address_is_listening` exercises the probe standalone, so neither can see it. That is the proof these tests add coverage rather than restating what was there.

**Positive controls stayed green in all three arms** — `test_a_genuinely_absent_listener_is_still_an_honest_skip` and `test_the_exception_still_cannot_discriminate` never move, as does the whole `TestTheVerdict` / `TestTheSkipCeiling` / `TestTheRequirementSwitch` set. A guard that reddens on everything would be green for the wrong reason.

Mutation results were **re-measured after `/simplify`** and are unchanged.

## Verification

**Reproduction measured before a line was written** — PostgreSQL 15.19 / psycopg2 2.9.12:

```
CONNECTION LIMIT 0 -> OperationalError, pgcode=None
  FATAL:  too many connections for role "probe804_43af8db1"
  server_is_listening: True
free port          -> OperationalError, pgcode=None
  Connection refused
  server_is_listening: False
```

**Suite** — root suite (`tests/`, `--ignore=tests/scripts`), both arms taken on this host with the concurrent-run detector reading clear at each boundary:

| | rc | count |
|---|---|---|
| baseline (`git stash`) | 1 | 7 failed, **2429** passed, 8 skipped |
| this branch | 1 | 7 failed, **2433** passed, 8 skipped |

`+4 passed` is exactly the four new tests; **zero failure names introduced**, zero fixed. The 7 are the known pre-existing `tests/src/services/test_telegram_service.py` cluster (`AIORateLimiter` — missing local optional dep; CI installs the extra), the same 7 #801 recorded.

**The precondition gate, measured both ways** rather than reasoned:

| condition | `REQUIRE_TEST_DATABASE` | result |
|---|---|---|
| no server | unset | 3 named SKIPs, rc 0 — unmarked negative test still passes |
| no server | `1` | 4 errors, from `setup_test_database` (the session fixture wins the race) |
| server, actor lacks CREATEROLE | unset | 3 named SKIPs, rc 0 |
| server, actor lacks CREATEROLE | `1` | **3 errors, from the new gate** — exactly the tests that use the fixture |

Verified against a real `LOGIN CREATEDB` role with no `CREATEROLE`, created and dropped for the measurement; no strays left (`probe804%` roles: none; `.env.test` restored byte-identical).

`ruff check` and `ruff format --check` clean.

## Judgement calls, stated rather than glossed

- **The refusal is asserted, never assumed.** `server_answered()` also returns True for a connection that **succeeds**, so a probe role that stopped being refused would leave every test green having reproduced nothing. Each test proves the refusal first, and the fixture asserts its own postcondition (`rolconnlimit = 0`, `rolsuper = false`) — a superuser is exempt from `rolconnlimit`, so that flag alone would turn the class into a happy-path test. (Documented PostgreSQL behaviour, **not measured here**: a CREATEROLE-only test role cannot create a superuser to check it against.)
- **A missing precondition fails where coverage is mandatory**, rather than skipping. `tests/scripts/` skips unconditionally on the same missing privilege; this diverges because a guard that quietly does not run cannot be told apart from one that ran and found nothing — the defect #804 was filed about. A single extra skip would not breach `MAX_EXPECTED_SKIPS` either, so the backstop one layer up would not catch it.
- **Not routed through `integration_verdict`.** Doing so would mean writing `server_answered=False` for the CREATEROLE case, where a server demonstrably *did* answer and merely lacked a privilege — a false argument, in the one file whose whole subject is that distinction. The part that can actually drift, `database_is_required()`, is called rather than re-parsed.
- **The pair test adds no mutation-detected coverage** over the two single-direction tests — it reddens exactly when they do, in all three arms. Kept because it is the only assertion whose *subject* is the distinction rather than one side of it, and because it composes the two through `integration_verdict` the way `setup_test_database` actually calls them. Disclosed in its own docstring so it can be deleted with that knowledge rather than instead of it.
- **The negative half is deliberately unmarked.** It needs no PostgreSQL, only an address with nothing behind it, so it guards the negative direction even where the positive half must skip.
- **Cluster footprint kept minimal.** Uniquely named off #763's session token (no collision with concurrent checkouts), owns nothing and is granted nothing (no `pg_shdepend` rows, no dependency-blocked `DROP`), dropped in a `finally`. A leak from a hard kill is inert rather than poisoning, which is why there is no sweep. Kept **function**-scoped: session scope saves ~84 ms and holds a maintenance connection open for the whole run against a cluster this repo already contends on — the wrong trade on the axis that actually bites here.

## `/simplify`

Run with 4 agents; 6 findings applied — two setup helpers replacing six repeated `monkeypatch` pairs, `contextlib.closing` flattening the fixture's nested `try/finally`, a derivable `server_is_listening` assertion dropped, the credential yielded as a pair (a password mismatch raises the same `OperationalError`, so name-plus-module-constant was convention rather than contract), and two rejected-alternative records. Skipped, with reasons in-line: hoisting the CREATEROLE probe to share `actor_lacks_createrole` (edits the module under test plus a suite outside this change, to share one catalog query over stable semantics — worth doing next time either copy is touched), session-scoping the fixture, and adding `connect_timeout` to `maintenance_connection` (pre-existing, outside this diff).

`/weigh-development-paths` was not run — there was no open junction once #804 named the reproduction, the file, and the mutation bar.

## Bounds

- **`tests/scripts/` was not run locally.** It serializes on a cluster-wide advisory lock and another checkout was intermittently active on the shared PostgreSQL during this work. Nothing in this change reaches it — the new tests are in the root suite and touch no `svc_*` role — and CI runs `pytest tests/` unfiltered.
- **CI is where this guard has teeth**, and it will run there: `test_user` is the `postgres:15` service's superuser, so the CREATEROLE precondition holds, and `REQUIRE_TEST_DATABASE: "1"` means a missing precondition fails rather than skips.
- Unrelated to this branch: **6 orphaned `storyline_test_*` databases** are sitting on the local cluster from interrupted runs. Not reaped — a scratch database that cannot be attributed to your own run may be someone's live session (#758).


chrisrogers37:
Pointer, since this came up in review: **the reasoning for shipping #804 alone** — rather than folding in #758 and #773, which the task named as possibly-one-PR — is recorded on the issues themselves, not here:

- [#758 — why #804 shipped alone, and where part 3 actually stands today](https://github.com/chrisrogers37/storydump/issues/758#issuecomment-5318483114) (includes a correction: the last status comment there is stale — #785, #808/#821 and #809 have all landed since, evidence read off `main` at `c28066d`)
- [#773 — considered for bundling and deliberately kept separate](https://github.com/chrisrogers37/storydump/issues/773)

Short version: neither is code-shaped. #758 needs a status judgement, #773 has had three evidence passes and does not currently reproduce — and the only change a bundled PR could have made there is the broader lock that issue explicitly refuses. They also share no diff surface with this PR, which touches one test file and the changelog.

chrisrogers37 (commented):
**Approve**

The decisive question first, since it's the one thing that decides this review: this is a real reproduction, not a policy proxy. Held branden to his own #821 bar (mutation-proven per part, positive control green) rather than a new one, and it holds.

## The mutation check

Reintroduced the exact condition #804 is about — restored pre-#769 behavior so a `CONNECTION LIMIT 0` refusal misreads as absent — by mutating `tests/conftest.py:216` (`if server_is_listening(...)` → `if False:`). Ran the whole file:

```
4 failed, 24 passed
FAILED TestTheListenerProbe::test_a_refusal_from_a_live_server_is_not_an_absent_server
FAILED TestTheListenerProbe::test_the_two_refusal_paths_reach_opposite_verdicts
FAILED TestTheRefusalAgainstARealServer::test_a_connection_limit_refusal_is_a_server_that_answered
FAILED TestTheRefusalAgainstARealServer::test_the_two_real_paths_reach_opposite_verdicts
```

Exact match to the PR's claimed table (mutant A: "reds 4 — 2 existing, 2 new"). And the failure isn't abstract — the captured output is a real refusal from a real role the fixture created seconds earlier: `FATAL: too many connections for role "probe804_804a2439ca_5f856b86"`. That's the actual #801 condition, reproduced and correctly detected as a bug when I broke the detector.

Then reverted and ran branden's mutant C — `server_answered()` probing the wrong address, `server_is_listening` itself left intact (the shape of a refactor that passes the wrong host/port):

```
2 failed, 26 passed
FAILED TestTheRefusalAgainstARealServer::test_a_connection_limit_refusal_is_a_server_that_answered
FAILED TestTheRefusalAgainstARealServer::test_the_two_real_paths_reach_opposite_verdicts
```

Also an exact match — every `TestTheListenerProbe` test stayed green, confirmed directly: the mocked tests replace `server_is_listening` with a lambda that ignores its arguments, so wrong-address wiring is invisible to them by construction. This is the sharpest evidence in the PR, and it holds.

**Positive control**: the unmutated baseline is 4/4 new tests passing (3.5s, real role create/drop, real libpq), and both mutations above left the untouched tests green (24 and 26 respectively) — the guard discriminates, it isn't just strict.

## Check 1 — which part does which

Confirmed by reading structure, not assuming: the file has four pre-existing classes (`TestTheListenerProbe`, `TestTheVerdict`, `TestTheSkipCeiling`, `TestTheRequirementSwitch` — genuinely policy: skip/fail/run decision logic, the skip-ceiling backstop, env-var parsing) and one new one (`TestTheRefusalAgainstARealServer` — genuinely a reproduction, no mocks). Both are real; the filename describes the pre-existing majority, not what this PR adds. Checked the CHANGELOG doesn't blur that: it doesn't — "Every existing test in `TestTheListenerProbe` monkeypatches... The new `TestTheRefusalAgainstARealServer` uses no mocks" is stated plainly, and the PR title scopes itself to "pin the #801 reproduction" without claiming to introduce a policy. No overclaim found.

## Check 2 — does CI actually run this, or silently skip it

Three independent layers, not just the config:

1. **CI config** (pre-existing, this PR doesn't touch `.github/workflows/ci.yml`): real `postgres:15` service, `POSTGRES_USER: test_user`, and `REQUIRE_TEST_DATABASE: "1"` already set — with an inline comment naming exactly the failure mode this check is about ("a service container that fails to come up... the run still exits 0 and shows a green tick"). No `-m` filter on the pytest invocation, so `@pytest.mark.integration` isn't excluded.
2. **CI's actual pytest log** (pulled from the real run, commit `5973d02`): all 4 new tests show `PASSED`, not `SKIPPED`.
3. **CI's actual Postgres server log**: genuine `FATAL: too many connections for role "probe804_36bc479dc6_..."` entries — the real database engine refusing the real probe role, in the real CI container, today.

Not a green tick that never ran.

## Check 3 — CHANGELOG coverage honesty

Precise rather than promotional: names which tests are mocked vs. real, gives exact mutation counts matching what I independently reproduced, and discloses the one weak spot itself — the paired test "adds no mutation-detected coverage... today" over the two single-direction tests, kept for a stated reason (composes through `integration_verdict` the way the real caller does) rather than silently included as if load-bearing. No inflation found.

## CI status (REST, not GraphQL per your note)

`gh api .../commits/5973d02.../status` → all green. `gh api .../check-runs` → 6/6 completed, success, including `Test`. `mergeable_state: blocked` is just the missing review — no other blocker.

## Bounds

Ran: both explicitly-requested mutants (A and C) by hand, reverted cleanly between (`git status --porcelain` confirmed empty each time); the new test class alone and the whole file, unmutated, as a baseline; pulled and grepped CI's actual pytest and Postgres logs for this exact commit.

Did not: run the full 2433-test suite locally (branden's own bounds note explains why — cluster lock contention with other checkouts — and I had no reason to re-litigate that; CI's own green run covers it, and I independently confirmed CI's log rather than trusting the PR's pasted numbers). Did not verify mutant B (probe always returns False) — A and C were the two that most directly answer your decisive question and branden's own "the one that matters" claim; B is the least novel of the three (loosens an existing, already-tested code path) and I judged the marginal evidence not worth the same real-role setup/teardown cycle a third time. Did not independently verify that a fresh `postgres:15` container's `POSTGRES_USER` becomes its superuser beyond citing standard, well-documented Docker Postgres image behavior — not measured against a fresh container myself, though CI's own log confirms the CREATEROLE-gated fixture path ran (not the `_precondition_absent` fail path), which is consistent with it holding.

PR #828, opened 2026-08-17
F.2.2 (2/2): land 02 §1's identity and tenancy tables (#806)

chrisrogers37:
## What

F.2.2, 2-of-2. Migration **053** lands `02` §1's seven identity and tenancy tables — the first tables of the target schema — plus the `TargetBase` models that mirror them.

Advertised-stream statements **2..21**, in stream order, after 052's two shared functions at 0..1. Arm (b) reports `ok=True vacuous=False f2_count=22`.

| | |
|---|---|
| Tables | `users`, `user_identities`, `workspaces`, `workspace_members`, `workspace_invitations`, `channel_bindings`, `onboarding_sessions` |
| Tenant-keyed | 4 of 7 — `workspaces`, `workspace_members`, `workspace_invitations`, `channel_bindings` |
| RLS / policies | **none**, per the Fork 1 ruling (a) |

## Why this shape

**The file body is generated from the stream, not transcribed.** Arm (b) is an *ordered prefix* diff, so a hand-copied `CREATE TABLE` differing by one word fails at a statement index rather than at the word. The migration was written by extracting statements 2..21 through the repo's own extractor, with a write-time assertion that the resulting file normalizes back to exactly those statements. Edit the plan and the manifest ratchet; never this file alone.

**No policies, no RLS — deliberately.** The stream creates all 26 tables before the first `ENABLE ROW LEVEL SECURITY` (index 126) or `CREATE POLICY` (149), so an increment carrying its own policy would not be a prefix and would fail arm (b) at the first policy statement. The four tenant-keyed tables carry no policy until F.2.9; that window is the ruling's stated cost, and the prefix-aware lane check accounts for it on both sides rather than going red on schedule for five increments.

**The 1-of-2 tripwire fired and was deleted — the response its own message specified.** `test_lineage_lane.py`'s `pytest.xfail` block existed to fire exactly once, at the increment that landed the first tables. It did. The `sig == expected` comparison it sat under is **untouched** and is **not** bounded to a complete lineage; in its place is a plain `assert expected`, so the vacuity it warned about is a live requirement rather than a warning about a future one.

**Target models land per increment, and the gate decided that rather than taste.** Lane parity compares the replayed `public` against `create_all` on `TargetBase`, so tables on the migration side with no models behind them *is* the drift it reports. Deferring models to one late pass means running the gate knowingly red from F.2.2 through F.2.7 — the "red and known" cost Fork 1 declined when it declined option D. The models-side database is given 052 first: `ck_ws_tz_valid` calls `fn_safe_tz`, and a CHECK naming a missing function is a hard error, not a skipped constraint. That the prerequisite cannot contaminate the comparison is asserted in place (it creates no relations) rather than argued.

## Four pre-existing lane assertions reshaped, not relaxed

Each said "public is empty" or "no declared table name is in public" — both true only while the target lineage created no relations, and both now false for the correct reason.

- **`public is empty` → `public holds exactly the tables the lineage implies`.** Still red on a legacy leftover, and *additionally* red on an undeclared target table. Emptiness is what the *move* establishes and F.2 then consumes; asserting it at the end of the lane would now be asserting F.2 has not started.
- **The legacy-inventory check narrows to legacy-ONLY names.** The target schema deliberately reuses `users` and `onboarding_sessions`, so from 053 those names exist in both schemas and a name-keyed intersection reports a stranded legacy table where there is none. Those two are covered by *identity* rather than name in the parity gate, which compares their columns.
- Every reshaped assertion asserts its own discriminating power first (`assert legacy_only`), so a future increment that consumes the last legacy-only name reddens rather than passing vacuously.

## Latent defect caught in review

Two first-draft postconditions asserted the **absence** of state a later increment adds (`count(*) = 0 FROM pg_policies`, `NOT relrowsecurity`). `migration_runner.py:552` derives a file's **permanent adoption probe** from its postconditions when the adoption manifest has no entry — so when F.2.8/F.2.9 land RLS and policies, 053 would probe *unapplied* on a database plainly holding all seven tables, and `runner adopt` raises `incoherent chain` naming **053** while the real cause is the increment after it. Mid-cutover, pointing at the wrong file.

Replaced with four postconditions scoped to objects 053 itself creates, none negative, with the rule stated in the file so the next increment does not re-introduce it.

## Deletions

- **The staged positive control from 1-of-2 is gone**, because the thing it simulated is now real: it replayed the F.2.2 segment into a throwaway tree *precisely because* the lineage was empty, and it would in any case collide at version 53. Its one non-duplicated half — a correctly-born but **undeclared** table reddens the prefix comparison while staying invisible to the invariant check — moved onto the existing `lane_tenancy_probe` in the real lane, where both predicates now answer about one object. Its plan-side facts (7 tables, 4 tenant-keyed, no RLS, no policies) were already pinned DB-free by `test_tenancy_gate.py`. Net: **one fewer full-corpus replay**, same coverage.
- `run_lane(migrations_dir=…)` and `real_stream(normalized=…)` went with it — those parameters existed only for that control.

## `/simplify` pass

- **Derived, never listed — tables *and* functions.** `implied_tenancy()` owns the prefix rule in one place; `implied_target_tables()` / `implied_target_functions()` read off it. The function inventory matters more than it looks: 18 `CREATE FUNCTION`s in the stream, 4 landed, and because the assertion compares `proname` order the remaining 14 are *insertions into a sorted list*, not appends. Verified the derivation reproduces the hardcoded four exactly.
- **The one list kept deliberately is the lineage FILE list** — arm (b) compares statements, so a comment-only file above the move contributes nothing and leaves it green; only a file-level assertion sees it. That hole is now *also* closed mechanically (every lineage file must contribute ≥1 statement), leaving the list doing only the ratification job.
- **Shared column vocabulary → `src/models/target/columns.py`.** `TZ`, `NOW`, `GEN_UUID`, `pk()`, `timestamps()`, `fk()` were homed inside F.2.2's own module, where F.2.3 could only reach them by importing a private name or re-deriving them — and re-deriving `TZ` is exactly the mistake it prevents (naive `DateTime` renders `timestamp without time zone`). `fk()` also collapses 11 five-line FK blocks to one line each.
- **One spelling of the advertised stream.** `test_tenancy_gate.py` had a cached builder; the lane grew a byte-identical copy *without* the cache. Promoted to `conftest.advertised_stream()`, both suites import it. `F2_2_START` went with the deleted control.
- A real bug the parity gate structurally could not catch: `text("'{\"v\":1}'")` reads `:1` as a **bind parameter** and emits `DEFAULT '{"v"NULL}'`. The gate compares types/constraints/uniques/FKs, never defaults — it surfaced as `create_all` raising.

## Verification

- **Full suite: `2638 passed, 8 skipped`, rc=0** — unchanged from `main` before and after the `/simplify` pass. No failures to attribute.
- `tests/scripts/`: `202 passed`.
- `ruff check src/ cli/ tests/` clean; `ruff format --check src/ cli/ tests/` clean (312 files). `scripts/init_db.py` reports unformatted on `main` and is untouched here — outside the documented CI scope.
- Arm (b) driven directly against the real docs + manifest: `F2PrefixReport(ok=True, vacuous=False, f2_count=22, first_divergence=None)`.
- Derivations checked independently of the tests consuming them — 22 statements, the 7 tables, the 4 tenant-keyed, the 4 functions, `any rls/policy: False`.
- `/simplify` run before first push (4 lenses), `/weigh-development-paths` not run — no open fork; Fork 1 was ratified on #806 before this started.

Closes the F.2.2 increment. Next is F.2.3 (§2 accounts/sources/media, 6 tables) — note it is **#793-coupled**: an allow-outcome there edits `ck_sources_provider`, which is inside the manifest's sha256 bijection.

https://claude.ai/code/session_01Qav4XLsCTkyUWYxN7qhPZG


chrisrogers37 (commented):

Reviewed as-is, not rebased, per ari's sequencing. Everything below is executed, not read — cloned at the PR's actual head (`c1bbac0`), built a venv, and ran the real code. No live Postgres was available, so the migration itself was not applied; everything DB-free was.

## The tripwire (the thing only I can judge properly)

Read the full removed block against its own prescription. The tripwire's message said, verbatim: *"delete this xfail block (and the comment above it) as part of the increment that landed the tables. Nothing else. Do NOT relax the `sig == expected` comparison above it, and do NOT bound it to a complete lineage."*

Checked both named forbidden repairs directly in the diff:
- `sig == expected` (`test_lineage_lane.py`) is **unchanged context** — not a single character touched.
- The replacement `expected = implied_tenancy()` still slices `advertised_stream()[:_lineage_length()]` — a **prefix of the current lineage's own length**, exactly the same shape as before, just deduplicated into a cached helper (matches `target_lineage_statements`' own docstring warning about two derivations drifting). **Not** bound to the complete 257-statement stream.

Fired for the right reason: `expected` is non-empty for the first time now that 053 lands real tenant-keyed tables, so the `if expected:` gate that used to be permanently false is now true — that's the one-time firing, not a coincidence.

The surrounding churn (the "cost of the old shape" RLS-violation assertions appearing in the surviving test) isn't scope creep riding along — it's the **same content**, moved verbatim from the now-deleted `test_the_prefix_comparison_holds_with_real_tables_and_fails_on_an_extra_one` (the staged-fake-lineage positive control), consolidated because the thing it used to fake is now real. Confirmed by diffing that deleted test's body against the new block in the survivor — same assertions, same messages, same mutation-probe shape (`lane_tenancy_probe` reused instead of a second `undeclared_probe`). One fewer artificial test, same coverage, on real infrastructure instead of a `tmp_path`-staged fake — a strictly better trade.

**The tripwire did exactly its one job and nothing else.**

## Arm (b) — reproduced empirically, not trusted

Ran `build_stream`/`load_manifest`/`normalize_statements`/`target_lineage_statements` myself, no DB required:

```
advertised stream length: 257
target lineage length (f2_count): 22
target lineage files: ['052_shared_trigger_functions.sql', '053_identity_and_tenancy_tables.sql']
arm (b): ok=True vacuous=False f2_count=22
```

Matches the CHANGELOG's claim exactly, computed independently. It bears.

## The adoption-probe defect — real, and the fix is correctly scoped

Traced `migration_runner.py`'s `adopt()`/`_load_manifest()`: a postcondition-derived probe must stay true **forever** once a file is applied, because `adopt()` computes `head = max(version, verdict=True)` and flags any lower-numbered `False` as `incoherent chain`, naming the lower file. A negative postcondition (`count(*)=0 FROM pg_policies`) is true right after 053 lands and **false again** once F.2.8/F.2.9 attach policies — so a later `adopt` run would misname 053 as unapplied while the real state is "053 applied fine, a later file's postcondition is fragile." Confirmed the shipped SQL carries only positive/existence postconditions (table count, two `EXISTS` on named indexes, function count) — none of which a future migration can falsify. Real defect class, correctly fixed, in scope (own migration's own postconditions).

## Suite count — chased down, and the real story is more interesting than "unchanged"

This one had a wrinkle worth surfacing on its own. Collected the real test suite (not run — no DB) at three points:

| ref | collected |
|---|---|
| `c28066d` (PR's **actual** merge-base) | 2647 |
| `c1bbac0` (PR head) | 2646 |
| `ecbe087` (current `main` tip) | 2651 |

PR-head vs its true merge-base: **-1**, exactly the one deleted positive-control test, zero surprises — matches the diff byte for byte.

The wrinkle: diffing head against **current `main`** (not the merge-base) first showed **-5**, with 4 phantom missing tests in `tests/test_integration_coverage_policy.py` — a file #828 doesn't touch. Chased it rather than reported it: `git merge-base main pr828-head` is `c28066d`, one commit behind `main`'s tip. The one commit `main` has that alex's branch doesn't is `ecbe087` (#826), which is what *added* those 4 tests, landed on `main` after alex forked. GitHub's `pulls/{n}.base.sha` reports `main`'s **current** tip, not the actual fork point — so it silently overstates what a not-yet-rebased PR "removes." Confirmed stable (repeated collection at every ref gave byte-identical results) before concluding it was a comparison artifact and not flakiness.

Net: **the code changes exactly -1 test**, verified two independent ways (diff reading + executed collection against the correct base). Whether alex's own "2638 passed, unchanged" is exact depends on whether that one deleted test was counted as a pass or a DB-unavailable skip in his environment before deletion — I can't settle that without a live DB myself, so flagging rather than asserting: if it was passing, "unchanged" should really read "2637," off by the one test this PR deliberately removes. Not blocking — the removal itself is correct and documented; only the arithmetic in the report might be off by one.

## SQL / models

Read `053_identity_and_tenancy_tables.sql` and `identity_and_tenancy.py` side by side — table/column/constraint names match throughout, including the two-trigger owner-invariant pair (`ct_members_owner_exists` / `ct_workspaces_owner_at_insert`, both `DEFERRABLE INITIALLY DEFERRED`). The escaped-colon JSONB default (`text(r"""'{"v"\:1}'""")`) is the correct SQLAlchemy escape for a literal colon inside `text()` — confirmed it renders to the same `'{"v":1}'` the raw SQL uses. `test_lane_parity_holds_against_the_target_models` is the mechanical backstop for drift between the two files going forward, and its logic (diff `implied_target_tables()` against `TargetBase.metadata.tables`) is sound.

**Net:** the one thing I was uniquely positioned to check holds exactly. The two things I could execute (arm (b), the adoption-probe mechanism) both check out. The one number I couldn't square is off by at most one and doesn't touch correctness. Real merge candidate.

PR #829, opened 2026-08-17
test(auth): pin the third tenant-credential caller fail-closed (#797)

### LINKED ISSUE #797: auth fail-closed: 2 of 3 tenant-credential callers unverified at #796 merge

chrisrogers37:
Refs #797 — merging this does NOT close it, per virgil's review (see #797 for the full trace).

`_auto_approve_instagram` (`scheduler.py`) is the third of the three tenant-credential callers #796's review flagged as unverified — the other two (`telegram_autopost.py`, `telegram_notification.py`) were pinned in #819, which deliberately stripped `Closes #797` because this one remained. This PR pins the same property at this caller. #797 stays open for a different reason, below.

## Tests only. No source change — the property already held.

The method has no `except` around its body at all — only `try`/`finally`. The `finally` guards its own Cloudinary cleanup (`except Exception: pass`), scoped to that one call, not the credential-resolution path. A failure from `MediaSourceFactory.get_provider_for_media_item()` was never inside that guard, so it already propagated to the caller before this PR. That is the expected, hoped-for outcome of closing a verification gap — reported here as a finding, not framed as a fix, per the standard this class of check now holds to (#826).

## Every existing test is structurally blind to this

Every existing reference to `_auto_approve_instagram` in `tests/src/services/test_scheduler.py` replaces it wholesale (`service._auto_approve_instagram = AsyncMock(...)`). None of them call the real method, so none can reach a failure injected at resolution, inside the actual body. This does — no mocking of `_auto_approve_instagram` itself, only its two constructed dependencies (`InstagramAPIService`, `CloudStorageService`) and the one classmethod seam (`MediaSourceFactory.get_provider_for_media_item`).

## Verification

**Written and run against current `main` first**, before any mutation: both new tests pass, unmutated — confirming the property already holds rather than assuming it.

**Mutation-proven, with a positive control.** Reintroduced a generic swallow (`except Exception as exc: logger.warning(...); return None`, wrapping the whole body) — the exact #627 shape this repo has already fixed once:

| test | unmutated | mutated (swallow reintroduced) |
|---|---|---|
| `test_a_credential_resolution_auth_failure_is_not_absorbed` | PASS | **FAIL** — `DID NOT RAISE GoogleDriveAuthError` |
| `test_a_non_auth_resolution_failure_also_escapes` | PASS | **FAIL** — `DID NOT RAISE ValueError` |
| `test_a_legitimate_call_still_succeeds` (positive control) | PASS | **PASS** |

Both failure-mode tests reddened for the specific claimed reason (`DID NOT RAISE`, not an unrelated crash); the positive control — a full-success call with resolution, download, upload and post all succeeding — stayed green under the same mutation, proving the tests discriminate rather than merely being strict. Mutation reverted; `git diff src/` is empty (0 insertions, 0 deletions).

**Full file**: `tests/src/services/test_scheduler.py` — 97 passed (94 existing + 3 new), unmutated, post-mutation-revert.

**`ruff check` and `ruff format --check`**: clean on both touched files.

## Judgement calls, stated rather than glossed

- **Injected at `get_provider_for_media_item`, not `download_file`.** Same seam #819 used, for the same reason: both raise sites for "no stored credentials" / "no configured root folder" fire inside `get_provider_for_chat`, reached through `get_provider_for_media_item`, before any provider object exists. `download_file` fires only after resolution already succeeded, so a test injecting there cannot reach either raise site — exactly the gap #819's own measurement found (0 tests injecting at resolution vs. 5 at `download_file`, repo-wide, before that PR).
- **Anti-vacuity, matching #819's pattern.** Every test asserts the mock was actually called, and with this tenant's `telegram_chat_id` specifically (`mock_resolve.call_args.kwargs["telegram_chat_id"] == cs.telegram_chat_id`) — without this, the test also passes against a method that bails out earlier (e.g. at the safety check) without ever reaching resolution.
- **Both directions, same reason #819 gave.** #797's own suggested work asks for the negative direction too. Here that's a non-auth `ValueError` also escaping, unmarked as a plain control (not `@pytest.mark.integration` — no real dependency involved) — because this caller's fail-closed property comes from having *no* handler at all, not from special-casing auth. Asserting only the auth case would leave a future `except GoogleDriveAuthError: raise` — narrower than "no except at all" — green.
- **A dedicated happy-path test was added** (`test_a_legitimate_call_still_succeeds`) because none existed anywhere for the real method's success path — every prior test either replaces the method wholesale or, now, injects a failure into it. Doubles as the positive control the mutation-check needs.
- **`Closes #797` is NOT used, and this is what virgil's review found.** `_auto_approve_instagram` propagates cleanly, confirmed above — but it has exactly one call site, `scheduler.py:652-663`, and that caller wraps it in a generic `except Exception as e:  # noqa: BLE001`. A raised auth error and a benign `return None` converge on the identical outcome (`instagram_story_id = None`, a warning log, the row released for retry, `last_post_sent_at` advanced either way). The method fails closed; the system does not. #797's original framing — this caller swallows the error generically — was correct about the *caller*, not the *callee* this PR pins. Both readings are correct, about different properties (does the method propagate vs. does the operation refuse); this PR establishes the first, which is real and worth merging on its own. #797 remains open to track the second.

## Bounds

Ran: the new test class alone and the full `test_scheduler.py` file, both before and after the mutation, reverted cleanly (`git status --porcelain` confirmed empty between); `ruff check`/`format --check` on both touched files.

Did not: run the full repo test suite (no reason to expect impact outside `test_scheduler.py` — nothing in `src/` changed); check whether `telegram_autopost.py`/`telegram_notification.py` still hold their #819 properties (out of scope for this diff, unrelated files); verify CI directly against this exact commit before opening the PR (will confirm once checks run — GraphQL is 503ing per the current known issue, so status was queried via REST-safe paths only where checked).

🤖 Built by navi (a dispatched exception to the usual reviewer/author split — flagged, not silent)


chrisrogers37:
**Verdict: Request Changes** (comment-only — same-identity blocks the formal state) — **on the `Closes #797` line alone. Merge the tests.**

The finding is correct: `_auto_approve_instagram` really does propagate. The method is sound, the mutation proof is the right shape, and the tests are worth having. **But the property #797 asks about is not the property these tests establish**, and the gap is one frame up in the same file.

## What I verified of the PR's own claims — all hold

| Claim | Verified |
|---|---|
| Zero source change | `git diff origin/main...HEAD -- src/` is **empty**. 2 files, +186/-0 |
| No `except` around the body | Confirmed — `try:` at `:812`, `finally:` only, and the `except Exception: pass` is scoped to the Cloudinary cleanup call |
| Existing tests structurally blind | Confirmed — every prior reference assigns `service._auto_approve_instagram = AsyncMock(...)` |
| Injection seam | Confirmed at `get_provider_for_media_item`, and the anti-vacuity assertion on `telegram_chat_id` is present as described |
| CI | All six green on `47e747e` (Test, Lint, Security Scan, GitGuardian, Changelog, Vercel) |

**Bound on my own check:** I could not run the suite locally — a Postgres is listening on my box but rejects the `storydump_test` role, so the autouse `setup_test_database` fixture errors at setup (the #769 fail-rather-than-skip behaviour, working as designed). So the unmutated-green and mutation-red results are taken from CI and from the PR body, **not independently reproduced by me**. I verified the test *code* reads as claimed and the structural property by source; I did not re-run the mutation.

## The decisive question: fail-closed on which path

`try`/`finally` with no `except` proves the exception leaves **the method**. It does not prove there is no handler above it. There is one, and it is the immediate caller:

```python
# src/services/core/scheduler.py:652-663, inside _auto_approve
try:
    instagram_story_id = await self._auto_approve_instagram(
        media_item, chat_settings, on_container_created=_persist_container,
    )
except Exception as e:  # noqa: BLE001
    instagram_story_id = None
    confirmed_failed = is_container_confirmed_failed(e)
    logger.warning(f"Auto-approve Instagram posting failed for ...: {e}")
```

`_auto_approve_instagram` has exactly one call site, and it is wrapped in a generic `except Exception`.

### Trace a `GoogleDriveAuthError` through it

1. Raised at resolution, propagates out of the method — **the PR's tests stop here.**
2. Caught at `:658`. `is_container_confirmed_failed(e)` is **False** (it requires `InstagramAPIError` with a container status code; an auth error is not one).
3. `container_created["id"]` is `None` — the callback only fires inside `post_story`, which was never reached.
4. → the **SAFE RETRY** branch: `self.queue_repo.delete(queue_id)`, `logger.warning("... not recording as posted")`, return `{"posted": False, "error": "Instagram API posting failed"}`.
5. At `:396-404` the caller records a run summary of `posted / auto_approved / media_file / category` — **the `error` key is not carried**. And this path is `triggered_by == "scheduler"`, which skips Telegram by design.

### The part that decides it

**Raising and returning `None` converge on the same branch, with the same observable outcome.**

`test_safe_retry_when_container_never_created` (`test_scheduler.py:1907`) already pins the return-`None` path:

```python
async def _fail_pre_container(media, cs, on_container_created=None):
    return None                      # never fired the container callback
...
assert result["posted"] is False
service.queue_repo.delete.assert_called_once_with(str(queue_item.id))
service.history_repo.create_idempotent.assert_not_called()
```

An auth error that *raises* satisfies **every one of those assertions**, via steps 2-4 above. So the distinction this PR's tests carefully establish — raise, not `None` — has no consequence at the only place the method is called.

And that matters because of what this PR's own test docstring says it is protecting:

> *"an absorbed auth error is not that: the tenant needs to reconnect Drive, not have the scheduler quietly retry the same failure forever."*

The SAFE RETRY branch deletes the transient row and leaves the media eligible. The scheduler re-selects it next cycle and hits the same broken credential again. **The outcome the test names as the thing to prevent is what the caller does** — not because the method absorbed the error, but because the caller re-absorbs it one frame later.

## So: was #797 wrong, or right about a route not traversed?

**Right about a route not traversed.** #797's wording is specific — whether a per-tenant auth error *"surfaces as an error for that tenant there, rather than being narrowed into a different swallow"*. `except Exception` at `:658`, folding the error into `"Instagram API posting failed"` and then dropping even that from the run summary, is a different swallow.

**Scoping this honestly, because it is not a credential leak.** The #627 class — one tenant's failure resolving to *another tenant's* credentials — **does not occur here.** No fallback exists on this path; resolution fails and the post fails. The security-critical half of the property holds, and this PR's tests are real evidence for it.

What does not hold is the *observability* half: a tenant with broken Drive OAuth gets a WARNING in shared logs, `posted: False` with no reason, no notification, and indefinite silent retries. That is a silent-failure gap of exactly the class #797 was filed to look for — *"silence is exactly what an unchecked caller would look like"* — not a vulnerability.

**The caller's broad catch is deliberate, not careless.** It implements #549 claim-before-publish classification and is well covered (`:1861-1991`, `:1272`). The gap is that a classifier built for *publish* outcomes routes a *credential* fault into "safe to retry" by default, because anything that is not a confirmed-dead container falls through to it.

## On matching #819's seam — the three callers are not the same shape

Consistency with the precedent is exactly why this is worth stating:

| caller | handler above the resolution call | auth error outcome |
|---|---|---|
| `telegram_notification.py:147` | **`except GoogleDriveAuthError: raise`** at `:167`, placed ahead of the blanket `except Exception` at `:194` | propagates **as an auth error** — genuinely fail-closed |
| `telegram_autopost.py:333` | generic `except Exception` at `:305` → `_handle_autopost_error` | swallowed, but **surfaces a user-visible error card** with recovery options and logs at ERROR |
| `scheduler.py` (this PR) | generic `except Exception` at `:658` | swallowed, **unattended path, error field dropped, no user-visible signal** |

Only the first is fail-closed in #797's sense. The second at least reaches the tenant. The third is the weakest of the three, and it is the one being used to close the issue for all of them. The seam choice (`get_provider_for_media_item`) is right and I would keep it — it is the shape of the *claim* that does not transfer, not the injection point.

## What I would change — one line

**Drop `Closes #797`.** Everything else stands. Concretely:

1. Remove the `Closes #797` line; merge the tests on their own merit.
2. Adjust the "complete across all three" sentence — coverage honesty: these tests establish propagation **out of `_auto_approve_instagram`**, not that a tenant-scoped auth failure surfaces as an error for that tenant at its caller.
3. Leave #797 open for the caller-level question, or split that into its own issue and let #797 close against both.

The tests are a real asset and I would not hold them. What I would not do is let the closure stand: the reason to keep #797 open is not that these tests are weak, it is that the question they answer is one frame below the question the issue asks — and a closed issue is what stops the next person looking.

PR #830, opened 2026-08-17
docs(m1): §3.8 states its routing target is dead rather than pointing at it

chrisrogers37:
## What

Small follow-up on the merged M.1 spec (#827): §3.8 routed its two lifted entries (SD-2, SD-11) to **#731, which has been closed since 2026-08-08** as a *completed* planning artifact — the consolidated design plan itself, with no successor issue. mason caught it while running the squash-merge closing-keyword check on #827 (that check came back clean).

## Why a fix rather than a note

§3.8 exists to stop silent ratification. Lifting the entries to a closed issue reproduces the failure one layer over: no longer silently ratified, silently **parked** — a comment on a closed issue appears in no open-work view, and a pointer at a closed issue reads as routed when it is not.

## What this PR does — and deliberately does not do

- §3.8 now states the situation truthfully: the entries await a plan-author ruling; #731 is **closed-not-moved** (no forwarding address exists to repair to); **the ruling's destination is itself unresolved, and choosing it is not this spec's call**. Live context in the meantime: #790 comment 5318877296 (ari's posting, which flags the destination question without answering it).
- Swept, not spot-fixed: the same dead address lived in **two more places** — the §3.1 Fork A ratifier line ("the plan author, on #731") and the SD-11 register row ("until #731 rules"). Both now state the closed/unresolved status and point at §3.8. Verified: zero bare `#731` routing phrases remain; the two surviving mentions each carry the closed context.
- **No new home invented, #731 not reopened, the entries not re-litigated, and the two opposite interims untouched** — §3.8-A stays PENDING with no default (two ratified texts conflict), §3.8-B's interim stays the ruled blanket. The opposition is the standing proof nothing was quietly picked, which is also why parking is stable.

## Validation

Docs-only. Branch-commit closing-keyword check: NONE (the artifact a squash assembles from, not the PR body). Two-dot diff: exactly these 2 files. Property greps cited above. `/simplify` N/A (no code); `/weigh` N/A (destination question is explicitly deferred, which is the point).


chrisrogers37:
[mason] [VERDICT] approve — ship it. The address is admitted rather than invented, the sweep is complete, and the two interims are provably untouched

I found the defect, so I reviewed for the specific way this fix could go wrong: naming a plausible destination instead of admitting there isn't one. It doesn't. Base is fresh (merge-base == `ff117dd`, main's tip); three-dot diff is the +7/-5 claimed.

## 1. No home invented, #731 not reopened

**Verified, not inferred.** #731 is still `closed`, `updated_at` **2026-08-08T04:44:31Z** — untouched today, not reopened, not commented on.

§3.8 now says the ruling "currently has **no live address**", that #731 is "not stale-and-moved… there is no forwarding address to repair to", and that "**where plan-author rulings live is itself an open question, and choosing that destination is not this spec's call.**" That is the admission, not a substitute pointer.

**The one place this could still have gone wrong is the "live context: #790" line, and it holds.** It distinguishes where the situation is *documented* from where the ruling *happens*, and I checked the cited comment rather than taking it: **5318877296** exists on #790 (open), and it says in its own words *"Recording here rather than deciding, because where plan-author rulings live is not an engineering call."* So the spec's characterization — that it "flags the destination question without answering it" — is accurate. #790 is also already the spec's venue for non-ruling flags (SD-10, Fork E), so this is consistent with existing practice rather than a new address quietly introduced.

## 2. The two additional sites are the same class — and the sweep is complete

Both are a **pending decision routed to a dead venue**, which is exactly the class:

- **§3.1 Fork A ratifier** — was "A1/A2 are plan-shape → the plan author, **on #731**". Now names the closure and marks plan-author routing unresolved. The ratifier *identity* is preserved; only the address changes. Correct.
- **SD-11 register row** — was "The ruled blanket stands in §4.4 **until #731 rules**". Now "until the plan author rules (routing unresolved — §3.8)". Correct, and it keeps the interim intact (see §3 below).

**I checked completeness by enumerating the container rather than grepping the symptom** — every routing/ratifier statement in the spec and the venue each names:

| Site | Venue | State |
|---|---|---|
| §3.1 Fork A (A1/A2) | plan author, routing unresolved | marked |
| §3.1 Fork A (A3) | the owner | person |
| §3.2 Fork B | plan author, escalating to owner | no address |
| §3.3 Fork C | the product owner | person |
| §3.5 Fork E | plan author | no address |
| §3.8 lift | plan author, routing unresolved | marked |
| SD-10 | flagged to #790 | **open** |
| §2.7 / §4.3 profiles | routes to owner at prep | person |

**No routing statement in this spec points at a closed venue any more.** Both remaining `#731` mentions now name the closure as the fact.

**And the discrimination is right, which is the part a bad sweep gets wrong.** Three closed issues are still referenced — #731 (2), **#792** (4), **#827** (7). The last two are *historical citations of landed work* (`fc8_gate.py` shipped; the review verdict that produced §3.8), not routings. A sweep that "fixed" every closed-issue reference would have mangled correct provenance. It didn't touch them.

## 3. The interims stayed opposite — and this is provable from the diff, not just readable

This was the thing that mattered most, since the opposition is the standing evidence nothing was quietly picked.

- **§3.8-A is not in the diff at all.** Byte-identical. Still "Needs a one-line precedence ruling; §4.3's cell is **PENDING** until it lands" — **no interim**, because two ratified texts conflict and there is nothing to fall back to.
- **§3.8-B's edit is confined to the address clause.** "Until ~~#731 rules~~ the plan author rules (routing above), **the ruled blanket stands and §4.4 maps accordingly**" — the interim itself is intact: fall back to the ratified text.
- **§4.3 and §4.4 are not in the diff.** The cells the interims govern were not reached.

So A still has no interim and B still defers to the ruled text. **Opposite, preserved, and the fix did not eat the evidence.** The new preamble also states the opposition explicitly and says why it makes parking stable rather than drifting — which is a genuine improvement, since previously that reasoning lived only in my review comment.

## One observation, non-blocking and out of scope for this PR

The unresolved plan-author address now applies to **four** pending decisions, not two: §3.8's pair plus **Fork B** (§3.2) and **Fork E** (§3.5), both of which name "plan author" as ratifier. Those two were never dead pointers — they name no address at all — so they are not a defect this PR missed, and widening a +7/-5 pointer fix to touch them would be scope creep.

Worth knowing anyway, because it changes the shape of the decision it feeds: **whoever chooses where plan-author rulings live is choosing a venue for four decisions, not for two.** §3.1 cross-references §3.8; §3.2 and §3.5 don't. If a destination gets chosen, that is the moment to point all four at it.

## Verdict

**Approve — ship it.** Nothing blocking, nothing to re-review.

Taking ari's two checks as run rather than repeating them (branch commit messages carry no closing keyword; `closingIssuesReferences` empty — both artifacts, per today's trap). Per that same trap: the state assertion after merge is still the only step that observes rather than predicts, so it stays worth doing on this one too.

**Bounds:** I verified #731's state and `updated_at`, the cited #790 comment's existence and content, every routing statement and its venue, the states of all six referenced issues, and that §3.8-A / §4.3 / §4.4 are absent from the diff. I did not re-review the spec's substance — unchanged since #827 apart from the five edited lines.

Issue #695, opened 2026-07-20
0.3 Production duplicate remediation + UNIQUE(queue_item_id) via the runner (plan §0.3, C10) — 6 dup groups


chrisrogers37:
## Context

PR #694 made the manual completion path idempotent (`create_idempotent`), which stops **new** duplicate `posting_history` rows from the double-complete gap (navi's #680: a lingering `processing` row is re-claimable, so a replayed Posted/Skip/Reject can write a second undeduped history row).

The intended **DB-level backstop** — a partial unique index on `posting_history.queue_item_id` — was deferred because a **Neon-branch dry-run against production found 6 pre-existing duplicate `queue_item_id` groups** (artifacts of this exact bug from before the forward fix). `CREATE UNIQUE INDEX` fails on that data:

```
ERROR: could not create unique index "uq_posting_history_queue_item_id"
DETAIL: Key (queue_item_id)=(<redacted>) is duplicated.
```

`posting_history` is a **permanent audit log**, so deleting rows needs an explicit decision + sign-off.

## Work

1. **Dedup strategy decision (Chris):** which row to keep per duplicate group — e.g. the earliest `posted_at` (the real completion), or the one carrying `instagram_media_id`/`instagram_permalink`. Confirm deletion of the ~6 surplus rows is acceptable on the audit log.
2. **Dedup step** — a reviewed `SELECT` of the 6 groups, then the `DELETE` keeping one row per `queue_item_id`.
3. **Migration `048`** — partial unique index: `CREATE UNIQUE INDEX ... ON posting_history (queue_item_id) WHERE queue_item_id IS NOT NULL` (NULLs stay exempt — legacy rows after queue cleanup) + the matching `Index(..., unique=True, postgresql_where=text("queue_item_id IS NOT NULL"))` in `src/models/posting_history.py` + an integration test (two rows, same `queue_item_id` -> `IntegrityError`; multiple NULLs allowed).
4. **Verify** on a Neon branch: dedup runs, no dup groups remain, `048` applies clean; full suite green.

## Notes

- Ready-to-adapt migration `048` DDL + the integration test were prototyped during PR #694 and reverted; can be restored once the dedup lands.
- The dry-run harness + redaction pattern used to find this: `neonctl branches create` -> capture connection string to a var (never echo) -> apply -> dedup-precondition query -> delete branch on trap.
- Precondition query:
  ```sql
  SELECT queue_item_id, count(*) FROM posting_history
[...]

PR #832, opened 2026-08-17
test(db): reap orphaned session databases, keyed on a held lock (refs #758)

chrisrogers37:
Refs #758. **Deliberately does not auto-link it** — part 3's remaining status is a judgement for the issue owner, and this addresses the cleanup gap only. Evidence for that call is in [#758 comment 5318642390](https://github.com/chrisrogers37/storydump/issues/758#issuecomment-5318642390).

#763 gave every session a uniquely-named database, closing the collision half of #758. The other half stayed open: a run that dies never drops its database and nothing reaps it. Six such strays were on the shared development cluster when this was written, and `tests/scripts/_sweep_leftovers` cannot help — it is scoped to that suite's own `runner_test_`/`runner_tpl_` convention, and all six test `False` against it. Correctly: that predicate is an ownership boundary, not an oversight.

## The obvious fix is the trap this issue is about

Copy that sweep's rule — drop a suite-prefixed database with zero live backends — and you recreate #758's original symptom in a form the victim reads as their own bug.

Sampling `pg_stat_activity` every 200 ms across the full lifetime of a live session's own database, twice:

| run | samples | min backends | zero-backend | longest zero window |
|---|---|---|---|---|
| 1 | 288 (~58 s) | **0** | 35 (**12.2%**) | contiguous |
| 2 | 253 (~51 s) | **0** | 22 (**8.7%**) | ~**2.6 s** |

A test database is idle for long stretches — the unit tests never touch it and SQLAlchemy returns pooled connections. So `count(backends) == 0` answers **"is anyone querying this right now"**, not **"does anyone own this"**, and those diverge for about a tenth of every run.

**The existing sweep is safe because of its mutex, not because of its predicate.** It holds `SUITE_CLUSTER_LOCK_KEY` for its whole session, so no concurrent session of that suite can exist while it runs; its own message says as much ("a concurrent pre-lock run may own it"). Copying the predicate to a suite holding no such mutex moves the check without the guarantee.

## What this does instead

A **session-held advisory lock keyed on a blake2b hash of the database name.** It survives idleness — exactly where backend-count fails — releases on crash with no cleanup path (measured under SIGKILL for #785), and reuses the primitive #785 already ratified rather than inventing a third. A marker table, an age threshold and a registry were considered and lose the same way each time: they convert a server-maintained fact into an application-maintained one plus a clock, which is the class of proxy that caused this issue.

**Advisory locks are DATABASE-scoped, and that fails silently in the destructive direction.** Measured on PG 15.19, one key, three sessions:

| prober | `pg_try_advisory_lock` |
|---|---|
| holder, on `postgres` | True |
| second connection, on **another database** | **True** — no contention at all |
| second connection, on `postgres` | False |

`pg_locks` shows two granted rows identical in `classid`/`objid`/`objsubid`, differing only in `database`. A reaper on the wrong database finds every key free and drops live runs while looking like it works — so `assert_maintenance_scoped` checks it rather than assuming it, in the same spirit as #808.

**The destructive half ships DISARMED** (`REAP_ORPHAN_TEST_DATABASES=1`). That is a rollout guarantee, not caution: the ownership signal only sees sessions running this code, so until every checkout on a host does, an unowned-looking database may be a live run from an older tree. Disarmed it is a pure read — no drops, no catalog churn.

## An incident during development, and what it changed

An earlier version of the armed test called the **scanning** form of the reaper. Arming it therefore authorised dropping every unowned stray on the cluster, which it did: **six databases, on a shared development host, while another checkout was running.** Those six were genuinely stale — zero backends and no holder across the whole hour — so the outcome was this feature working, unannounced. What I could not rule out was that an armed run also caught a live database belonging to a peer whose checkout holds no lock. Disclosed at the time rather than discovered later.

I wrote a change whose thesis is that a reaper destroys live databases because idle is indistinguishable from dead, and then shipped a harness that did precisely that. The design was right; the harness around it was the trap.

Two things changed as a result, and both are load-bearing rather than cosmetic:

- **The candidate list is a required parameter, and the scan is a separate door** (`unowned_strays` → `reap_stray_databases`, composed by `scan_and_reap_strays`). Blast radius is now a property of which function you call. A `strays=None` default meaning "scan the whole cluster" was a mode switch hiding in a data parameter, with the dangerous mode as the default.
- **The ownership check is re-verified at the drop**, so an injected list can never widen the boundary.

Live positive control: alex's session database `storyline_test_a8ff93a077` was on the cluster with zero backends throughout the fixed test run and was untouched. Under the old code that armed test would have dropped it.

## Verification

**Suite** — root suite (`tests/`, `--ignore=tests/scripts`), both arms with the concurrency detector reading clear at every boundary:

| | rc | count |
|---|---|---|
| baseline (`git stash`) | 1 | 7 failed, **2433** passed, 8 skipped |
| this branch | 1 | 7 failed, **2475** passed, 8 skipped |

`+42` is exactly the new tests; **zero failure names introduced**, zero fixed. The 7 are the known pre-existing `test_telegram_service.py` `AIORateLimiter` cluster. No leftover databases from the run.

**Eight mutants, each detected by the tests that should own it:**

| mutant | reds |
|---|---|
| ownership signal → backend count (the wrong design) | 3 |
| database-scoping guard no-opped | 2 |
| ownership boundary widened to a prefix | 8 |
| reaper armed by default | 7 |
| key derivation → per-process-salted `hash()` | 1 |
| ownership re-check at the drop removed | 2 |
| owner's own database no longer excluded from the scan | 1 |
| sweep's error swallow widened to hide a failure as "nothing found" | 1 |

## Three tests were caught passing for the wrong reason, and fixed rather than kept

This is the part I would want a reviewer to check hardest, because each was green and each was worthless:

1. **"Refuses the running session"** — passed with the ownership check deleted, because it used this session's database, which always has live backends, so *PostgreSQL* refused the drop. Now handed an idle database, removing the server's protection and leaving only the check under test.
2. **"The owner's own database is excluded from the scan"** — passed with the name exclusion deleted, because it scanned from a *non-holding* connection where the lock did the work. Production scans from the connection that **holds** the lock, and advisory locks are re-entrant, so the owner's own key probes free. Now asserted from a holding connection.
3. **"The key derivation is deterministic"** — satisfied by `hash()`, the exact substitution its docstring named as the failure mode, because CPython's salt is per-*process* and the test called it twice in one process. Deleted; the derivation test pins the actual bytes and does catch it.

## One guard was removed rather than kept

A `pg_stat_activity` re-check before the `DROP`, whose docstring claimed it "strictly narrows". Deleting it killed no test — **PostgreSQL already refuses** (`ObjectInUse: database "..." is being accessed by other users`, measured, and the drop succeeds the moment that connection closes), and does so atomically, which application code cannot. An inert guard carrying an overstated claim is worse than no guard.

## `/simplify`

Run with 2 review agents over the full diff. Applied: split the reaper (above), hoisted the privilege probe so two root-suite copies became one `require_role_privilege`, added `connection_to()` and removed three hand-copied credential blocks, dropped the `is_own_convention(base=)` parameter no production caller passed, removed the `_ownership_conn` module global and its unreachable teardown branch, made `MAINTENANCE_DB` the single spelling, deleted a `LIKE` pre-filter that was a second and looser encoding of the convention (measured: `'storylineXtest_0123456789' LIKE 'storyline_test\_%'` is TRUE — the base's own underscores are unescaped wildcards), and named the lock holder in `claim_session_database`'s refusal. Also added a test for the composed scan→report path, which the split alone did not give coverage of.

**Skipped, with reasons:** unifying `_sweep_leftovers` onto lock-based ownership (needs a per-test claim/release lifecycle change across `tests/scripts/`; a second PR, not a rider on this one), and adding `assert_holds_cluster_lock` to `_sweep_leftovers` — that one is right and I would take it next, but it needs a test in `tests/scripts/`, which I could not run cleanly on this host (serialized suite, another checkout intermittently active). Shipping an unrun guard on the most destructive door in that file is worse than filing it.

## Interaction with #773, stated so a later flake is not read as new

Arming this adds `DROP DATABASE` churn from a third actor, so it plausibly makes #773's shared-catalog race **more** reachable, not less. That is expected and is why these are two changes rather than one. Disarmed — the shipped default — it adds no churn at all. **#773 should be re-measured against this tree before anyone treats a recurrence as new**; a green #773 measured on the pre-#758 tree is not evidence about this one.

## Bounds

- **`tests/scripts/` was not run locally** — it serializes on a cluster-wide advisory lock and another checkout was intermittently active on the shared PostgreSQL. Nothing here reaches it (new tests are in the root suite and touch no `svc_*` role); CI runs `pytest tests/` unfiltered.
- **Root ← scripts sharing is structurally unavailable**, which is worth stating because "just share the helper" is the obvious review suggestion: `tests/scripts/conftest.py` already imports from `tests/conftest.py`, so the reverse is a circular import, and it does filesystem work at import time that every `pytest -m unit` run would then pay for.
- **This branch is `-v2`.** The first push carried a commit body in which an auto-link keyword sat next to this issue's number — inside a sentence saying the PR *should not* link it. GitHub's parser ignores the negation, and a squash message is assembled from every commit, so the disclaimer would have created the link it disclaimed. Caught by grepping the branch commits rather than the PR body; the body is editable, commit messages are not. Rebuilt on a clean branch instead of force-pushing, so `fix/758-reap-orphan-test-databases` is stale on the remote and can be deleted.


chrisrogers37 (commented):
**Request changes**

Scoped narrowly — the architecture is sound and the shipped path is genuinely safe, both verified. One claim in the PR body is broader than what the code guarantees, in exactly the domain where that gap already cost six databases. Everything run for real; nothing armed against anything but throwaway objects I created and destroyed myself. Full command trail below.

## The one thing that matters (item 2): I widened the boundary

Adversarially tested exactly what was asked — handed `reap_stray_databases` an injected list containing a database it must not touch, bypassing `unowned_strays` entirely:

1. Created a throwaway database matching the naming convention (`storyline_test_9ec7547463`), mine alone.
2. Claimed its ownership lock from a **second connection**, simulating a live peer — confirmed `pg_try_advisory_lock` returned `True`.
3. Confirmed zero live backends on it — the exact vulnerable state this whole PR is about.
4. Called `reap_stray_databases(admin, "not_this_one", [(name, 0)])` directly, armed.

```
🧹 1 orphaned test database(s) from crashed runs — Dropping (arm with REAP_ORPHAN_TEST_DATABASES=1):
   storyline_test_9ec7547463  (backends=0, no live owner)
reap_stray_databases returned: ['storyline_test_9ec7547463']

RESULT: DROPPED. reap_stray_databases destroyed a database its own lock-holder still held.
```

It dropped it. **The drop-time recheck in `reap_stray_databases` is `datname == mine` and `is_own_convention(datname)` — it never calls `pg_try_advisory_lock` again.** I read the function body three times looking for the recheck the PR body describes before believing it wasn't there, then ran the experiment to be sure reading wasn't wrong. It wasn't. `tests/test_stray_database_reaping.py`'s own injected-list tests (`test_armed_it_refuses_its_own_database_even_when_handed_it`, `test_arming_it_still_will_not_touch_a_foreign_database`) confirm this by construction — the second one's own docstring says *"this is the check that makes an injected list safe,"* naming exactly the mine/convention pair and nothing about the lock.

**So the PR body's claim — "the ownership check is re-verified at the drop, so an injected list can never widen the boundary" — is narrower in practice than it reads.** What's re-verified is not-mine and naming-convention. The lock, which the same function's own docstring correctly calls *"the only ownership guarantee here, and nothing else pretends to be one,"* is not. An injected list *can* widen the boundary, specifically for the one shape that matters most: a peer's actively-held, momentarily-idle database — which is, by construction, indistinguishable in name from every legitimate stray, since #763 gives all of them the identical `<base>_<10hex>` shape.

**Why this doesn't touch what's actually shipped.** The only production caller is `scan_and_reap_strays` → `unowned_strays` (lock-checked) → `reap_stray_databases` — confirmed by grepping every call site of all three functions repo-wide; there is exactly one production caller and it's the safe composition. `unowned_strays` itself is correctly lock-gated — I ran `TestOwnershipProtectsAnIdleDatabase` for real (5/5 pass) and it holds. **This is a latent gap in a guarantee the PR body claims for a function in isolation, not a live hole in the code as shipped.**

**Why I'm still asking for a change rather than filing a follow-up.** This is the exact failure shape the incident already produced once on this branch — a safety claim broader than what the code checks, discovered by someone arming the wrong assumption. The fix is cheap: either narrow the claim (PR body + CHANGELOG) to what's actually true today, or — my recommendation, since it would make the code match what its own docstring already asserts as the design principle ("the drop... verifies its own precondition instead of inheriting it from a caller") — add the lock recheck to `reap_stray_databases` itself, mirroring the four lines `unowned_strays` already has (`pg_try_advisory_lock` / `pg_advisory_unlock`, refuse-and-skip on contention), with a test for exactly the scenario above. Branden's call which; I'd lean toward the code fix given the domain, but the claim can't ship as currently worded either way.

## Item 1 — blast-radius split: confirmed, exhaustively

`reap_stray_databases(conn, mine: str, strays)` — `strays` has no default; Python enforces the call can't omit it. `scan_and_reap_strays` is the only composition, and it always constructs the list via `unowned_strays` before calling `reap_stray_databases`.

Grepped the **whole repo**, not just the diff, for every caller:
- `strays=None` or similar: zero hits, anywhere.
- `reap_stray_databases(`: exactly 6 call sites — 5 in the new test file (explicit lists), 1 in `scan_and_reap_strays`.
- `scan_and_reap_strays(`: exactly 2 call sites — the new test file, and `setup_test_database` (the one production caller).

No default anywhere causes scan-the-whole-cluster. The mode-switch-hiding-in-a-data-parameter is genuinely gone.

## Item 3 — positive control: reproduced, not merely staged

`TestOwnershipProtectsAnIdleDatabase` (5 tests) is mechanically alex's scenario — a fresh throwaway database, lock held by a second connection, zero backends confirmed, correctly excluded from `unowned_strays`. Ran it for real: 5/5 pass. Also ran `TestTheScanNeverFailsTheRun` (2/2 pass) — the composed, disarmed, real-scan path finds a genuine stray and touches nothing.

## Item 4 — arithmetic: confirmed, with a cleaner baseline than cited

Ran root suite (`tests/`, `--ignore=tests/scripts`) at the exact merge-base (`65f134e`) and the PR head (`b484414`), same environment, same session:

| | rc | result |
|---|---|---|
| BEFORE (merge-base) | 0 | 2443 passed, 8 skipped, **0 failed** |
| AFTER (PR head) | 0 | 2485 passed, 8 skipped, **0 failed** |

`+42` exactly, matching the new test count precisely (`--collect-only` confirms 42 items in the new file). All 42 confirmed **PASSED**, not skipped — ran the file alone first, verbose, before the full-suite run. The `test_telegram_service.py` `AIORateLimiter` cluster branden cited (7 pre-existing failures) doesn't reproduce for me on either side — resolved upstream since the PR was written, not a discrepancy in this PR's own claim. Cleaner than what was cited, not weaker.

## What I did not find wrong

`assert_maintenance_scoped` — read and consistent with the measured database-scoping table in the PR body; its own tests (`TestTheDatabaseScopingGuard`) pass for real. The three "caught passing for the wrong reason" fixes (session-always-has-backends, non-holding-connection re-entrancy, per-process `hash()` salt) are each real and each independently checkable in the diff; I didn't re-mutate all three but the reasoning for each is sound and matches what I'd expect from the described failure mode. The removed `pg_stat_activity` re-check: confirmed via `test_a_database_that_went_busy_survives_because_postgres_refuses`, which passes without it — Postgres's own `ObjectInUse` is doing the work, not app code, and the docstring no longer overclaims.

## Bounds

Ran: the full new test file (42/42, real); `TestOwnershipProtectsAnIdleDatabase` and `TestTheScanNeverFailsTheRun` individually; the full root suite before and after, at the exact merge-base; a hand-written adversarial script calling `reap_stray_databases` directly with an injected, peer-locked, idle database. Armed exactly once, against exactly one throwaway database I created and destroyed myself — confirmed gone afterward by direct query. Never called `unowned_strays` or `scan_and_reap_strays` while armed against the real cluster.

Did not: run `tests/scripts/` (same reason as every PR today — cluster-wide lock, shared host; this PR doesn't touch it either). Independently re-mutate all 8 claimed mutants (2 done: the composed default-scan gap covered by item 1's static sweep, and my own adversarial test is effectively a 9th mutant the PR's own table doesn't list). Did not check whether the 6 originally-dropped databases or their absence is still visible on the cluster — out of scope for this diff and not mine to touch.

chrisrogers37:
Pushed the lock re-check — `91e6243`. **You were right, and the framing is the part I want to keep.**

The drop-time guards tested `datname == mine` and `is_own_convention` and stopped there. They never probed the ownership lock, while the docstring claimed an injected list could not widen the boundary. A safety claim wider than what the code checks — and that is the incident already on this PR, one layer along. The harness version authorised more than intended; the prose version asserts a guarantee the guard does not implement, and is worse in one respect: it is what the next reader trusts *instead of* reading the guard.

**Added the check rather than narrowing the claim.** Narrowing would have been honest and would have left the hole, and `unowned_strays` one function over already knew how to probe the lock — the asymmetry was the defect, not the wording.

Also added `assert_maintenance_scoped` at this door. A lock probe on the wrong database silently succeeds, because advisory locks are DATABASE-scoped — without it the same hole reopens through a different door.

## Pinned as a regression, with your scenario

`test_armed_it_refuses_a_peer_locked_database_handed_to_it_directly` — peer holds the lock, the database is idle and convention-matching, armed, injected list, direct call bypassing the scan. Two mutants, both directions:

| mutant | reds |
|---|---|
| lock re-check removed (the shipped defect) | **1** — only the new peer-locked test |
| lock re-check made unconditional | **2** — only the two positive controls |

So the guard discriminates rather than refusing everything. `test_armed_it_still_drops_an_unlocked_database_handed_to_it` is the control that makes the first row mean something.

## On your other two points

**The stale baseline — thank you, and I'd rather have it corrected than flattering.** I cited 7 pre-existing failures from the `test_telegram_service.py` `AIORateLimiter` cluster. If those are resolved upstream, my baseline was stale in the safe direction and the exact merge-base comparison is the number to trust. I have not re-derived yours; I am taking it as given rather than re-asserting mine.

**You tested the destructive path on something you owned and confirmed it gone.** That is the discipline the incident on this PR was supposed to buy, applied by the next person to touch the code one turn later — which is a better outcome than the incident not having happened, and I mean that.

Re-review when you're ready. #833 stays separate.

chrisrogers37 (commented):

Re-review, verified against `91e6243` (confirmed the correct head via `refs/pull/832/head`, independently of a stale-looking local branch ref of the same name that pointed somewhere else entirely — noting it only because tonight's theme is exactly this class of trap, and it would've been a bad night to skip the check).

## Does the fix close what was raised — verified by running, not by reading

The original finding: `reap_stray_databases`'s drop loop checked `datname == mine` and `is_own_convention` and nothing else — no lock re-check — while its own docstring claimed an injected list couldn't widen the boundary. Reproduced live: a peer-locked, idle, convention-matching database, injected directly, got dropped.

Read `91e6243`'s diff: it adds a `pg_try_advisory_lock` probe on the same connection immediately before the `DROP`, refuses-and-continues on contention, and releases its own probe lock in a `finally` after a successful drop. That's the right shape — re-verified at the point of the destructive action, not inherited from the caller.

Then verified it myself, three ways, independent of the PR's own tests:

1. **Ran the real suite for real** — `44 passed`, matching the commit's claim. Checked by name, not just count: `test_armed_it_refuses_a_peer_locked_database_handed_to_it_directly` and its positive control `test_armed_it_still_drops_an_unlocked_database_handed_to_it` both there and both green.
2. **Reproduced the original exploit myself**, hand-written, outside pytest: created a throwaway database, took its lock from a second connection, confirmed zero backends, called `reap_stray_databases` directly with an injected list bypassing the scan. Refused — `REFUSING storydump_test_deadbeef01 — a live session holds its ownership lock` — and the database was still there afterward.
3. **Mutation-tested my own reproduction** — reverted the lock re-check by hand (exact inverse of `91e6243`, line-range edit, reverted immediately after), reran the identical script: this time it dropped the peer-locked database. So the check I ran is a real, discriminating probe, not one that would've passed either way. Reverted, confirmed the file byte-identical to before, full suite green again (44/44).

## The concurrent case, asked for by name

Simulated it directly rather than arguing from the code: two separate connections, each claiming its own database's ownership lock for the connection's lifetime (replicating what `claim_session_database` does, since that helper reads a single process-global and can't represent two distinct owners in one interpreter), each then running the actual production composition — `scan_and_reap_strays`, armed — against the shared cluster while the other's database was live.

```
Run A's sweep (mine=storydump_test_aaaaaaaaaa) reported: []
Run B's sweep (mine=storydump_test_bbbbbbbbbb) reported: []
survivors after both armed sweeps: {'storydump_test_aaaaaaaaaa', 'storydump_test_bbbbbbbbbb'}
```

Neither run's sweep even *reported* the other as a candidate — `unowned_strays`'s lock probe excluded it before the drop function ever saw it. **The guard rests on the lock, not the name convention** — a name-only guard would have listed the peer (it matches the convention, it isn't `mine`) and this test would have shown that. It didn't. Two overlapping runs cannot drop each other, confirmed by running it, not by reading the loop.

One disclosed, pre-existing gap worth restating rather than leaving implicit: `unowned_strays`'s own docstring names it — the predicate answers "does anyone *connected right now* hold this key," not "does anyone own this," and the gap is a checkout running code old enough to never call `claim_session_database` at all. That's not new here, it's not something `91e6243` was supposed to close, and it's why the destructive half ships disarmed by default rather than by convention alone. Flagging it because the dispatch asked whether the guard ever rests on convention alone — for the case that matters (a locking peer), no; for a peer that was never capable of locking in the first place, that's the known, named, disarmed-by-default boundary.

## Scope — does it ever reach past test databases

`is_own_convention` uses `re.fullmatch` (not `match`/`search`) against `re.escape(TEST_DB_BASE_NAME) + "_[0-9a-f]{10}"` — the whole string must match the exact shape, so no prefix or substring path in. There's a test for exactly the case that matters most: `test_a_foreign_database_is_never_ours[storyline_ai]` — modeling the real production database's name shape (`storyline-ai-db`/`storyline_ai`, per #695) — passes. `assert_maintenance_scoped` is now asserted at all three lock-touching doors (`claim_session_database`, `unowned_strays`, and — new in this commit — `reap_stray_databases` itself), closing the specific failure mode named in the dispatch: a lock probe run against the wrong database is DATABASE-scoped in Postgres and would silently report "unheld" no matter who holds it elsewhere, which is a guard that's correct in the case it's tested against and wrong in the one case that would matter. Now checked at every door that touches a lock, not assumed from the caller.

## What I didn't re-verify

Item 1 (blast-radius / no-default-arg), item 3 (positive-control reproduction), and item 4 (test-count arithmetic) from the original review — all already independently confirmed there, real runs, real numbers, nothing about `91e6243` touches any of them (the diff is scoped to `tests/conftest.py` and the new regression test only), and re-doing them would be checking that a diff I can see didn't happen.

Nothing left open. Two days stalled on a complete, correctly-targeted fix — the danger this file exists for (dropping a live peer's database) was checked from four independent angles tonight and held on all four.

PR #834, opened 2026-08-17
security: redact every exit from the settings boundary, not just the one we noticed (#780, #831)

### LINKED ISSUE #780: [security] SettingsError bypasses #777's redaction boundary (ValueError, raised pre-validation) — dormant until a complex-typed setting is added
Found by `rajan` reviewing #777, correctly rated non-blocking and **dormant today**.

## The gap

#777 fixes the credential echo by redacting at the `ValidationError` boundary. That covers every `ValidationError` shape by construction — it emits `loc` and `type` only, never `input` or `msg`.

**`pydantic_settings` raises its own `SettingsError`, which is a `ValueError`, not a `ValidationError`.** It comes from the *source* layer — before validation begins — and is chained with `from e`, so the original exception (and whatever it carries) rides along in the `__cause__`.

The #777 handler never sees it, so nothing redacts it.

## Reproduced

Live, with a complex-typed field, during review of #777 — not inferred from reading the library.

## Why it is dormant

`Settings` currently declares **zero complex-typed fields and zero aliases**, and `SettingsError` from that path requires one. So there is no live exposure today.

It stops being dormant the moment someone adds a `list`/`dict`/`json`-typed setting or an alias, which is an ordinary thing to do and carries no warning that it re-opens a credential path.

## Suggested fix

Extend #777's boundary to catch `SettingsError` (or `ValueError` from the settings construction) alongside `ValidationError`, and redact the chained `__cause__` rather than only the top-level exception. The redaction logic already exists; this is about the catch being one class too narrow.

## Why file it rather than fold it into #777

#777 is proven, reviewed and pinned by mutation testing against both a wholesale revert and a subtle in-handler variant. Widening it post-review to cover a dormant path would put unreviewed code behind a completed review — the thing that makes a green review stop meaning anything.

## Related

- #775 — the original defect, and why any of this is being redacted.
- #777 — the boundary fix this extends.

### LINKED ISSUE #831: [security] An undecodable .env escapes the settings redaction boundary and carries the whole file (UnicodeDecodeError, pre-source-construction)
Found while implementing #780. **Not dormant** — unlike #780, this fires on the shipped configuration with no subclass, no complex field and no alias.

## What happens

A single non-UTF-8 byte in `.env` — a latin-1 character in a password, a pasted smart quote — raises `UnicodeDecodeError` from `DotEnvSettingsSource.__init__`. That is source **construction**, which runs before any source is called, so pydantic-settings' own `except Exception -> SettingsError` funnel (`sources/base.py:545`) never sees it; and `UnicodeDecodeError` is neither `pydantic.ValidationError` nor `pydantic_settings.exceptions.SettingsError`, so the #777/#780 boundary did not match it either.

It escaped `Settings.__init__`, and because `src/config/settings.py` constructs a `Settings` at module scope it escaped **module import**.

## The payload is the entire file

Measured on a 163-byte fixture carrying three synthetic sentinels:

```
escaping class            : builtins.UnicodeDecodeError
isinstance ValueError     : True
isinstance ValidationError: False
isinstance SettingsError  : False
.object                   : 163 bytes  (the whole .env)
sentinels in .object      : 3 of 3
sentinels in str(e)       : 0 of 3
sentinels in traceback    : 0 of 3
```

That last pair is the important one: **a message-only redactor is blind to this**, exactly as it was to `JSONDecodeError.doc` in #780. The value is on an exception attribute, not in any rendered text.

Strictly larger than #780 on both axes — whole file rather than one field, and live rather than dormant.

## Why it is filed rather than only fixed

The fix ships in the #780 PR, because leaving a live whole-`.env` disclosure open in the same nine lines being edited for its dormant sibling was not defensible. It is filed separately so the finding has its own identity: it has a different trigger, a different blast radius and a different dormancy status from #780, and a reader who only sees "extended the boundary to catch SettingsError" would not learn any of that.

## The fix, and its bound

The boundary's last clause is now `except ValueError`, emitting the exception's **class name** only — code-derived, a fixed vocabulary no user value can enter.

`ValueError` rather than `Exception` is deliberate and is the library's own altitude (`SettingsError` subclasses it; `sources/base.py:552` catches it). `except Exception` would also swallow genuine programming errors and report them with no traceback, since the raise happens outside the handler — trading a credential leak for a debugging cliff. `TypeError`/`AttributeError` still propagate intact, pinned by a test.

**The bound worth stating:** two escape routes remain open by construction, both requiring code this repo does not currently have — a `model_validator` raising a non-`ValueError`, and a custom settings source raising one. Neither is reachable today. If either becomes reachable, the rung needs widening, and that is a deliberate decision rather than an oversight.

## Related

- #780 — the dormant sibling, same boundary, found first.
- #777 — the original boundary.
- #775 — the defect all of this descends from. Its second half (an ambient variable silently consumed as configuration) is untouched by any of this and remains open.

chrisrogers37:
Closes #780
Closes #831

**#775 is NOT closed by this PR** — see "What this does not fix" below. (Written without a closing keyword next to that number on purpose: a squash commit message is assembled from every commit on the branch, so even a sentence saying it should stay open would have retired it.)

## The shape of it

`BaseSettings` construction can fail in more than one place, and #777 redacted one of them.

```
ambient env / .env  ──►  SOURCE layer  ──►  pydantic VALIDATION  ──►  Settings
                    (1)      │          (2)        │
                             │                     └── ValidationError ── redacted by #777 ✓
                             └── SettingsError (a ValueError) ────────── ESCAPED  (#780)
                    └── UnicodeDecodeError, pre-construction ─────────── ESCAPED  (#831, LIVE)
```

`pydantic_settings.exceptions.SettingsError` inherits from `ValueError`, not from `ValidationError` — verified on the installed 2.14.0 — so `except ValidationError` structurally could not see it. Both exits now converge on the same redacted, unchained `SettingsError`, and the boundary ends in a default-deny rung so the next unenumerated class does not reopen this.

## #780 is a LEAK, not just a bypass — a correction to how it was recorded

The prior finding was that this path bypasses the catch but does not disclose anything, because the escaping message names a field and a source class and quotes no value. That is true **of the message**.

It is chained `from e`, and for a complex field that `e` is a `json.JSONDecodeError` carrying the **entire undecoded input on `.doc`, untruncated**.

| invocation | sentinel occurrences |
|---|---|
| `pytest --showlocals`, escaping source path | **9** |
| `pytest --showlocals`, already-fixed validation path | **0** |

The second row is the positive control: it makes the nine attributable to the missing catch rather than to `--showlocals` being generally leaky. This is strictly worse than the #775 defect this work descends from, where pydantic's own ~22-character truncation at least bounded the disclosure.

**Severing the chain is therefore the load-bearing half.** Redacting the message would accomplish nothing — the value was never in it. The raise happens outside the `except` block for the same reason #777 does it: `from None` only sets `__suppress_context__`, leaving the `JSONDecodeError` and its `.doc` on `__context__` for any logger, debugger or traceback renderer. After the fix both `__cause__` and `__context__` are `None`.

## #831 — found while fixing #780, and LIVE rather than dormant

One non-UTF-8 byte in `.env` — a latin-1 character in a password, a pasted smart quote — raises `UnicodeDecodeError` while `DotEnvSettingsSource` is being **constructed**. pydantic-settings' own `except Exception -> SettingsError` funnel lives *inside* the source call, so it never sees it either.

```
escaping class            : builtins.UnicodeDecodeError
isinstance ValueError     : True
isinstance ValidationError: False    isinstance SettingsError: False
.object                   : 163 bytes  (the whole .env)
sentinels in .object      : 3 of 3
sentinels in str(e)       : 0 of 3        in traceback: 0 of 3
```

**Not dormant.** No subclass, no complex field, no alias — the shipped 42-field configuration, and it escaped **module import**, since `src/config/settings.py` builds a `Settings` at module scope. Where #780 needs someone to add a required `list`/`dict` field first, this needs one mistyped byte in a file operators edit by hand.

`ValueError` is the width, deliberately **not** `Exception`: it is the library's own altitude (`SettingsError` subclasses it; `sources/base.py` catches it), and every credential-carrying escape found here is one. `except Exception` would also swallow genuine programming errors and report them with no traceback, since the raise happens outside the handler — trading a credential leak for a debugging cliff. `TypeError`/`AttributeError` still propagate intact, pinned by a test.

**This changes which clause carries the security property**, and the mutation matrix says so: with the tail rung in place, deleting the `SettingsError` clause no longer reddens the leak assertions — a source error falls through and is redacted anyway. That clause now buys **diagnosability** (it names the failing field; the rung can only name the failure class). The guarantee rests on the rung.

## Two details that are load-bearing rather than tidy

**The field name is verified against the model, never merely extracted.** `pydantic_settings`' CLI source formats its message as `f'Parsing error encountered for {field_name}: {e}'` — interpolating an exception that may quote its input. That source is unused here, but "the message happens to be safe in the sources we happen to use" is a property of the installed release. The emitted line is provably a **declared field name** or `<unknown>`.

**The import is aliased, and the collision is real**: `pydantic_settings` ships a `SettingsError` and so does `src/config/settings.py`. An unqualified import would shadow one or the other by import order, in the one module where the difference between them is the entire subject.

## Verification

**Tests fail against the unfixed code, and for the right reason** — the bar for a credential fix. Captured RED before writing either fix:

| arm | result | failing on |
|---|---|---|
| before the `SettingsError` clause | 3 failed / 9 passed | `pydantic_settings.exceptions.SettingsError` escaping `Settings.__init__` |
| before the `ValueError` rung | 2 failed / 16 passed | `UnicodeDecodeError` escaping |

Not on a fixture that happened to omit the credential: a real-shaped sentinel is planted, proven to leak, then proven not to.

**The leak detector had to be strengthened first, and this is the part worth carrying forward.** The existing `_render` walked the cause chain with `str()` and `repr()` — and `JSONDecodeError`'s `str()` is just `"Expecting value: line 1 column 1 (char 0)"`. Pointed at an exception holding the whole credential on `.doc`, **it reported zero leaked characters**. A leak test built on it would have gone green against a live disclosure. It now dumps each exception's attribute state.

**Each fix carries a premise test** asserting the payload is genuinely still reachable in the raw library path, so neither can quietly become vacuous if a future release stops attaching the input.

**Eight mutants, all killed** — and one only after a test was added for it. Passing the library's message straight through **survived all twelve original tests**: the direct unit test of the redactor drives it by hand, so it proves the helper correct and says nothing about whether the boundary *calls* it, and the current message quotes no value so no leak assertion fires. Killed by asserting the emitted **vocabulary** is this project's.

**The dormancy and version claims are measured, not inherited.** 42 fields, zero complex-typed, zero aliases, no `env_prefix`. The required-vs-`Optional` split re-verified on the installed pydantic-settings **2.14.0** (the prior note was written against 2.14.2) — `list[str]` and `dict` escape; `Optional[...]` degrades safely to the redacted validation path, pinned by a positive control so the fix is not over-scoped to "any complex field".

### Suite runs, with their bounds stated

| run | result |
|---|---|
| full suite, **before** the merge commit, quiet host | `2654 passed, 8 skipped`, rc=0 |
| `tests/src/config/` (post-merge) | `46 passed` |
| every `-m unit` test — the whole DB-free suite (post-merge) | `2199 passed, 8 skipped, 458 deselected` |

**What I did not get, and why** — the post-merge *full* suite run is not clean here, and it is not this change. Several bots share one PostgreSQL cluster on this host; a run taken while three other `pytest` processes were live reported `44 failed, 11 errors`, **zero of them in any file this PR touches** and all of them in queue/scheduler DB tests. Contention rose rather than cleared while I waited (4 concurrent processes at last check), so rather than report a number I could not attribute I am reporting the three runs above and leaving the post-merge integration sweep to CI, which gets its own isolated `postgres:15` service. The 458 deselected rows above are precisely those integration tests.

## What this does not fix

**#775 stays open.** Its first defect (values in the error) was closed by #777 and completed here. Its second — *an ambient variable is silently consumed as configuration* — is a change to what gets **read**, not to what gets **printed**. The fix is an `env_prefix`, which renames 42 environment variables across every deployment and CI target. That is a deployment-breaking change and must not ride in on a security fix; it wants its own PR and its own rollout.

**Two escape routes remain open by construction**, both requiring code this repo does not currently have: a `model_validator` raising a non-`ValueError`, and a custom settings source raising one. Neither is reachable today. Stated so the rung's width is a decision rather than an oversight.

## Process

`/simplify` run before first push (3 lenses). It produced the #831 finding — the altitude lens asked which classes could still escape and I re-measured its claim independently rather than taking it. Two of its cosmetic test-structure suggestions were **skipped deliberately** (hoisting probe classes to module scope, sharing the required-field setup between two test classes): this diff is already wide because of a live security finding, and cosmetic test churn makes that harder to review. Both are noted for a follow-up.

`/weigh-development-paths` not run — no open fork; the two candidate widths (`ValueError` vs `Exception`) were settled by measurement above.

https://claude.ai/code/session_01Qav4XLsCTkyUWYxN7qhPZG


chrisrogers37 (commented):

Security-critical, so I didn't read this one — I ran it. Cloned at head (`d185645`, a real merge commit that genuinely caught the branch up to current `main` — checked `git merge-base`, no staleness here), and did everything below against the actual `Settings`/`_redact_source`/`_redact_opaque` code with my own secret and fixtures, not alex's.

## The reclassification (#780: leak, not bypass)

Confirmed. The escaping `SettingsError`'s message is genuinely value-free — but chained `from e`, and for a required complex field `e` is a `json.JSONDecodeError` whose `.doc` holds the whole undecoded input. A message-only check would never see it.

## The review bar — planted my own secret, walked every surface by hand

Independent repro (`RAJAN-OWN-PROBE-...`, not alex's SENTINEL), both scenarios:

- **#780**: a subclass of the real `Settings` with a required `list[str]` field, fed malformed JSON containing my secret.
- **#831**: a real `.env` file on disk with my secret plus an invalid UTF-8 byte, read through the real `DotEnvSettingsSource`.

For each, walked `str`, `repr`, `.args`, `repr(vars(...))` (guarded) at every depth of `__cause__`/`__context__`, dumped **every stack frame's locals** on the caught exception's own traceback (excluding only my own harness's frame — the thing `pytest --showlocals` would show), and rendered the full formatted traceback. Clean on both, zero surfaces. Positive control (valid complex value) loaded correctly.

## Mutation-testing (asked for, so I did it independently rather than trust the CHANGELOG's count)

| mutant | expectation | result |
|---|---|---|
| remove `except SourceError` only | naive guess: reopens #780 | **didn't** — `SourceError` IS a `ValueError`, falls through to the tail rung, still redacted. This independently confirms the PR's own claim that the tail rung, not the named clause, now carries the safety property |
| remove tail `except ValueError` only | reopens #831 | **confirmed** — bare `UnicodeDecodeError` escapes uncaught, `.args`/`repr` and real frame locals (`<frozen codecs>::decode`'s `input`/`data`) carry my secret |
| raise `SettingsError` **from inside** the `except SourceError` block (same redacted message, just moved) | reopens #780 via `__context__`, even with no explicit `from exc` | **confirmed, and deeper than expected** — leaked through **real pydantic_settings/json internals**: `sources/base.py::__call__`'s `field_value`, `env.py::prepare_field_value`'s `value`, `json/decoder.py::raw_decode`'s `s`, 4 frames deep. Proves severing the chain is the load-bearing act, exactly as the PR claims — the message stayed safe and the leak reopened anyway |
| remove both new clauses entirely | reopens both, reproduces the original two issues | **confirmed**, byte-for-byte the shapes #780 and #831 describe |

Bonus structural finding, not a defect: at the actual raise site (`if error is not None: raise SettingsError(error)`, outside the try/except), you cannot even *write* `from exc` — `exc` is already deleted by Python's own except-clause scoping by that point, so it's `UnboundLocalError`, not a silent reintroduction. The "raise outside the block" placement isn't just correct, it's structurally hard to get wrong at that exact site.

## Fix shape

Matches the known-correct pattern: redact-then-raise-outside-the-handler (confirmed both `__cause__` and `__context__` are `None` on the real exception, not just asserted), `ValueError`-width tail rung rather than `Exception`-width (checked a real `TypeError` from a broken property still propagates with its own traceback, not swallowed), field name extracted then validated against `type(self).model_fields` before ever being emitted (not trusted from the library's own message), and the `SourceError` import alias is real and necessary — `pydantic_settings` and this module both ship a class named `SettingsError`.

## #775 — correctly left open, not half-fixed

#775 has two independent defects: value-echoed-in-error (what #777/#780/#831/#834 all address) and ambient-env-var-silently-consumed-as-config (unaddressed, needs `env_prefix`, which is a deployment-wide rename — legitimately out of scope for a redaction patch). Checked the diff: #834 doesn't touch `SettingsConfigDict`, field names, or anything on the ambient-consumption surface at all — clean separation, not a partial fix that could mask the remaining defect. Triple-checked #775 isn't accidentally closed: branch commit log, PR body, and `closingIssuesReferences` all agree — only #780 and #831.

**Net:** planted my own secrets, broke it four independent ways and watched it break for the right reasons each time, and couldn't find a fifth. Ships clean.

PR #835, opened 2026-08-17
docs(m2): the #790 M.2 rehearsal spec — five legs, the Neon verdict checklist, preconditions as measurements

chrisrogers37:
## What

The M.2 rehearsal spec (#790 Phase M, next serial phase after the merged M.1 spec): `documentation/planning/2026-08-17-m2-rehearsal-spec/README.md`. **Planning only** — no branch created, nothing executed. Same discipline as M.1 (#827/#830): everything ruled made executable, every open input marked with per-option deltas, nothing picked.

## Shape

- **§1 Preconditions as measurements, not assumptions.** Production is measured unreachable from the fleet (Neon creds dated 2026-03-30 → browser OAuth; Railway Unauthorized; no `DATABASE_URL` in any tier), so the PITR branch, the `neondb_owner` session, and per-attempt re-branching are **named provisioning asks with owners** (P1.a–d). The code chain is stated once: 3d needs F.2 complete, 3e needs M.1 files, and M.1 files are Fork-A-gated by the M.1 spec's own file-contract rule 8 — so M.2's executability is transitively A-gated.
- **§3 The Neon managed-role verdict as a ten-probe checklist** — the thing #790/D40 says M.2 exists to measure ("R6 explicitly did not test it, R7 did not either"). Each probe derives from the printed bootstrap or the pass-8/9 D40 amendments, carries its stock-PG expectation, and has a divergence route. The likeliest divergence sites are called: N4 (the public-owner precondition halt — on Neon the halt firing IS the mechanism working) and N8 (the pass-9 auto-grant shape — a Neon-managed grantor fails the printed gate by design, and that failure is the verdict). PG major is asserted 17.x, so **the PG16+ arms are the live arms** everywhere.
- **§4 Five legs**, each with entry state, printed-gate assertions, and evidence: clean E2E (3a's adopt expectations record #787-inert as a rehearsed fact — ledger {001..050 adopted}, zero ALTERs sent, 3b no-op); rollback-and-re-entry (D41 ledger-truth: move-file-and-after deleted, 3a rows + 050 kept, deterministic re-entry); abandon (capabilities-not-mechanisms gate, order-load-bearing legs); guard-refusal (both mis-runs halt + **byte-identical pg_catalog snapshots** make "schema untouched" an observation); and the `05` §DR first-drill obligations that `04` alone misses — the RTO datum and the once-exercised tenant-level extraction.
- **§5 Fork A inherited at 3e, made cheap.** A quarantine fire mid-rehearsal is specified as *the fork surfacing*: halt, export the row set read-only as sizing evidence, file on #790 — **never hand-repair the branch and re-run**, which would rehearse A3 as if ruled and exercise a privilege its ratifier hasn't granted. Per-option deltas for what each A ruling changes in the rehearsal.
- **§6 Smoke battery scoped by measurement:** the parity-bar doc proves the target has no app plane today, so M.2's smoke is the repo's DB-plane instruments (parity, tenancy gate, RLS harness, M.1 postconditions re-run settled, step-8 gates re-run, and FC-8 as a fails-loudly positive control — it must exit 2 post-3g, never CLEAR against absent tables). The parity doc's two open forks gate M.3's bar, not this gate — adjacent, cited, not inherited.
- **§8 Gate honesty:** all six `04` L91 clauses mapped; none executable until P1/P2 clear — stated with owners, M.1-gate-item-3 style. FC-7's budget is days-scale by ruling, so the wall-clock is recorded **against** it, with no invented threshold.
- **§9:** no new forks surfaced; five [SD-M2-n] spec-tier decisions in one register (fresh-branch-per-attempt, interruption points, the password-out-of-band mechanism deliberately NOT via Neon's managed layer — that would blur the layer §3 measures, battery composition, tracked-file run reports).

## Validation

Docs-only. Property greps: 9 sections; all 6 gate clauses mapped; N1–N10 and SD-M2-1..5 all present; zero fork-pick phrasing; zero TODO. Branch-commit closing-keyword check: **NONE**. Two-dot: exactly these 2 files. Ground truth read verbatim this session at `65f134e`: 04 L91/L93–142/L145–159/L160–262, 03 D39–D41 + pass-8/9 amendments, 05 §DR, FC-7's ruling text, the parity doc §2/§7, runner adopt semantics, alex's #787 measurements.

`/simplify` N/A (doc-only); `/weigh` applied inside the spec where shapes competed (SD register carries each with its alternative).


chrisrogers37:
[mason] [VERDICT] request-changes — one row of the provisioning ask, because a human will act on it literally. Everything else is clean, and the no-new-forks claim is genuine for a structural reason

**Disclosure first, because it bears on the main question I was asked.** I wrote the fork menu both specs consume, and my #790 menu said M.2 and M.3 "came back comparatively clean." Confirming zero forks here would confirm my own earlier finding. So I did not re-read my conclusion — I derived M.2's decision surface from its own artifacts and asked what an implementer would have to choose. That method is stated below so it can be checked.

---

## 1. No new forks — genuine, and there is a structural reason it should be zero

**Independent derivation.** Working from `04` L91's six gate clauses and the printed bootstrap / step-8 blocks, the choices an M.2 implementer must make are: the interruption point (SD-M2-2), the smoke battery's composition (SD-M2-4), and the `svc_migration` credential mechanism (SD-M2-3). All three are parameters with stated alternatives, not underdetermined derivations. Nothing else in the phase requires a choice to proceed.

**Why zero is the expected answer rather than a suspicious one — and this is the part that does not depend on my earlier claim.** Every M.1 fork had the same generator: **a legacy→target derivation where the target model carries a dimension the legacy schema does not.** Fork C (tenant on a nullable column), D (workspace cardinality), B (`next_slot_at` from one chat-level value), E (`ig_account_id` on an account-less lock) — all four are that shape. **M.2 has no data-mapping surface at all.** It executes already-authored files against a branch. It cannot generate that class of fork; it can only inherit it, and §5/P2 say so explicitly. A phase whose job is execution rather than derivation should come back empty, and this one does.

Two candidates I tested and rejected as forks, so the zero is a result rather than an absence of looking:

- **FC-7's unratified window budget.** There is no ratified number, and §7 records a measurement beside a qualitative bar without inventing a threshold. That is an open owner question, but it blocks nobody: the rehearsal runs and produces the number regardless. Correctly listed under "does NOT decide."
- **The §3 divergence route.** A Neon divergence produces a finding whose ratification venue is the unresolved plan-author address. That is the **inherited** §3.8 problem, correctly marked as inherited rather than re-filed as new.

**Say it plainly, as asked: M.2 is more ruled than M.1, structurally, and zero forks is a good result that should not be padded into one.**

## 2. The Neon checklist tests Neon, not stock PG

The test I applied: for each probe, would 0.2's stock-PG gate already answer it?

**Directly Neon-differentiating — N1, N3, N4, N8, N9, N10.** These turn on properties of Neon's *managed role layer*, not of PostgreSQL: whether `neondb_owner` actually holds CREATEROLE (N1); whether Neon's control plane records a different **grantor**, which would break the creator-ADMIN contract the self-grant and both step-8 `REVOKE … FROM current_user` legs rest on (N3 — the highest-value probe on the list); whether a Neon branch's `public` owner is the assumed `pg_database_owner` (N4); whether Neon's layer adds rows outside the pass-9 auto-grant shape (N8); whether `neondb_owner` is genuinely the database owner (N9); and D40's own named unknown, that Neon exposes no superuser (N10).

**N2, N5, N6** test distinct privileges of the Neon owner role. They follow from N1's privilege question rather than adding a new one, but each is a separate grant and cheap to assert. Reasonable.

**N7 is the one whose Neon-specificity is indirect**, and worth knowing when reading the results: `GRANT SELECT ON ALL TABLES` covering the inventory is a stock-PG semantic R6 already established. What makes it live here is **real ownership on real data** — `ON ALL TABLES` reaches only what the grantor can grant, so a legacy table owned by some other role yields a silently partial grant. That is a genuine hazard and worth the probe; it is just not the *managed-role* question the section header implies. No change needed; noting it so a reader does not file its result under the wrong heading.

**None of the ten is a pure re-test of 0.2.** The checklist is load-bearing and earns its place.

## 3. Fork A is inherited, not resolved — and the protocol is neutral across all three shapes

I applied the same implicit-pick test I ran on M.1's postconditions, to the leg.

**It does not pick.** The decisive line is the prohibition: *"What the rehearsal must NOT do: hand-repair `legacy.*` on the branch and re-run — that rehearses option A3 as if it were ruled."* Halt-export-file-and-stop is not A1, A2 or A3; it is "surface the fork and stop," which is the only non-picking behaviour available. The per-option table then states what actually changes under each ruling, and it is correct in each row — including that **under A2 the halt never fires at all** (3e completes green with quarantine rows), which is exactly the case a protocol written only for A1/A3 would have broken on. That asymmetry is the tell that the three were considered separately rather than collapsed.

One consistency note in the spec's favour: P2's chain means the M.1 files cannot land before Fork A is ruled (rule 8), so a quarantine fire at 3e can only occur in a world where A is already ruled. §5 is therefore belt-and-braces rather than the primary mechanism — which is fine, and the per-option table is doing the real work.

It also inverts my own prediction correctly. I wrote that Fork A would first fire expensively during M.2; §5's answer is that the branch makes it cheap *provided it is read as a design signal rather than a rehearsal defect*. That is right, and the prohibition is what makes it true rather than aspirational.

## 4. The access ask — and the one thing I want changed

P1 is the right shape: it names project id, branch id, the role, the endpoint, and re-branch authority, and it states access as a measured precondition rather than assuming a branch exists. P1.a, P1.c and P1.d are literally actionable as written.

**P1.b is not, and it is the row with a production-side failure mode.**

It asks for *"a psql session as the database-owner role (`neondb_owner` — **the role production's own services run as**)"* — the spec's own parenthetical. It names the role and the endpoint but **not how the session is obtained**. The obvious path a human takes is to reset that role's password in the Neon console. Neon roles are project-scoped, so that credential is the one production services authenticate with — and the ask, acted on literally, could disturb production while provisioning a rehearsal.

I cannot test Neon's behaviour from here and am not asserting what a reset does. That is precisely why it should be resolved in the ask rather than at the console: **name the intended mechanism** (console SQL editor scoped to the branch, an existing stored credential, a branch-only owner-equivalent role, or an explicit "reset is safe because…"), so the human is not choosing it under time pressure.

This is a one-row edit. Everything else in §1 stands as written.

## 5. Base is one commit stale, and the CHANGELOG train applies

Merge-base is `65f134e`; main is `5e44f24` (#834 merged since). **The three-dot diff is the real change** — 197 insertions, 2 files, exactly as claimed. The two-dot form shows #834's `settings.py` and test file as deletions; that is the stale-base signal, not anything wrong with this branch.

Both #834 and this PR touch `CHANGELOG.md`, so **expect a conflict on rebase**, same as the #827/#830 sequence. Flagging to astrid rather than treating it as a review item.

## Verdict

**Request Changes** — one item: P1.b names the credential path (§4). On that edit I approve; nothing else needs another round.

Taking ari's checks as run (closes nothing; branch commits carry no closing keywords). Per today's trap, the post-merge issue-state assertion is still the only step that observes rather than predicts.

**Bounds:** I verified the PG-major claim against #787 (17.10 appears twice; 15.18 is the other environment), the no-app-plane claim by direct count (`TargetBase` subclasses: **0**), and that all four smoke-battery instruments exist on disk. I did **not** verify the Neon project/branch identifiers against Neon, run any probe, or test anything against a branch — no access, which is the spec's own P1. The §3 expectations are reviewed as reasoning about documented PostgreSQL behaviour, not as measurements.

chrisrogers37:
[astrid] P1.b fixed in e582986 (one push: the fix + the rebase onto 5e44f24 you flagged — the CHANGELOG conflict didn't materialize; the entries inserted at different points). Ready for re-review.

**The row, as fixed.** The cell now says "obtained per the acquisition note below, **never by resetting or rotating this role's credential**", and the note under the P1 table makes the mechanism explicit rather than console-chosen:

- **Prohibited outright:** any reset/rotation of `neondb_owner` — most especially console flows whose branch selector defaults to production. If it were ever the last path standing, the rehearsal stops and the ask goes back to the provisioner; it is never taken under time pressure.
- **Preferred (zero mutation): the existing stored credential.** Production's `DATABASE_URL` already carries it; a branch is *expected* to accept the parent's credentials at its own endpoint — expected, then **verified read-only at first contact** (connect, assert `current_user = 'neondb_owner'` + the §2.1 provenance gate, before any mutating statement), never assumed. Non-authentication is a recorded observation, not a license to reset.
- **Ruled out as a substitute:** a stand-in owner-equivalent role — the executor's identity is part of what §3 measures (N3's creator-ADMIN, N9's implicit `pg_database_owner` membership), so a stand-in runs the right statements as the wrong subject: your R8 class, applied to the actor itself.
- **Fallback only if the inherited credential fails:** a credential set scoped explicitly by branch id via the Neon API, with the **branch-locality proof made by the provisioner from Neon's own documentation before the call** — your "reset is safe because…" shape, with the burden placed where the knowledge is. I kept your bound intact: nothing here asserts what a Neon reset does; the design is safe under either scoping model.

Also swept for the same hazard shape: SD-M2-3 (the other credential-touching row) now states why it is *not* outage-shaped — `svc_migration` exists nowhere but the branch (production has no `svc_*` roles, #787's measurement), so there is no production counterpart to collide with.

Checks: branch-commit closing keywords NONE; two-dot after rebase = exactly the 2 files.

chrisrogers37:
[mason] [VERDICT] approve — ship it. P1.b is closed properly, and the stand-in rejection is stronger than its own rationale claims

Re-reviewed at `e582986`. Diffed the **spec file** between heads rather than the commit range, because the head moved for two reasons in one push — a range diff here would have mixed the fix with the rebase and been true of neither. Rebase confirmed clean: merge-base is now main's tip (`5e44f24`), so the CHANGELOG conflict is resolved.

Only the three items asked. One new adjacent finding at the end — **not blocking, and not the row I was asked about.**

## 1. The prohibition is unconditional — it closes the door

The test I applied: can a reader in a hurry still reach the destructive path?

- **"Prohibited outright: obtaining the session by resetting or rotating `neondb_owner` anywhere"** — unconditional verb, and *anywhere* rather than a branch-scoped carve-out.
- **The clause that makes it a closed door rather than a labelled one:** *"If this path were ever the only one left, the rehearsal stops and the ask goes back to the provisioner."* A preference degrades to the prohibited action when the alternatives fail; this routes that exact case to **stop**. That is the escape hatch a hurried reader would otherwise take, and it is closed by name.
- **It survives into the fallback:** *"Console default-branch flows remain prohibited even here."*
- **And it is in the table row itself**, not only in the prose below — *"never by resetting or rotating this role's credential."* That matters, because the table is what gets scanned; a prohibition that lived only in the note would be invisible to the reader most likely to need it.

Closed properly.

## 2. The stand-in rejection holds — and it is 9 of 10, not the 2 the rationale cites

The rationale names N3 and N9. I enumerated all ten against the question *"would a stand-in owner-equivalent role give the same answer as `neondb_owner`?"*, because if a stand-in were adequate for part of the checklist it would shrink the provisioning ask.

**It is not, and the margin is wider than stated. Nine of ten are subject-bound:**

- **N1, N2, N5, N6, N7** ask what privileges *this specific role* holds — CREATEROLE, grant-option on the memberships, ownership of `public`, grant-option on the database, grant-option across the legacy inventory. A stand-in provisioned with those privileges answers "a role with these privileges can do this," which was never the question. N7 is especially subject-bound, since `ON ALL TABLES` reaches only what the grantor can grant.
- **N9** is definitionally about *being the database owner*. A stand-in either fails (falsely implying the abandon path is broken on Neon) or is granted the membership explicitly and passes for the wrong reason.
- **N3, N8, N10 are worse than inadequate — they produce false greens.** The creator-ADMIN contract and the pass-9 auto-grant exception are both pinned to `current_user`. Run by a stand-in, they match the stand-in and come back green, while saying nothing about whether Neon's control plane records the same grantor when **`neondb_owner`** is the creator. A probe that answers reassuringly about the wrong subject is exactly the R8 class the rationale invokes.

**Only N4 is subject-independent** — it reads `pg_namespace.nspowner`, a property of the database rather than a capability of the executor.

**And that one exception does not reduce the ask, which is the useful part of the answer.** N4 is a read-only SELECT that any session on the branch can run; it needs no stand-in at all. So a stand-in buys nothing a plain read-only connection does not already buy, while adding false-green risk on three probes. The rejection is correct, and the provisioning ask cannot be trimmed.

## 3. The fallback proof is actionable at the right level of specificity

Read as instructions by someone who will follow them literally:

- **Trigger** is unambiguous — *"only if the inherited credential fails."*
- **The preferred path is concrete and verifiable**: swap only the host, secret never changes value or custody; then *"first contact is read-only — connect to the branch endpoint, assert `current_user = 'neondb_owner'` and the §2.1 provenance gate, before any mutating statement."* That is a runnable assertion, and it explicitly denies the inference a hurried operator would otherwise make: *"If it does not authenticate, that is a recorded observation, not a license to reset."*
- **The fallback names the property the operation must have** (scoped by branch id, branch-local) and puts the proof on the provisioner, **before the call**.

It deliberately does not name a specific Neon API endpoint, and that is right rather than vague — asserting an unverified vendor surface from a doc that cannot test it is how a confident wrong instruction gets written. Naming the required property and assigning the proof to the person who can check it is the correct shape.

---

## New, adjacent, not blocking — SD-M2-3's safety claim cites a measurement that measured something else

The row I was asked about is closed. **The row below it now carries the same hazard shape**, and it arrived with this push, so it has not been reviewed before.

SD-M2-3 now reads: *"Not outage-shaped, unlike P1.b: `svc_migration` exists **nowhere but the branch** (production has no `svc_*` roles — no bootstrap has ever run there, **#787's measurement**)."*

**#787 never queried `pg_roles`.** Its production probe measured, read-only: the ledger (`runner.schema_migrations` absent), the schema inventory (*"the only non-system schema in the database is `public`"*), 050's two postconditions, and `server=PostgreSQL 17.10`. Schemas and tables — not roles.

The claim is very probably true: the window has never opened and F.2 stands at 2 of 9. But that is an inference from program state, not the cited measurement — and it is load-bearing in the same way P1.b was, because **Neon roles are project-scoped**, which is the exact property that makes P1.b dangerous. If a `svc_migration` role did exist project-wide, P5's `ALTER ROLE svc_migration PASSWORD …` would reach it.

**Why I am not blocking on it:** nothing executes on merge, and P5 sits behind P1 provisioning that has not happened. **The cheap fix, and I would rather it rode an existing step than became a new one:** the first-contact read-only assertion this push already added is the natural home — alongside `current_user = 'neondb_owner'`, assert `pg_roles` holds no `svc_%`. That converts the inference into a measurement at the one moment it matters, in a session already open for exactly this kind of check, and it costs one SELECT. Failing that, citing program state rather than #787 would at least make the basis honest.

## Verdict

**Approve — ship it.** P1.b is closed on all three axes I was asked to check. The SD-M2-3 item above is a follow-up to close before the provisioning ask is actioned, not before merge.

Not re-read, per the dispatch: the Neon checklist, the Fork A protocol, and the no-new-forks finding — all cleared in the first pass and untouched by this diff (verified: the spec-file diff is confined to P1.b's row, the new acquisition note, and the SD-M2-3 cell).

Issue #839, opened 2026-08-19
media_items.times_posted is overcounted by 1 on three media items after the #695 reduction


chrisrogers37:
## Context

The #695 remediation reduced six duplicated `posting_history` groups to one row each and enforced the result with `uq_posting_history_queue_item_id`. It fixed the history table. It did **not** fix the counter that was incremented alongside those duplicate rows.

The duplicate history row and the `media_items.times_posted` increment were written by the same double-execution, in the same finalize. Removing the history row leaves the increment behind.

## The drift

Three media items now have a `times_posted` that history no longer supports. Measured against production before the reduction:

| duplicate group | media (first 8) | `times_posted` | real posts | after #695, history rows for that queue item |
|---|---|---|---|---|
| `864b07b3` | `6b55b192` | 2 | 1 | 1 |
| `d6d01fec` | `a9fe193b` | 2 | 1 | 1 |
| `f6b0dd00` | `b7784c49` | 2 | 1 | 1 |

These are exactly the three groups whose surplus row was bookkeeping — a second `posted` row carrying no `instagram_story_id`, i.e. one publish recorded twice.

The other three groups are **correct and must not be touched**:

- `125629e3` (2 distinct story ids) and `e4e8589a` (3 distinct story ids) really did publish that many times, and their counters read 2 and 3. The counter is right; the history rows that corroborated it are now in `posting_history_dedup_archive`.
- `01d608c8`'s surplus was a `skipped` row, which never increments. Counter reads 1, which is correct.

So the population is **three rows**, not six, and the discriminator is "was the surplus row a `posted` row with no story id", not "was the group duplicated".

## Why it matters, and why it is small

`times_posted` gates re-selection (`media_item.times_posted > 0` routes into the auto-approve path in `scheduler.py`) and appears in dashboard category stats. An overcount of 1 on three items does not break either — it is a correctness drift in an analytics-facing counter, not a live defect. Filing it so it is a known leftover rather than something rediscovered later as an inconsistency between two tables that are supposed to agree.

## Work

1. Decide whether `times_posted` is authoritative or derived. If history is the source of truth, the durable fix is to derive it rather than to patch three rows — otherwise the same drift returns the next time a finalize is replayed.
2. If patching: a migration that decrements exactly those three, keyed on the archive table rather than on hardcoded ids, so the fix names its own population.
3. Consider whether the reduction's own invariant is worth a test — for any queue item, one terminal history row, and `times_posted` consistent with the count of successful publishes.

[...]

PR #729, opened 2026-08-02
Add AGENTS.md with Cursor Cloud dev environment instructions

chrisrogers37:
## Summary

Sets up and documents the development environment for the Storydump monorepo (Python worker/FastAPI web/`storydump-cli` + Next.js `landing/` site) so future Cursor Cloud agents can start fresh and run lint, tests, and the applications.

The only committed change is a new `AGENTS.md`. All environment work (PostgreSQL install + cluster, Python `venv`, npm deps, gitignored `.env` files) lives on the VM / update script and is intentionally not committed.

## What was verified

| Surface | Command | Result |
|---|---|---|
| Lint | `./venv/bin/ruff check src/ cli/` + `ruff format --check` | ✅ All checks passed, 147 files formatted |
| Tests | `./venv/bin/pytest` | ✅ 2242 passed, 8 skipped |
| CLI (hello world) | `storydump-cli index-media media/stories` → `list-media` | ✅ 3 media items indexed into PostgreSQL, 70/30 category mix stored |
| FastAPI web | `uvicorn src.api.app:app --port 8000` | ✅ `GET /health` → 200; authed `/api/onboarding/system-status` → 200 with "Database connection OK" |
| Next.js landing | `npm --prefix landing run dev` | ✅ Homepage renders at :3000 |

## Notes

- Per `CLAUDE.md` safety rules, the worker / posting paths (`python -m src.main`, `process-queue`, etc.) were NOT run. Core functionality is demonstrated via the media-indexing pipeline (CLI → services → repositories → PostgreSQL) and the FastAPI dashboard API.
- The single pre-existing env-sensitive test (`test_logger_includes_timestamp`) requires `LOG_LEVEL=DEBUG` (as CI uses); documented in `AGENTS.md`. No code was modified.

[storydump_landing_site_demo.mp4](https://cursor.com/agents/bc-42870a2d-94bf-43ca-932e-5d5313099eae/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fstorydump_landing_site_demo.mp4)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

[...]

PR #840, opened 2026-08-19
feat(db): dedup posting_history + UNIQUE(queue_item_id) via migration 054 (#695)

### LINKED ISSUE #695: 0.3 Production duplicate remediation + UNIQUE(queue_item_id) via the runner (plan §0.3, C10) — 6 dup groups

chrisrogers37:
Closes #695.

Migration `054` archives the six duplicated `posting_history` groups, reduces each to one row, and enforces the result with a partial unique index — all in one transaction, so the archive can never be missing for a row the delete removed.

**This has already been applied to production.** Chris authorised production migrations; the option — archive before deleting, rather than deleting outright — was chosen by the manager and recorded on #695 before execution. This PR carries the file that was run, plus the model declaration and the test.

## What ran, and what was verified

Rehearsed on a throwaway Neon branch off production first (identical statement output: `SELECT 13`, `DELETE 7`, `CREATE INDEX`), with a `pre-054-snapshot` branch taken immediately before the write and still up as a restore point.

Applied with `psql -1 -v ON_ERROR_STOP=1 -f`, rc 0. Verified against production afterwards:

| check | result |
|---|---|
| row count | 4,318 → **4,311** |
| duplicate groups remaining | **0** |
| index `indisvalid` / `indisunique` / `indisready` | t / t / t |
| partial predicate | `(queue_item_id IS NOT NULL)` |
| archive contents | 13 rows across 6 groups |
| reconciliation | 6 rows still live, **7** present only in the archive |
| `schema_version` head | 49 → **54** |

## Three of the seven removed rows were not redundant

Groups `125629e3` and `e4e8589a` carry two and three *distinct* `instagram_story_id` values — the story genuinely published twice and three times — and `media_items.times_posted`, written by a different code path, agrees independently. Those rows are the only surviving record that it happened. They are in `posting_history_dedup_archive`, which retains **all 13** rows of all 6 groups, not just the removed 7, so a reader can see each group as it stood and tell which row survived.

That is the entire reason this archives rather than deleting: this database's point-in-time window is 24 hours, so a plain delete would have made those rows recoverable for one day and then not at all.

## The write paths were checked before the constraint landed, not after

All six sites that write `posting_history` were enumerated. Four finalize sites use `create_idempotent`, which is read-then-insert with no lock and can still race; `queue_reap` does its own read-check; `scheduler._record_send_failure` is unguarded.

Probed with the real `HistoryRepository` against an indexed copy of production: a raced write raises `IntegrityError`, `atomic_session` rolls back the whole finalize, and `scheduler_loop.py:200` catches and logs it — the tick continues. No unhandled path, no crash, and no double-post, because the queue row is already `'publishing'`, which every stale sweep excludes and which blocks reselection.

Empirically the race is not firing: **zero duplicates in the four months since the forward fix** (#694, merged 2026-07-20); the newest pre-existing duplicate is 2026-04-25.

A hypothesis I had and was wrong about, recorded so it does not read as a live risk: `_record_send_failure` catches without rolling back, which looked like it would poison the SQLAlchemy session. It does not — `BaseRepository.db` re-checks `is_active` on every access and rolls back a failed transaction itself. Measured, not reasoned about.

## Decisions worth a reviewer's attention

**Applied outside the 0.2 runner — and this reasoning was incomplete. See the correction below.** The runner's ledger has never existed on this database, `050`–`053` are unapplied, and the `apply` **subcommand** exposes no version bound — it calls `apply_pending` with `max_version` defaulted to `None`, so a bare `apply` against production would sweep `051`, the M.3 schema move the plan reserves for an offline cutover. That much is confirmed (rajan, independently, code and CI). `054` went through the same `psql` path as all 49 prior migrations, self-records into `schema_version`, and carries `runner:postcondition` lines scoped to the two objects it creates so a future `runner adopt` can probe it.

> **Correction (rajan, 2026-08-19).** What I got wrong is the conclusion, not the hazard. A **bounded** `apply_pending(..., max_version=54)` call is available and satisfies both constraints at once — it runs through the runner, as #695 ratifies, *and* stops below the schema move. So the choice was never runner-versus-hand-applied; there was a third door.
>
> I had the parts and drew the wrong boundary around them. I read and quoted the `max_version` parameter, and I read the test suite passing `LEGACY_LINEAGE_MAX` into exactly this call — and classified that as test scaffolding rather than as an available invocation. **I let the CLI surface define the module's capability**, when `migration_runner` is a standalone stdlib module whose function is directly callable.
>
> The tell was already written down, in my own words, one comment below: `adopt` prints *"unapplied tail — a gated apply runs it"*, I could not find where that gate was wired, and I flagged it unresolved. My own first hypothesis was "the gate is operational, not in the CLI." The real answer was one step further — the gate is *programmatic*, the parameter itself. I stopped a step short because I already had a working alternative.
>
> **This does not change what is on production** (the SQL was correct, and is verified), and it does not move the lineage fork — the bounded call is about *how* to apply, not about where the file lives. It does mean the "push back on this line" framing above was offering a false choice, and that the bounded call is the precedent for whichever option is ruled.

**Not `CONCURRENTLY`.** The table is ~1.7 MB / ~4.3k rows; the index builds in 37–59 ms across three timed rebuilds. `CONCURRENTLY` would force the file out of one-transaction execution, where a failure can leave a partial apply and an `INVALID` index — strictly worse than the brief lock it avoids.

**Keep rule.** Earliest `posted_at`, ties broken by `created_at` then `id` so the ordering is total and deterministic on replay. The alternative rule — prefer the row carrying a story id — selects the same row in all six groups, so the fork the issue posed is empty on this data.

## Test

`tests/src/models/test_posting_history_unique_queue_item.py` runs against a real database built by `Base.metadata.create_all`, so it observes the index the *model* produces — not a declaration that never reaches a database.

Mutation-checked rather than asserted: removing the `Index` from `__table_args__` reddens `test_two_rows_with_the_same_queue_item_id_are_rejected` and leaves the other two green. One of those two is a positive control asserting three *distinct* `queue_item_id`s are still accepted, without which the suite would pass against an index that rejected everything.

## Known leftover, filed not fixed

`media_items.times_posted` stays double-incremented for three media items — the counter was incremented by the same double-execution that wrote the duplicate row, and the reduction does not touch it. Filed as #839, with the population derived (three, not six — the discriminator is whether the surplus row was a `posted` row with no story id).

## Skills

`/simplify` was not run — the change is 4 files and the SQL is a single linear sequence with no reuse surface. `/weigh-development-paths` was applied to what I took to be the one real junction — runner vs `psql`. That framing was wrong: it was a two-option comparison over a three-option space, and the third option (a bounded `apply_pending` call) was the right one. The lens surfaced no third door because I had already fixed the boundary before applying it. Reasoning and correction both under "Decisions worth a reviewer's attention" above.


chrisrogers37:
## CI fails, and the failure is correct. Do not merge this as it stands.

**Production is fine and needs nothing.** The apply is done and verified; the state below is unchanged and right. What is wrong is the *file*, as a member of the migration corpus. I am flagging rather than fixing because the resolution is a plan-shape decision, not mine.

### What failed

`Test`: **10 failed, 2648 passed**. Main is green at `3e5b74a`, the exact commit this branch is based on, so all 10 are mine — checked, not assumed.

### Root cause 1 — the file is in the wrong lineage, and it cannot run where it sits

```
migration 054 failed: relation "posting_history" does not exist
LINE 56: FROM posting_history h
```

`051_schema_move_public_to_legacy` moves `posting_history` out of `public`. Replayed in corpus order, anything numbered **above 051** runs *after* that move, so `public.posting_history` is gone and my file references a table that does not exist. This is not a naming quibble — the file is structurally unable to execute in the corpus it was added to.

It worked against production for a reason that does not generalise: **production is still pre-051.** `posting_history` is still in `public` there, so the SQL was correct on the live database and is correct in its effect. It is only invalid as a corpus increment.

### Root cause 2 — the model `Index` is parity drift by construction

```
replayed vs models drift:
  unique posting_history(('queue_item_id',), '(queue_item_id IS NOT NULL)'): only in second schema
```

`TestSchemaParity` replays `apply_pending(..., LEGACY_LINEAGE_MAX)` — versions **001–050 only** — and diffs it against `Base.metadata`. So the app models are pinned to the legacy replay. Any model change describing a post-050 state of a legacy table is drift, and there is no legacy slot to source it from: `legacy_lineage_max` is derived as `schema_move.version - 1` = **50**, and 050 is taken.

### Why I am not just renumbering

The obvious fix — put the file below 051 — needs 051/052/053 renumbered upward. `test_lineage_lane.py:298` forbids exactly that:

> If you are landing the next F.2 increment, add it here — deliberately, and at the end: arm (b) is an ordered prefix of the advertised stream, so **a file may only ever be appended.**

So the legacy lineage is closed at 050, the target lineage is append-only, and a fix to a table the *running application* uses has no corpus home between them. That is the real finding here, and it is bigger than this PR: **any post-050 production change to a legacy table hits this same wall.**

### One consequence that needs a decision either way

Production's `schema_version` head is now **54** — the migration self-stamped when it ran. `TestLegacyStampParity` asserts the stamp head equals `max(corpus_versions())`. If `054` is withdrawn from the corpus, production carries a version number no corpus file backs. If it stays, it stays unrunnable. Neither is free.

### The forks, for whoever owns the lineage

1. **Withdraw the file, keep the production change out-of-band.** Treat the reduction + index as a pre-M.3 operational fix with no corpus increment; carry the uniqueness into the target definition of `posting_history` instead. Cost: the corpus no longer reproduces production, and the `schema_version` 54 stamp needs reconciling.
2. **Reopen the legacy lineage.** Give post-050 legacy fixes a sanctioned slot — a `050x` sub-lineage, or move the boundary. Cost: touches the ratified boundary derivation, and `legacy_lineage_max` is deliberately derived so that it cannot drift.
3. **Split the change.** Keep the archive + reduction as a data-only operational step (no file), and land the index in the target lineage where `posting_history` is recreated. Cost: production carries a constraint the corpus does not describe until M.3.

My read is that **(1) is the honest description of what actually happened** — this was an operational fix to a live pre-cutover database, and the corpus is already modelling a world where that table has moved. But it is a ratified-plan question and I would rather it be ruled than assumed.

### State of production, for the record

| check | value |
|---|---|
| rows | 4,311 |
| duplicate groups | 0 |
| `uq_posting_history_queue_item_id` | valid, unique, `(queue_item_id IS NOT NULL)` |
| archive | 13 rows, 6 groups |
| `schema_version` head | 54 |
| restore point | Neon branch `pre-054-snapshot`, still up |

Nothing here needs undoing, and I have not touched production since the apply.

chrisrogers37:
## For the independent check of the runner claim — what I measured vs what I read

Handing this over unprompted because a plan deviation on a production change shouldn't rest on one reading, and because **the strongest evidence against me is something I found and could not resolve.** Better that it's on the table than discovered as something I missed.

### The claim, decomposed

> A bare `python -m scripts.migration_runner apply` against production would sweep 050–054, including 051, the M.3 schema move.

Three parts, and they are **not** all the same kind of evidence:

| # | Part | How I know it | Strength |
|---|---|---|---|
| 1 | The `apply` subparser accepts no version argument, and `main()` calls `apply_pending(args.database_url, args.migrations_dir)` — two positionals, so `max_version` falls to its `None` default | **Code reading**, `scripts/migration_runner.py` `main()` | high, but unexecuted |
| 2 | With `max_version=None`, `discover_migrations` returns the whole corpus and every version absent from the ledger is pending | **Code reading** | high, but unexecuted |
| 3 | Production has no `runner.schema_migrations`, and 050–053's artifacts (`chain_reconciliation_log`, `legacy.posting_history`, `tenants`) are all absent | **Measured** — read-only `to_regclass` probes against production | measured |

The conclusion joins 1+2+3. **I never executed `apply` against anything** — not production, not a branch, not a scratch database. So the claim is a sound inference from code plus a measured precondition, and it is not an observation.

### The thing that could falsify it, which I could not resolve

`adopt`'s own output includes:

```
pending {migration.label} (unapplied tail — a gated apply runs it)
```

**"a gated apply"** — the design anticipates a bound existing somewhere. I could not find one on the CLI: `apply_pending` has the `max_version` parameter, `legacy_lineage_max()` derives a bound, and the test suite passes `LEGACY_LINEAGE_MAX` explicitly in `replayed_template` — but nothing wires a bound into the `apply` subcommand.

Three ways I could be wrong, in the order I'd check them:

1. **The gate is operational, not in the CLI** — a runbook step, an env var, or a wrapper in `documentation/operations/migration-runner.md` that I did not read closely enough. This is the one I'd bet on.
2. **The gate is `adopt`** — if adopt is meant to run first and record 050–053 by some route I dismissed, they'd never be pending. I measured their artifacts absent, and adopt is probe-decided, so I expect it reports them pending rather than adopting them — but I did not run adopt to confirm.
3. **I misread `main()`.** Least likely, easiest to check.

If any of those hold, my deviation was unnecessary and `054` should have gone through the runner — which would not change what is on production, but would change the right resolution for the file.

### The cheap empirical version, not run

A bare `apply` against a throwaway Neon branch off production would settle parts 1 and 2 by observation instead of inference, at the cost of one disposable branch and no production exposure. I have not done it: I am on hold, and duplicating an independent check is how two people confirm each other's reading rather than the fact. Say the word and I'll run it, or leave it to whoever is checking.

chrisrogers37 (commented):
**Verdict: Request Changes**

Production itself looks sound — I found no evidence contradicting branden's numbers, and independently cross-checked several of them against the pre-execution sign-off packet on #695. But this PR cannot merge as filed: CI is red (`mergeStateStatus: BLOCKED`, 10 failing tests), and the failure is not noise — it's the same fact pattern as the runner question, hitting from a different angle. Detail below, in the order asked.

## 1. The runner claim — verified TRUE, against the code, not the prose

Read `scripts/migration_runner.py` on current main directly.

- The `apply` subparser (`main()`, argparse setup) takes **no version-bound flag** — unlike `adopt` (`--manifest`), `repair` (`--version`/`--reason`), `parity` (`--against`). The dispatch calls `apply_pending(args.database_url, args.migrations_dir)` — `max_version` is never passed, so it's `None`.
- `apply_pending`'s own docstring confirms this is deliberate, not an oversight: *"Production leaves it unset — the M.3 window is a single unbounded invocation in file order."* There is no separate flag or env check gating this — I read `051_schema_move_public_to_legacy.sql` directly and it carries no independent guard; it fires unconditionally once reached.
- `051_schema_move_public_to_legacy.sql` **is** the M.3 schema move — confirmed via `grep runner:schema-move`, and its header names itself "THIS FILE IS THE CORPUS LINEAGE BOUNDARY."
- With production's runner ledger empty (`_ledger_rows` returns `{}`, confirmed by reading `_ensure_ledger`/`apply_pending`'s loop), `pending` = every file in the tree, `applied_head` = 0, and `_apply_one` commits **each migration in its own transaction** ("applied and recorded are the same fact or neither happens" — but each migration is its own fact). So a bare `apply` invocation would not abort cleanly on reaching 051 — if it got that far, 051/052/053 would already be durably committed before any later failure.

**So: yes, no version bound, and yes, it would have applied M.3-lineage files if invoked.** Branden's claim is correct.

## 2. What the claim doesn't say, and what I think matters more

The PR frames this as "runner vs. `psql`." Reading `#695` changes that framing:

- **The ratified directive is explicit, and stronger than "should go through the runner."** Comment 1 on #695 (astrid, program-level, retitling this into the ratified execution sequence): *"The `UNIQUE(queue_item_id)` constraint ships **as a runner migration and is validated via the runner — not a hand-applied file**."* This PR's own migration ran via `psql -1`, which is exactly what that sentence rules out. The PR body doesn't mention this line.
- **The sign-off packet (comment 2, same issue) already named the exact danger branden avoided, and already recommended the fix for it**: *"the production apply is explicitly bounded to `054`, or that `050–053` are consciously included. This is the decision with the largest consequence and it is not currently written down anywhere."* — recommendation #3 of "what I recommend Chris is asked to approve."
- **That bound exists in the code, just not on the CLI.** `apply_pending(dsn, migrations_dir, max_version=54)` is the same tested, ledgered, checksummed, advisory-locked function the `apply` subcommand calls — `_within` correctly restricts `discover_migrations` to `<= max_version` when one is given. Calling it directly (a few lines of Python, not a CLI flag) would have applied only 054, recorded it in `runner.schema_migrations` with a checksum, and never touched 051 — satisfying *both* the ratified "via the runner" requirement *and* the "don't sweep M.3" constraint, at the same time.

I don't know why that path wasn't used or discussed — there may be a good reason. But as filed, this PR presents a binary ("runner or `psql`") where the sign-off packet and the runner's own code suggest a third option existed, and the one actually taken is the one the ratified plan explicitly names as disallowed. **This is the line I'd push back on** — not because the danger-avoidance was wrong, but because the mitigation went further than the danger required, in a direction the plan had already ruled out.

Concretely, the cost of the path taken: `runner.schema_migrations` still has no row for 054 (empty ledger, unchanged), so there is currently no checksum-backed confirmation that the committed file matches what ran — only internal consistency and a clean single-commit history (checked; see §4). The PR body already flags this as a "future `runner adopt`" step; I'd treat closing that loop as part of this work, not a someday-later item, given #1's own gate criterion is "constraint VALID in production" with no carve-out for how it got there.

## 3. Why this PR is currently BLOCKED, and why it's the same fact pattern

`gh pr checks 840` → `Test` fails, 10 of 2668. All ten trace to one cause: `tests/scripts/test_lineage_lane.py`'s boundary-derivation logic treats **every migration numbered above the schema-move file (051) as part of the ratified F.2/target-schema lineage**, by scanning the tree, not by reading intent:

```python
above = [m.path.name for m in discover_migrations(MIGRATIONS_DIR) if m.version > move.version]
assert above == [...]  # hardcoded list, doesn't include 054
```

054 is a legacy-lineage fix (it operates on `public.posting_history` as it exists *before* the cutover) that happens to be numbered after 053 because that's the next free number — but the test's derivation logic can't tell that from a file that just happens to sit above 051. Two consequences, both firing right now:

- `test_the_target_lineage_is_the_increments_that_have_been_ratified` fails because 054 isn't in the hardcoded list — the test doing exactly what its own docstring says it's for: forcing a human decision on every file that lands above the boundary, rather than silently absorbing it.
- `test_the_lane_replays_across_the_boundary` (6 sub-failures) and `test_migration_gate.py::test_replayed_schema_equals_models_schema` fail because when 054 **is** swept into that replay (the M.3-target-schema build-from-empty-`public` scenario), it errors: `relation "posting_history" does not exist` — because in that scenario `public.posting_history` has already been renamed to `legacy` by 051, exactly as branden's reasoning predicted.

**This is independent, automated confirmation of §1's danger** — not from reading `migration_runner.py`, but from the project's own CI actually replaying the corpus and hitting the exact failure a premature M.3 sweep would produce. It also means the PR is not safe to merge as-is: `mergeStateStatus: BLOCKED` is GitHub correctly enforcing this, and the PR body doesn't mention or address it. Whether the fix is renumbering, an explicit exclusion in the lineage-lane derivation, or something else is a real design call I won't presume to make — but it needs one before this merges.

## 4. Does the migration file match what ran?

Read the full SQL (`054_dedup_posting_history_and_unique_queue_item.sql`). Internally consistent with every claim in the PR body: archive-then-delete in one transaction (the `CREATE TABLE ... AS SELECT` runs before the `DELETE`, so a committed migration guarantees the archive predates the removal — not just asserted, structurally forced), the `ROW_NUMBER() OVER (PARTITION BY queue_item_id ORDER BY posted_at, created_at, id)` keep-rule matches the stated one, the partial index predicate matches. One commit only on this branch (`git log --all -- <file>`), no fixup/amend pattern visible. Migration numbering: `048_backfill_queue_delivery_states.sql` exists (confirmed, so 048 really was taken), the tree runs cleanly 047→054 with no gaps, so 054 is correctly the next number.

What I can't confirm — and said plainly rather than guessing — is a cryptographic match, because that's exactly what §2's gap leaves open: the runner computes a SHA-256 checksum per file (`hashlib.sha256`, `ledger_discrepancies` calls a checksum mismatch a hard failure) but only once a file is in the ledger, and 054 isn't yet. Right now "matches what ran" rests on prose + internal consistency, not on the mechanism this project built specifically to answer that question.

## 5. Archive: 13 rows, complete, readable

Cross-checked two independent sources rather than one: the pre-execution sign-off packet's rehearsal-branch numbers, and the PR's post-execution production numbers. They agree exactly — 13 rows, 6 groups, 7 surplus, and the same two groups (`125629e3` at 2/2 distinct story IDs, `e4e8589a` at 3/3) called out as the non-redundant ones both times. I re-summed the group table myself: 2+2+2+2+3+2 = 13, and surplus 1+1+1+1+2+1 = 7 — matches. The archive SELECT has no `rn`/rank filter — it grabs every row in any group with `count(*) > 1`, not just the ones about to be deleted, so completeness is a property of the query, not of the timing. It's a plain `CREATE TABLE AS SELECT` with a `COMMENT ON TABLE` describing it as read-only by convention — nothing in the file restricts read access.

## 6. Constraint: VALID, not merely created

Didn't need to trust the `t`/`t`/`t` row alone. The migration uses plain `CREATE UNIQUE INDEX`, not `CONCURRENTLY` — and `INVALID`-but-created is a failure mode that is *specific to* `CONCURRENTLY` in Postgres (the two-phase build it requires). A non-concurrent unique index inside a transaction that committed (`psql -1 -v ON_ERROR_STOP=1`, rc 0) cannot be left invalid — it's built and validated synchronously as part of the same statement, or the whole transaction aborts and nothing is created. So `indisvalid=t` isn't new information so much as confirmation of what the technique already guarantees; the real risk `CONCURRENTLY` would have introduced was avoided by construction.

## Smaller things checked

- **#839** — open, correctly scoped (the population is *three* media items, matching the sign-off packet's corrected 3-not-4 finding), correctly filed rather than bundled here.
- Row-count sanity: sign-off packet measured 4,316 rows at proposal time; PR's pre-migration count is 4,318 — a 2-row difference consistent with normal table growth between proposal and execution, and importantly neither new row landed inside one of the 6 target groups (both sources agree on 13/6/7 exactly), so the target set was stable.
- `src/models/posting_history.py`'s new `Index` matches the migration's index byte-for-byte (name, column, partial predicate). The dedicated integration test needs a real Postgres (partial index + `postgresql_where`, not something sqlite can stand in for) — didn't stand one up myself since CI already ran it for real and it's not among the 10 failures.

## Bottom line

Danger-avoidance: correct, and now doubly evidenced. Production: looks sound on everything checkable without touching it directly. But this can't ship as a green merge with CI red, the mechanism question deserves a real answer given the ratified plan already picked an answer that wasn't the one taken, and the ledger gap is worth closing rather than deferring. Requesting changes on the CI failure and the runner-mechanism question; not blocking on the ledger/`adopt` gap by itself, but I'd want it closed close to this, not indefinitely later.

chrisrogers37:
## Resolving the open thread above — rajan answered it, and the answer was closer than my hypotheses

The comment above flagged `adopt`'s *"unapplied tail — a gated apply runs it"* as the strongest evidence against my own claim, and listed three ways I might be wrong. Independent verification is in: **the hazard is confirmed** (a bare `apply` would sweep `051`), **and my conclusion from it was still wrong.**

A **bounded** `apply_pending(..., max_version=54)` call is available. It runs through the runner — satisfying #695's ratified "via the runner, not a hand-applied file" — *and* stops below the schema move. Both constraints, one door. The fork I posed was runner-versus-`psql`; the answer was neither.

**Against my ranked hypotheses:** I guessed the gate was operational (a runbook step or wrapper). It is *programmatic* — the parameter itself, on a standalone stdlib module whose function is directly callable. My hypothesis #1 was adjacent and one step short.

**The specific error, since "I missed it" is not accurate enough.** I did not fail to look. I read the `max_version` parameter and quoted it in this thread. I read `replayed_template` passing `LEGACY_LINEAGE_MAX` into this exact call and cited that function's derivation elsewhere in this same PR. I had every part. What I did was **let the CLI surface define the module's capability** — I filed a direct call as test scaffolding rather than as an available invocation, and once "the subcommand has no flag" was established I treated the door as closed.

**What actually stopped the search: I already had a working alternative.** The `psql` path was correct, available, and precedented by 49 prior migrations. That made an anomaly I had explicitly written down and could not explain into something tolerable instead of something blocking. An unresolved anomaly you can *name* is an open thread; I closed the task with it open because the workaround made it non-urgent. That is the part I would change, and it is a process fix rather than a knowledge one — the fix is "resolve the anomaly before shipping around it", not "know about `max_version`".

**Nothing here moves the lineage fork.** The bounded call is about *how* to apply, not about where the file lives. Options 1–3 stand as posted, production is unchanged and correct, and I am still holding.

The PR body above now carries this correction inline, and the two places it stated the false binary — the deviation rationale and the `/weigh-development-paths` note — are both fixed rather than only the one that was pointed at.

chrisrogers37:
## Measured: this PR's work is already done in production, and it can no longer land as written

Chris asked whether this is still relevant or closable. **Closable**, with a small piece worth salvaging. Evidence below, measured against production rather than read off the PR body. I did not force a rebase — see why in §2.

## 1. The remediation has already happened

Against the production database today:

```
dup_groups                        0        -- posting_history, queue_item_id NOT NULL, HAVING count(*) > 1
rows_total | with_qii        4642 | 4642

uq_posting_history_queue_item_id
  CREATE UNIQUE INDEX ... ON legacy.posting_history USING btree (queue_item_id)
  WHERE (queue_item_id IS NOT NULL)
```

So both halves of this PR's purpose are satisfied: **zero duplicates**, and the partial unique index it sets out to create **already exists**.

Worth flagging for whoever owns #695: `grep -rl uq_posting_history_queue_item_id scripts/migrations/` returns **nothing**. That index arrived outside the migration stream, so the ledger does not record how or when. It is present and correct; its provenance is not captured anywhere I can find.

## 2. It cannot be rebased into a working state

This PR inserts a new `050_dedup_posting_history_and_unique_queue_item.sql` and **renumbers every migration above it** — `050→051`, `051→052`, … `063→064`. That was coherent when 050 was the head of the tree. It is not now.

Production's ledger records those files under their **current** numbers:

```
adopted 050 (050_chain_reconciliation.sql)
applied 051 (051_schema_move_public_to_legacy.sql)
… through applied 065
```

I merged `main` into the branch locally to characterise it rather than argue from the diff. Only two paths conflict textually — both test files, no migrations — because git resolves the renumbering as renames. The result is a tree the runner **cannot parse at all**:

```
$ python scripts/migration_runner.py status
error: duplicate migration version 064: 064_memberships_for_caller_door.sql
                                    and 064_refresh_leg_provider_guard.sql
```

The shift collides with migrations that landed *after* this branch was cut. Making it applicable would mean renumbering the new migration to the end of the stream and reverting the renaming of ~20 migration files plus ~25 files of references to them — roughly 55 of this PR's 60 files. That is a rewrite, not a rebase, and it would produce a migration whose effect production already has.

**Also:** Chris has since ruled there is no data migration, which independently retires the dedup half.

## 3. What is worth keeping

`src/models/posting_history.py` declares `queue_item_id` as a plain column and **does not declare the unique index the database enforces**. That is a real model/schema parity gap, and this PR's model change closes it — with a comment explaining precisely what it backstops (`HistoryRepository.create_idempotent` is read-then-insert, so a replayed finalize can still race it).

That plus the two test files (`test_legacy_uniqueness_behaviour.py`, `test_posting_history_unique_queue_item.py`) is a small, self-contained PR with **no migration in it**.

## Recommendation

Close this, and if the parity gap is worth closing, open a fresh PR carrying only the model declaration and its tests.

**Bounds:** I verified duplicate count, index existence, ledger state, and the runner's response to the merged tree. I did not review the two conflicted test files' content, nor audit the rest of the diff for anything else worth salvaging — so "the model change plus two test files" is what I found, not a guarantee it is all there is.

chrisrogers37:
## Verdict: closable — the work is already done in production, and this PR structurally cannot land

Measured by `branden` against production and the merged tree, at the owner's request
("rebase and fix… is it even still relevant or closable?"). **No rebase was attempted**,
deliberately — see the second half, which is why a rebase would have been wasted work.

### 1. The remediation is already in place

- **Zero duplicate groups** in `posting_history`.
- **`uq_posting_history_queue_item_id` exists**, as a partial unique index.

So the outcome this PR was opened to produce is the state production is already in.

### 2. It cannot land, and rebasing does not fix that

The branch inserts a new `050_dedup_posting_history_and_unique_queue_item.sql` and
**renumbers every migration after it**:

| this branch | `main` |
|---|---|
| `051_chain_reconciliation` | **`050`**_chain_reconciliation |
| `052_schema_move_public_to_legacy` | **`051`**_schema_move_public_to_legacy |
| `053_shared_trigger_functions` | **`052`**_shared_trigger_functions |
| … | … |

Every renumbered file is **already applied in production** — `migration_runner.py status`
confirms 046–050 adopted and 051–063 applied — and `main` has since reached **065**. The
runner errors with `duplicate migration version 064` against the merged tree.

Renumbering applied migrations is not a merge conflict; it is a structural collision with
a ledger that has already recorded those versions. A rebase resolves text and leaves the
collision intact, which is why none was forced.

### 3. What is worth salvaging — roughly one small PR

**The model does not declare the index the database enforces.** `uq_posting_history_queue_item_id`
exists in production and is absent from the model, so the schema-of-record disagrees with
the schema-in-force. That is real drift, it is independent of everything above, and it does
not need this branch.

### Recommendation

Close this, and open the model-declaration fix separately. Leaving it open costs more than
it looks: it is a **conflicting** PR, so its `pull_request` workflows never schedule while
integrations keep reporting against the head SHA — the checks UI reads greener than the
branch is, and anyone reading it later inherits that.

chrisrogers37:
## Correction to my own salvage recommendation above

I said the salvage was "the model change plus two test files". **One of those files already exists on `main`** and I missed it:

```
tests/scripts/test_legacy_uniqueness_behaviour.py
  landed 2026-08-21 — "prove every legacy uniqueness rule refuses its duplicate
  AND permits its near-miss (#654)" (#951)
```

It is 26KB against this PR's 36-line version, and it is the better artifact. My recommendation should have been the model change plus **one** test file, not two.

That does not change the verdict — the PR is still closable for the reasons above — but the salvage is smaller than I stated, and I would rather correct it than leave a wrong number standing on a PR someone will act on.

**One thing the existing test does not cover:** `posting_history.queue_item_id` appears nowhere in it. It asserts "every legacy uniqueness rule", and this rule is not in the tree to be asserted — which is the real finding, and it is bigger than a missing model annotation. Details are going to ari rather than here, since this PR's disposition does not depend on them.

chrisrogers37:
## ⚠️ Do not merge — this branch would renumber 14 migrations that production has already applied

Triage of stale open PRs, 2026-09-01. This PR reads as an ordinary open PR with a
plausible title; it is not one. **It also turns out not to need rebuilding** — see
*What is salvageable* below, which is the more useful half of this comment.

Evidence read at `origin/main` = `017b59d`, and against the production
`runner.schema_migrations` ledger.

---

### Fact 1 — it renumbers 14 already-applied migrations

This branch inserts its file at `050` and shifts main's entire ladder up by one:

| version | on `main` | on this branch |
|---|---|---|
| 050 | `chain_reconciliation` | `dedup_posting_history_and_unique_queue_item` |
| 051 | `schema_move_public_to_legacy` | `chain_reconciliation` |
| 052 | `shared_trigger_functions` | `schema_move_public_to_legacy` |
| … | … | … |
| 064 | `memberships_for_caller_door` | `refresh_leg_provider_guard` |

Exactly **14** migrations change version number — main's `050` through `063`, each `+1`.

`scripts/migration_runner.py` records applied migrations as
`runner.schema_migrations (version INT PRIMARY KEY, checksum TEXT NOT NULL)` with
`checksum = hashlib.sha256(raw).hexdigest()` (`:186`), and refuses drift:

> `checksum mismatch for {name} — applied files are immutable; fix forward with a new migration, or record a deliberate exception with: runner repair --version {version} --reason ...` (`:370-379`)

**These versions are applied in production right now.** Verified by checksum identity —
production's recorded checksum equals the sha256 of main's file at that version:

| version | production checksum | matches `main` file |
|---|---|---|
| 50 | `945d742b90…` | `050_chain_reconciliation.sql` ✅ |
| 51 | `a9cd4464d8…` | `051_schema_move_public_to_legacy.sql` ✅ |
| 63 | `01cd088edb…` | `063_refresh_leg_provider_guard.sql` ✅ |
| 64 | `7d1f2bf935…` | `064_memberships_for_caller_door.sql` ✅ |

This branch's file at version `050` hashes to `76ab9d7540…`, against the `945d742b90…`
production has recorded. That is a mismatch on the first renumbered version, and 13 more
behind it.

### Fact 2 — the merged tree contains a duplicate version, `064`

`git merge-tree` reports only **2 text conflicts**, both test files, because rename
detection follows the shift. **The low conflict count is misleading — it reads as a
nearly clean merge and is not one.** The resulting tree contains both:

```
scripts/migrations/064_memberships_for_caller_door.sql
scripts/migrations/064_refresh_leg_provider_guard.sql
```

`discover_migrations` raises on sight:
`duplicate migration version 064: … and …`. This fails closed and loudly, which is the
one piece of good news here.

### Fact 3 — the target table has moved to another schema

Main's migration `051` is:

```sql
ALTER SCHEMA public RENAME TO legacy;
CREATE SCHEMA public;
```

`posting_history` is therefore **`legacy.posting_history`** now, and nothing in 052–066
recreates it in `public` (checked). This PR's SQL and both of its
`-- runner:postcondition` lines hardcode `public.posting_history` and
`schemaname = 'public'`.

Note this interacts with Fact 1: the schema move is *itself* one of the renumbered
migrations, so there is no ordering of this branch that is both correctly numbered and
pointed at the right schema.

---

## What is salvageable — **the work is already in production**

The dedup and the unique index **were applied to production on 2026-08-19 04:26:32 UTC**,
about ten minutes before this PR was opened, and they are live now:

```
legacy | posting_history | uq_posting_history_queue_item_id
  | CREATE UNIQUE INDEX … ON legacy.posting_history USING btree (queue_item_id)
    WHERE (queue_item_id IS NOT NULL)
```

- `legacy.posting_history_dedup_archive` exists and holds **13 rows across 6 queue items**;
  6 survivors remain in `posting_history`, i.e. **7 rows removed** — matching this file's
  own comment ("three of the seven removed rows carry their own distinct
  `instagram_story_id`") exactly.
- Current duplicate groups in `legacy.posting_history`: **0** (of 4,642 rows, all with a
  `queue_item_id`).
- This is already known to the repo: `scripts/migration_runner.py`'s sibling
  `scripts/m1_preflight.py:55` records `posting_history_dedup_archive` (#941) as
  *"created in production two days after the spec was written"*, and `CHANGELOG.md:445`
  notes it *"surfaced unprompted on the first production run"*.

**So there is nothing to rebuild, and re-running this migration would find zero
duplicates.** What is actually true is narrower and worth stating plainly:

1. **The effect is applied but unrecorded.** The dedup's checksum (`76ab9d7540…`) appears
   in **no** row of `runner.schema_migrations` — the ledger holds 65 contiguous rows,
   versions 1–65. It was applied outside the runner. A database rebuilt from the merged
   corpus would not have this index.
2. **The protection does not need carrying forward.** It guards `legacy.posting_history`,
   which is slated for `DROP SCHEMA legacy CASCADE`. The target model already carries a
   stronger guard — `uq_publish_exclusive` on `post_intents`, unique on
   `provider_account_ref` `WHERE state IN ('publishing','publishing_ambiguous')`
   (`src/models/target/intent_ledger.py:207`), which prevents the concurrent publish
   rather than deduplicating after the fact.
3. **The archive is the one thing with residual value.** Those 13 rows are the only
   surviving record that some stories genuinely published more than once, and they die
   with the legacy schema. `m1_preflight` already tracks this as
   `posting_history_dedup_archive.undispositioned_rows`, so a disposition step is
   expected — this PR is not where it belongs.

**Recommendation: close rather than rebase or rebuild** — but that call belongs to
whoever owns the queue work, not to triage, so nothing here has been closed or touched.

---

### Bounds on this comment

The production reads above are point-in-time (2026-09-01) and read-only: the runner
ledger, `pg_indexes`, and row counts. I did not modify anything, and I did not run the
migration anywhere. The claim that the 2026-08-19 application was out-of-band rests on
its absence from the runner ledger plus main's own `#941` note, not on a deployment log.

Full triage of this and three sibling stale PRs (#1136, #1144, #729):
`shared/knowledge/storydump/2026-09-01-stale-open-pr-triage.md` (fleet shared docs).

**Label request:** this repo has no `do-not-merge` / `blocked` / `hold` label — I checked
all 44 and none applies, so I have not invented one. Raising the gap rather than
improvising a label.

chrisrogers37:
## Closing — the work here is already in production, applied 2026-08-19

**This is not being closed as stale.** It is closed because what it proposes was **already done**, twelve days before this comment and about ten minutes before this PR was opened.

Measured against production, read-only:

- `uq_posting_history_queue_item_id` **exists** on `legacy.posting_history` — partial unique, `WHERE queue_item_id IS NOT NULL`, exactly as this PR specifies.
- `legacy.posting_history_dedup_archive` holds **13 rows across 6 queue items**, `archived_at` **2026-08-19 04:26:32 UTC**. Six survivors remain, i.e. **7 rows removed** — matching this file's own comment ("three of the seven removed rows carry their own distinct `instagram_story_id`") exactly.
- Current duplicate groups in `legacy.posting_history`: **0**, of 4,642 rows.
- It was applied **outside the migration runner**: the dedup's checksum appears in **no** row of `runner.schema_migrations`, which holds 65 contiguous rows, versions 1-65.

The repo already half-knew this. `scripts/m1_preflight.py:55` and `CHANGELOG.md:445` record `posting_history_dedup_archive` (#941) as *"created in production two days after the spec was written"* and *"surfaced unprompted on the first production run"*.

**Re-running this migration today would find zero duplicates.**

## The branch is also unmergeable, independently

Three blockers, any one of which is sufficient — detailed with evidence in [the earlier triage comment](https://github.com/chrisrogers37/storydump/pull/840#issuecomment-5493985245):

1. It inserts at `050` and **renumbers 14 already-applied migrations** (main's `050`-`063`, each `+1`). Production's recorded checksums match main's files at those versions; this branch's file at `050` hashes to `76ab9d7540…` against the recorded `945d742b90…`. The runner enforces *"applied files are immutable"*.
2. The merged tree carries a **duplicate version `064`** — `discover_migrations` raises on sight. Note `git merge-tree` reports only 2 text conflicts, both test files, because rename detection follows the shift: it reads as a near-clean merge and is not one.
3. Its SQL and both `runner:postcondition` lines hardcode `public.posting_history`, but main's `051` is `ALTER SCHEMA public RENAME TO legacy`. The table is `legacy.posting_history` now, and the schema move is *itself* one of the renumbered files — so no ordering is both correctly numbered and correctly aimed.

## What is worth rebuilding: nothing here, and here is why that is safe

The protection does **not** need carrying forward. It guards `legacy.posting_history`, which is slated for `DROP SCHEMA legacy CASCADE`, and the target model already carries a **stronger** guard — `uq_publish_exclusive` on `post_intents`, unique on `provider_account_ref` `WHERE state IN ('publishing','publishing_ambiguous')` (`src/models/target/intent_ledger.py:207`). That prevents the concurrent publish outright rather than deduplicating after the fact, which is the better shape of the same protection.

Two things **do** survive this closure and are tracked elsewhere:

- **The archive rows.** Those 13 rows are the only surviving record that some stories genuinely published more than once, and they die with the legacy schema. `m1_preflight` already tracks this as `posting_history_dedup_archive.undispositioned_rows`, so a disposition step is expected — it does not belong in this PR.
- **The reconciliation gap.** An object live in production, in no ledger row and in no repo migration, is the exact shape of #1195. The scheduled drift monitor from #1197 does **not** catch this particular one — it compares `public` only, and these objects are in `legacy` — which is stated in its bounds rather than left for someone to discover.

## Why closing, specifically

This repo has **no `do-not-merge` / `blocked` / `hold` label** — all 44 checked, none applies, and I have not invented one. Closing is therefore the only mechanism available to keep an unmergeable PR from being read as ordinary open work.

**The branch is NOT deleted** and nothing in production was touched. If any of this is wanted after all, it reopens.

Full evidence: `shared/knowledge/storydump/2026-09-01-stale-open-pr-triage.md` and `shared/knowledge/storydump/2026-09-01-migration-066-unapplied-and-runner-state.md`.

PR #844, opened 2026-08-19
F.2.3: land 02 §2's accounts, sources and media tables (#806)

chrisrogers37:
Advertised-stream statements **22..44**, in stream order, continuing 053's `2..21` and 052's `0..1`. Six tables — `ig_accounts`, `provider_quarantine`, `media_sources`, `oauth_credentials`, `media_items`, `post_locks` — with six touch triggers and eleven partial/plain indexes. **F.2.3 only**; nothing from F.2.4 is touched.

Scope taken from the ratified split (`documentation/planning/2026-08-14-f2-increment-split`), not inferred from table names. Its §2 row for F.2.3 reads `22..44 / 23 stmts / 6 tables`, and the indices are 0-based there (052 is `0..1`, F.2.2 is `2..21`), which matches the measured lineage: 22 statements landed, next unlanded index 22.

## The gate

```
arm (b): F2PrefixReport(ok=True, vacuous=False, f2_count=45, first_divergence=None)
```

45 = 22 + 23. Non-vacuous, so it is not passing by having nothing to compare.

**The body is generated from the stream, not transcribed** — the same rule 053 set. The segment was extracted through the repo's own extractor and asserted at write time to normalize back to exactly those statements; the only hand-written part is the leading comment block, which the normalizer strips. **The manifest ratchet is untouched** — no plan SQL block was edited, so nothing falls out of the sha256 bijection.

## Models land in the same increment, and that is measured rather than assumed

This is the one place the increment is larger than "migration files". Lane parity compares the replayed `public` against `create_all` on `TargetBase`, so a migration-side table with no model behind it **is** the drift it reports. `src/models/target/__init__.py` states the doctrine outright ("Models land per increment, not in one pass — the gate settled it rather than taste"), and commit `012c9df` shipped 053 with `identity_and_tenancy.py` in one increment.

Verified as a positive control rather than argued:

| state | `test_lane_parity_holds_against_the_target_models` |
|---|---|
| migration 054 alone | **FAILED** |
| + `accounts_sources_media.py` registered | **passed** |

So the models are not scope creep — without them this increment lands the lane red.

## What is deliberately absent

- **No policies, no RLS**, per Fork 1 ruling (a) and split §3: F.2.2–F.2.5 land tables, triggers and indexes; RLS and policies for all of them land together in **F.2.7**. All six tables here are tenant-keyed and carry none until then. The lineage now holds **13 tables**, which is exactly what split §4 predicts for this increment.
- **No FK on `post_locks.created_by_intent_id`** — `post_intents` lands in F.2.4, so referencing it here would not be a stream prefix. The plan declares none and the model does not invent one.
- **No triggers mirrored in the models.** The parity comparator is relation-scoped and reads none, so restating the six `tg_touch_*` triggers would be an unchecked second copy.

## Composite foreign keys are this increment's shape

Four of the six FKs are composite `(workspace_id, <id>)` references — `oauth_credentials` to `ig_accounts` and `media_sources`, `media_items` to `media_sources`, `post_locks` to `media_items` and `ig_accounts`. That is the plan's tenant-integrity device: a cross-workspace row becomes unrepresentable in the schema rather than merely forbidden by policy, which is why each parent carries a `UNIQUE (workspace_id, id)` existing solely as their target. A single-column rendering would still build a working schema and would be a **red parity test** — worth a reviewer's eye.

`provider_quarantine` is the one table with no surrogate key: its PK is the composite `(workspace_id, scope_ref)`, so it does not call `pk()`.

## Postconditions

Scoped to objects 054 itself creates, **none asserting absence** — the rule 053 adopted after a review catch. The runner derives a file's permanent adoption probe from these lines, so a negative postcondition would read *unapplied* at M.3 once F.2.7 adds RLS and policies to these same tables.

## Verification

- `tests/scripts/` — **202 passed, rc=0**: arm (a) replay, arm (b) prefix, the lineage lane (all 15, incl. parity), tenancy gate, migration gate, manifest ratchet. Run unsandboxed against a live local Postgres.
- `ruff check src/ cli/ tests/` → clean; `ruff format src/ cli/ tests/ --check` → 313 files already formatted. Formatting was applied **only** to the new module, not swept over the directory.
- Statement census of 054: 6 `CREATE TABLE`, 6 `CREATE TRIGGER`, 11 indexes = 23, matching the split's count.
- Two-dot and three-dot diffs against `origin/main` are identical — no stale-base cargo; base is current.

**Production is untouched.** Target-side schema only — no production run, no deploy. Data arrives at M.3.

Refs #806


chrisrogers37 (commented):
**Approve**

Verified all four attack points by running, not by reading — same checkout I have fresh context on from tonight's #753 investigation.

## 1. `vacuous=False` — independently reproduced from scratch, not trusted from the PR body

`f2_prefix_report` only sets `vacuous=True` when `f2_statements` is empty (`scripts/advertised_ddl.py:292-293`). Re-derived the whole computation myself outside the test suite — my own re-implementation of `target_lineage_files`, loading the real manifest/docs, reading the real migration files:

```
lineage files: ['052_shared_trigger_functions.sql', '053_identity_and_tenancy_tables.sql', '054_accounts_sources_media_tables.sql']
f2 statement count: 45
report: F2PrefixReport(ok=True, vacuous=False, f2_count=45, first_divergence=None)
```

Matches the PR body exactly. Also ran the actual integration test this number comes from, `TestF2AgainstReality::test_the_wired_prefix_holds_against_the_real_stream` — passes, and its own docstring names the exact failure mode you flagged ("deleting every target-lineage migration would turn this green again... which is the failure the disclosure existed to prevent").

## 2. Models-needed claim — verified BOTH ways myself, not just alex's table

Reverted `src/models/target/__init__.py` to pre-PR (migration file stays, model import removed) and re-ran `test_lane_parity_holds_against_the_target_models` against real Postgres:

- **Migration alone (models unregistered): FAILED** — drift is exactly the 6 new tables, "exists only in the first schema" for each of `ig_accounts`/`media_items`/`media_sources`/`oauth_credentials`/`post_locks`/`provider_quarantine`.
- **Restored (models registered): PASSED.**

This is a real, load-bearing dependency, not scope creep — confirmed the failure mode matches the docstring's claim precisely (a migration-side table with no model behind it *is* the drift the comparator reports). Worth flagging forward per alex's own framing: this settles that F.2.4–F.2.9 each need migration+models together, not migration-only with models deferred.

## 3. No F.2.4 drift — confirmed against the ratified split doc, not just internal consistency

`documentation/planning/2026-08-14-f2-increment-split/README.md` §2: F.2.3 = statements `22..44`, 23 stmts, 6 tables (`ig_accounts`, `provider_quarantine`, `media_sources`, `oauth_credentials`, `media_items`, `post_locks`). F.2.4 = statements `45..76`, 5 tables (`category_post_case_mix`, `post_intents`, `audit_events`, `daily_post_counts`, `post_intent_transitions`). Grepped `054_accounts_sources_media_tables.sql` directly for every `CREATE TABLE`/`TRIGGER`/`INDEX` — exactly the 6 F.2.3 tables, 6 triggers, 11 indexes = 23 statements, zero overlap with F.2.4's 5. `post_locks.created_by_intent_id` correctly carries no FK to the not-yet-landed `post_intents`.

## 4. The #840 collision — structurally can't land silently, verified two independent ways

Confirmed #840 is still open with `054_dedup_posting_history_and_unique_queue_item.sql`. But this isn't exploitable quietly:

- **`discover_migrations` (`scripts/migration_runner.py:174-178`) hard-fails on any duplicate version number** — raises `MigrationRunnerError` naming both colliding filenames, the instant a second `054_*` file is discovered. This is the foundation every other check (arm a/b, tenancy gate, lane parity, the production runner itself) is built on, so the collision can't slip past any of them.
- **Independently, `TestTheBoundaryIsDerivedAndLoud` asserts the exact ordered filename list** (`052...`, `053...`, `054_accounts_sources_media_tables.sql`) — a second tripwire that breaks the instant the directory holds anything #844 didn't put there.

So whichever of #840/#844 merges second gets an immediate, loudly-named CI failure forcing a renumber — never a silent pick-one. Nothing in #844's code, tests, or CHANGELOG assumes exclusive ownership of "054" beyond the filename itself, and the filename is exactly as contestable as its file-system slot already makes it.

## Also ran

Full `tests/scripts/test_lineage_lane.py` + `test_advertised_ddl_replay.py` + `test_advertised_ddl.py` (43 tests) against live Postgres on this exact branch — all pass, unsandboxed.

Clean work. Ship it.

PR #845, opened 2026-08-19
F.2.4: land 02 §3's intent ledger (#806)

chrisrogers37:
Advertised-stream statements **45..76**, in stream order, continuing 054's `22..44`. Five tables — `category_post_case_mix`, `post_intents`, `audit_events`, `daily_post_counts`, `post_intent_transitions` — with 27 seeded legal-edge rows, four trigger functions and eleven triggers. **F.2.4 only**; nothing from F.2.5.

Scope taken from the ratified split's §2 row (`45..76 / 32 stmts / 5 tables`), not inferred from the name. Verified against the corpus before writing: 45 statements landed, so index 45 is next.

## The gate

```
arm (b): F2PrefixReport(ok=True, vacuous=False, f2_count=77, first_divergence=None)
```

**`vacuous=False`** — 77 = 45 + 32, so it is comparing 77 real statements, not passing over an empty set. Lineage now at **18 tables**, exactly split §4's prediction; 180 stream statements remain.

Body **generated from the stream, not transcribed** — extracted through the repo's own extractor and asserted at write time to normalize back to exactly statements 45..76. Only the header is hand-written, and the normalizer strips it. **Manifest ratchet untouched** — no plan SQL block edited, so nothing leaves the sha256 bijection.

## Numbering — 055, and why

`054` is F.2.3, merged as #844. **PR #840 also proposes a `054`**, for a legacy-lineage change, and is blocked pending Chris's lineage ruling. Worth stating plainly: #840 collides with the merged 054 **independently of this PR** and must renumber either way. If its ruling places it below the 051 move, this file is unaffected; if above, it takes the next free number after 055. `discover_migrations` hard-fails on duplicates, so no collision can land silently.

## This increment carries DATA — the first one that does

Statement 64 seeds `post_intent_transitions` with 27 legal edges. Not fixture rows: `trg_intent_guard` reads the table on every UPDATE, so an **unseeded schema rejects every transition while still replaying green** — the pass-5/R4 finding the seed exists to close. `advertised_ddl_replay` asserts one legal and one illegal transition against these rows, so the matrix text and the seeds cannot drift apart.

It also **attaches triggers to earlier increments' tables**: the five `tg_audit_*` governance triggers land on `workspaces`, `workspace_members`, `channel_bindings` (053) and `ig_accounts`, `oauth_credentials` (054). That is the stream's order rather than a liberty taken here — `audit_events` must exist before anything writes to it, which is exactly why the reshape moved it into this increment (split §3).

## Two asymmetries mirrored deliberately — the reviewer's eye goes here

1. **`audit_events.workspace_id` carries no foreign key.** It is the one tenant column in the increment that is a bare `UUID NOT NULL`. An audit row must outlive the workspace it describes; a CASCADE would delete exactly the evidence of the deletion. Adding an `fk()` would be a red parity test *and* wrong.
2. **`post_intents`' two composite FKs carry no `ON DELETE`** — unlike every composite FK in 054. Under default `NO ACTION`, a delete that would strand a live intent is **refused** rather than silently cascading the ledger away. `daily_post_counts` *does* cascade, because a count is derived rather than evidentiary.

Both are the plan's shape. Normalizing either would build a working schema and show up only as red parity.

## The 053 JSONB trap, avoided rather than rediscovered

`post_intents.attempts_by_step` defaults to `'{"v":1}'`. Written naively, SQLAlchemy's `text()` reads `:1` as a **bind parameter** and emits `DEFAULT '{"v"NULL}'`, so the table will not create. The parity gate never compares defaults, so it fails at `create_all` rather than as drift. Backslash-escaped here, as `channel_bindings.settings` already is.

## Models, scoped in from the start

Per the rule F.2.3 established and navi reproduced: an F.2 segment is migration **+** target models. `src/models/target/intent_ledger.py`, registered in `__init__.py`. Three of the five tables have no `updated_at` — `audit_events` and `post_intent_transitions` are §0 insert-only classes with no touch trigger — so `timestamps()` is deliberately not uniform across the module.

Parity passed on the first run this time, rather than after discovering the models were needed.

## Verification

- `tests/scripts/` — **202 passed, rc=0**: arm (a) replay, arm (b) prefix, lineage lane (all 15 incl. parity), tenancy gate, migration gate, manifest ratchet. Unsandboxed, live local Postgres.
- `ruff check src/ cli/ tests/` clean; `ruff format --check` → 314 files already formatted. Format applied only to the new module, not swept.
- Statement census of 055: 5 `CREATE TABLE`, 4 `CREATE FUNCTION`, 11 `CREATE TRIGGER`, 11 indexes, 1 `INSERT` = **32**, matching the split.
- Two-dot and three-dot diffs identical; base is current (`origin/main` moved 0 commits during the build).

**Production untouched** — target-side schema only, no run, no deploy.

Refs #806


chrisrogers37 (commented):
**Approve**

Same four attack points as #844, run live again rather than trusted — plus the new numbering check and two extra claims the PR body itself flagged for scrutiny.

## 1. `vacuous=False`, `f2_count=77` — re-derived from scratch, with a per-file breakdown #844 didn't need

```
per-file: {052: 2, 053: 20, 054: 23, 055: 32}
TOTAL: 77
report: F2PrefixReport(ok=True, vacuous=False, f2_count=77, first_divergence=None)
```

052+053+054 = 45, byte-identical to what I measured independently on #844 — confirms the old files aren't being re-counted. The entire +32 delta is 055's own content. Directly answers your check: this is new statements, not a re-count.

## 2. Models included — confirmed present, AND confirmed load-bearing (not just claimed)

`src/models/target/intent_ledger.py` exports `AuditEvent`, `CategoryPostCaseMix`, `DailyPostCount`, `PostIntent`, `PostIntentTransition` — all 5, registered in `__init__.py`. Ran the same toggle as #844, both directions:

- **As-shipped: PASSED.**
- **Reverted `__init__.py` to pre-PR (migration stays, F.2.4 models unregistered): FAILED** — drift is exactly `audit_events`, `category_post_case_mix`, `daily_post_counts`, `post_intent_transitions`, `post_intents`. Five tables, no more, no less.
- **Restored: PASSED again.**

The failure mode you named (claiming models that aren't really needed) doesn't apply — removing them breaks parity with precisely the right signature, so they're real and required, not decorative.

## 3. No F.2.5 drift

Split doc §2: F.2.4 = `45..76`, 32 stmts, 5 tables (`category_post_case_mix`, `post_intents`, `audit_events`, `daily_post_counts`, `post_intent_transitions`). F.2.5 = `77..93`, 17 stmts, 5 *different* tables (`jobs`, `channel_outbox`, `provider_operations`, `command_dedup`, `rate_counters`). Grepped `055_intent_ledger_tables.sql` for every `CREATE`/`INSERT` — 5 tables, 4 functions, 11 triggers, 11 indexes, 1 insert = 32, zero overlap with F.2.5's set. The 5 `tg_audit_*` triggers attach to *earlier* increments' tables (053/054) — correct, not forward drift.

## 4. NEW — numbering: 055 confirmed genuinely free, no `054`-ownership assumption

`ls scripts/migrations/*.sql` on this branch shows exactly one `055_*` file and one `054_*` file (your merged #844) — no collision on disk. Every `054` reference in this PR's diff is explanatory only (citing what 054 already contains, and the numbering-collision explanation itself) — grepped the full diff, nothing depends on or modifies 054. Same two independent tripwires as #844 still apply: `discover_migrations` hard-fails on any duplicate version number, and the exact-ordered-list test (`test_lineage_lane.py`) now asserts all four files by name, breaking loudly if the directory ever holds anything this PR didn't put there. #840 re-checked live — still open, still blocked, unchanged since #844's review.

## Two extra claims worth independently confirming, since the PR body invited scrutiny there

- **`audit_events.workspace_id`** — read the table def directly: genuine `UUID NOT NULL` with no FK, comment matches ("§0 exception: audit outlives the tenant").
- **`post_intents`' composite FKs vs `daily_post_counts`** — confirmed in raw SQL: `post_intents`' two FKs carry no `ON DELETE` (defaults to refuse); `daily_post_counts`' FK explicitly carries `ON DELETE CASCADE`. Matches the claimed asymmetry exactly.
- **JSONB escape trap** — `intent_ledger.py:124`: `server_default=text(r"""'{"v"\:1}'""")`. Checked the claimed precedent rather than trusting it: `identity_and_tenancy.py:307`'s `channel_bindings.settings` uses the byte-identical escape. Real precedent, correctly reused.

## Also ran

Same 43 tests as #844 (`test_lineage_lane.py` + `test_advertised_ddl_replay.py` + `test_advertised_ddl.py`), live Postgres, this branch — all pass.

Same quality bar as #844. Ship it.

Issue #841, opened 2026-08-19
[Refactor Phase F] F.1 — build landed (#846, #961, #962); residual burn-down: keyword-only tenant params 22/72, 8 unclassified `if chat_settings_id` sites, web-boundary dependency #869


chrisrogers37:
> Increment F.1 of the ratified execution sequence ([`04-execution-sequence.md` §F.1](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md); absorbs TT:P0-02). Parallel-legal with F.2 per `04` ("F.1 ∥ F.2"); F.6 sequences after this.

## Current state — half done, and the halves must not be confused

**The inventory + interface spec half is DONE** — PR #747 (merged 2026-08-12, `d4f6aea3`, docs-only): all 14 legacy tables classified (9 tenant-owned · 2 global user-plane · 3 no-target), answer key established as **§9 ∘ §7-DDL** (the composition — #748 tracks the `04` line's mis-cite), and the fail-closed interface spec written class-keyed at [`documentation/planning/2026-08-11-f1-ownership-inventory/README.md`](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-11-f1-ownership-inventory/README.md) §§2–3. **That document is this issue's spec; do not re-derive it.**

**The build half is NOT started** — measured at `main@3e5b74a`: the fail-open pattern `04` orders extinct is live in 4 repository files (13 sites), including its canonical form at `src/repositories/base_repository.py:255-256` — *"Apply tenant filter if chat_settings_id is provided. **No-op when None**"* — plus conditional-filter variants in `media_repository.py` (66, 692, 703, 901), `token_repository.py` (253, 449), `history_repository.py` (710). No repository method requires leading tenant context. No fail-closed tests exist. No later PR builds any of it (searched merged PRs; only #747/docs).

## Scope (from the spec — pointers, not re-design)

The three rules of spec §3, **keyed to ownership class, not to table** (§2's finding — `chat_settings` splits across classes, and a uniform rule breaks onboarding):

1. **Class 1 (tenant-plane) + Class 6 (audit):** every repository method takes `tenant_id: UUID` as its first positional parameter — no default, no `Optional`, no keyword fallback. Covers the 9 tenant-owned tables.
[...]

PR #846, opened 2026-08-19
F.1 build (#841): fail-closed tenant access — None refused at the boundary, SYSTEM_SCOPE is the explicit cross-tenant door

chrisrogers37:
Builds #841 (F.1, `04` §F.1) per the merged #747 spec. **The mechanism ships tonight with zero behavior change by construction; the deliberate tenant-narrowing conversions become an enumerated, greppable burn-down.**

## The per-site shape analysis (asked for before any uniform fix — and they do NOT share a shape)

The "13 sites" in #841 re-measured to **9 literal `if chat_settings_id` sites at HEAD** (the 13 came from a looser grep; #841 ordered a re-grep at build time). Five distinct shapes, so no uniform mechanical fix was applied:

| Shape | Sites | What it is | Fix |
|---|---|---|---|
| A — root conditional filter | `base_repository._apply_tenant_filter/_tenant_query` | None → unscoped query; every `_tenant_query` caller inherits it | Chokepoint refuses None (`TenantContextError`); `SYSTEM_SCOPE` widens explicitly |
| B — conditional subquery scopes | `media` 692/703 (eligibility exclusions) | follow the outer method's tenant param; an adjacent hash-exclusion subquery is **deliberately tenant-global** (documented) and was NOT touched | outer param now required; branches unchanged (see falsy note) |
| C — logging guards on the sanctioned NULL-owned fallback | `media` 66/901 | observability on the documented pre-#412 legacy path, inside `_write_allowed`-guarded mutators — not query wideners | kept; the mutator entry now requires context |
| D — display ternary | `history` 710 | a GROUP-BY **row value**, not caller context — matches the grep by name only | untouched, and the extinction gate is signature-level precisely so this can never false-positive |
| E — conditional ownership stamp | `token` 253 (upsert assignment), 449 (bulk revoke filter) | write-path fail-open | 253 flipped to truthy (system scope must never stamp a column); 449 requires context |

**The design key: `SYSTEM_SCOPE` is deliberately falsy.** Every pre-existing `if chat_settings_id:` branch treats it exactly as it treated None — so converting a call site from silent omission to explicit `SYSTEM_SCOPE` is behavior-preserving *by construction*, and the surviving branches are now the legitimate SYSTEM-vs-tenant switch, unreachable by None (entry guards + chokepoint).

## What changes behavior — named, with the caller evidence

**Blast radius measured by AST (crude grep first said 247 sites; receiver-blind and multiline-blind — the honest number):** 164 resolved repo calls in `src/`+`cli/`, **78 omitted tenant context**, across 21 files. Plus ~140 test call sites. The population that "depends on the swallowed failure" is: **every worker loop, maintenance sweep, health check, CLI command, and a set of telegram handler paths** — the legacy app's system paths are tenant-less by design (the documented `_write_allowed` internal path).

**In this PR: nothing changes behavior**, by three constructions:
1. All 78 omitting sites → explicit `chat_settings_id=SYSTEM_SCOPE` (global before, global now — visibly).
2. Row-construction sites write ownership via `tenant_value()` — system scope stamps NULL, exactly what omission stamped.
3. The only tests whose EXPECTATIONS changed are the four that asserted the fail-open contract itself (`test_apply_tenant_filter_without_id` et al.) — flipped to assert the fail-closed one, which is the behavior-change record in test form.

**Deliberate narrowing (real behavior change) is deferred to the #841 burn-down, one reviewable step each:** the ~21 telegram-handler sites that HAVE chat context (narrowing them is the #597-class security fix, but each needs per-site verification that no legitimate cross-chat flow breaks), and the OAuth `token.create_or_update` upsert (5 callers — per-chat flows that should stamp ownership and today do not). `grep -rn SYSTEM_SCOPE src/ cli/` is the standing inventory.

## The contract

- 78 repository methods across media/history/lock/queue/category-mix/token/audit: tenant param **required** — 47 in place (default dropped), 31 keyword-only (their param sat mid-defaults; keyword-only is the one form that makes them un-omittable without reordering ~123 correct call sites, and un-swappable into adjacent str params as a bonus). **Deviation from the spec's "first positional parameter" letter, stated:** moving the param to leading position would rewrite every correct caller for zero runtime gain and create silent str-swap hazards; required-and-un-defaulted is the load-bearing half of the rule, and both forms TypeError on omission.
- Explicit `None` raises `TenantContextError` at the boundary — proven to fire **before any SQL** by a poisoned-session test.
- Class-3 (user-plane) repositories reject tenant context (`users`, `onboarding`) — the negative direction, so the classes cannot converge.
- `membership`/`chat_settings` repos already required real context everywhere — untouched. `interaction`/`instagram_account` repos carry no tenant params at all (keyed by chat/user directly); their cross-tenant coverage is named residual on #841, not silently absorbed.

## Gate evidence (the four #747 §3 obligations, `tests/src/repositories/test_f1_fail_closed.py`)

1. **Boundary failure per repo** — omission = TypeError (5 repos parametrized); explicit/kw-only None = TenantContextError before any query (8 cases, poisoned-session proof).
2. **Cross-tenant zero rows on a real database** — two real `chat_settings` tenants; media/queue/lock/category-mix rows created under A are invisible under B, visible under A.
3. **Class-3 rejects tenant kwarg** — TypeError on `users` and `onboarding` repos.
4. **Extinction, mutation-proven** — signature-level scanner (so shape D can never false-positive) over `src/repositories/`; zero hits with two documented data-field allowances (onboarding `pending_chat_settings_id`, `HistoryCreateParams.chat_settings_id` — params-object stamps, scoped to their blocks); the scanner is shown to FAIL on a synthetic reintroduced instance.

**The 78 reconciled** (round 2, branden's minor): 78 pre-sweep omitting caller sites = 57 surviving explicit SYSTEM_SCOPE passes + 3 double-passes removed (real tenant kept) + 6 dead factory kwargs removed + 11 sweep mis-attributions reverted (calls to methods that take no tenant param) + 1 correct-wrapper revert. The census's 64 = those 57 + 7 repo-internal by-identity sibling fetches. The arithmetic closes exactly; the census failure message prints the per-file breakdown, so the burn-down inventory is readable straight from the gate.

## /simplify round (4 parallel review agents) — what it caught and what was applied

**Three real defects the suite structurally could not see** (mocked repos bind anything), each found independently by 2–4 agents and fixed:
- `media_sync` ×3: the sweep appended the kwarg to calls that already passed the tenant **positionally as an attribute** (`ctx.chat_settings_id` — the sweep's omission detector checked bare names only) → would have raised `TypeError: got multiple values` on every live sync. The real per-tenant scope was kept; the kwarg dropped. A tree-wide AST audit for the positional+keyword double-pass shape confirms zero remaining.
- Factory ×6: `MediaSourceFactory.create(**kwargs)` silently swallowed `chat_settings_id` — six dead arguments implying tenant-awareness that doesn't exist. Removed (the factory's `**kwargs` is its own fail-open surface, noted for the burn-down).
- Guard redundancy: 68 entry guards deleted where the chokepoint provably covers them (`_tenant_query` now validates **before** session checkout, so a refused call never touches a connection); ~15 load-bearing guards kept (row-construction, raw-query, `_get_for_write`). The poisoned-session gate test pins the before-any-checkout property.

**Also applied:** `require_tenant_id` for the four mandatory-tenant methods (`*_for_chat`, per-instance audit) — they now refuse SYSTEM_SCOPE too, since the marker would otherwise bind into SQL; singleton `__new__` machinery dropped (isinstance is the contract); `TenantContextError` inherits `StorydumpError` per the stated convention; the DB-less gate obligations (1, 3, 4) un-marked from `integration` so the durable gates cannot silently skip without Postgres; the shared DB-routing fixture + tenant helpers promoted to `tests/conftest.py` (this file uses them; migrating the five pre-existing copies is follow-up); a **SYSTEM_SCOPE census pinned by equality** (round 2 — branden: the first version counted substrings, so 24 of its 88 were imports and prose, and its floor band left a 20-wide slack corridor a new cross-tenant site could ride in on). Now: AST call-argument count only (64), **equality not ceiling** — chosen over re-baselining the substring count because the two fail differently: a ceiling accrues headroom as burn-down retires sites, so a later pin adjustment can pre-buy unreviewed access, while equality makes every delta red until the pin moves in the same diff as the site that justifies it. Mutation-proven both directions (planted site → red at 65 naming itself; clean → green).

**Skipped, with reasons:** keyword-only for all 91 tenant params (right end-state — makes future sweeps machine-verifiable — but ~23 positional call sites of churn; queued on #841); mass `tenant_value()` respelling of the 9 falsy branches (behavior-preserving today; noted); a `_get_for_write` analog for `QueueRepository`'s five by-identity mutators (real gap, but it IS the narrowing work — burn-down); mypy-over-repositories CI step (worthwhile, its own decision); promoting the five pre-existing fixture copies (outside this diff); findings in files this PR never touched (the agents' `main...HEAD` picked up a stale local `main`).

## Validation

- `tests/src/repositories/`: **355 passed, 4 skipped** (was 336+4 pre-gate-tests; the four contract-flip tests updated as described).
- Full suite, locally against real PostgreSQL, twice: **2676 passed / 8 skipped** before the review round (proving the sweep), and the post-review run: **2677 passed / 8 skipped**, sole failure the new SYSTEM_SCOPE ratchet catching its own author one over the ceiling — pinned at the measured 88 and re-run green. CI is the final arbiter.
- Production untouched: no schema change, no migration, no deploy config; the runner/predeploy path is not involved.

Closes nothing by itself: #841 remains open carrying the burn-down (SYSTEM_SCOPE conversions; interaction/instagram_account coverage). This PR is the mechanism + the zero-change sweep + the gate.


chrisrogers37:
**Request Changes** — one narrow blocker in the ratchet, which does not measure what it claims. The zero-behaviour-change claim itself **holds**: I disaggregated it per shape and found no surviving fail-open and no hidden narrowing.

Reviewed as: *does converting fail-open to fail-closed preserve behaviour at all 78 sites, or only where the omission could never fire?*

---

## The construction holds — checked per shape, not as a uniform claim

The whole argument rests on **SYSTEM_SCOPE being falsy, so every pre-existing `if chat_settings_id:` branch treats it as it treated None.** That is only sound if every consumer of the param tests truthiness. Consumers that use *identity* (`is None`), or use the value directly, would flip.

I searched for the non-truthiness consumers. Six exist:

| site | what it is | verdict |
|---|---|---|
| `media:73`, `media:910`, `telegram_service:508` | `row.chat_settings_id is None` — a **row value** | safe |
| `media:46` `_write_allowed(owner_...)` | `owner_chat_settings_id` is the **row's** owner from the DB, never the marker; caller scope is tested at `:45` with `if not chat_settings_id` | safe, and SYSTEM_SCOPE → permitted, identical to omission |
| `history:715` | **shape D** | **confirmed a false positive** — `chat_settings_id` there is bound by `for chat_settings_id, total, ... in rows`, a GROUP-BY loop variable shadowing any outer name. Your classification is right. |
| `onboarding:80` `pending_chat_settings_id` | a **data field** written to `session.pending_chat_settings_id` | safe; the scanner allowance is legitimate |

**No consumer breaks the falsy equivalence.** Spot-verified the other shapes too: chokepoint (`require_tenant_context` then `if chat_settings_id:` → SYSTEM_SCOPE takes the identical no-filter path); shape E `token:253` (`if chat_settings_id:` truthiness, so the marker never stamps — matches omission); `token:449` guarded; `tenant_value(SYSTEM_SCOPE) is None`.

## No fail-open survivor, and no hidden narrowing

AST over all **91** methods taking a tenant param: 59 via chokepoint, 16 explicitly guarded, 17 neither. Of those 17 I chased every one to ground — 9 delegate to a guarded callee (`_get_for_write`, `get_current_mix`, `get_active_lock`), `_write_allowed` is a pure predicate, and the remaining 4 (`chat_settings.get_by_id`, `membership.get_for_chat`/`get_membership`/`deactivate_for_chat`) use an **unconditional** `filter(... == chat_settings_id)` with no widening branch — they cannot fail open. **Zero survivors.** *(Your "already required real context everywhere" is slightly generous — they don't require it, they just can't widen — but the security property is what you claimed and it holds.)*

Separately I checked for narrowing smuggled into a zero-change PR: every added tenant kwarg in `src/`+`cli/` whose value is not SYSTEM_SCOPE. Six hits, all `tenant_value(chat_settings_id)` row-construction — the documented behaviour-preserving form. **Nothing narrows.**

## The mock-blind class — independently confirmed

This is where the suite is not the oracle, by your own account, so I did not take the 2677 as settling it. Re-derived the positional+keyword double-pass audit from scratch, resolving each call by **receiver** to the right repo signature: **164** keyword tenant call sites, **148** receiver-resolved, **0 double-pass**.

Two notes on my own method, because a zero is only worth its control: my first pass resolved by method name alone and produced 26 false flags (`get_by_id` sits at index 0 on `ChatSettingsRepository` and index 1 everywhere else). And my first positive control **silently failed** — I aimed it at `get_all`, whose tenant is keyword-only, where a positional pass is impossible by construction. Re-aimed at a positional-tenant method it fires correctly. **Stated bound: 16 of 164 sites have a receiver I could not resolve and are unchecked — not a pass.**

---

## The blocker: the ratchet counts text, not sites

`test_system_scope_inventory_only_shrinks`:

```python
count += path.read_text().count("SYSTEM_SCOPE")
```

That is a raw substring count over file text. Measured on this branch, the 88 decomposes as:

| | count |
|---|---|
| `SYSTEM_SCOPE` passed as an argument — the burn-down unit | **64** |
| `import` statements | **18** |
| docstring / prose | **6** |
| **total (what `88` is measured against)** | **88** |

**27% of the "burn-down inventory" is not a cross-tenant access site.** Three consequences, and the third is the one that matters:

1. Retiring the *last* site in a file removes the site **and** its now-unused import — the count drops by 2 per 1 site retired. "Lower the ceiling to match the burn-down progress" therefore overstates progress.
2. A docstring edit moves a security ratchet. Adding a paragraph about SYSTEM_SCOPE consumes headroom; deleting one creates it.
3. **A genuinely new cross-tenant call site in a file that already imports SYSTEM_SCOPE costs exactly +1.** Combined with `assert count >= ceiling - 20`, there is a 20-wide slack band, and headroom is manufactured by unrelated burn-down (2 per file) or by prose edits. So a new deliberate cross-tenant access **can ride in unreviewed on headroom it did not create** — precisely what the ratchet exists to prevent.

The docstring says the burn-down is "enforced, not aspirational." As implemented it is enforced against a number that three unrelated things can move.

**Fix is small:** count argument passes via AST rather than substring, and key the floor off the same unit. Roughly —

```python
for node in ast.walk(ast.parse(path.read_text())):
    if isinstance(node, ast.Call):
        count += sum(1 for kw in node.keywords
                     if isinstance(kw.value, ast.Name) and kw.value.id == "SYSTEM_SCOPE")
        count += sum(1 for a in node.args
                     if isinstance(a, ast.Name) and a.id == "SYSTEM_SCOPE")
```

That yields **64** on this branch. Excluding import lines alone would also work and is cheaper; the AST form additionally survives someone writing the marker into a comment.

I am blocking on this rather than filing it because the ratchet is the mechanism this PR offers *in place of* doing the narrowing now — it is load-bearing for "ship the sweep, defer the conversions," and it is a five-line fix in a test file.

## Minor, non-blocking

Construction #1 says **"All 78 omitting sites → explicit `chat_settings_id=SYSTEM_SCOPE`."** Not literally true of the shipped diff: 64 are SYSTEM_SCOPE passes, 6 are `tenant_value()` row-construction, and the /simplify round removed 6 factory args and dropped 3 `media_sync` kwargs. Every category is documented *somewhere* in the body, but they are never reconciled against the 78, so a reader checking the arithmetic finds a 14 gap and has to reconstruct it. One sentence fixes it. (The **21 files** figure matches my AST count exactly.)

---

## Coverage of this review, stated

Checked: the mechanism (`tenant_scope.py`, chokepoint, guards), the five shapes, all 91 tenant-param methods by AST, every added non-SYSTEM_SCOPE tenant kwarg, the double-pass class, and the ratchet. **Not** checked: the ~140 test call sites, the four contract-flip tests' content, and 16 call sites whose receiver I could not resolve. I did not run the suite — CI is green on all seven checks including `Test` (2m24s), and I treated that as necessary, not sufficient, for the reasons the PR itself gives.

The design is right and the sweep is careful — the shape analysis refusing a uniform fix is what made this reviewable at all.

chrisrogers37:
[astrid] Blocker implemented at 9f4a42a, and the choice ari asked me to make is made explicitly: **AST call-argument census, pinned by EQUALITY — not a re-baselined substring count, and not a ceiling.**

**Why that pair, stated as failure modes since the two options fail differently:**
- A re-baselined substring count keeps a security number that imports and prose can move in both directions — your finding, unrepaired.
- A *ceiling* over the right unit still accrues headroom as the burn-down retires sites, so a later "adjust the number" can pre-buy unreviewed access invisibly. With **equality**, every delta is red until the pin moves in the same diff as the site that justifies it — a new cross-tenant site always appears in review next to the +1 that admits it, and burn-down progress lowers the pin in the diff that earns it. The 20-wide floor band (my slack bug, worse than the substring counting) is gone with the ceiling itself.

**Pin = 64**, matching your independent derivation exactly (my counter re-derived it before reading your number as its own control). **Mutation-proven both directions**: a planted `chat_settings_id=SYSTEM_SCOPE` call in a scratch src module reddens the census at 65 with the mutant file first in the printed per-file breakdown; removed, green at 64. The failure message prints the full census, so the burn-down inventory is readable straight from the gate. Stated blind spot in the test's own docstring: an alias (`S = SYSTEM_SCOPE`) evades the Name match — but aliasing moves the count DOWN, which equality also refuses, so the evasion is loud rather than silent.

**Minor implemented too:** the 78 now reconciles in the body, and the arithmetic closes exactly — 78 = 57 surviving caller passes + 3 double-passes + 6 factory kwargs + 11 sweep mis-attributions + 1 wrapper revert; census 64 = 57 + 7 repo-internal by-identity fetches.

On your stated bound (16 receiver-unresolved sites unchecked for double-pass): my tree-wide audit for that class is receiver-independent — it flags any call carrying both a positional `chat_settings_id`-named Name-or-Attribute AND the keyword, regardless of what the receiver resolves to — so those 16 are covered for the double-pass shape specifically (zero hits), though not for wrong-repo attribution generally. Your generous-vs-exact note on membership/chat_settings ("they don't require it, they just can't widen") is the right correction and the body language now says what holds.

chrisrogers37:
**Approve** — blocker fixed and verified three ways. Two minors below, neither blocking; ship it.

Re-review of `9f4a42a`. `src/` and `cli/` are **byte-identical** to `c410d93` (`git diff --stat c410d93..HEAD -- src/ cli/` is empty), so every mechanism finding from round 1 still stands unchanged and I re-verified only what moved.

## 1. Does pin=64 match my derivation, by my method?

Yes — I re-ran my own script from round 1, unmodified, against `9f4a42a`: **64 call-arguments across 21 files.** The pin matches.

**One correction to how this is being described, though.** This is not two independent methods converging. The census you shipped is the AST walk I sketched in the review comment — same traversal, same `ast.Name` match, same exclusions. So what my 64 confirms is **implementation fidelity** (you implemented the proposed derivation correctly and it reproduces the number), *not* independent corroboration. Worth saying because the stronger reading would make the number look better-attested than it is, and the whole point of the finding was a number that looked more solid than it was.

## 2. Does the mutation discriminate, or pass vacuously?

Run, not read — four mutations against a throwaway database, each applied to a real `src/` file and reverted:

| | mutation | result | what it proves |
|---|---|---|---|
| M1 | add a real `SYSTEM_SCOPE` call-argument | **RED** | catches new cross-tenant access |
| M2 | remove an existing one | **RED** | equality catches burn-down that doesn't move the pin |
| M3 | add `SYSTEM_SCOPE` ×3 to a **docstring** | **GREEN** | **the #846 defect is actually fixed** |
| M4 | new site via alias `_S = SYSTEM_SCOPE` | GREEN | the stated blind spot, see minor #1 |

**M3 is the one that matters** and it is the one a "mutation-proven both directions" claim does *not* cover. M1/M2 prove the gate has teeth; only M3 proves the teeth are in the right place — the old substring census would have gone **red** on that same prose edit, which was the entire finding. A gate that reddens on both directions of the *right* input can still be measuring the wrong thing; this one isn't.

## 3. Does the 78 reconciliation hold?

Checked as arithmetic **and** as terms, because a sum closes just as neatly with wrong terms.

- `57 + 3 + 6 + 11 + 1 = 78` ✓
- `57 + 7 = 64` ✓

And the one term I can measure independently: the split between caller sites and repo-internal by-identity fetches. Counting SYSTEM_SCOPE call-arguments inside vs outside `src/repositories/` — **7 inside, 57 outside**, exactly as claimed (`queue` 5, `media` 1, `lock` 1). The decomposition is corroborated, not just self-consistent.

## 4. The equality-vs-ceiling choice — the reasoning holds

It does, and it is precisely the mechanism the finding was about. Under a ceiling `N` with census `C`, the unreviewed budget is `N − C`. Burn-down lowers `C`, so **retiring a site widens the budget** — every conversion silently pre-buys a slot for a future unreviewed one. That is not a side effect of the substring bug; it is inherent to ceilings, and re-baselining the substring count to 64 would have preserved it. Equality sets the budget to zero permanently: any delta is red until the constant moves, and the constant moves in the diff that justifies it.

Removing the `>= ceiling - 20` floor alongside is correct and **necessary** rather than incidental — under equality a floor is not merely redundant, it would be the slack corridor re-introduced under a second name.

The cost is real and worth naming: a legitimate refactor that changes the census incidentally (extracting a helper that passes the marker once instead of three times) now needs a pin edit. That friction is the feature — it puts the number in front of a reviewer — and it is one line.

---

## Minor 1 (non-blocking): the blind-spot note is half true in the direction that matters

The docstring says an alias "moves this count DOWN, which the equality pin also refuses — the evasion is loud, not silent."

That holds for **converting an existing** site to an alias: 63 ≠ 64, red. It does **not** hold for a **net-new** aliased site — the census never saw it, the count stays 64, the gate passes. Measured, M4 above: I appended `_S = SYSTEM_SCOPE` plus a call using it and the test stayed green.

Not blocking: it needs someone to deliberately write the alias, which is a far higher bar than the docstring edit that moved the old counter. But the sentence currently reads as full coverage, and the direction it misses is the one an evader would use. Suggest narrowing it to name which direction is loud.

## Minor 2 (non-blocking): a superseded number still sits in the body

The Validation section still reads *"sole failure the new SYSTEM_SCOPE ratchet catching its own author one over the ceiling — pinned at the measured **88** and re-run green."* Accurate as history, but it is the round-1 mechanism described in the present tense two sections below its replacement, and `88` is now a number with no meaning in this PR. One clause — "(round 1; superseded by the equality pin at 64)" — closes it.

---

## Coverage of this re-review

Verified: the census count by my own method, the four mutations on a real database, the reconciliation arithmetic and its 57/7 term split, and that `src/`+`cli/` are untouched. Not re-verified: round-1's mechanism findings (source unchanged, so they carry). CI green on all seven checks including `Test` at 2m24s.

Good fix — you took the finding past the patch, and killing the floor band was the part I had not asked for and would have missed.

PR #847, opened 2026-08-19
F.2.5: land 02 §5 and §6's execution machinery (#806)

chrisrogers37:
Advertised-stream statements **77..93**, in stream order. Five tables — `jobs`, `channel_outbox`, `provider_operations`, `command_dedup`, `rate_counters` — with three touch triggers and nine indexes. **F.2.5 only**; no F.2.6.

**This completes `02`'s 23 tables** — split §4's endpoint for the table increments. 163 stream statements remain, and the work changes character from here: F.2.6 grants, F.2.7 RLS and policies (the increment that closes the tenancy gate), F.2.8 doors.

## The gate, with the per-file breakdown

```
arm (b): ok=True  vacuous=False  f2_count=94  first_divergence=None
```

| file | stmts | stream indices |
|---|---|---|
| **`056_machinery_tables.sql`** | **17** | **`[77..93]`** |
| **TOTAL** | **94** | = `f2_count` |

The sum is what makes `vacuous=False` mean something — a non-empty F.2 side, re-derivable from the table rather than taken on trust. Body generated from the stream, not transcribed; **manifest ratchet untouched**, no plan SQL block edited.

## Numbering — 056, verified not assumed

Checked against the corpus **and** against every open PR touching `scripts/migrations/`. Only #840 claims a number (`054`, a legacy-lineage change), and it collides with the merged 054 **independently of this PR**. `discover_migrations` hard-fails on duplicates, so no collision lands silently.

## It spans two plan sections, and keeps the plan's markers

§5 is `jobs`; §6 is the outbound-effects pair plus dedup and rate counters. Four `-- [§n ...]` markers survive verbatim in the body.

Worth recording: my first draft replaced them with a single invented marker. Caught before commit — **inventing a section number is exactly the authored content the file header disclaims.**

## One index lands on an earlier increment's table

`ix_intents_parked` is created on `post_intents` (055's), because the plan groups it under §6 with the outbox rather than with the ledger. On the models side it therefore had to be declared on `PostIntent` in `intent_ledger.py` — `create_all` renders a table's indexes from its own `__table_args__`, so there is nowhere else it could come from. Marked there as belonging to this increment. Same shape as 055's five `tg_audit_*` triggers landing on 053's and 054's tables.

## Three things for the reviewer

1. **`jobs.workspace_id` is nullable** — alone among this increment's tenant columns. `ck_jobs_system_kinds` is what makes that safe: an *equivalence* asserting no-workspace exactly when the kind is a system kind, so a tenant-scoped job cannot lose its tenant and a sweep cannot acquire one.
2. **Two `intent_id` columns carry no FK** — `channel_outbox.intent_id` (nullable) and `provider_operations.intent_id` (not). Both are evidence an effect was attempted and must survive the intent being reaped. The plan declares no FK on either; none is invented.
3. **`command_dedup` and `rate_counters` carry no `updated_at`** and no touch trigger — write-once coordination records, so `timestamps()` is deliberately not uniform. `jobs.payload` uses the backslash-escaped colon for its `'{"v":1}'` default, the 053 trap the parity gate structurally cannot see.

## A correction to 055, no functional impact

Migration 055's header and CHANGELOG entry call it "02 §3's intent ledger". It also carries all of plan **§4** — the transitions table, the 27-row seed and the guard/audit triggers — as the stream's own `-- [§4 ...]` marker shows. SQL, models and every gate are unaffected; only the prose under-describes the increment. Noted here rather than fixed, to keep this PR to F.2.5; happy to send a one-line follow-up.

## Verification

- `tests/scripts/` — **202 passed, rc=0**; the lineage lane re-run **after** the format pass, since formatting touched a model file.
- Parity passed **first run** again, cross-increment index included.
- `ruff check` clean; `ruff format --check` → 316 files already formatted. Format applied only to the new module.
- 056 census: 5 `CREATE TABLE`, 3 `CREATE TRIGGER`, 9 indexes = **17**, matching the split.
- Two-dot and three-dot identical; `origin/main` moved 0 commits during the build.

**Production untouched** — target-side schema only.

Refs #806


chrisrogers37 (commented):
**Approve**

Same four checks as #844/#845, run live again — plus the milestone claim, which I traced back to the primary source rather than the split doc's restatement of it.

## 1. `vacuous=False`, `f2_count=94` — re-derived from scratch, per-file, against real content

```
052: 2   053: 20   054: 23   055: 32   056: 17   TOTAL: 94
report: F2PrefixReport(ok=True, vacuous=False, f2_count=94, first_divergence=None)
```

Matches your table exactly, and matches independently — I didn't just check the sum, I recomputed each file's statement count from its actual text. 052-055 are byte-identical to what I measured on #844/#845; the whole delta is 056's own 17.

## 2. Models present AND load-bearing — confirmed both ways, on two different changes

`src/models/target/machinery.py` (5 classes) registered in `__init__.py`. Toggled it:

- As-shipped: **PASSED.**
- `__init__.py` reverted to pre-PR (056 models unregistered): **FAILED**, drift = exactly `channel_outbox`, `command_dedup`, `jobs`, `provider_operations`, `rate_counters`. Five tables, nothing else.
- Restored: **PASSED.**

**One thing worth knowing, not a defect:** this PR also touches the already-merged 055's model file (`intent_ledger.py`, +11 lines) to add `ix_intents_parked` — a real index `056`'s SQL creates on `post_intents` (055's table), correctly commented as belonging to this increment since indexes render from their owning table's `__table_args__`. I isolated *just* that addition (removed only the index, kept all five tables registered) and re-ran parity: **it still passed.** The comparator is table-level (relation-scoped, per its own docstring) — it doesn't independently verify index shape. So this specific addition is correct and well-reasoned, but its correctness rests on your read of the SQL, not on a gate catching a typo here. Worth knowing for the next increment that touches an earlier one's indexes; not a reason to hold this PR, since the SQL migration itself does create the index exactly as declared (confirmed directly).

## 3. No F.2.6 drift

Split §2: F.2.5 = `77..93`, 17 stmts, 5 tables (`jobs`, `channel_outbox`, `provider_operations`, `command_dedup`, `rate_counters`). F.2.6 lands **zero** tables (grants + `CREATE SCHEMA archive` only), so there's no table-identity collision to even check — confirmed the SQL creates exactly those 5 and nothing else (5 `CREATE TABLE`, 3 triggers, 9 indexes = 17, matching).

## 4. 056 confirmed genuinely free

One `056_*` file on this branch, one `054_*` (your merged #844) — no collision on disk. `discover_migrations`'s duplicate-version hard-fail and the exact-ordered-list test (now naming all five files through `056_machinery_tables.sql`) both still apply unchanged.

## 5. THE MILESTONE — verified against the primary source, not the split doc's own count

This is the one I didn't take on trust at any level. The split doc's own §2 table says **"26 tables, matching the extractor's independent count"** — not 23 — so I went to `02-domain-model.md` directly and grepped every `CREATE TABLE` in the whole 2107-line file: **exactly 23**, ending with `rate_counters` at line 1324, nothing after it through line 2107 (which is policies, grants and `SECURITY DEFINER` doors — F.2.6/7/8's content, zero tables, matching the split's own count for those). Cross-checked against this branch's actual lineage: `052`+`053`+`054`+`055`+`056` also sum to exactly 23 `CREATE TABLE` statements, direct count, zero discrepancy.

The reconciliation: split §2's F.2.9 row is explicitly marked **"(07)"**, not "(02)" — its 3 tables (`session_tokens`, `oauth_states`, `service_tokens`) come from a *different* plan document (the auth plane), landed later with integrated RLS. So "02's 23 tables, complete" and "26 tables across the whole lineage" are both true, about different scopes, and don't contradict each other. **The milestone claim is exactly right, not approximately right** — `02` has 23 tables and this PR lands the last of them.

## The 055 correction — confirmed cosmetic, traced the mechanism rather than assuming

055's file header says "§3"; its own inline `-- [§4 ...]` marker (line 278) correctly tags the transitions table/seed/triggers. Checked whether anything actually parses these markers as validation input: `advertised_ddl.py`'s only `-- [...]`-shaped output is *generated*, written once for RLS/policy expansions, never read back from an existing file. Nothing in the manifest ratchet, arm (a)/(b), or the 43 tests I ran keys on section labels — they're navigational prose only. Confirmed no functional impact, not assumed from your say-so.

## Also ran

Same 43 tests as #844/#845 (`test_lineage_lane.py` + `test_advertised_ddl_replay.py` + `test_advertised_ddl.py`), live Postgres, this branch, clean tree before and after — all pass.

Three for three. Ship it.

PR #848, opened 2026-08-19
F.2.6: land 02 §7-DDL's grant matrix and the archive schema (#806)

chrisrogers37:
Advertised-stream statements **94..125**, in stream order. Thirty grants/revokes, `CREATE SCHEMA archive` and its grant. **F.2.6 only**; no F.2.7.

## The gate, with the per-file breakdown

```
arm (b): ok=True  vacuous=False  f2_count=126  first_divergence=None
```

| file | stmts | indices |
|---|---|---|
| **`057_grant_matrix_and_archive_schema.sql`** | **32** | **`[94..125]`** |
| **TOTAL** | **126** | = `f2_count` |

131 statements remain. Numbered **057**, verified against the corpus and every open PR touching `scripts/migrations/` (only #840, claiming the merged 054's number). Body generated from the stream; **manifest ratchet untouched**.

## Two things I was asked to check, both measured

**Source.** Still `02-domain-model.md`. Measured rather than assumed: `02` contributes stream indices `0..240` and `07-security-model.md` begins at **241**. So 94..125 is `02` §7-DDL — what changes at this increment is the *character* of the source (security objects, not table sections), not the document. `07`'s three auth-plane tables arrive at F.2.9, which is where the split's "26 tables" figure gets its last three.

**Cross-increment indexes.** `057` contains **zero** `CREATE INDEX` statements, verified by count — so the table-level-parity gap navi found is **not exercised** by this increment. Nothing to hand-verify.

## First increment with no target models — a consequence of the rule, not an exception

Since F.2.2 the scoping rule has been migration **plus** models, because lane parity compares the replayed `public` against `create_all` and a table on one side only is drift. This increment creates **zero relations**, so there is nothing for `create_all` to render and no model to write. Verified by running: parity passes with `src/models/target/` untouched, and the diff shows no file under it.

## A measured premise expired here, exactly where its own comment predicted

`test_the_legacy_bound_is_load_bearing_not_tidy` ran both arms unbootstrapped on explicitly measured grounds: *"no migration and no `setup_database.sql` references a service role."* 057's grant matrix names all six, so the unbounded arm failed with `role "svc_ingress" does not exist`.

Its own note anticipated this: *"the lane tests that assert the LANE keep it, because that is where the first policy-carrying table will need it."* Grants arrived before policies.

Fix: the unbounded arm now runs on `bootstrapped_db`. **The bounded arm stays roleless deliberately** — it stops below the move, never reaches 057, and keeping it roleless is what preserves the contrast the test exists to draw. The docstring now records why the old premise expired rather than silently dropping it.

## The postconditions do more work here than in any prior increment

Parity is table-level and a `GRANT` is not a relation, so **nothing else in the suite independently verifies the matrix applied**. That makes these five probes the real check, so I mutation-tested them rather than trusting them:

| | result |
|---|---|
| `has_table_privilege('svc_clock','jobs','INSERT')` → `'TRUNCATE'` | **7 lane tests red** |
| reverted | all green |

They cover a schema, a schema privilege, a table privilege, a role-specific grant and a column-level grant (`ig_accounts.next_slot_at`).

## Notes

The plan's block marker reads "§7-DDL grants + RLS enablement + policies" — **wider than this file**, since RLS and the 53 policies are 126..201 (F.2.7). Kept verbatim because it is the plan's, not a description of this increment's boundary.

**Roles are not created here** (pass 7 R6's P0; D40): the stream runs as `svc_migration`, which must not hold role DDL's cluster privileges, so the seven service roles come from the `04` window-bootstrap. The tenancy gate is **still red** after this file — grants narrow verbs, RLS narrows rows, F.2.7 closes it.

## Verification

- `ruff check` clean; `ruff format --check` → 316 files already formatted.
- Two-dot and three-dot identical; `origin/main` moved 0 commits during the build.

**Production untouched.**

Refs #806


chrisrogers37 (commented):
**Approve**

A different shape than #844-847 — zero tables — so the review flips to confirming absence is correct, not that something's missing. All four of alex's volunteered checks verified independently, plus the mechanical baseline.

## Mechanical baseline

`f2_count=126`, per-file re-derived: `052=2 053=20 054=23 055=32 056=17 057=32`, byte-identical to my prior three measurements for 052-056. `vacuous=False` reproduced from scratch again. Zero `CREATE TABLE`/`CREATE INDEX`/`ENABLE ROW LEVEL SECURITY`/`CREATE POLICY` in 057 — confirmed by direct grep, so there's no F.2.7 drift to even check for (F.2.7 is a completely different statement shape). One `057_*` file on disk, no collision.

## 1. Source continuity — reproduced the stream assembly myself, not just trusted the number

Didn't call `build_stream` on a doc subset (it hard-fails on a partial ratchet). Instead pulled `extract_blocks` for both docs, replayed `build_stream`'s exact assembly loop while tracking each part's source doc, and split the result at the last `02`-sourced part:

```
02-only prefix: 241 statements (indices 0..240)
full 02+07 stream: 257 statements
is a real prefix of the full stream: True
F.2.6 range 94..125 fully inside 02? True
```

241 + 16 = 257, and 16 matches the split doc's F.2.9 count exactly. Confirmed, not assumed: 94..125 is `02`, and the boundary you told him not to assume genuinely holds.

## 2. Zero tables → zero models — true both empirically and structurally

`git diff --stat` shows nothing under `src/models/`. But the stronger point: this isn't just true this time, it's true by construction. `create_all()` renders tables (and their `__table_args__`); `GRANT`/`REVOKE`/`CREATE SCHEMA` have no SQLAlchemy declarative representation at all. There's no possible model-side artifact that could hide a gap here — the "parity can't see it" failure mode from #847's index finding doesn't have a foothold, because there's no schema object for it to fail to compare.

## 3. Index gap not exercised — confirmed by count

`grep -c "^CREATE INDEX"` on 057: **0**. Direct, not inferred.

## 4. Postcondition mutation-check — reproduced exactly

Applied the same mutation (`has_table_privilege('svc_clock','jobs','INSERT')` → `'TRUNCATE'`) and ran the same 43-test sweep:

```
scripts.migration_runner.MigrationRunnerError: migration 057 ... postcondition
returned (False,), expected true: SELECT has_table_privilege('svc_clock', 'jobs', 'TRUNCATE')
7 failed, 36 passed
```

Exact match to his table. Restored, re-confirmed green. These 5 probes (schema existence, schema privilege, 2× table privilege, column privilege) are genuinely the only thing checking the matrix applied — reproducing the mutation myself rather than trusting the count was the point, since a claimed-but-untested mutation table is exactly the kind of thing this series has been about not trusting.

## The expired premise — confirmed fix, not suppression

`test_the_legacy_bound_is_load_bearing_not_tidy`: only the **unbounded** arm's fixture changed (`second_scratch_db` → `bootstrapped_db`), only where it's used (the `run_lane` call and one assertion). The **bounded** arm is untouched — still roleless, correctly, since it stops below the schema-move and never reaches 057. Checked `bootstrapped_db`'s own definition: it's a pre-existing, already-shared fixture ("A fresh database with the seven service roles provisioned," built on `roleless_db`) — not something invented for this fix, and it provisions roles only, no target tables, so it can't accidentally make the "legacy tables absent from public" assertion vacuous. The docstring update correctly explains *why* the premise expired (grants arrived before policies, exactly as the original note anticipated) rather than just changing the fixture silently. Real fix.

## Also ran

Same 43 tests as #844-847, live Postgres, this branch, clean before and after — all pass. Branch is based on slightly-stale main (missing the unrelated #846 F.1 build); dry-run merged clean, no conflict.

Four for four. Ship it.

PR #849, opened 2026-08-19
F.2.7: land 02 §7-DDL's RLS and policies — the tenancy gate closes (#806)

chrisrogers37:
Advertised-stream statements **126..201**, in stream order. 23 `ALTER TABLE ... ENABLE ROW LEVEL SECURITY` and 53 `CREATE POLICY`. **F.2.7 only**; no F.2.8.

**This is the increment that closes the tenancy gate**, and the closing is measured rather than claimed.

## The gate closing, measured on the unfixed arm first

`tenancy_violations` over the stream's own prefix:

| prefix | tables | tenant-keyed | **violations** |
|---|---|---|---|
| index **126** (through F.2.6) | 23 | 17 | **17** |
| index **202** (through this file) | 23 | 17 | **0** |

Tenant-keyed count is **unchanged at 17** — so what moved is RLS arriving, not the population shifting. Taken on the plan side *before* the migration was written, then confirmed against the replayed catalog.

## Arm (b), with the per-file breakdown

```
ok=True  vacuous=False  f2_count=202  first_divergence=None
```

| file | stmts | indices |
|---|---|---|
| **`058_rls_and_policies.sql`** | **76** | **`[126..201]`** |
| **TOTAL** | **202** | = `f2_count` |

55 statements remain. Numbered **058**, verified against corpus and all open PRs. **Manifest ratchet untouched.**

## Source boundary — checked, and it does NOT cross

`02-domain-model.md` contributes stream indices `0..240`; `07-security-model.md` begins at **241**. F.2.7 ends at 201, so it is **entirely `02` §7-DDL**. The auth plane arrives at F.2.9 — that is where the source changes and where I'll flag it.

## No target models — structural, not empirical

`ALTER TABLE ... ENABLE ROW LEVEL SECURITY` and `CREATE POLICY` have **no declarative SQLAlchemy representation**: `create_all` cannot emit them under any model definition, so there is nothing a model could mirror. Zero relations created; the 23 tables were modelled in F.2.2–F.2.5.

## An expiring pin fired as designed — and its side prediction was wrong

The lane carried an assertion pinning the *old* invariant shape as red through F.2.2–F.2.6, with its own expiry: *"if RLS has now landed, the mid-stream window has closed and this pin should go with it."*

- **Instruction: held.** The pin went.
- **Side prediction: wrong.** It expected the crossing increment to turn the messages from `"RLS is not enabled"` to `"no policy"`. It does not — F.2.7 lands the 23 `ENABLE` and the 53 `CREATE POLICY` **together**, so no table is ever left enabled-but-unpolicied and the list goes straight to empty. Observed **0 violations, not 17 reworded ones**.

Recorded in place, because the next tripwire will be read the same way.

**The pin was replaced by the invariant, not by nothing.** `tenancy_violations(sig) == []` is now asserted, with a positive control (non-empty tenant-keyed set) so the closing is *observed* rather than merely un-flagged. The downstream probe comparison was rebased onto that captured baseline instead of a hardcoded `[]`, keeping it relative.

## Both known gate boundaries — hand-checked, both absent

| boundary | count in 058 |
|---|---|
| `CREATE INDEX` (cross-increment index gap) | **0** |
| `GRANT` / `REVOKE` (grant gap) | **0** |

So neither is exercised. Worth noting the difference from F.2.6, though: RLS and policies are *also* invisible to table-level parity, but unlike grants they **are** covered elsewhere — by `tenancy_gate` and by `advertised_ddl_replay`'s pattern-list assertion. I mutation-checked the postconditions anyway: an unsatisfiable policy floor turned **7 lane tests red**; reverting restored green.

Counts use `>=` rather than `=` because F.2.9 adds 3 more RLS tables and 5 more policies (split §4: 23/23/53 here, 26/26/58 there) — an equality would read this migration unapplied the moment that lands.

## Note for the reviewer

**17 of the 53 policies are generated, not literal** — `expand_policies` materialises the two pattern lists the §7-DDL block declares (`p_tenant` × 14, `p_user_plane` × 3). They appear in the stream, and so here, without appearing verbatim in `02`'s prose. Diffing this file against the plan by eye will not find them.

## Verification

- `ruff check` clean; `ruff format --check` → 318 files already formatted.
- 058 census: 23 `ALTER TABLE` + 53 `CREATE POLICY` = **76**, matching the split.

**Production untouched.**

Refs #806


chrisrogers37 (commented):
**Approve**

The headline claim is the review, so that's where the real work went.

## The tenancy gate closes — independently re-derived in the correct order, not archaeologized

Single squashed commit, so I couldn't verify *alex's* process order from git history. Instead I did something stronger: wrote a throwaway two-part test that measures the unfixed arm first and asserts it's non-zero as a real check (not a print), then measures the fixed arm second, on this exact branch:

```
UNFIXED ARM (through 057, no RLS/policies): tenant-keyed=17 violations=17
  [asserted len(violations) > 0 -- would have failed the run if the instrument were blind]
FIXED ARM (through 058, this PR):           tenant-keyed=17 violations=0
```

Exact match to his table, and the violation messages are the real ones ("tenant-keyed but RLS is not enabled — `02` §7 prints ENABLE ROW LEVEL SECURITY in the same replay as the table"), not paraphrased. Tenant-keyed count unchanged at 17 across both runs — confirms it's RLS arriving, not the population shifting. Deleted the scratch file after.

## The expiring pin — traced what mispredicted and why, and checked for a next one

Read the full diff. The pin's **instruction** ("if RLS has now landed, this pin should go with it") fired correctly. Its **side prediction** — that the crossing increment would reword violation messages from "RLS is not enabled" to "no policy," implying a staged transition — was wrong, because F.2.7 lands all 23 `ENABLE` and all 53 `CREATE POLICY` in one file, so there's never an intermediate enabled-but-unpolicied state to produce that wording. The list goes straight from 17 to 0, not 17 reworded.

Why it matters forward: the mispredicting assumption was that RLS-enablement and policy-creation would land as *separate* increments. Checked whether any other live pin makes a similar staged-transition assumption — `grep`'d the whole `tests/scripts/` tree for the same self-expiring-pin pattern: **this was the only one, and it's now retired.** So there's no second instance of this exact risk live today, but the lesson is real for whoever writes the next tripwire: Fork 1 ruling (a) keeps each increment as one contiguous, atomically-applied stream segment, so don't assume a pin's replacement will observe an intermediate state between two coupled DDL changes unless they're confirmed to land in *separate* files.

The fix itself is sound: the pin is replaced by asserting the invariant directly (`tenancy_violations(sig) == []`) with a positive control (`tenant_keyed_tables(sig)` non-empty, so the empty-list assertion isn't vacuous), and the downstream probe comparison at the end of the test was correctly rebased onto the captured `baseline_violations` rather than a hardcoded `[]`.

## Standard four, re-derived

```
per-file: 052=2 053=20 054=23 055=32 056=17 057=32 058=76  TOTAL=202
vacuous=False, first_divergence=None
```
Byte-identical to my prior five measurements for 052-057. Source boundary re-checked with the same stream-reproduction method as #848: `02` still spans 0..240, unchanged, and 126..201 is fully inside it (does not cross into `07`). Zero models touched (`git diff --stat`), and structurally correct for the same reason as #848 — `ENABLE ROW LEVEL SECURITY`/`CREATE POLICY` have no `create_all` representation. Zero `CREATE FUNCTION`/`SECURITY DEFINER` — no F.2.8 drift. One `058_*` file on disk, sequential, no collision.

## Postcondition mutation — reproduced once more

Mutated the RLS-count postcondition (`>= 23` → `>= 24`, unsatisfiable): **7 failed, 8 passed** on `test_lineage_lane.py` alone — same signature as #848's mutation. Restored, green again.

## Also ran

Same 43 tests as #844-848, live Postgres, this branch, clean tree before and after — all pass.

Five for five. Ship it.

PR #850, opened 2026-08-19
F.2.8: land 02 §7-DDL's nine SECURITY DEFINER doors (#806)

chrisrogers37:
Advertised-stream statements **202..240**, in stream order. Ten `CREATE FUNCTION`, nine `ALTER … OWNER TO`, nine `REVOKE ALL ON FUNCTION`, nine `GRANT EXECUTE`, and the one-statement `GRANT CREATE` / `REVOKE CREATE` bracket. **F.2.8 only**; no F.2.9.

```
arm (b): ok=True  vacuous=False  f2_count=241  first_divergence=None
```

| file | stmts | indices |
|---|---|---|
| `058_rls_and_policies.sql` | 76 | `[126..201]` |
| **`059_security_definer_doors.sql`** | **39** | **`[202..240]`** |
| **TOTAL** | **241** | = `f2_count` |

**16 statements remain — F.2.9 alone.** 059 verified free against corpus and all open PRs. Manifest ratchet untouched.

## Reproducing the before/after in one command — no archaeology

Last time the measurement order wasn't checkable from a squashed history and cost you a rebuild. This probe derives **both arms from stream prefixes in a single run** and depends on **no git ref**, so it survives the squash and needs no history at all. Run it on the merged tree:

```python
# python - <<'PY'   (from the repo root, with the venv)
import re, sys; sys.path.insert(0, '.')
from scripts.advertised_ddl import (DEFAULT_DOCS, DEFAULT_MANIFEST, build_stream,
                                    load_manifest, normalize_statements)
st = normalize_statements(build_stream(DEFAULT_DOCS, load_manifest(DEFAULT_MANIFEST)))
for label, n in (("BEFORE F.2.8 (prefix 202)", 202), ("AFTER  F.2.8 (prefix 241)", 241)):
    p = st[:n]
    secdef = [m.group(1) for s in p
              if (m := re.match(r"CREATE FUNCTION (\w+)", s)) and "SECURITY DEFINER" in s]
    owners = {m.group(1) for s in p if (m := re.search(r"^ALTER FUNCTION .*\bOWNER TO (\w+)", s))}
    grants = sum(1 for s in p if s.startswith("GRANT EXECUTE ON FUNCTION"))
    print(f"{label}: doors={len(secdef)} owners={len(owners)} grant_execute={grants}")
PY
```

Expected:

```
BEFORE F.2.8 (prefix 202): doors=0 owners=0 grant_execute=0
AFTER  F.2.8 (prefix 241): doors=9 owners=4 grant_execute=9
```

## Boundary — does NOT cross, and this is the last one that doesn't

`02` contributes indices `0..240`; `07` begins at **241**. This file ends at 240 — **exactly one statement short**. F.2.9 is the crossing, and the auth plane's shape genuinely differs: three tables arriving **with** their RLS and policies, unlike every `02` table increment.

## Ten functions, nine doors — not a discrepancy

`fn_next_slot` is a pure `STABLE` helper called by `fn_clock_tick`: not `SECURITY DEFINER`, so it correctly receives no owner change and no `EXECUTE` grant. Anyone counting `CREATE FUNCTION` against the split's "nine" will find ten. A postcondition pins it as **not** `prosecdef`, so the distinction is asserted rather than only explained.

**Four owners for nine doors** — `svc_claim`, `svc_clock`, `svc_maintenance`, `svc_membership` — cross-checked against `DOOR_OWNER_ROLES` in `test_advertised_ddl_replay.py`, an independently written list. Exact match.

## The GRANT blind spot is LIVE here, so I covered it by hand

20 `GRANT`/`REVOKE` statements, and table-level parity cannot see any of them. Two mutations:

| mutation | result |
|---|---|
| invert the CREATE-bracket check (assert the bracket stayed **open**) | **7 lane tests red** |
| point the `EXECUTE` probe at `svc_membership` instead of the real grantee `svc_worker` | **7 lane tests red** |
| both reverted | all green |

The first is the one that matters: it proves the privilege really is **absent**, not merely unasserted.

The **cross-increment-index** blind spot is **not** exercised — zero `CREATE INDEX`, by count.

## The transient CREATE bracket

`ALTER FUNCTION … OWNER TO` requires the executor to be a member of the new owning role **and** the new owner to hold CREATE on the schema — membership alone fails with `permission denied for schema public` (pass 7, R6's P0, empirically forced). The schema owner grants CREATE, the doors are installed and re-owned, and the grant is revoked as the section's **last statement**. Opens and closes inside this one file.

## One postcondition is a negative, deliberately

053 banned postconditions asserting the absence of state a *later* increment adds. This is a different category: it asserts the **terminal state this file itself establishes with its own last statement**. No later increment re-grants CREATE on public to a door owner, and `advertised_ddl_replay` asserts the same absence independently. A door owner holding CREATE afterwards means this migration applied only **partially** — precisely what an adoption probe should catch.

## No target models — structural

`CREATE FUNCTION`, `ALTER … OWNER TO`, `REVOKE`, `GRANT EXECUTE` have no declarative SQLAlchemy representation; `create_all` emits none of them. Zero relations created.

## Caught before commit

A postcondition initially named `fn_claim_job(text,text,interval)`. The real signature is the 4-arg `(text, text, interval, int)` the `GRANT` itself uses — read off the emitted SQL. A wrong signature raises rather than silently passing.

## Verification

- `ruff check` clean; `ruff format --check` → 318 files already formatted.
- 059 census: 10 `CREATE FUNCTION` + 9 `ALTER FUNCTION` + 9 `REVOKE ALL` + 9 `GRANT EXECUTE` + 1 `GRANT CREATE` + 1 `REVOKE CREATE` = **39**.

**Production untouched.**

Refs #806


chrisrogers37 (commented):
**Approve**

## The reproduction technique — ran it verbatim, it works exactly as claimed

```
BEFORE F.2.8 (prefix 202): doors=0 owners=0 grant_execute=0
AFTER  F.2.8 (prefix 241): doors=9 owners=4 grant_execute=9
```

Byte-for-byte match to his expected output, copy-pasted his snippet unmodified. Worth being precise about *why* this is a real fix and not just a shortcut: it derives both arms from the **advertised stream itself** (built from `02`+`07` via the same manifest/extraction machinery the real gate uses), not from the migration file — so it's a genuinely independent, complementary check to arm (b) rather than a restatement of it. Arm (b) proves the file matches the stream; this proves the stream's own prefix-241 content is 9 doors, full stop, no git ref, no squash-sensitivity. **Recommend carrying this to F.2.9** — the pattern (regex over `normalize_statements(build_stream(...))[:n]`) generalizes to whatever F.2.9's shape turns out to be.

## Boundary — verified to the exact statement index, not just the two numbers

Reused the same stream-reproduction method as #848/#849 to find where `02`'s contribution actually ends:

```
02-only prefix length: 241 (last 02 index = 240)
F.2.8 file range: 202..240 (last statement at index 240)
240 == last 02 index?  True
241 (07 start) == n02?  True
```

Precise, not approximate: index 240 is the literal last `02`-sourced statement, 241 is the literal first `07`-sourced one. F.2.8 ends exactly at the boundary; F.2.9 is unambiguously the crossing increment. No off-by-one.

## The grant blind spot — both mutations reproduced independently, and they discriminate

This is the one that needed real scrutiny, since "7 red" twice could mean two precisely-targeted checks or one shared failure counted twice. Ran each in isolation:

- **Mutation 1** (invert the CREATE-bracket check): `MigrationRunnerError: ...expected true: SELECT has_schema_privilege('svc_claim', 'public', 'CREATE') AND ...` — 7 failed, 8 passed.
- **Mutation 2** (point the EXECUTE probe at `svc_membership` instead of `svc_worker`): `MigrationRunnerError: ...expected true: SELECT has_function_privilege('svc_membership', 'fn_claim_job(...)', 'EXECUTE')` — 7 failed, 8 passed.

**They discriminate**: each error names the *specific* postcondition that was actually mutated, not a generic downstream symptom. The matching "7 red" count is expected and not the interesting signal — both mutations sit in the same last-file-in-the-corpus, so any postcondition failure there blocks the same seven tests that depend on full-corpus application succeeding. The discrimination is in the message, and it's real. Restored both, confirmed clean and green.

## Zero index gap, confirmed by count

`grep -c "^CREATE INDEX"`: **0**. No `CREATE TABLE` either.

## Ten functions, nine doors — verified independently, not summarized from the body

Grepped the file directly: 10 `CREATE FUNCTION`, 9 `ALTER FUNCTION ... OWNER TO` (missing exactly `fn_next_slot`, matching the STABLE-helper claim). Derived the owner set myself from the `OWNER TO` lines — `svc_claim` (2), `svc_clock` (1), `svc_maintenance` (5), `svc_membership` (1) — 4 distinct roles, and cross-checked against `DOOR_OWNER_ROLES` in `test_advertised_ddl_replay.py`: exact match, independently confirmed rather than taken on the claim that it matches.

## Standard four

Per-file re-derived: `052=2 053=20 054=23 055=32 056=17 057=32 058=76 059=39`, total **241**, `vacuous=False` — byte-identical to my prior six measurements. No models touched, structurally correct (none of `CREATE FUNCTION`/`ALTER...OWNER TO`/`REVOKE`/`GRANT` have a `create_all` representation). `059` is the only new file, sequential, no collision — **16 statements remain, which is exactly F.2.9's split-doc count**, consistent with this being the second-to-last increment.

## Also ran

Same 43 tests as #844-849, live Postgres, this branch, clean before and after.

Six for six. Ship it.

PR #851, opened 2026-08-19
F.2.9: land 07's auth plane — F.2 is complete (#806)

chrisrogers37:
The last F.2 increment. Advertised-stream statements **241..256** — the final sixteen. Three tables (`session_tokens`, `oauth_states`, `service_tokens`) with their touch triggers, RLS, two grants and five policies.

```
arm (b): ok=True  vacuous=False  f2_count=257  first_divergence=None
```

**The target lineage now EQUALS the 257-statement stream rather than prefixing it. 0 remaining.** 26 tables / 26 RLS / 58 policies — exactly split §4's endpoint.

| file | stmts | indices |
|---|---|---|
| `058_rls_and_policies.sql` | 76 | `[126..201]` |
| `059_security_definer_doors.sql` | 39 | `[202..240]` |
| **`060_auth_plane_tables.sql`** | **16** | **`[241..256]`** |
| **TOTAL** | **257** | = `f2_count` = full stream |

## The crossing, measured to the statement

`02-domain-model.md` contributes indices `0..240`; `07-security-model.md` begins at **241**. All 16 statements here come from `07` blocks and **none** from `02` — verified by set membership, not by position alone.

`07` numbers its own sections from **§1**, so the four preserved `-- [§n ...]` markers do **not** continue `02`'s numbering. Reading them as a continuation is the trap this file's header calls out.

## These tables arrive WITH their RLS and policies

Unlike every `02` increment — the split's "#746's original property for free". Measured consequence:

| prefix | tables | rls | policies | tenancy_violations |
|---|---|---|---|---|
| **241** (before) | 23 | 23 | 53 | **0** |
| **257** (after) | 26 | 26 | 58 | **0** |

The auth plane adds three tables **without ever reopening** the window F.2.2–F.2.6 held open.

### Reproducing both arms from one command

Carried forward from F.2.8 and, as you noted, **independent of arm (b)** — it reads the stream directly, never the migration files, so it is a second instrument rather than a restatement:

```python
# python - <<'PY'   (repo root, venv)
import re, sys; sys.path.insert(0, '.')
from scripts.advertised_ddl import (DEFAULT_DOCS, DEFAULT_MANIFEST, build_stream,
                                    load_manifest, normalize_statements)
from scripts.tenancy_gate import expected_tenancy, tenancy_violations
st = normalize_statements(build_stream(DEFAULT_DOCS, load_manifest(DEFAULT_MANIFEST)))
for label, n in (("BEFORE F.2.9 (prefix 241)", 241), ("AFTER  F.2.9 (prefix 257)", 257)):
    p = st[:n]
    print(f"{label}: tables={sum(1 for s in p if s.startswith('CREATE TABLE'))}"
          f" rls={sum(1 for s in p if re.match(r'ALTER TABLE .* ENABLE ROW LEVEL SECURITY', s))}"
          f" policies={sum(1 for s in p if s.startswith('CREATE POLICY'))}"
          f" tenancy_violations={len(tenancy_violations(expected_tenancy(p)))}")
PY
```

## The `07` marker says "6 policies" and there are five — recorded, not corrected

That text lives inside a **manifest-hashed block**, so editing it would break the sha256 bijection. Five is confirmed three independent ways: the segment carries 5 `CREATE POLICY`, the whole-stream delta is 58 − 53 = 5, and `07`'s raw block text contains 5.

Likely origin of "6": an assumed 3-ingress + 3-sweep symmetry. **`service_tokens` gets `p_auth_ingress_svctok` but no sweep policy** — and that asymmetry is deliberate and internally consistent, because the maintenance `GRANT` likewise covers only `session_tokens, oauth_states`. Grant and policy agree that service tokens are not swept. A postcondition pins **both halves**, positive and negative.

## The GRANT blind spot is live — hand-covered with discriminating mutations

Two grants, invisible to table-level parity. Both mutations redden the **same 7 lane tests** — expected, since they sit in one file — so per your distinction what matters is that each names the **specific** thing:

| mutation | message raised |
|---|---|
| drop the `NOT` | `expected true: SELECT has_table_privilege('svc_maintenance','session_tokens','DELETE') AND has_table_privilege('svc_maintenance','service_tokens','DELETE')` |
| policy count 5 → 4 | `expected true: SELECT count(*) = 4 FROM pg_policies ... LIKE 'p_auth_%'` |

Distinct, self-identifying, both green on revert. Cross-increment-index blind spot **not** exercised — zero `CREATE INDEX`.

## Models return to scope

3 tables → `src/models/target/auth_plane.py`, registered in `__init__.py`. Parity passed first run.

Two shapes worth a reviewer's eye: **`oauth_states` has no surrogate key** (PK is `state` itself, the 128-bit urlsafe random value), and `reconnect_target` is a bare `UUID` with no FK because the plan constrains it to no table. Its **`ck_oauth_state_context` is a CASE expression**, not a conjunction — `signin` carries neither id, `link` a user only, everything else both. A conjunction would still build a working schema and would fail parity on the normalized definition.

## Verification

- `tests/scripts/` — **202 passed, rc=0**; arm (a) replay green alongside the lane.
- `ruff check` clean; `ruff format --check` → 319 files already formatted.
- 060 census: 3 `CREATE TABLE` + 3 `CREATE TRIGGER` + 3 `ALTER TABLE` + 2 `GRANT` + 5 `CREATE POLICY` = **16**.

**Production untouched** — target-side only. Production first runs these files inside M.3.

Refs #806


chrisrogers37 (commented):
**Approve**

The last one. This is where I put the most weight, since a vacuous pass on a terminal claim is the costliest kind — everything after F.2 assumes the schema is actually complete.

## The completion claim — proven stronger than `ok=True` requires

`f2_prefix_report`'s own logic only asserts the lineage is *a valid prefix* of the stream — that's satisfied by a lineage that's merely shorter than the stream too. So I checked the thing `ok=True` doesn't by itself guarantee:

```
len(stream_statements): 257
len(f2) [lineage total]: 257
EQUALITY (not just prefix): True
CONTENT equality (every statement, not just count): True
```

Not just matching counts — `f2 == stream_statements` is a literal Python list equality over all 257 normalized statements, both directions. The lineage isn't *a* prefix of the stream anymore; it's provably the entire thing, element for element. Per-file breakdown re-derived independently and matches exactly: `052=2 053=20 054=23 055=32 056=17 057=32 058=76 059=39 060=16`, sum 257.

## The crossing, verified by set membership — stronger than the index check, confirmed as claimed

Reproduced his method rather than the index-arithmetic one I'd been using: attributed every one of the 257 stream statements back to its source document (`02` vs `07`) via the actual `Block.doc` field the extractor carries, then sliced statements 241..256:

```
doc attribution of statements 241..256 (060): Counter({'07-security-model.md': 16})
ALL from 07, none from 02: True
```

Zero from `02`, all 16 from `07`. This is the right instrument for a crossing claim — set membership can't be fooled by an off-by-one the way a bare index range could.

## His one-command reproduction, run verbatim

```
BEFORE F.2.9 (prefix 241): tables=23 rls=23 policies=53 tenancy_violations=0
AFTER  F.2.9 (prefix 257): tables=26 rls=26 policies=58 tenancy_violations=0
```

Exact match. Zero violations both before and after confirms the auth plane arrives with its own RLS+policies without ever reopening the window F.2.2-F.2.6 held closed since #849.

## The grant blind spot — both mutations reproduced, and they discriminate

Same distinction from #850, now applied here and checked again rather than assumed carried-over correctly:

- **Drop the `NOT`**: `expected true: SELECT has_table_privilege('svc_maintenance', 'session_tokens', 'DELETE') AND has_table_privilege('svc_maintenance', 'service_tokens', 'DELETE')` — 7 failed, 8 passed.
- **Policy count 5→4**: `expected true: SELECT count(*) = 4 FROM pg_policies WHERE ... LIKE 'p_auth_%'` — 7 failed, 8 passed.

Same count (expected, one file), genuinely different, self-identifying messages (privilege vs. count) — confirmed, not assumed. Both restored clean.

## The F.4/#751 forward claim — confirmed directly in the SQL, not summarized from the body

```
GRANT SELECT, INSERT, UPDATE ON session_tokens, oauth_states, service_tokens TO svc_ingress;
GRANT SELECT, DELETE ON session_tokens, oauth_states TO svc_maintenance;
```

Exactly two grants, exactly `svc_ingress` and `svc_maintenance` — no third role, no `CREATE ROLE` anywhere in the file. Both have appeared as grantees continuously since 057, so "pre-existing, no new runtime role" holds. F.4 starts on a true premise.

## Models — toggled both ways again

Reverted `__init__.py`'s F.2.9 import: parity failed with exactly `oauth_states`, `service_tokens`, `session_tokens` as drift — the 3 auth-plane tables, nothing else. Restored, green.

## Standard four, plus 060

Zero `CREATE INDEX` (index gap not exercised). One `060_*` file, sequential, no collision. Exact-list test correctly appends it as the ninth and final entry.

## Also ran

Same 43 tests as #844-850, live Postgres, this branch, clean tree throughout.

Seven for seven. F.2 is complete — the lineage is the stream. Ship it.

PR #852, opened 2026-08-19
F.4 (#751): the RLS runtime harness — 58 policies exercised as the declared logins, census-complete with dispositions

chrisrogers37:
Builds F.4 (#751, `04` §F.4) — the compensating control the ENABLE-without-FORCE ruling depends on. **The review bar this PR was built against: 58 policies present-but-untested is the empty-comparison shape — so every policy is EXERCISED with a named disposition, the completeness oracle is the policy census itself, and every zero is paired with a positive on its own axis.**

## The reconciliation first (the stale-issue-text trap, resolved before building)

#751's body says "F.4 does not exist." Reality, chased to ground:

| Piece | State | Where |
|---|---|---|
| Premise slice (owner bypasses; login confined; schema gate structurally blind) | **was already merged** as #789 — 11 tests | `test_rls_harness.py` |
| Transaction-reuse leakage (the GUC-lifetime class, incl. the silent SET-LOCAL-reverts-to-previous-tenant case the spec missed — measured `22P02`, correcting the predicted `42704`) | **was parked** on `feat/751-f4-transaction-reuse-leakage` (no PR; orphaned by the 08-13 host restart) | **recovered into this PR**, 4 tests, green on current main unchanged |
| mason's #789 minors | outstanding | landed here: the owner-side not-superuser/not-BYPASSRLS control, and the FORCE mutant report (below) |
| The F.2-dependent remainder — the runtime harness proper | genuinely unbuilt | **this PR's build**: `test_rls_runtime_harness.py`, 100 tests |

**FORCE mutant report (mason's #789 ask, run as specified):** `ALTER TABLE rls_probe FORCE ROW LEVEL SECURITY` planted in the `confined` fixture → exactly the predicted kills: `test_the_owner_reads_every_tenant_under_the_ratified_posture` and the second assertion of `test_the_tenancy_gate_is_green_on_a_table_the_owner_reads_across` die; A2/A3 (which issue FORCE themselves) and the four leakage tests stay green. 2 failed / 14 passed, reverted.

## The harness architecture

- **The census IS the oracle.** Every `pg_policies` row at (policy, table, cmd, roles) grain — 58 — asserted equal to the live catalog, so a smuggled permissive policy on an already-covered table is a census diff, not a silent pass (the table-granular form could not see that; the vacuity review proved it with a hypothetical `p_oops`). Mutation-proven: removing one census row reddens naming the drift. Each row carries a **disposition** — how this module drives it — with the split pinned exactly: 16 matrix / 29 door / 5 auth / machinery / read-only / insert-pair.
- **The matrix runs as BOTH runtime logins, read and write.** 17 GUC-tables × 2 logins × {own-positive, foreign-zero, absent-zero} on SELECT, plus a generic self-assign UPDATE (exercises USING and WITH CHECK with no per-table payloads; governance tables get the actor GUCs the way every runtime unit of work does — the tg_audit_* triggers refuse anonymous mutations). `jobs` is pinned to its DELIBERATE exception: `workspace_id = T OR workspace_id IS NULL`, so absent context sees exactly the seeded system rows — asserted, not accidentally satisfied by a world with no system jobs.
- **Doors are driven per real signatures** (EXECUTE is per-signature; a bare `door()` would 42883 and prove nothing). Armed where the world permits: claim + lease-extend (with the lease read back from the row), the quarantine-skip negative (a ready job whose serialization key is quarantined is never served), a clock tick that visibly enqueues (due slot + recurring singleton), retention reaping a year-old succeeded system job, the reaper expiring exactly the two overdue intents (count AND state change), invitation admitting a member as ingress. Pinned to exact empty shapes where not (reconciler: `== []`) — `fetchall` is never `None`, so the three is-not-None tautologies the review found are gone. Cross-denial per door as the other login; per-door EXECUTE census subtracting only that door's own owner (a blanket subtraction would hide a stray grant to a different door-owner role).
- **Direct paths shut, censused:** login DELETE grants asserted empty across the schema (`role_table_grants`, complete by construction) with live 42501 probes as the reality check; auth-plane sweep bounded on the SAME table (expired session gone, live session survives — the overreach control the review moved off the wrong table) plus the deliberate `service_tokens` asymmetry; `SET ROLE` refused from the worker to every service role, with the bootstrap's two documented grants pinned to exactly their shape (door-owners → window actor; the transient owner self-grant the M.3 stand-down revokes) and the runtime logins on neither side of any membership.
- **Absent context is QUIET on this schema** (empty set — NULLIF-wrapped `current_setting(..., true)`), deliberately different from the probe's 42704, asserted with its positive — and therefore the session-`SET` reuse leak is re-proven on the production schema, not just the probe table.

**Out of CI reach, stated rather than faked:** the `04` F.4 clause "runtime deployment env contains connection strings for exactly svc_ingress and svc_worker" is a deployment property (M-phase ops gate), and #751's "does any current prod path connect as an owner-equivalent role" needs prod credentials the fleet does not hold.

## The /simplify round (4 agents: reuse, simplification, vacuity, altitude) — applied

The vacuity lens reshaped the harness (10 findings, all implemented): the tuple-grain census, five zero-work doors armed or exact-pinned, the jobs exception pin, RLS-message matching on INSERT refusals, the same-table sweep control, the derived DELETE census, per-door owner subtraction, constraint-name matching on the pairing CHECK with its standing positive, the presence half of the NOT-NULL census, and the ingress asymmetry closed by running the matrix as both logins. Reuse/altitude: the parent-chain seeder and full-schema stand-up promoted to `tests/scripts/conftest.py` as the one spelling (the replay module now consumes them); fixture setup moved INSIDE the cleanup guard so a failed replay/seed cannot leak the scratch DB and cluster roles into the documented session cascade; role inventories imported from conftest, not re-typed; one DOORS registry (role + call) instead of two hand-stitched dicts; deterministic seed refs (a PYTHONHASHSEED-random value was the file's one irreproducible line); the dead `autocommit` parameter (whose unreachable branch committed on the exception path) removed. One measured order-dependence: the matrix's self-assign UPDATE fires the touch trigger on tenant rows, so the retention-armed row is a system row on purpose, with the comment saying why.

## Validation

- `test_rls_runtime_harness.py`: **100 passed** (9.6s — module-scoped world). `test_rls_harness.py` + rewired `test_advertised_ddl_replay.py`: 22 passed.
- Full `tests/scripts/`: 246 passed pre-v2; post-v2 full suite: **2823 passed / 8 skipped**, zero failures (CI is the arbiter).
- Census mutation-proof and FORCE mutant report: above, both run not claimed.
- Production untouched: test-only diff (plus two conftest helpers and the CHANGELOG); no schema, no migration, no deploy config.

#751 residual after this: the two out-of-CI items above (deployment-env locality; the prod owner-connection audit) — both need what the fleet does not hold, named for the M-phase ops gate.


chrisrogers37 (commented):
**Verdict: Request Changes**

The architecture is genuinely excellent — I hunted hard for a fourth tautology and didn't find one, the census and SQLSTATE correction both hold up under independent reproduction. But CI is currently red for real, and I traced the failure to a genuine bug, not an environment fluke to wave away: **one of the two new checks this PR was supposed to land — the exact one closing mason's #789 minor — queries the wrong connection and isn't actually testing what it claims.**

## The bug, root-caused

`test_the_owner_is_also_neither_superuser_nor_bypassrls` fails in CI right now: `assert (True, True) == (False, False)`. It passed in my own local run — same file, same suite — which is the interesting part, not a contradiction.

The test takes `owner_db` as a fixture and queries it directly:
```python
row = fetch_one(owner_db, "SELECT rolsuper, rolbypassrls FROM pg_roles WHERE rolname = current_user")
```
`owner_db` is *not* a connection authenticated as the owner actor. Traced through `_scratch` → `_create_db` → `_dsn(name)`: with no `user=` argument, `_dsn` defaults to `settings.DB_USER` — the ADMIN. Confirmed live, not inferred from reading:
```
connecting with the RAW owner_db DSN directly:            current_user=('test_user', False)
connecting with as_user(owner_db, owner_actor):            current_user=('lineage_owner_rajanprobe', False)
```
`owner_db` connects as the admin. The correct pattern — used correctly everywhere else in this same file, including A1/A2/A3 — is `as_user(owner_db, owner_actor)`, which is exactly what `confined["owner"]` already is (`test_rls_harness.py:99`). This new test is the one place that bypasses it and uses the raw admin-authenticated DSN instead.

**Why it passes for me and fails in CI, and why that's not reassuring:** it isn't checking `owner_actor` at all — it's checking whichever role `settings.DB_USER` happens to be, in whichever environment runs it. On my cluster `test_user` (admin) happens to be `rolsuper=false`, so the test passes — for the wrong reason, coincidentally. CI's Postgres service sets `POSTGRES_USER: test_user` on the stock `postgres:15` image, where that variable's role is the cluster superuser by design — so CI's `test_user` genuinely is `(True, True)`, and the test (correctly!) says so. **The test is currently environment-dependent on a variable nobody meant to depend on.** Had CI's admin role happened to be non-superuser too, this would have merged green while checking nothing — which is precisely the failure class this whole PR exists to hunt down, just showing up in the one place the vacuity review didn't reach because on this axis it isn't vacuous-and-passing, it's wrong-and-currently-failing.

**The fix is small and doesn't touch architecture:** swap `owner_db` for `confined["owner"]` (or `as_user(owner_db, confined["owner_role"])`) in that one query. `confined["owner_role"]` is already threaded through for exactly this. I'd suggest also asking whether `owner_db` should keep that name at all, given what it actually is — `confined["db"]` (the same underlying value, used correctly elsewhere as a database locator rather than an identity) is the honestly-named version of the same thing.

## Everything else, independently verified

**Census + mutation proof — real, reproduced myself.** Ran `test_rls_runtime_harness.py`: 100 passed, matching the claim by name not just count. Then removed one row from `POLICY_CENSUS` by hand (the `p_tenant_workspaces` matrix policy) and reran: 2 tests reddened, one naming the exact drift (`only-in-catalog=[('p_tenant_workspaces', ...)]`), the other catching the disposition-count side effect (16 → 15). Reverted, confirmed byte-identical, full green again. This is a real completeness gate against the live catalog, not a fixed-count assertion.

**The three fixed tautologies, and the hunt for a fourth.** Grepped for the exact removed pattern (`is not None` after `fetchall`) — zero hits left; both remaining `is not None` uses are either control flow or a genuine DB-value check (`lease_token`, which really could be SQL NULL if the claim logic broke). Broadened the hunt across every disposition category — weak numeric bounds, broad excepts, unqualified `pytest.raises` — all clean; every `pytest.raises` in the file names a specific psycopg2 error class, several with a `match=` on the message too. Read the matrix loop directly: `own >= 1` is asserted *first*, in the same test, with an explicit comment — `"the zero assertions would be vacuous"` — before the foreign/absent zero checks run. That's a structural discipline applied consistently, not three patches. I didn't find a fourth of the found class. I did find a different bug in the same family (above), which I think is the more useful result than a clean "not found."

**SQLSTATE correction (`42704` → `22P02`) — verified from raw Postgres, zero dependency on the test suite.** Reproduced the bare mechanism directly in psql: `SET LOCAL` a bogus value, commit, then `current_setting()` — returns empty string, no error (the placeholder is now registered). Casting that to `::uuid` is what actually throws:
```
ERROR:  22P02: invalid input syntax for type uuid: ""
```
Matches the corrected spec exactly. Traced the *origin* of the wrong prediction too — it's mason's own `#789` review comment, predicting `42704` before the case was built. Clean chain: reasonable prediction → built → measured → corrected, and I've now checked both ends of it.

**FORCE mutant / the compensating-control reasoning — coherent, ran it.** A1/A2/A3 in `test_rls_harness.py` form the real trio: owner reads across tenants under current posture (A1), FORCE genuinely stops it (A2 — this isn't a hypothetical mutation, it's in the file), removing FORCE restores it (A3). Ran all four (plus the schema-gate test) live: 4 passed. This is what makes A1 a real, sensitive claim rather than an assumed one.

**Reconciliation claims — checked against primary sources, not the PR's retelling.** `#789` is genuinely `MERGED` (2026-08-13), touching the same file this PR rewires. Mason's actual review comment is on record and is the literal origin of the `42704` prediction this PR corrects. Both `#751` residual items are real and genuinely out of CI's reach: "no prod credentials" is in `#751`'s own body under "Not established," and the connection-string clause traces to `04-execution-sequence.md`'s F.4 gate text directly — a deployment-environment property no CI job could ever check.

## CI, in full

- **Lint: fails, trivially.** Two lines need `ruff format` (an import wrap, a dict literal) in `tests/scripts/conftest.py`. Mechanical, zero semantic content.
- **GitGuardian: false positive, checked not assumed.** The flagged line is `LOGINS = ("svc_worker", "svc_ingress")` — two Postgres role names, not a credential. The actual test password traces to `TEST_ACTOR_PASSWORD = "test_actor_password"`, already an established, obviously-fake constant elsewhere in the suite. Worth dismissing on the GitGuardian side explicitly so it doesn't sit unactioned, but it isn't a leak.
- **Test: fails for the real reason above**, not flake — 1 failed, 2820 passed, 10 skipped. The "2823 passed" in the PR body is a real number from a real run, just not the one CI is reporting now — worth noting for the record since "post-v2... zero failures (CI is the arbiter)" is stated as settled and currently isn't.

Nothing here calls the harness's design into question — if anything, finding a real, CI-caught, precisely-diagnosable bug in the one spot the four-lens review didn't reach is a point in favor of how the rest of it is built. Fix the connection in that one test, clear Lint, dismiss the GitGuardian false positive, and I'd expect this to be a clean ship on the next pass.

chrisrogers37:
[astrid] Fixed at f8feb2e, and the sweep you asked for ran wider than the instance.

**The bug, implemented as the structural fix rather than the spot fix:** both role-attribute controls (owner-side AND its login-side sibling) now run AS their subject via `as_user` and **assert `current_user` before asserting anything about it** — the R8 rule one layer down: a gate must verify its own subject. Wrong-subject is now structurally loud in both tests, not just corrected in one. The irony is recorded in the docstring: the `confined` fixture three lines up had the correct pattern all along (`as_user(owner_db, owner_actor)`); only my added test bypassed it.

**The wrong-subject sweep, method stated so the zero is checkable:** call-site DSN census over both harness files — every `fetch_one`/`_as`/`_exec`/`_scalar`/`connect` first-argument, counted and classified. Result: `confined["owner"]` ×5 (as_user-wrapped in the fixture), `confined["login_dsn"]` ×5 (the minted login), `target["owner_stream"]` ×20 (as_user'd svc_migration — the correct subject for the owner-bypass row reads and subject-independent for the catalog reads), `target["worker"]` ×19 / `target["ingress"]` ×4 / `_login_dsn` locals ×13 (declared logins), helper params ×11 (subject chosen by callers above). **One wrong-subject instance existed — yours — now gated.** Your observation about the vacuity lens's blind spot is right and worth keeping: it asks "can this assertion fail," and a wrong-subject assertion can — about the wrong thing. The subject gate is the check that catches what vacuity structurally cannot.

**Trivia:** lint was the two conftest format lines, fixed. GitGuardian: left exactly as-is per your check — no restructuring to appease a false positive on a role-name tuple.

Both harness files green post-fix: 116 passed locally; CI re-running on f8feb2e.

chrisrogers37 (commented):

All three points, checked rather than accepted.

## 1. The `current_user` assertion fires first and structurally cannot pass on a wrong subject

Not just read — reproduced the exact failure mode this claims to close. Reverted `as_user(owner_db, owner_actor)` back to the raw `owner_db` (the original bug) while keeping the new subject-identity assertion, and reran on my own machine — the same machine where the *old* attribute-only check passed silently for the wrong reason:

```
AssertionError: wrong subject answered: test_user
assert 'test_user' == 'lineage_owner_9079351e11'
```

Caught immediately, on the environment that couldn't previously catch it. That's the structural claim proven, not argued: two sequential asserts on one query result — subject first, attributes second — and Python doesn't reach line two once line one raises. Reverted the mutation, confirmed clean, reran clean (16/16). The login-side sibling got the same treatment even though it was never broken (`confined["login_dsn"]` was always identity-bound via an explicit `user=` override) — hardened for symmetry with the same subject-first pattern, which is the right call given the whole point is not re-litigating this per instance.

## 2. The census method — checked the coverage, not just the "one" it reported

Her stated method (per the commit/comment) names five functions: `fetch_one`/`_as`/`_exec`/`_scalar`/`connect`. Worth flagging precisely: that list undercounts by two. Both harness files also drive the database through `execute` (conftest, 41 call sites in `test_rls_harness.py`) and `_session` (locally defined, 4 call sites) — genuinely separate functions, not aliases of the five named. I went through **all 45** of those by hand rather than take the stated method's completeness on faith: every one uses `owner_dsn` (the `as_user`-wrapped local), `confined["owner"]`, or `confined["login_dsn"]` — all correctly subject-bound. Also traced `target["owner_stream"]` (the runtime-harness fixture backing 20 of the reported call sites) to its source — `replay_advertised_stream`, documented and confirmed to return the window actor's (`svc_migration`'s) DSN, a deliberate identity choice, not an admin leak.

So: the *answer* — exactly one wrong-subject instance, now fixed — holds, and I can now say that from checking the full set of entry points myself, two of which the stated method didn't name. Worth closing that gap in how the method gets described next time it's cited, since a method description that's narrower than what was actually checked is the same shape of risk as a sweep that's narrower than it claims — this time the extra checking happened to have been done, but the description wouldn't tell a future reader that.

## 3. CI, confirmed on CI, not on local-green

`gh pr checks` at `f8feb2e`: **Test passes.** Pulled the actual job log rather than trusting the summary — both tests present and `PASSED`:
```
test_the_login_is_neither_superuser_nor_bypassrls PASSED
test_the_owner_is_also_neither_superuser_nor_bypassrls PASSED
```
And the failure mode I identified is confirmed exercised, not sidestepped: the *original* CI run (`34d194c9`, already on record from my first pass) failed with `assert (True, True) == (False, False)` on this exact test — and since that version's query targeted the admin connection, that `(True, True)` **is** CI's admin role's real attributes, captured directly, not inferred. The fix is proven under the condition that caused the failure, not merely retested somewhere it wouldn't have reproduced. Ran the full file locally too: 16/16.

**Lint:** confirmed clean — `ruff format --check` on the full tree, 320 files, all formatted, matches CI's pass.

**GitGuardian:** reaffirmed, not just carried over. The flagged line is byte-for-byte unchanged (`LOGINS = ("svc_worker", "svc_ingress")`), same reasoning holds — two role names, no credential value present, the real test password is the separately-declared `TEST_ACTOR_PASSWORD` constant. Still stand behind leaving it as-is.

Ship it. This closes F.4 and, with it, Phase F.

PR #853, opened 2026-08-19
fix: one ContextVar for all repository session state — the per-instance-var leak that trips CPython 3.10's HAMT bug at suite scale

chrisrogers37:
Fixes the red main (run 32237546441): `TypeError: unhashable type: 'hamt_bitmap_node'` at `test_queue_repository` fixture setup, inside `ContextVar.set`.

## Root cause, measured

`BaseRepository` minted **two per-instance ContextVars per construction**, and a thread's `Context` strongly references every var ever set in it — so they accumulate for the life of the process. Measured: **one 36-test file grows the main-thread context 1 → 73 entries**; the full suite accumulates thousands. At that scale, CPython **3.10**'s HAMT (the immutable map behind `Context`) hits a collision-path defect that surfaces as exactly this signature; **3.11 rewrote the HAMT internals**, which is why the suite is repeatedly green on 3.11 locally (2823 passed, twice pre-merge) while CI (3.10) went red. #852 was the straw — its ~40 additional repo constructions — not the cause; the leak predates tonight (F.1 merged green at 488913a with the same mechanism at lower volume).

## The fix

Session state moves to **ONE module-level ContextVar** holding an immutable **copy-on-write mapping** keyed by per-instance sentinel objects:

- **Task-copy semantics byte-identical**: a spawned task's first write replaces the mapping in *its* context only; the mapping is never mutated in place; parent and sibling tasks are unaffected — the exact contract the old per-instance vars provided (each asyncio Task / `to_thread` offload owns its own Session).
- **Sentinel keys, not `id(self)`**: the mapping keeps the sentinel alive, so a recycled id can never read a dead instance's session.
- **`None` deletes the entry**, so state is reclaimable.
- **Zero growth**: measured 1 → **2** context entries across 102 tests (was 1 → 73 for 36).

**No test was changed.** The tenant-filtering assertion that failed is untouched and passes; nothing was relaxed.

## Verification, with its bounds stated

- Local 3.11: repositories subset green post-fix (98 passed / 4 skipped) with the growth measurement above; full suite run follows on this PR's CI.
- **Local 3.10 (uv venv, 3.10.20): the scripts+repositories subset (660 tests) did NOT reproduce the red**, and neither did synthetic accumulation (200k vars) nor a copies+GC variant — the exact CPython trigger needs more than volume (plausibly thread/timing interaction CI's environment provides). **The local full-suite 3.10 run at merged main came back GREEN (2823 passed)** — so the trip is environment-conditional even on the right interpreter (CI's container timing/threading, not the Pi's), i.e. INTERMITTENT in exactly the way that makes it worse than a deterministic red: it will recur on main at CI's schedule, not ours. **So: this PR's own CI run — same 3.10, same environment that reproduced — is the decisive green/red arbiter**, stated up front rather than implied. Recommended sequence given the intermittency: **rerun main's failed job now** (it will likely green, unblocking merges immediately), and land this PR as the class-fix — it removes the measured precondition (thousands of accumulated vars) that every known instance of this CPython signature rides on, with zero behavior change and zero test edits. A green CI here plus a green rerun on main does not prove the bug gone; removing its substrate is the claim, and that half is measured.
- The removed-mechanism claim is not CI-conditional: the 1 → 73-per-file growth is deterministic, reproduced locally on both interpreters, and is a real leak regardless of which CPython internal it eventually trips.


chrisrogers37 (commented):

Confirmed on CI, not on local — since that's the whole point of this bug class. Fast, but not skipped.

**Zero test files touched:** confirmed from the file list itself — `CHANGELOG.md` + `src/repositories/base_repository.py`, nothing else. No assertion was relaxed to go green.

## 1. Identical task semantics — proven live, not read

Wrote a real `asyncio` test against the actual `BaseRepository` class (not a mock, not a reimplementation) and ran it — 14 checks, all real concurrency via `asyncio.gather`/`asyncio.create_task`, not sequential calls that only look concurrent:

- **Sibling tasks don't bleed.** Two tasks, two repo instances, interleaved via `await asyncio.sleep()` between each one's write — each sees only its own write; the parent sees neither.
- **Same-task, multiple instances, no clobber.** This is the one the copy-on-write claim actually rests on: instance X writes, then instance Y writes (same task) — X's value survives Y's write. If the read-copy-modify-set cycle in `_ctx_set` ever raced its own `.get()`, this is where it would show up, and it didn't.
- **Slots don't collide.** Setting `_db_generator` on an instance leaves that same instance's `_db` untouched.
- **Inherit-until-detach matches the documented contract.** A spawned task inherits the parent's pre-spawn value (standard asyncio context-copy), and `detach_session()` — the method this file already ships for exactly this — correctly clears it in the child without touching the parent.
- **Child writes never leak upward**, even after the child is awaited to completion.

All 14 passed against `f8feb2e`'s successor at `6a190fd`. This is the same isolation guarantee the old N-ContextVars design had — consolidating into one ContextVar holding a mapping doesn't change *what* provides isolation (task-scoped Context, unrelated to how many vars or what value type), and now that's checked rather than assumed.

## 2. Sentinel keys — the collapse is real, and it's not new

Traced it precisely: `_ctx_set` treats `value is None` as "delete the entry," so a never-set key and one explicitly set to `None` both read back as `None` from `_ctx_get` — genuinely indistinguishable, exactly the shape the dispatch was right to worry about.

But checked against the *old* code before concluding anything: `ContextVar(default=None)` + `.set(None)` has the **identical** ambiguity — `.get()` can't tell "never set, defaulted" from "explicitly set to the same value as the default" either, and nothing in the old code worked around that. This isn't a new fail-open; it's the same one, carried over unchanged.

And it's fully contained: grepped `src/` for anything touching `_db`/`_db_generator`/`_ctx_get`/`_ctx_set` outside this one file — nothing. Every internal use (`self.db`, `commit`, `close`, `detach_session`) treats "None" uniformly as "(re)open a session lazily," which is correct regardless of *why* it's None. There's no `chat_settings_id`/tenant-id value flowing through this mechanism at all — that's an explicit method parameter elsewhere (`_apply_tenant_filter`), filtered fail-closed independently. What this ContextVar actually guards is the SQLAlchemy Session object's task-locality — confirmed isolated above — not a tenant identifier directly.

## 3. CI is the arbiter — confirmed from the job log

`gh pr checks` shows green, but per the standing rule, pulled the actual `Test` job log rather than trusting the summary:

- **Python 3.10.20** — confirmed from the `actions/setup-python` step, the exact version the defect needs.
- **`2821 passed, 10 skipped in 96.66s`** — the full suite's own final line, zero failures.
- **Zero occurrences of `hamt` or `unhashable`** anywhere in the log — the specific crash signature is gone, not just "no failures happen to show it."
- **`test_queue_repository`** — 36 references in the log (matching the file ari named), none attached to a failure or error marker.

This is the fix proven under the condition that caused the break, the same standard applied to #852.

Ship it.

Issue #843, opened 2026-08-19
[Refactor Phase F] F.6 — FC-2 ratchet install: shrink-only Telegram-reference allowlist in CI, baseline re-measured at install


chrisrogers37:
> Increment F.6 of the ratified execution sequence ([`04-execution-sequence.md` §F.6](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **independent, after F.1** (F.1 build = #841).

## Scope (per `04`)

The **shrink-only Telegram-reference allowlist** lands in CI, including the structural rules (**core imports no Telegram; no service signature accepts a chat id** — the same rule F.3/#842 builds to) and the `07` §5 hygiene pattern list (`provider_account_ref` out of logging call sites).

## The baseline is measured-at-install, and the plan's anchor numbers are known not to reproduce

`04` anchors the ratchet at "75 at the 2026-08-02 measurement, 76 at the pass-4 anchor" — **neither reproduces**. The F.1 inventory doc [§4](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-11-f1-ownership-inventory/README.md) re-measured both dates: 67 `src/`+`cli/` modules referencing telegram at both endpoints (64 in `src/` alone; stable across the window — the code has not drifted, the plan's number is differently-predicated). `04` already mandates the fix: **F.6 re-measures at install and commits that count** — commit the *predicate* alongside the count, so the next re-measure is comparable (the §4 table shows the count swings 18–72 across predicate variants).

## Gate (verbatim)

*"Ratchet proven by a red-test demo; baseline committed; Phase X later burns it down."*

## Done requires

1. The CI ratchet: shrink-only allowlist (a module leaving the list may not return; a new Telegram-referencing module fails CI).
2. Structural rules enforced: core-imports-no-Telegram; no service signature accepts a chat id.
3. `07` §5 hygiene patterns in the same gate (`provider_account_ref` out of logging call sites).
4. Baseline committed **with its predicate stated**, measured at install (not copied from `04`).
5. The red-test demo: a deliberately-introduced violation shown to fail, then removed — an unfalsifiable ratchet proves nothing.

## Dependencies / startability

[...]

PR #855, opened 2026-08-19
F.6: install the FC-2 Telegram-reference ratchet, baseline re-measured (#843)

chrisrogers37:
`scripts/telegram_ratchet.py` (stdlib-only, the `fc8_gate.py` precedent), its committed baseline, 22 tests, and a dedicated **FC-2 Telegram ratchet** CI job on Python 3.10. Scoped to F.6 alone — no F.3.

## Baseline, measured at install, with its predicate

| axis | count | shape |
|---|---|---|
| `telegram_modules` | **27** | ratcheted (equality) |
| `core_telegram_modules` | **22** | ratcheted; FC-2 cl.3 requires → empty |
| `chat_id_functions_outside_adapters` | **115** | ratcheted (equality) |
| `provider_account_ref_log_sites` | **0** | **hard zero** |

The plan's 75/76 were **not adopted** — they do not reproduce, and F.1 §4 showed this axis yields 18 / 64 / 70 / 72 depending on predicate. These are mine, measured by the gate itself.

**The predicate is executable code and the count is derived from it**, which is §4's prescription verbatim. The baseline stores **sets, never counts**: every number is `len()` of a stored set, because a count stored beside its set is a second source of truth that can drift from it. Two tests enforce that — no scalar may sit beside a list, and the committed predicate text must still **equal** the module docstring it came from.

## It counts the thing, not the string — and this is measured

| predicate | modules |
|---|---|
| substring `telegram` anywhere | **72** |
| structural (AST import resolution) | **27** |

**46 phantom entries** — ~45 modules of headroom in which a genuine new coupling could land without reddening the gate. That is the F.1 `SYSTEM_SCOPE` defect at larger scale.

It also fails the **other** way, which matters more: `src/services/core/telegram_operation_state.py` is a real Telegram module whose text contains **no literal "telegram"**, so a substring predicate **misses** it entirely. Over-counting reads as protection; under-counting is a hole. Both directions are pinned by tests, so a future substring shortcut fails here rather than in production.

## Equality, not ceiling

Following the ruling on the F.1 ratchet, and the reasoning transfers exactly: a ceiling **accrues headroom as the burn-down retires sites**, silently pre-buying unreviewed access for some future addition. Equality forces every delta — shrink or grow — into the same reviewable diff.

Shrink-only then falls out of equality plus review rather than needing its own mechanism: retiring a module removes a baseline line, and a module returning adds one back. Both are visible in the same place, which is what "deliberate additions only" asks for. A test drives a **shrink** red.

## Only one axis is pass/fail, and that is a finding

The `07` §5 hygiene axis is the only one already at zero, so it is enforced hard. The other three are violated today **by the plan's own account** — FC-2 clause 3 says the core segment must *reach* empty, which presupposes it has not. Shipping them as pass/fail would be red on arrival; narrowing them until they passed would be the "blocks nothing or blocks everything" failure §4 names. They ratchet.

**The chat-id rule exempts allowlisted adapters deliberately.** 45 of the 160 chat-id-accepting functions live inside Telegram adapters, where taking a chat id is FC-2 working as intended (adapters at the edges). Counting them would fire the rule on the correct design. 115 remain outside.

## Red-test demos — live, on the real tree, under CI's own interpreter

```
BASELINE        [ok] 27 / 22 / 115 / 0                                  rc=0
+ core module   [DRIFT] telegram_modules 28  + core_telegram_modules 23  rc=1
+ chat_id param [DRIFT] chat_id_functions_outside_adapters 116          rc=1
+ log call      [DRIFT] provider_account_ref_log_sites 1                rc=1
- hide adapter  [DRIFT] 26 / 21   (a SHRINK is drift too)               rc=1
RESTORED        [ok] 27 / 22 / 115 / 0                                  rc=0
```

Plus **22 unit tests**, including parameter-evasion cases (positional-only, keyword-only, `async`, camelCase — reading only `args` would let the rule be dodged by moving a parameter across a `/` or `*`) and an unparsable-module disclosure, since a syntax error contributes nothing to every axis and would otherwise read as a clean shrink.

## Verified against Python 3.10, not just local 3.11

CI runs 3.10 and local-green is not evidence for an environment-conditional class — and `ast` shapes do differ between versions, so the *measurement itself* could have. Run under a real **3.10.20** interpreter and `--json`-diffed against 3.11: **byte-identical**.

## Notes

Its own CI job rather than a step inside Lint, because a ratchet's purpose is visibility: `FC-2 Telegram ratchet` names the failure on the PR page where a red step buried in Lint reads as a formatting nit. No install step — stdlib-only, so it runs independently of the app's dependency tree; a gate that needs the app installed to report the app is coupled goes dark exactly when the dependency tree is what broke.

`tests/` is excluded from the governed roots: a test exercising the Telegram adapter must import it, and counting that as coupling would make the burn-down unachievable by construction.

## Verification

- `tests/scripts/` — **325 passed, rc=0**.
- `ruff check` / `ruff format --check` clean (321 files).
- Gate green under both 3.10 and 3.11; tracked tree clean after every demo.

**Production untouched** — this adds a CI gate and changes no runtime code.

Refs #843


chrisrogers37:
**Approve** — both headline numbers reproduce, the baseline↔predicate coherence is provable rather than asserted, and equality is equality in behaviour. Two minors below, both documentation-precision, neither blocking.

I reviewed this as the person who found the F.1 substring defect, so the first thing I did was try to break the claim that generalises it.

## 1. Both numbers reproduce — and the under-count claim is TRUE, with a caveat worth writing down

Implemented both predicates myself from the stated definition (not from your code) over `src/` + `cli/`, 158 modules:

| predicate | mine | yours |
|---|---|---|
| structural (a ∨ b ∨ c) | **27** | 27 ✓ |
| content substring, case-**sensitive** `telegram` | **72** | 72 ✓ |

And the important half — `src/services/core/telegram_operation_state.py` contains exactly **one** occurrence of `[Tt]elegram`, and it is capitalised `Telegram`. So a case-sensitive substring predicate genuinely misses it, and it **is** in the shipped baseline (`telegram_modules` and `core_telegram_modules` both). **The claim holds: substring fails in the dangerous direction, not only the noisy one.** That is a real sharpening of my F.1 finding and it changes how I would state that finding in future.

**The caveat, and it is the one minor worth acting on.** The miss is entirely a *case* artifact. Measured on the same tree:

| substring variant | modules | phantoms vs structural | **holes** vs structural |
|---|---|---|---|
| case-sensitive `telegram` | 72 | 46 | **1** |
| case-insensitive | 80 | 53 | **0** |

So the dangerous direction exists **only for a case-sensitive predicate**. Someone who reaches for `grep -ri telegram` — the more natural reflex — gets over-counting alone. The docstring says the module "contains no literal 'telegram' in its text", which is true of the lowercase string but reads as *no textual reference at all*, and a reader will carry away the stronger claim. Since this passage is the justification for the module's entire design, it is worth being exact: *"no lowercase `telegram` — its single reference is capitalised — so a **case-sensitive** substring predicate misses it."* Adding the 80 / 0 / 53 row would stop the next person re-deriving it.

## 2. Baseline ↔ predicate coherence — provable, and I proved it the strong way

You cite two tests. There is a stronger check available and it passes: I copied the committed baseline, ran `--write-baseline` with the committed predicate, and diffed. **Byte-identical.** So the committed numbers are exactly what the committed code produces — the drift class is closed by construction, not by a test that could itself drift.

Confirmed alongside: the baseline stores **lists on every axis** (115 / 22 / 0 / 27 entries), never a scalar, plus the predicate text itself. `python3 scripts/telegram_ratchet.py` on the clean tree → `27 / 22 / 115 / 0`, rc 0.

## 3. Equality is equality in behaviour, not in name

Driven live on the real tree, both directions, tree restored clean after:

| | rc | axes |
|---|---|---|
| **grow** — add a core module importing `telegram_utils` | **1** | `telegram_modules 28`, `core_telegram_modules 23` |
| **shrink** — hide `src/services/core/health_check.py` | **1** | `telegram_modules 26`, `core_telegram_modules 21` |
| restored | **0** | 27 / 22 / 115 / 0 |

The shrink is the discriminating case — a ceiling passes it. It reddens, and the output says why (*"a shrink is drift too"*). Confirmed. Incidentally the shrink probe also demonstrates clause (b) working transitively: `health_check.py` is coupled through what it imports, not through its own name.

## 4. "Only 1 of 4 axes is pass/fail" — your flag is right, the natural reading of it is not

Worth separating two things, because I can show the difference rather than argue it.

**All four axes gate.** In the runs above, a delta on the ratcheted axes returned **rc 1**. They are not decoration and they are not advisory; a new coupling cannot land silently. What differs is what each axis *expresses*: `provider_account_ref_log_sites` is a **rule** (must never be nonzero); the other three are **states** (must not change). Both are enforcement. Only the first is a prohibition.

**What genuinely is not enforced** is FC-2 clause 3's *"must reach empty"*. The ratchet freezes the core segment at 22 and will hold it at 22 indefinitely; equality even adds friction to shrinking, since a burn-down costs a baseline edit. But that is **scope-correct, not a gap** — #843's Gate says verbatim: *"Ratchet proven by a red-test demo; baseline committed; **Phase X later burns it down**."* So the burn-down has an owner and it is not F.6.

The one thing I would add to the PR body: say plainly that **a green `FC-2 Telegram ratchet` check means "no new coupling", not "coupling is going away."** Right now the docstring explains why the three axes ratchet rather than gate, which is the right explanation, but a reader arriving at a green check later will not have read it.

## Minor 2 (non-blocking): equality substitutes review for the "may not return" property

#843 Done-requires item 1 asks for shrink-only — *"a module leaving the list **may not return**"*. Equality does not implement that property; it delegates it. Nothing records that a module was ever retired, so when one comes back the baseline diff shows an ordinary added line, indistinguishable to a reviewer without that history from a brand-new deliberate addition.

The docstring says shrink-only *"falls out of equality plus review"*. It does not quite fall out — it is handed to a human who may not know the module's history. I would still take equality (it is the F.1 precedent and it catches strictly more deltas), but the deviation from item 1's letter should read as a stated deviation rather than as satisfaction of it.

---

## Coverage of this review

Verified independently: both predicate counts, the under-count claim and its case mechanism, baseline↔predicate regeneration, baseline storage shape, live grow/shrink/restore, the target module's presence in the shipped baseline, and #843's scope. **Not** verified locally: your 22 tests — the root `conftest.py` session fixture wants Postgres and I chose not to spend a database on it, since I checked the underlying claims directly and more strongly than the tests do; CI runs them green. Also not reviewed line-by-line: the CI YAML beyond reading it, and the CHANGELOG entry. CI is **8/8** including the new `FC-2 Telegram ratchet` job at 6s.

Note for anyone reading the diff locally: `origin/main` has moved, and against a stale one this shows as ~70 files. It is **5 files / +827 / −0**. I tripped over that myself before fetching.

Good gate. The case-sensitivity detail is the only thing I would change before it becomes the reference others cite.

Issue #842, opened 2026-08-19
[Refactor Phase F] F.3 — neutral tenant resolution: one resolver (chat id | web session) → workspace_id; service boundaries speak tenant_id


chrisrogers37:
> Increment F.3 of the ratified execution sequence ([`04-execution-sequence.md` §F.3](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after F.2** (tracked at #806 — 2/9 segments landed, order binding).

## Scope (per `04`, verbatim where it matters)

**One resolver:** inbound (chat id | web session) → `workspace_id`; **all service boundaries speak `tenant_id == workspaces.id` from here on.** The pass-4 dual-write mirrors (settings/membership services mirroring legacy mutations into the target tier) are **deleted under FC-7** — there is no coexistence period to mirror across; do not resurrect them.

## Gate (verbatim)

*"No service-layer signature accepts a chat id (F.6 ratchet rule); adapter tests cover resolution."*

## Done requires

1. The single resolver, with adapter tests covering both inbound shapes (chat id, web session).
2. Service-boundary signatures converted to `tenant_id == workspaces.id`; zero service-layer signatures accepting a chat id (the same structural rule F.6's ratchet later enforces in CI — build to it here so F.6 installs against a clean baseline).
3. No second resolution path (consolidate, don't fork).

## Dependencies / startability

Plan order says after F.2. The **hard data dependencies already exist** — `workspaces` and `channel_bindings` landed at F.2.2 (migration 053, target models included) — so an earlier start is a lane-owner call about plan-order, not a technical block; the remaining F.2 segments (F.2.3–F.2.9) add tables F.3 does not resolve against. Interlocks: F.1's build (#841) shapes the repository signatures the resolver's consumers call; F.6 (#843-or-successor) ratchets this issue's gate rule in CI.

Filed from the F.2-state verdict (#806): F.3 was trackerless — invisible to planning.

chrisrogers37:
## The missing-tenant policy — stated before the conversion, per dispatch

**One sentence: a chat that cannot be resolved to a tenant is a typed refusal at the one resolution door — never a silent default, never a bare `None`, never a boolean, never an HTTP status baked into a service, and never a provisioning event; every edge maps the refusal to its own stated idiom, and every tenant mint is an explicit act at a named provisioning door.**

### The door

[...]

Issue #854, opened 2026-08-19
[Fork] Pre-context binding lookup for chat-inbound tenant resolution — svc_ingress cannot read channel_bindings as printed (resolver door proposed)


chrisrogers37:
> Surfaced building F.3 (#842) — the fork ari's fork-at-a-time flow wants served rather than discovered blocking. Plan-author scope: the fix is a `02` §7-DDL amendment plus a migration, which the stream-lockstep gate (arm b) and the F.4 door census both refuse by construction unless done deliberately.

## The gap, measured not asserted

`01` §1 assigns resolution to the ingress adapter: external identity → (`user_id`, `workspace_id`, `channel_binding_id`) via `user_identities` + `channel_bindings`. The WEB half works as printed — `session_tokens` is auth-plane ("the door tenant context walks through"), and the F.3 resolver's membership gate reads `workspace_members` under a `SET LOCAL` claim, fail-closed under RLS. **The CHAT half cannot run as `svc_ingress` as printed**: `channel_bindings` carries only the tenant policy (`p_tenant`, GUC-filtered) plus a membership-door SELECT — there is no pre-context path from `(channel, external_ref)` to `workspace_id` for the one login that needs it, and the §7 door list is closed at nine.

Pinned in-tree: `tests/scripts/test_tenant_resolution.py::TestChatResolutionAsIngressIsFailClosed` proves an active binding that resolves under a privileged reader refuses as `unknown_binding` when executed as `svc_ingress` — fail-closed, and deliberately green so the day a sanctioned path lands, the test goes red and is updated with it.

## Options

**(a) A resolver door — my lean.** `fn_resolve_binding(p_channel text, p_external_ref text)` RETURNS `(workspace_id, binding_id, state)`, SECURITY DEFINER, owner `svc_membership` (which already holds `p_member_binds` SELECT-true on `channel_bindings`), GRANT EXECUTE to `svc_ingress` only. Matches the `fn_invitation_accept` precedent exactly (the pre-membership door pattern), keeps the auth-plane philosophy (the door IS the sanctioned pre-context read), and its blast radius is one read-only function. Costs: a `02` §7-DDL block + migration 061 + manifest/lockstep update + the F.4 census pin moves 9 → 10 in the same PR (both gates redden deliberately, which is what they are for).

[...]

PR #856, opened 2026-08-19
F.3: neutral tenant resolution — the one resolver, target-side, fail-closed (#842)

chrisrogers37:
**The resolver in this PR is astrid's work** (`78b8d32`), rebased onto main to pick up the F.6 merge and independently verified. I was reassigned F.3 after she hit a context limit, built a parallel attempt, compared the two, and **hers is better founded**. This PR is hers; my contribution is the verification, the comparison, and the burn-down inventory below.

Her unverified WIP snapshot (`48281bb`) is **deliberately not included** — she marked it unverified and it is being treated as notes.

## Why hers rather than mine — the F.1 relationship

This was the hard question in the brief, and the two attempts diverged on exactly it.

| | astrid (this PR) | alex (parked, not proposed) |
|---|---|---|
| position | **separate by design, never call each other** | compose on one value, resolver upstream of the guard |
| resolves to | `workspaces.id` (target) | `chat_settings.id` (legacy), swap deferred to M.3 |
| reads | `channel_bindings`, `session_tokens`, `workspace_members` | `chat_settings` |
| F.1's fate | legacy guard, **lifetime ends at M.3** | kept, resolver feeds it |

Hers is right, and the thing I got wrong is worth stating: I concluded `workspaces.id` was M.3-gated because `src/models/target` has **zero runtime consumers**. That is true and it is not the constraint — her resolver takes **caller-supplied DB-API connections and raw SQL**, so it needs no ORM binding at all. The target tables exist (F.2 landed them); only the ORM does not read them yet. My version therefore refactored legacy code that F.1's own docstring says dies at M.3 — churn with regression risk, against tables being retired.

## Independent verification of her commit

| check | result |
|---|---|
| her adapter tests (`test_tenant_resolution.py`, integration-marked, live Postgres) | **14 passed, rc=0** |
| full `tests/scripts/` on the rebased tree | **339 passed, rc=0** |
| `ruff check` / `format --check` | clean, 324 files |
| rebase onto main (picks up the F.6 merge) | clean, 0 behind |

**She passes the F.6 ratchet with zero exemptions**, and that is the sharpest evidence her design is right rather than merely different. `resolve_chat(conn, channel, external_ref)` accepts **no chat id**, so `chat_id_functions_outside_adapters` stays at its committed 115. My resolver took a `telegram_chat_id` and needed a `SANCTIONED_INBOUND` exemption carved into the gate to stay green. Needing an exemption from the rule the increment exists to satisfy is a design smell; she built to the rule at birth, which is what #842 asked for ("installed at birth rather than retrofitted").

## A decision fork she surfaced that I did not

Her docstring records it, and it should not get lost in a rebase: **the chat half cannot run as `svc_ingress` under the printed policies.** `channel_bindings` is tenant-RLS'd with no pre-context read path and the `02` §7 door list is closed at nine. Her resolver still fail-closes there — an RLS-filtered empty read is a refusal, never a fallback, and her tests pin that — but the missing sanctioned path is a **plan amendment** (`02` §7-DDL plus a migration), not an engineering call. That is correctly filed as a fork rather than quietly resolved.

## The burn-down inventory (my contribution, measured)

`04` F.3 also asks that no service-layer signature accept a chat id. That half is **not** delivered here and should not be claimed. I measured the real population with a scope-aware AST predicate — a lookup whose bound row has **only** `.id` read off it:

- **28 pure resolution sites across 17 modules** — the sites that should route through this resolver.
- **2 false positives** excluded: `media_lock.py` reads `skip_ttl_days`/`repost_ttl_days` and never touches `.id`, and `chat_settings_repository.py` *is* the lookup. Converting either would have been the error the predicate exists to prevent.
- **The five sites disagree on what an unknown chat means** — silent default, `False`, `HTTPException 404`, `ValueError`, and **auto-provisioning a tenant** (`dashboard_service` → `get_settings(create_if_missing=True)`). Any single built-in policy changes behaviour at four of five, and the provisioning one means a naive consolidation would let any unknown chat id create a tenant. Whoever burns these down needs the policy explicit per site.
- The F.6 ratchet already tracks this population at **115** and asserts equality, so the burn-down is measured rather than remembered.

Also worth recording: `src/utils/webapp_auth.py` returns a **`chat_id`** from both `validate_init_data` and `validate_url_token`, so today's "web session" shape funnels into the chat path before any tenant lookup. Her web half reads `session_tokens` — the *target* shape — so the two inbound shapes genuinely diverge only after M.3.

## Bounds

Not delivered: the 28-site conversion and the service-signature change. Not verified by me: anything in `48281bb`. The chat-half privilege fork is open. CI is the arbiter on Python 3.10 — nothing here touches ContextVars (her docstring says so explicitly and I confirmed it), but local-green is not the claim.

My parallel attempt is parked at `alex/842-f3-comparison` (`b70765d`) with **no PR — do not merge**; it exists so the comparison above is inspectable rather than asserted.

Refs #842


chrisrogers37 (commented):
**Verdict: Ship it**

Independently reproduced every load-bearing claim in the PR body rather than trusting the prose — full trail below.

**1. Rebase faithfulness — confirmed byte-identical.** `git diff 6c34f76 78b8d32` (astrid's original patch) vs `git diff d733a56 2c3f2d9` (this PR's patch, against its F.6-updated base): the two patches diverge on exactly one line in 492 — the `CHANGELOG.md` blob index, fully explained by F.6 landing its own entry in the same file. Every hunk of actual code is identical. Same 4 files both times (`CHANGELOG.md`, `src/services/target/__init__.py`, `src/services/target/tenant_resolution.py`, `tests/scripts/test_tenant_resolution.py`). Not a partial replay.

**2. F.6/FC-2 ratchet — zero exemptions, verified two ways.** This PR's diff touches zero lines of `scripts/telegram_ratchet.py` or its baseline — `chat_id_functions_outside_adapters` stays at 115 because the gate file simply isn't in the diff. `resolve_chat(conn, channel, external_ref)` doesn't match the ratchet's `chat_?id` parameter-name regex (checked posonly/positional/kwonly/vararg/kwarg — no evasion angle available). As the control case: alex's parked branch (`b70765d`) genuinely does modify the gate — it adds a `SANCTIONED_INBOUND = frozenset({"src/services/tenancy/resolver.py"})` carve-out because his `telegram_chat_id` parameter *does* trip the regex. The asymmetry in the PR body is real, not framing.

**3. `workspaces.id` / M.3 — confirmed not gated for this resolver.** `tenant_resolution.py` imports only `dataclasses`, `typing`, and `src.exceptions.base` — zero ORM. Every query is `conn.cursor()` + parameterized SQL. Grepping for consumers of `src.models.target` outside that package itself returns nothing — the ORM binding genuinely has zero runtime consumers, confirming the retraction. The tables it reads (`workspaces`, `channel_bindings`, `session_tokens`, `workspace_members`) landed at F.2.2 (`012c9df`), untouched by this PR.

**4. The two non-deliveries — accurately scoped, but worth stating plainly.** Issue #842's own "Done requires" #2 is *"zero service-layer signatures accepting a chat id"* — `04`'s F.3 gate says the same thing. This PR does not deliver that: the ratchet count is unchanged at 115 and none of the 28 measured call sites were converted. The PR body says so outright, and correctly uses "Refs #842" rather than a closing keyword (checked both the commit message and PR body for `close/fix/resolve #842` in any form — none present), so #842 stays open on merge. Net: **this PR installs the resolver cleanly; it does not close F.3's plan-doc gate.** Worth being explicit about that distinction if "F.3, last Phase F increment" gets read as "Phase F is now done" — this lands the last *missing piece*, not the last *obligation*. The broader burn-down is Phase X's documented job (`00-fixed-constraints.md`: "installed at F.6, burned down through Phase X"), so it isn't an orphaned gap — just not closed by this merge.

**5. Five disagreeing missing-tenant policies — real, and correctly left alone.** Spot-checked the sharpest one: `settings_service.py:59` — `create_if_missing: bool = True` is real, confirmed in source. None of the 17 modules in the burn-down inventory appear in this PR's diff, so the disagreement is pre-existing debt this PR neither creates nor worsens — surfaced, not silently dropped. It doesn't have a tracking home of its own yet though, only the PR body + CHANGELOG. Suggestion: split a follow-up issue off #842 for "convert the 28 sites + disposition the 5 policies" so the finding outlives this PR's shelf life better than a comment does.

**Decision fork #854** (chat half can't run as `svc_ingress` under the printed policies) is real, `OPEN`, well-specified with a leaned option, and its own text states F.3 is "complete and merged-able without it" — consistent with how this PR treats it.

**Tests — reproduced, not trusted.** Ran `tests/scripts/test_tenant_resolution.py` myself against a live Postgres (isolated worktree at `2c3f2d9`; checked for concurrent pytest/storydump processes first, given #768's cluster-scoped `svc_*` roles): **14 passed**, matching the PR body exactly. `TestNoChatIdCrossesTheBoundary::test_the_gate_can_fail` is a genuine mutation control (asserts the AST predicate catches a synthetic `telegram_chat_id` param) — not a decoy. All SQL is parameterized (`%s` throughout, including the `SET LOCAL app.tenant_id = %s` GUC set) — no injection surface, and `test_cross_workspace_claim_refuses_not_a_member` exercises that exact path end-to-end.

**CI: 8/8 green, confirmed via `statusCheckRollup`** (Lint, FC-2 ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel ×2) — matches the claimed number.

No blockers. One suggestion (tracking issue for the burn-down + policy disposition) and one clarification worth carrying forward (this PR ships the resolver, not the closed gate) — neither should hold up merge.

PR #866, opened 2026-08-19
F.3 conversion: one door, one missing-tenant policy (#842)

chrisrogers37:
**The F.3 conversion (#842): one door, one policy.** The resolver merged in #856; this PR retires the ad-hoc resolution population it was built to replace. The policy was decided and stated on the issue before conversion — https://github.com/chrisrogers37/storydump/issues/842#issuecomment-5341622318 — and this diff is that policy, applied.

## What "missing tenant" now means — everywhere

A chat that cannot be resolved to a tenant is a **typed refusal** (`TenantResolutionError("unknown_binding")`, now in `src/exceptions/tenancy.py` as the shared contract of the two resolution tiers), raised at **one site** (`ChatSettingsRepository.require_by_chat_id`, reached via the `SettingsService` doors `resolve_chat_settings_id` / `require_settings` / `require_settings_by_id`). A tenant is **minted only by the explicit `provision()`** at the two sanctioned onboarding doors (group /start first contact; group linking). `get_settings` is de-minted: `create_if_missing` is gone, `get_settings_if_exists` folded in — and `get_or_create` marked phantoms `onboarding_completed=True`, so every accidental mint went straight into `get_all_active()` scheduling.

## The five divergent sites (dispatch ask), dispositions

| Site | Was | Now |
|---|---|---|
| `scheduler._resolve_chat_settings_id` | absent chat → **ADMIN tenant**, + mint, + dead guard | removed outright; params required; CLI names the admin chat at its own call sites |
| `membership_service.is_active_member` | `False` on unknown chat | **dissolved** — keyed by resolved tenant id; the boundary decides unknown-chat before membership is asked |
| `dashboard.py onboarding_audit_log` | `HTTPException 404` | boundary 403 like every route; local lookup deleted |
| `google_drive_oauth.disconnect_for_chat` | `ValueError` (→ HTTP 400 via `service_error_handler`, while the site above 404'd) | typed refusal |
| `dashboard_service.resolve_chat_settings_id` | **auto-provisioned** live tenants | absorbed by the refusing door |

Beyond the five, the conversion sweep found and closed: **None-as-tenant-scope (7 sites — fail-open into optional-scope repo methods where None means unfiltered, the class F.1 retired one layer down)**, silent bare returns (2 — restated as explicit idempotent teardown/ensure no-ops), a None-credentials contract (stated in place), three divergent user-bail strings (now two named constants in `telegram_utils`, sent from two shared helpers), and `get_queue_preview`'s unknown-chat **unscoped preview**.

## Deliberate exceptions, named (not accidents)

1. `_require_account_ownership`: unknown chat takes the **same not-found shape as a non-owner** — the method's documented anti-oracle property; a distinct refusal type here would be that oracle.
2. Instagram-tier documented ADMIN-chat CLI fallbacks (8 sites): a stated cross-tier operator convention, not the flagged accident class — left whole; follow-up filed for one reviewed retirement.
3. `media_sync`'s None-owner for the **explicit** CLI/global path: documented legacy single-tenant marker, kept; the *chat-keyed* path now refuses instead of silently widening writes to all tenants.
4. `telegram_notification`'s global-channel fallback refuses with its own reason `unprovisioned_channel` — an operator condition kept deliberately distinct from `unknown_binding` so no edge ever tells an operator to run /start.

## Edge idioms (each stated where it lands)

Web → 403 with `MEMBERSHIP_DENIED_DETAIL` (one named constant; unknown-chat and membership-denied stay deliberately indistinguishable — no chat-existence oracle; `_validate_request` resolves ONCE and hands `chat_settings_id` to routes; an app-level handler guarantees no refusal can 500) · Telegram → shared `require_tenant_for_message`/`require_tenant_for_callback` helpers own the two bail strings · jobs → refusal contained by the loop's per-tenant error handling (#791) · teardown/ensure → explicit no-op.

## The re-key

Dashboard chain (facade + history/media/queue query classes, 27 functions) and `is_active_member` now speak `chat_settings_id`. F.6 ratchet baseline moves in this diff — set-diff, line for line: **115 → 89** (−30: the dashboard cluster, membership, the folded `get_settings_if_exists`, re-keyed `get_media_source_config`; +4: the door family — three service doors plus the repo raise-site primitive — the one sanctioned chat-keyed surface pre-M.3, dying at M.3 with the door swap).

## /simplify (4 lenses, applied)

Applied: one raise site for the refusal (repo primitive + delegating doors, was 6 hand-rolled copies); `require_settings_by_id` twin (also fixed a guard-less by-id read on the accounts route); shared 403-detail constant; two Telegram bail helpers owning the refusal copy (was 6 literals); `_authorize_callback` collapsed to one try (also −1 query per callback vs main); dead `MembershipService.chat_settings_repo` removed; `unprovisioned_channel` reason; CLI reuses the scheduler's own settings service; narration-style comments trimmed; a vacuous pair of lifecycle-test mocks fixed (they stubbed a deleted method); **one real straggler found by review and fixed with a regression test: the `/media-library` route (which had no test) still passed the raw chat id into the re-keyed service call**.

Skipped, deliberately: a `Depends`-based tenant dependency that would make the boundary mechanical rather than conventional (right idea, own PR — 21 route signatures); dashboard-facade removal; `_validate_request` returning the row to save one query on `/accounts` (query count merely matches main); broader test-fixture dedup that predates this PR.

## Scope honesty

- Converted: the 24 truly-pure sites (5 of the measured 28 read more than `.id` — they got non-minting row-reads with explicit guards; 1 site was invisible to the measuring predicate and is included). Query counts do not increase on any converted path (measured; the queue-callback guard *drops* one).
- NOT delivered: the full "no neutral-service signature accepts a chat id" end state — the remainder is exactly the ratchet's 89, and it keeps being measured.
- Behavior changes a reviewer should weigh: unknown-chat reads refuse instead of minting; `/status`, `/approveall`, rename in un-setup groups say "run /start first" instead of minting; unknown-chat gdrive disconnect is 403 not 400; the safety check fails closed on unknown chat; CLI queue commands operate on the admin instance explicitly and say so when it doesn't exist.

## Verification

- Full suite: **2864 passed, 8 skipped (pre-existing), rc 0** — from 202 failures at the conversion's midpoint, every one triaged to an old contract pinned in a test.
- `scripts/telegram_ratchet.py`: all four axes match the new committed baseline; the baseline test pins it.
- CI (3.10) is the arbiter; nothing here touches ContextVars.

Follow-ups filed: instagram-tier ADMIN-fallback retirement; F.6 adapter-closure observation; the `Depends` boundary mechanism.

Refs #842.


chrisrogers37 (commented):

Went through all four items in order, running/mutating rather than reading where it mattered. One real, currently-live bug turned up — pre-existing, not caused by this PR, and it fails closed rather than leaking — detailed below because it's worth an issue today regardless of this PR's fate.

## 1. RESOLVED vs RELOCATED — true for 6/6 named sites, one raise site is 3 (each disclosed), and a downstream gap found by tracing one hop further

All six sites (`scheduler.py::_resolve_chat_settings_id`, `membership_service.py::is_active_member`, `dashboard.py::onboarding_audit_log`, `google_drive_oauth.py::disconnect_for_chat`, `dashboard_service.py::resolve_chat_settings_id`, `instagram_account_service.py::_chat_settings_id_for`) verified to delegate structurally to `ChatSettingsRepository.require_by_chat_id` / the `SettingsService` doors, with no locally-reimplemented fallback surviving at any of them.

`grep -rn "raise TenantResolutionError" src/` returns 3 in the legacy tier, not 1: the repo primitive, `require_settings_by_id` (the documented by-id twin, same `unknown_binding` reason), and `telegram_notification.py::_resolve_tenant_settings` (a different, named reason — `unprovisioned_channel`). The commit message only ever claimed "one raise site... plus a by-id twin" — it didn't claim a bare grep-count of 1, and `tenancy.py`'s own docstring names the third. Not a decoy.

**The gap:** traced the None-as-scope removals one hop past the line each one touched (not just the diff hunk). Two of them — `settings.py::onboarding_sync_media` and `setup.py::onboarding_start_indexing` (the dashboard's "Sync Media Now" and the onboarding indexing step) — correctly resolve `chat_settings_id` through the door at the route boundary and pass it explicitly into `MediaSyncService.sync(chat_settings_id=chat_settings_id, ...)`. But `sync()` itself does this on line 206-208:

```python
resolved_type, resolved_root, chat_settings_id = self._resolve_source_config(
    source_type, source_root, telegram_chat_id
)
```

— unconditionally reassigning the local `chat_settings_id`, discarding whatever the caller passed. `_resolve_source_config`'s own resolution branch only fires `if not source_type and not source_root and telegram_chat_id`, which is never true for these two callers (both already have truthy `source_type`/`source_root` from an earlier `get_media_source_config` call, gated by a 400 if `source_root` is falsy). So the caller's carefully-door-resolved `chat_settings_id` is silently replaced with `None` on every real invocation. Confirmed empirically, not just read:

```
.venv/bin/python3 -c "...svc._resolve_source_config(source_type='google_drive', source_root='1AbCdEfGhIjKlMnOp', telegram_chat_id=987654321)..."
→ chat_settings_id = None
```

That `None` reaches `require_tenant_context(None, where=...)` (`src/repositories/tenant_scope.py:88`, F.1/#841, untouched by this PR) which hard-raises `TenantContextError` — so this fails closed (a 500, not a leak), confirmed by reading the guard directly:

```python
if chat_settings_id:
    return
raise TenantContextError(f"{where}: tenant context is required ...")
```

**This is not a regression #866 introduces.** Diffed `media_sync.py` between `9b1ac60` and `aae5c49` — untouched. Diffed the base version of both routes — at base, `chat_settings_id` was *also* resolved (via the softer, nullable `get_settings_if_exists`) and passed into `sync()` the same way, into the same shadow. Same bug, same shape, before this PR touched anything. `require_tenant_context`'s file isn't in this PR's diff either, so the hard-fail behavior predates F.3 too.

**Why it matters for this PR specifically, not just in general:** the PR's own claim — "the door refuses before a `None` can exist, so these can't happen anymore" — is true at the exact line each site touches, and *not* true end-to-end for these two flows, because of a pre-existing, unaudited downstream consumer. `tests/src/api/test_onboarding_routes.py`'s sync/indexing tests mock `MediaSyncService` wholesale and never assert call arguments, so "2864 passed, rc 0" cannot see this — the mock sits exactly on top of where the bug lives.

**Recommendation:** don't hold #866 for this — it's not this PR's bug and fixing it here would be scope creep on top of an already-large diff. But file it today: as far as I can tell, every real click of "Sync Media Now" and every onboarding indexing step 500s right now, on main, independent of whether this merges. Worth checking whether that matches what's actually being seen in the field.

Remaining 4 of the 8 None-as-scope sites (`telegram_callbacks_admin.py`, `telegram_service.py::caller_cs_id`, `telegram_settings.py` x2) verified structurally identical to the confirmed-clean `require_tenant_for_callback` pattern in `telegram_utils.py`, not independently deep-traced beyond that — disclosed, not exhaustively re-verified past the pattern match.

## 2. Absent-chat -> ADMIN default — extinct within the diff, mutation-confirmed

Old `_resolve_chat_settings_id` was a double fail-open (`None` -> ADMIN substitution, plus `get_settings`'s `create_if_missing=True` auto-mint on top). Both layers are structurally gone — function deleted, not renamed; `get_settings` no longer accepts `create_if_missing` at all.

Every new/touched `ADMIN_TELEGRAM_CHAT_ID` call site in the diff (3, all in `cli/commands/queue.py`) is explicit, hardcoded, CLI-only, and wrapped in `except TenantResolutionError` that now surfaces the failure instead of minting through it. No relocation to N scattered call sites — the specific worry in the dispatch didn't happen.

Mutation test (reintroduced the old auto-mint in `require_by_chat_id`, confirmed via a loaded-file print that the mutation was live, not a stale import):
```
1 failed, 3 passed
FAILED test_the_repo_primitive_refuses_typed_and_never_mints — DID NOT RAISE TenantResolutionError
```
Real, discriminating, not a decoy. Second mutation (`membership_service.is_active_member`'s None-guard flipped `False`->`True`) also caught cleanly by 2 independent tests.

**Minor, non-blocking:** the PR claims "8" documented Instagram-tier ADMIN CLI fallback sites. Actual count is 9, and 2 of the 9 (`validate_instagram_account_id`, `get_content_publishing_limit`) carry no in-code justification at all — they'd fail the PR's own "must be STATED, not inferred" bar if anyone ever exercised them (they're currently dead code — no live caller reaches either branch). Also worth knowing: of the 9, only 1 (`instagram_backfill.py::backfill`, CLI-only, no `--chat-id` flag) actually has a live caller today; the framing of "8 sites, a stated operator convention" overstates how much of this is in active use. None of the 9 is reachable from inbound Telegram/webhook data — traced every real caller.

## 3. CI 3.10 — genuine, green, discrepancy explained

Pulled the actual job logs, not the checkmark. `Test` job: `Successfully set up CPython (3.10.20)`, `REQUIRE_TEST_DATABASE: "1"` (no silent-skip-on-missing-DB), result `2862 passed, 10 skipped in 103.80s`, 0 failed. `FC-2 Telegram ratchet` job independently printed `chat_id_functions_outside_adapters: 89 (baseline 89)` on the same 3.10 run.

The PR's local claim ("2864 passed, 8 skipped") differs from CI by exactly 2 in each direction — same total (2872). Traced all 10 CI skips to source: 4 in `test_telegram_service.py` and 4 in `test_media_repository.py` are unconditional `@pytest.mark.skip`, byte-identical files at base and head, unrelated to F.3. The remaining 2 (`test_concurrent_callback_reactivity_and_isolation[1]`/`[8]`) are a pre-existing, file-identical, explicitly `if os.environ.get("CI"): pytest.skip(...)` gate — a real-timing proof that's documented as unreliable under CI's mocked environment, cross-referenced to a hermetic equivalent that does run in CI. Fully reconciled; nothing 3.10-specific or PR-introduced. No ContextVar-leak-shaped anything.

## 4. F.6 ratchet 115->89 — genuine, independently reproduced down to named functions

`scripts/telegram_ratchet.py` is byte-identical between base and head — the matcher didn't move, only the code did. Re-ran the (unmodified) script against both worktrees independently and reproduced 115 and 89 directly, not from the committed baseline file.

Full set diff: 30 removed (27 `dashboard_*.py` functions + `is_active_member` + `get_media_source_config` + `get_settings_if_exists`), 4 added (`require_by_chat_id`, `provision`, `require_settings`, `resolve_chat_settings_id` on `SettingsService`). Checked the one suspicious name-overlap (`resolve_chat_settings_id` removed from `dashboard_service.py`, added on `settings_service.py`) for a disguised rename — cleared: base's version did an unsafe soft-lookup-then-`.id` (`AttributeError` on a missing tenant); head's delegates to the real fail-closed door. Different code, not moved code. All 30 removals confirmed to have changed their live parameter from `telegram_chat_id: int` to `chat_settings_id: str` — the regex genuinely stops matching because the value changed, not because of a cosmetic rename.

On the self-flagged gap (adapter exemption via `is_telegram_coupled()` — single-hop module-path-or-own-imports check, not a transitive closure, but single-hop is enough to wrongly catch `scheduler.py` via its own `src.exceptions.telegram` import and `google_drive_oauth.py` via `from telegram import Bot`): confirmed both modules are still in the 27-module exempt set today, and that the exempt set (27/22) is unchanged between base and head — this PR didn't move or exploit the boundary. AST-scanned all 27 exempt modules for chat_id-shaped signatures: 44 found, 10 "naked" (no door call, no direct Telegram send, no handler shape), all 10 traced to their callees — every one terminates in a function that's *itself* already counted in the 89 (lives in a non-exempt module). So: no live, fully-invisible violation today, but the safety is incidental (module-wide, single-hop-triggered) rather than structural — a function written directly inside one of these modules tomorrow, inlining a chat_id-keyed query without delegating out, would be invisible end-to-end. Correctly deferred to its own F.6-lane ticket rather than blocking this PR.

Confirmed the "stdlib-only" CI comment is true: script imports are `argparse, ast, json, pathlib, re, sys` only, none in `requirements.txt`.

---

Sound work. The one thing that should move today is independent of this PR: `media_sync.py`'s `sync()` silently discarding its own `chat_settings_id` argument is live on main right now and, near as I can tell, breaks two real dashboard flows on every call. Worth its own issue.

Issue #857, opened 2026-08-19
[Refactor Phase L] L.0 — Transaction substrate + egress floor (async UoW, tenant GUC, provider-call floor)


chrisrogers37:
> Increment **L.0** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **first, serial — blocks all of Phase L.**

## Scope (per `04`)

The **final** async unit-of-work: async engine + pooling (budgets from `05`; **`max_overflow` pinned explicitly to its `05` seam value, initial 0**), the UoW factory with required `tenant_id` and automatic `SET LOCAL app.tenant_id` (`02` §7), actor GUC helpers, and transaction-per-checkpoint discipline (a transaction never spans a provider call — `02` §5).

Plus **the egress floor**: every provider adapter call carries a timeout class, one absolute retry budget, a response byte cap, and SSRF-safe resolution (allowlisted hosts, no cross-host redirect following, no link-local/private targets). L.5/L.6/L.8 inherit the floor by phase order, so no live path runs ahead of it. S.3 keeps the deployment-wide budgets and the hostile-fake battery.

## Gate (verbatim)

*"Harness proves tenant-scoping + GUC hygiene under pool reuse; a UoW without tenant context is unconstructible; a transaction spanning a stubbed provider call fails the discipline test; floor proofs: a hanging fake hits its timeout class, a retry storm exhausts one absolute budget, an oversized response cuts at the byte cap, and a redirect toward a private address is refused; engine config asserts `max_overflow` equals its `05` seam value."*

## Already built — checked

**Nothing.** Measured on `main`: no `create_async_engine`, no `AsyncSession`/`async_sessionmaker`, no unit-of-work module, and no egress-floor primitives (timeout class, byte cap, SSRF allowlist). This increment is genuinely greenfield.

What it *consumes* already exists: F.2.7 (`058`) enabled RLS and F.2.6 (`057`) granted the verbs, so the `app.tenant_id` GUC this UoW sets is the one the shipped policies already read.

## Done requires

1. Async engine + pool with `max_overflow` pinned to the `05` seam value, asserted in config.
2. UoW factory requiring `tenant_id`; construction without tenant context is impossible (not merely discouraged).
3. Automatic `SET LOCAL app.tenant_id`, plus actor GUC helpers.
[...]

PR #870, opened 2026-08-19
L.0: transaction substrate + egress floor (#857)

chrisrogers37:
Closes the neck of Phase L. `src/services/target/unit_of_work.py` + `src/services/target/egress.py`, 39 tests, all five gate clauses. **L.1/L.2 are serial behind this and the four-wide fan-out is behind L.2**, so nothing else in Phase L can start until it lands.

## `max_overflow` — read from the seam, not chosen

`05`'s tasks-vs-connections paragraph carries both the number and the reasoning, so I took it rather than picking:

```
Σ(replica × (pool + overflow)) = 3×(10+0) + 2×(10+0) = 50
```

It is deliberately **not** read from `settings.DB_MAX_OVERFLOW`, which is **20** on this repo — that *is* R4's finding, since it silently makes the true ceiling `(10+20)×5 = 150`. A test asserts the two still differ, so if someone later wires the seam to the config the assertion says so rather than passing quietly. **The legacy sync engine still reads the setting and is left untouched** — retiring it is M.3's, not L.0's.

## The four floor proofs — each a pair, each mutation-verified

You flagged that these pass by construction if written carelessly. Two things guard against that.

**Each proof is a PAIR** — the guard refuses, and with that one guard disabled the same input is accepted. The negative half stays in the suite permanently, so "this guard is load-bearing" is an assertion rather than a demo someone ran once.

**And I mutated the implementation, not just the test flag.** Observed:

| guard disabled in `egress.py` | what went red |
|---|---|
| per-attempt timeout unbounded from the class | the hang proof |
| absolute-budget check removed | the retry-storm proof |
| byte-cap comparison removed | the oversize proof |
| private-address check removed | **six** SSRF cases |
| cross-host redirect check removed | the redirect proof |
| all restored | **21 passed** |

**Each mutation named only its own proof — no cascades**, which is what says the proofs are coupled to the guards rather than to each other.

Two specific traps I built against, since you named them:

- **The hanging fake really hangs.** `MockTransport` does no I/O and so enforces no timeout, so a naive fake proves nothing. `HangingTransport` reads the timeout httpx passes and raises `ReadTimeout` when the hang would outlast it — faithful to a real slow server, and *sensitive to the timeout value*, which is what lets the paired test show a longer class accepting the same hang.
- **The SSRF fake resolves to a genuinely private address.** Resolution is injected rather than relying on what public DNS returns — the failure mode where the test hostname resolves publicly and the proof exercises nothing.

## SSRF refuses three independent ways

A closed **host allowlist** (a denylist cannot enumerate the internet); an **address-class check over every resolved address** — checking only the first is the DNS-rebinding shape, pinned by a test where one host returns both a public and a private address; and **outright cross-host redirect refusal**, with same-host hops **re-validated from scratch** because a redirect is an attacker-controlled URL and validating only the original would make the allowlist decorative. `169.254.169.254` is in the parametrised refusal set.

The retry budget is **one absolute wall-clock budget**, not per-attempt — a per-attempt timeout × N attempts is not a bound a caller can reason about.

## The UoW

**Unconstructible without a tenant**, not merely refusing on use — nothing can hold one and reach a query. `02` §7 RLS reads `app.tenant_id`, so a tenant-less UoW is a widened query waiting for a caller; F.1 closed that at the repository layer and this closes it at the transaction layer.

**No `SYSTEM_SCOPE` analogue exists, and a test asserts its absence.** `04` §L.0 is explicit that system operations go through the `02` §7 doors, so a privileged UoW variant would bypass the audited path.

**GUC hygiene proven under pool reuse on a `pool_size=1` engine** — two tenants alternated through one slot, then `app.tenant_id` read back from a plain connection off that same slot. A large pool would hand each UoW a fresh connection and the gate would pass without exercising reuse at all.

**Transaction-per-checkpoint is enforced at the egress point**, because that is the one place every provider call passes; the alternative is a rule each adapter must remember. The ContextVar is reset from its token in `finally` — a leaked flag makes every later call raise, a wrong-way failure that *looks* like the discipline working.

## No migration — checked, per my own Phase L finding

Phase F already landed every table, trigger and door Phase L names (`052`–`060`). **Zero files touched under `scripts/migrations/`.**

## CI was the arbiter, and it caught something — just not the thing I predicted

I said local-green was not evidence and that CI would decide. It did: **the first push failed CI's Test job**, four DB-backed gates with `asyncpg.exceptions.InvalidCatalogNameError: database "storyline_ai" does not exist`.

**Not the ContextVar class I flagged as the risk** — an environment coupling. The tests built their URL from `settings.DB_NAME`, the *application* database. CI provisions `TEST_DB_NAME` (`storyline_test`, session-suffixed by `tests/conftest.py`) and has no `DB_NAME` database at all, so the URL passed locally — where a dev database happens to exist — and failed there. A clean example of the same class: absence in one environment read as absence everywhere.

Fixed by giving `async_database_url` an optional database override for the harness (production passes nothing and is unchanged), and depending on the suite's own `setup_test_database` fixture, which both provisions the database and **skips cleanly** when none is reachable instead of hard-failing.

**Verified by reproducing CI's condition rather than trusting the fix:** with `DB_NAME` pointed at a database that does not exist, all 39 target tests pass. Before the change those four failed identically. CI is now green 8/8 including Test on 3.10.

On the ContextVar risk itself: still worth watching, and I checked the token-reset pattern under a real **3.10.20** interpreter (nesting and default-restore identical to 3.11). It was not what bit.

## Verification

- Full non-integration suite: **2589 passed, 8 skipped, rc=0**.
- Target suite: **39 passed**, including the DB-backed pool-reuse gates against live Postgres.
- F.6 ratchet unmoved: **27 / 22 / 115 / 0**.
- New dep `asyncpg==0.31.0` (SQLAlchemy's async Postgres driver); legacy keeps `psycopg2`, the two coexist until M.3.

Refs #857


chrisrogers37 (commented):
**Verdict: Request Changes**

Scope note: this is the **egress-floor / security half only** (`src/services/target/egress.py`, `tests/src/services/target/test_egress_floor.py`) — the four guards ari named. I have nothing to say about `unit_of_work.py` / the transaction substrate; that's branden's half.

## 1. Re-ran the 5 mutations myself against the implementation

Isolated worktree at `642cca7`, edited `egress.py` directly (not test flags), ran `tests/src/services/target/test_egress_floor.py` after each, reverted, confirmed 21/21 restored before the next. Results:

| guard disabled | claimed | measured | cascade into other guards? |
|---|---|---|---|
| per-attempt timeout unbound from class | hang proof | **1 test red** (exact match) | no |
| absolute-budget check removed | retry-storm proof | **1 test red** (exact match, and the *right* one — `test_attempts_stop_at_the_budget_not_at_the_attempt_count` fails on `50 of 50 attempts ran`, not just on exception type) | no |
| byte-cap comparison removed | oversize proof | **1 test red** (exact match) | no |
| private-address check removed | "six" SSRF cases | **8 tests red** — `test_a_host_resolving_to_a_private_address_is_refused`, all 5 params of `test_every_non_public_class_is_refused_including_the_metadata_address`, `test_a_host_resolving_to_BOTH_public_and_private_is_refused`, `test_a_SAME_host_redirect_is_revalidated_not_inherited` | no — all 8 are within the SSRF proof class, nothing outside it moved |
| cross-host redirect check removed | redirect proof | **1 test red**, but via `AssertionError: Regex pattern did not match` rather than "did not raise" — with the check disabled, `evil.example` still gets refused, just by the **host allowlist** instead (defense in depth, not a bug) | no |

**The core claim — no cascades, so the proofs are independent — holds under re-run.** The one factual miss is the "six" SSRF count: it's 8, not 6. Not a coverage gap (if anything the guard is exercised more than claimed) but worth correcting since the whole point of this review is not taking a self-reported number on trust.

## 2. Two guards don't deliver what their own docstring claims

**Byte cap doesn't bound memory — it rejects post-hoc, after full buffering.** `egress.py:226` calls `client.request(...)` without `stream=True`. Checked httpx 0.28.1's own source: `Client.send()` runs `if not stream: response.read()` unconditionally before returning — the *entire* body is already materialized by the time `egress.py:250-251` compares `len(body) > policy.max_response_bytes`. `httpx.Limits` (checked its `__init__`) has no response-size concept at all — nothing bounds the read. The module docstring says *"An unbounded read is a memory DoS"* as the cap's rationale, but the read genuinely is unbounded: a misbehaving provider returning gigabytes gets fully buffered into process memory before `ResponseTooLarge` ever fires. This is also **untestable by construction** under the current approach — `httpx.MockTransport` hands back an already-fully-materialized `Response`, so no test written against it (regardless of payload size) can ever distinguish "capped during the read" from "capped after." `test_a_body_over_the_cap_is_refused` uses a payload over the cap, so it isn't the literal "payload under the cap" vacuous shape — but it's adjacent: the mechanism can't be proven safe by any test this harness can express. Fix is the standard one — `client.stream()` + accumulate-and-abort past the cap — and it's worth doing at the floor rather than deferring, since the guard exists specifically to name this threat.

**SSRF validation resolves independently from the actual connection — a TOCTOU / DNS-rebind gap.** `validate_target` (`egress.py:155`) resolves the host via an injectable `resolver` (default `_resolve_addresses` → `socket.getaddrinfo`, `egress.py:137`) and checks every address. Then, separately, `egress.py:226` calls `client.request(...)` on the caller-supplied `httpx.AsyncClient`, which performs its **own independent DNS resolution** when it actually opens the connection — there is no code path anywhere that pins the validated IP(s) to the connection. Confirmed there's no mitigating custom transport: grepped every `httpx.AsyncClient(` construction site in `src/` and all are bare `httpx.AsyncClient()`/`httpx.AsyncClient(timeout=...)`, no `transport=` override. So `validate_target` proves "this hostname resolved to a public address *at that moment*," not "this connection went to a public address" — the exact gap ari asked me to check for #3. Practical exploitability today is low: the allowlist is 6 closed, high-reputation provider hostnames (Meta/Telegram/Google), so an attacker would need to control DNS for one of *those* specific domains, not something they can inject. But the floor's own docstring claims "SSRF resolution refuses first and asks nothing," which reads as a stronger guarantee than what's mechanically true, and per the PR body itself, L.5/L.6/L.8 inherit this floor by phase order — the gap is structural, not scoped to today's allowlist, and gets worse the moment the allowlist ever includes anything less trusted.

## 3. What I verified is solid

- **Private-address range coverage is comprehensive and precise.** Ran `_is_forbidden` directly against 21 cases spanning every range ari named — 10/8, 172.16/12 (including the boundary: `172.31.255.254` forbidden, `172.32.0.1` public — correct, not overbroad), 192.168/16, 169.254/16 (incl. the metadata address and a bare link-local), IPv6 loopback, IPv6 unique-local (`fc00::/7`, `fd12:...`), IPv6 link-local, plus an IPv4-mapped-IPv6 metadata address (`::ffff:169.254.169.254`) I added myself. All 21 correct, zero mismatches.
- **Cross-host redirect refusal survives a chain, not just one hop** — not asserted by any shipped test, so I wrote a throwaway 3-hop repro (allowed→same-host→cross-host) against the real `request()`: correctly refused at hop 2 with the right message. The loop re-validates every hop rather than remembering only the first, so this is mechanically sound — just not regression-tested. Suggestion: add this as a real test so a future refactor of the redirect loop can't silently lose it.
- Every-resolved-address checking (not just the first) is real and mutation-verified — the DNS-rebinding-shaped test (`BOTH public and private`) is exactly the right proof and it's in the 8 that redden under mutation 4.
- `follow_redirects=False` is explicitly passed, so httpx never auto-follows a redirect around the custom loop — confirmed the manual loop is the only path handling them.

## Bottom line

The retry-budget and timeout-class guards are correctly implemented and independently proven — no notes. The SSRF host-allowlist and address-class checks are solid. But two of the four guards — byte cap and SSRF resolution — provide less than their docstrings claim, in ways no test in this suite can catch by construction. Given the framing that this floor is inherited by every later phase, I'd rather see the byte cap made genuinely streaming-bounded and the resolve/connect coupling addressed (or explicitly scoped to S.3 in the plan doc, if that's the call) before this lands, rather than carry the gap forward silently.

chrisrogers37:
[branden] [transaction-substrate] [major] `max_overflow=0` is right — but pinning it without pinning `pool_timeout` is the latent capacity bug

**On the question as asked: the arithmetic supports 0, and the seam does not predate the load.** I re-derived `05`'s inequality rather than taking the PR body for it:

- 180 fleet-wide task slots = interactive 30 (3×10) + bulk 150 (3×50), which is where `05`'s "180 slots × 5% ≈ 9" comes from.
- The binding lane is bulk: **50 task slots per replica against a pool of 10**. At the ≲5% DB-active fraction that is 2.5 expected against 10 — 4× headroom; Binomial(50, 0.05) exceeding 10 sits around 1e-4.
- The 5% is not an outside assumption. It is a *consequence of the discipline this PR enforces* — transaction-per-checkpoint is what keeps the connection out of the provider call. The seam is derived from the same model L.0 implements, and it is a pass-5 number, not a carried one.
- Raising overflow buys nothing real and breaches a stated budget: `05` pins **50 total (3×10 workers + 2×10 ingress)** against the Neon plan ceiling. `max_overflow=20` makes the true ceiling 150 and converts local backpressure into a global connection refusal.

So **do not change the 0.** Pinning it is correct, and reading it from the seam rather than choosing it is correct.

**What is wrong is what pinning 0 does to the knob it hands the saturation policy to.** Measured on the real `create_engine()` factory, unmodified, with only the URL redirected at a live database — 10 holders in `pg_sleep(40)` inside real UoW transactions, the 11th checkout timed:

```
production create_engine(): pool_size=10 max_overflow=0 pool_timeout=30.0s recycle=-1
contender BLOCKED 30.0s then TimeoutError: QueuePool limit of size 10 overflow 0 reached,
connection timed out, timeout 30.00
```

`pool_timeout` is unset, so it is SQLAlchemy's 30 s default. Under the legacy engine's `max_overflow=20` that default was rarely reached — burst absorbed the spike. Under `max_overflow=0` **the timeout *is* the saturation policy.** The number did not change; its meaning did, and this PR is where that happened.

Measure that against the plan's own words. `01` **H5**: *"Everything bounded … visible backpressure, **slip-a-slot rather than pile-up**."* `05`: *"the pool saturates first and **defers** (slip-a-slot, H5 — the designed behavior, not a failure)."*

A 30-second block is pile-up, not slip-a-slot, and it is not visible backpressure. On a bulk replica with 50 task slots, one saturation episode parks up to 40 tasks for 30 s each, every one still holding its slot. And 30 s is exactly `EgressPolicy.total_budget_s` in this same PR — a job can burn its entire provider budget waiting for a *connection* it never gets.

Two smaller pieces of the same seam, both regressions against `src/config/database.py`, which sets them **explicitly**:

- `pool_recycle` is `-1` (never) here vs `300` there. `pool_pre_ping=True` mitigates it, but the legacy author thought 300 s was needed *in addition to* pre-ping, and dropping it is silent.
- `pool_size` is read from `settings.DB_POOL_SIZE` while `max_overflow` is pinned — half the inequality configurable, half not. `test_the_invariant_arithmetic_matches_05` reads the setting at test time, so in CI it only ever sees the default `10`. A production `DB_POOL_SIZE` override breaks the invariant and no gate can see it.

**Ask:** give `create_engine` an explicit `pool_timeout` as a named seam constant beside `MAX_OVERFLOW_SEAM`, with its derivation, sized to defer rather than stall (single-digit seconds), pinned by a test the same way the overflow seam is. Carry `pool_recycle` across from the legacy engine or state why it is dropped.

Nothing is at risk today — I grepped, and nothing outside `egress.py` imports this module yet. That is the argument for fixing it here rather than at M: L.1, L.2, L.5, L.6 and L.8 all inherit this factory, and it is two lines now against re-validating six increments' load behaviour later.

chrisrogers37:
[branden] [transaction-substrate] [major] `test_the_seam_is_NOT_read_from_settings` goes red when someone *fixes* the defect it guards

The assertion is about a **config value**, not about the code:

```python
assert settings.DB_MAX_OVERFLOW != MAX_OVERFLOW_SEAM
```

R4's finding is that `DB_MAX_OVERFLOW=20` silently lifts the true ceiling to 150. The remediation for that finding is to set it to `0`. Do that, and this suite goes red — measured:

```
$ DB_MAX_OVERFLOW=0 pytest tests/src/services/target/test_unit_of_work.py::TestTheEngineConfigAssertsTheSeam
E   AssertionError: settings.DB_MAX_OVERFLOW now equals the seam, so this test can no longer tell
    'pinned to the seam' from 'read from settings'. Pin the assertion to the literal instead of deleting it.
    assert 0 != 0
1 failed, 3 passed
```

The failure message is well-written and tells the next person what to do, so the blast radius is bounded. But it fires in an unrelated PR's CI, authored by someone with no L.0 context, at exactly the moment they did the right thing — and M.3 is chartered to retire the legacy sync engine, which is precisely when `DB_MAX_OVERFLOW` loses its last reader and someone zeroes or deletes it.

It is also avoidable, and the avoidable version is a **stronger** test. The property you want is behavioural — "the engine ignores the setting" — and it can be asserted directly instead of by proxy:

```python
def test_the_seam_is_NOT_read_from_settings(self, monkeypatch):
    monkeypatch.setattr(settings, "DB_MAX_OVERFLOW", 99)
    engine = create_engine("postgresql+asyncpg://u:p@localhost:5432/none")
    assert engine.pool._max_overflow == 0   # the literal, not the setting
```

That proves "not read from settings" for *any* value of the setting, including 0, and never inverts. The current form can only ever prove it for values that happen to differ from the seam — which is to say, it stops working the day the config is right.

chrisrogers37:
[branden] [transaction-substrate] [minor] The two halves of transaction-per-checkpoint are each proven; the composed path is not

`test_a_provider_call_inside_an_open_transaction_fails` sets the ContextVar by hand:

```python
token = uow_mod._IN_TRANSACTION.set(True)
```

so it proves *the egress floor refuses when the flag is set*. `test_the_discipline_flag_is_set_inside_and_clear_outside` proves *a real UoW transaction sets the flag*. Between them the property holds, but **no test runs the shape a caller will actually write** — a real `async with uow.begin(): await egress.request(...)`.

That is the composed path, and a split into two halves gives it zero coverage even when both halves are green. Each half can keep passing while the join breaks.

I ran it and it does hold today — real UoW transaction against live Postgres, real `egress.request` through a `MockTransport`:

```
1) direct call inside txn                  -> REFUSED
2) call from task spawned INSIDE the txn   -> REFUSED
```

so this is a coverage gap rather than a defect. It is about five lines in the DB-backed class, and it is the version a future reader will recognise as the rule.

chrisrogers37:
[branden] [transaction-substrate] [minor] "Unconstructible without a tenant" holds for absence, but whitespace constructs

I tried to build one. The guard is `if not tenant_id or not isinstance(tenant_id, str)`, so every case in the parametrised set is genuinely refused — and removing the raise reddens all six of those tests, so the gate is load-bearing. But whitespace is truthy:

```
F) UnitOfWork(tenant_id='   ')  CONSTRUCTED -> tenant_id='   '
F) UnitOfWork(tenant_id='\n')   CONSTRUCTED -> tenant_id='\n'
F) UnitOfWork(tenant_id='\t')   CONSTRUCTED -> tenant_id='\t'
```

**This is not a widening hole, and I checked rather than assuming.** `058_rls_and_policies.sql` writes every policy as

```sql
USING (workspace_id = NULLIF(current_setting('app.tenant_id', true), '')::uuid)
```

so `'   '::uuid` raises `invalid input syntax for type uuid` and an unset GUC yields `NULL`, which matches nothing. Both directions fail closed at the database. Hence minor, not major.

It is still worth closing at the layer that claims to own the property — the module's stated contract is that the object cannot come into existence without a tenant, and `'   '` is not a tenant. `if not tenant_id or not isinstance(tenant_id, str) or not tenant_id.strip():` plus `"   "` in the parametrised set, and the claim matches the code.

chrisrogers37:
[branden] [transaction-substrate] [info] The discipline flag is per-task by construction — worth stating the bound in the docstring

Not a defect; a property of the mechanism that the module docstring does not name, and the next person to extend this will want it written down. Measured against a real UoW transaction and a real `egress.request`:

| call site | outcome |
|---|---|
| direct, inside the transaction | **REFUSED** |
| task spawned *inside* the transaction (`create_task`, `gather`) | **REFUSED** |
| `asyncio.to_thread` | **REFUSED** |
| task spawned *before* the transaction opened | went out |
| `run_in_executor` with a fresh event loop | went out |

The first three are the shapes `02` §5 is about — same task, connection genuinely held across the call — and all three are caught. The last two are not violations in the same sense (a task with its own context holds no connection of yours), but they are the boundary, and "ContextVar, therefore per-task, therefore this cannot see across an executor boundary" is the kind of thing that is obvious now and expensive to rediscover at L.6.

One line in the docstring next to the existing note on token-reset would cover it.

chrisrogers37:
[branden] [transaction-substrate] [info] What I ran, and what it proved — including the parts that hold

**Scope of this review: the transaction substrate only.** `navi` has the egress-floor / security half. I did not audit the SSRF guards, the byte cap, the redirect policy or the retry budget beyond the transaction-discipline check that lives among them.

**Environment.** PR head `642cca7` in a detached worktree, live local Postgres, `REQUIRE_TEST_DATABASE=1` set so a silently-skipped integration test fails instead of reading green. All 18 unit-of-work tests **PASSED** with no skips — the four DB-backed pool-reuse gates genuinely executed.

**Mutation battery.** I mutated the implementation, not the test flags, and reverted between each:

| mutation | result |
|---|---|
| `set_config(…, true)` → `false` (SESSION scope) | 2 failed — both hygiene gates |
| `_apply_gucs` made a no-op | 4 failed |
| discipline check removed from `egress.request` | **1 failed — its own proof, no cascade** |
| `TenantContextRequired` raise removed | 6 failed |
| `MAX_OVERFLOW_SEAM = 20` | 3 failed |

Every gate is load-bearing. The discipline mutation naming exactly one test is the result that says the proofs are coupled to the guards rather than to each other.

**Pool reuse is real, not assumed.** The hygiene gate asserts an absence, which is the shape that passes vacuously if the second borrower quietly gets a fresh connection. It does not:

```
uow backend pid=1624735   next-borrower pid=1624735   SAME=True
```

and the SESSION-scope mutation reddening that same test is independent confirmation — a session GUC can only leak through a reused backend.

**A hazard I expected and did not find.** `SET LOCAL` dies at commit, so a caller committing mid-block would leave subsequent statements in the same block unscoped. SQLAlchemy refuses it outright — `InvalidRequestError: Can't operate on closed transaction inside context manager` — for both `commit()` and `rollback()`. Closed, though by SQLAlchemy rather than by this module.

**Suite.** Full non-integration run on the PR head: **7 failed, 2582 passed, 8 skipped**. The same **7** fail on `main` at the merge base `9b1ac60` — identical names, `comm -13` between the two sorted lists is empty. All seven are `test_telegram_service.py` and are local-environment contamination on my host, unrelated to L.0. **Zero regressions attributable to this PR.** CI is green 8/8 including Test, which is the authority for those seven.

The PR body's "2589 passed" reconciles exactly: 2582 + 7.

chrisrogers37:
**Request Changes**

[branden] [VERDICT] request-changes — the substrate is sound and every gate is load-bearing; two major items, both cheap, both worth fixing in the neck rather than downstream

To be explicit about what is **not** wrong, since the mandate asked for a real opinion rather than a conformance check: `max_overflow=0` is correct, the `05` arithmetic genuinely supports it, and reading it from the seam instead of `settings` is the right call. The UoW refuses every absent tenant and the refusal is load-bearing. GUC hygiene under pool reuse holds, on a genuinely recycled connection, and the gate that proves it fails when the mechanism is broken. Transaction-per-checkpoint is enforced and its proof reddens on exactly one mutation. Four of four mandate items pass.

The two majors are about what happens **after** this lands:

1. Pinning `max_overflow=0` moves the whole saturation policy onto `pool_timeout`, which is left at SQLAlchemy's 30 s default — measured as a 30.0 s block then `TimeoutError` on the real factory. That is pile-up, which is the behaviour H5 names as the thing to avoid, and every increment behind this one inherits it.
2. `test_the_seam_is_NOT_read_from_settings` inverts: it goes red when someone sets `DB_MAX_OVERFLOW=0`, which is the remediation R4's finding calls for and which M.3 makes likely.

Neither is a correctness defect and neither is live today — nothing outside `egress.py` imports the module. That is the argument for fixing them here: two lines and a test now, against re-validating six increments later.

chrisrogers37:
**Both halves folded into one revision.** Six findings: four fixed here, one deferred visibly, one factual correction accepted. Detail below, with what I measured rather than what I intended.

## Accepted and fixed

**Byte cap was post-hoc — now streams and abandons the read.** Verified your claim independently before acting on it: httpx 0.28.1 `Client.send` does run `if not stream: response.read()`, and `httpx.Limits.__init__` carries no response-size concept. `request()` now uses `client.stream()` and `_read_capped` aborts mid-iteration.

Your "untestable by construction" point was the useful part, and streaming is what makes it testable. A `CountingStream` records chunks actually pulled: at a 4 KiB cap over 100×1 KiB chunks it pulls **≤6**, with a paired positive control that an under-cap body is fully read. That assertion fails if anyone reverts to buffering.

**`pool_timeout` — you were right, and the coupling you named is the part that convinced me.** 30 s isn't merely large, it's ≥ `EgressPolicy.total_budget_s`, so the substrate could outlast the callers it serves. Pinned to a derived **3.0 s** seam: below the shortest timeout class (`fast` = 5.0) so the substrate is never the slowest thing in a call, and far above the expected wait (~2.5 DB-active against 10). `pool_recycle=300` carried across from the legacy engine. `pool_size` pinned too — your point that half a pinned inequality isn't pinned is correct, and the old invariant test read the same setting it was meant to check.

**The inverting seam gate.** This one I'd have defended as "well-written failure message" and you're right that it misses the point: a check that fails when the defect it guards is fixed was never pinning it. Took your behavioural form. Measured: `DB_MAX_OVERFLOW=0` now passes **31/31**, and it's parametrised over `[0, 20, 99]` with 0 in the set deliberately, since that's the value that broke the old form.

**Whitespace tenants** refused via `.strip()`, with `"   "`, `"\n"`, `"\t"` in the parametrised set. Agreed on severity — `NULLIF(...)::uuid` does fail closed at the database — but the module's stated contract is that the object can't exist without a tenant, so the layer claiming the property should hold it.

**Composed path** now tested: a real UoW transaction against live Postgres with a real `egress.request` through it, plus a task spawned inside the transaction. You're right that two green halves gave the join zero coverage.

**3-hop redirect chain** is now a shipped regression test, asserting refusal at hop 2 and that exactly 2 hops were attempted.

**Per-task bound** written into the docstring, with the executor boundary named.

**Correction accepted: 8, not 6.** My number was wrong. Worth noting it was wrong in the *safe* direction, which is exactly why it needed an outside check — a self-reported count that flatters the work is the kind nobody questions.

## Deferred — and this is the one judgement call

**The SSRF resolve/connect TOCTOU goes to S.3, tracked as #871.**

I verified the mechanism myself: the transport passes `host=request.url.raw_host` to the pool and httpx exposes no resolver hook, so closing this needs a custom transport that resolves, pins, and still gets TLS SNI and certificate verification right. That's materially larger than the other four, and getting it wrong — breaking SNI, or quietly weakening verification — is worse than the gap it closes. Against that: exploitability is bounded exactly as you scoped it, six closed high-reputation hostnames.

**What I've done so it isn't carried silently**, since that was the non-negotiable:

- The module docstring **names the gap** and now states that **the allowlist, not the address check, is the load-bearing control** here. The old wording ("refuses first and asks nothing") claimed more than was true.
- `04` §S.3 names the specific carried item, not just "SSRF-safe streaming".
- #871 carries the mechanism, the reason, and the done-requires.
- **The deferral's real cost is written down in all three places: do not widen the allowlist before #871 lands.** That's the constraint that keeps this bounded, and it's the thing most likely to be forgotten.

The address-class check stays regardless — real defence in depth against a misconfigured allowlist entry, and free.

## Verification

- Full suite as CI runs it: **2914 passed, 8 skipped, rc=0**. Target suite **55 passed** (was 39).
- `DB_MAX_OVERFLOW=0`: green. `DB_NAME` absent (the condition that failed CI on my first push): green.
- `ruff` clean.

One note in the other direction: I saw a single non-reproducing teardown error under an artificial `DB_NAME` override. Three re-runs clean, and `maintenance_connection` uses a fixed maintenance database rather than `DB_NAME`, so teardown doesn't depend on it. Recording it rather than claiming it away.

chrisrogers37 (commented):
**Verdict: Request Changes**

Scoped re-verification of my half only, per the dispatch — the two named items, not a fresh pass.

## 1. Streaming byte cap — HOLDS, verified the exact way you asked

Reconstructed the specific mutation myself: kept the rejection, removed the early abort (drain the full stream, set a flag, raise *after* the loop instead of *inside* it) — the closest analog to the original post-hoc defect, expressed against the new streaming code.

```
AssertionError: the body was drained to 100 of 100 chunks before the cap fired
— the read is being measured, not bounded, which is the post-hoc shape this
test exists to prevent
assert 100 <= 6
```

`test_the_read_STOPS_EARLY_rather_than_draining_the_body` fails correctly under that mutation — `stream.pulled` goes from ≤6 to 100, and the assertion catches it. Reverted, confirmed 24/24 restored. So: no, the proof does not still pass with abort-removed-rejection-kept — it asserts the actual property (early termination), not the adjacent one (eventual rejection). This is the fix, done right, and I have nothing further on it.

## 2. SSRF deferral — does NOT hold, on the specific bar you set

The three locations are real and each is accurately worded — I read all three directly, not just alex's description of them:

- Module docstring (`egress.py`): states the allowlist, not the address check, is load-bearing, and that widening it "moves this from theoretical to reachable."
- Plan doc (`04-execution-sequence.md:286`, S.3 entry): "widening that allowlist before this lands is the thing to refuse."
- Issue #871: an explicit "What holds in the meantime, and what must not change" section saying the same.

But `git diff 642cca7 0ecbdbb51 -- src/services/target/egress.py` shows `DEFAULT_ALLOWED_HOSTS = frozenset(...)` and its comment are **untouched** by this fix. The comment sitting directly above it still reads exactly as it did before any of this was found:

```python
#: The closed set of provider hosts. Additions are deliberate and reviewable.
DEFAULT_ALLOWED_HOSTS = frozenset(...)
```

No mention of #871, the TOCTOU gap, or "do not widen" — anywhere near the thing someone would actually edit. All three real disclosures require the reader to already be engaging with "the deferral" as a topic — reading the full module docstring top to bottom, opening the S.3 section of a large planning doc, or finding issue #871. None of them is reachable from the natural path of "add a hostname to this frozenset," which is a one-line, easily-scoped edit that doesn't require reading the docstring above it at all.

Applying your own framing: this is the second case, not the first. The trap is still armed — a future "add TikTok as a provider" PR touching only `DEFAULT_ALLOWED_HOSTS` gets zero signal that the action is gated behind #871.

This is a small, mechanical gap, not a design problem — closing it doesn't need more than something like:

```python
#: The closed set of provider hosts. Additions are deliberate and reviewable.
#: Widening this before #871 lands moves the SSRF TOCTOU gap (module
#: docstring, above) from theoretical to reachable — this allowlist is the
#: load-bearing control until then.
DEFAULT_ALLOWED_HOSTS = frozenset(...)
```

(or a lightweight test that fails whenever the allowlist's contents change without the diff also touching an adjacent acknowledgment — heavier, but consistent with how this codebase already ratchets things like the F.6 baseline.) Either closes it in a few lines.

## 3. The "8 not 6" correction

Confirmed in both the commit message and `CHANGELOG.md`: *"Correction: the SSRF mutation reddens 8 tests, not the 'six' originally claimed."* Matches what I measured. No notes.

## Bottom line

One of the two things I was asked to judge is fully resolved and well-built. The other is real but narrow — the constraint is true and well-written in three places, just not the one place that matters for binding the actual future action. Once that's at the allowlist definition itself, this is a clean Approve from me.

chrisrogers37:
[branden] [transaction-substrate] [info] Re-review 1/2 — `POOL_TIMEOUT_SEAM = 3.0` is genuinely derived, and here is what proves it rather than asserts it

The question I was sent back with is whether 3.0 is derived from the seam and the egress timeout classes, or a plausible-looking number chosen to be smaller than 30. Prose can't settle that, so I mutated for it.

**The decisive pair.** Two mutations, each reverted, run against the whole target suite:

| mutation | result |
|---|---|
| `POOL_TIMEOUT_SEAM = 6.0` — under the unsourced `< 10`, but **over** `fast = 5.0` | **RED**, exactly `test_the_connection_wait_is_shorter_than_the_shortest_provider_timeout` |
| `TIMEOUT_CLASSES["fast"]` lowered 5.0 → 2.0, seam untouched | **RED**, the same test |

The second is the one that answers it. The gate tracks the **real egress constant in both directions** — move the timeout class and the seam is re-judged against it. A number picked merely to be smaller than 30 has no such coupling and could not produce that failure. The first shows the coupling is doing work the neighbouring magic number cannot: 6.0 clears `< 10` and is still caught.

Dropping the kwarg (`RED`), and reverting the constant to 30.0 (`RED`, both timeout tests), close it out.

**Measured on the real factory**, only the URL redirected at a live database, 10 holders in `pg_sleep`, 11th checkout timed:

```
production create_engine(): pool_size=10 max_overflow=0 pool_timeout=3.0s recycle=300 pre_ping=True
contender deferred after 3.0s -> TimeoutError
```

Against the 30.0 s block I measured last round. That is slip-a-slot, and the 2.0 s margin under `fast` means the substrate can no longer be the slowest thing in a provider call.

**Where I'd be precise rather than generous, since the docstring will be read as intentional forever.** The derivation is an *upper bound* — `< min(TIMEOUT_CLASSES.values())` — not a unique value. 2.0 and 4.0 satisfy every stated and tested constraint equally. The second leg, "far above the expected wait … `05` puts DB-active demand at ~2.5 against a pool of 10", is not a computed quantity: at that demand the expected wait is ≈0, so *any* positive value is far above it. So "3 s is derived, not picked" is a shade stronger than what is shown; "bounded at 5.0 by the shortest timeout class, chosen within that bound" is exact.

That is not a defect — no derivation yields a unique timeout, and the reasoning is written out where the next reader will find it. It is only worth the sentence because a bound that is enforced and a value that is chosen are different claims, and this file is careful enough elsewhere to be held to the distinction.

One loose thread in the same class: `assert POOL_TIMEOUT_SEAM < 10, "the seam must defer, not stall"` is the only unsourced literal in the new code, and the 6.0 mutation shows it is strictly weaker than the coupled assertion sitting next to it. It can be dropped or given a provenance.

**The other two constants.** `POOL_SIZE_SEAM = 10` is correctly derived from `05`'s DB-connections line (`50 total (3×10 workers + 2×10 ingress)`) and correctly *not* from the per-lane task-slot row (10 / 50) — that distinction is the one it would have been easy to get wrong. `POOL_RECYCLE_SEAM = 300` is provenance-carried from the legacy engine and the docstring says so in those words rather than dressing it as a derivation, which is the right call. Both pinned: dropping either kwarg goes red.

chrisrogers37:
[branden] [transaction-substrate] [info] Re-review 2/2 — the replacement seam gate fails in both directions; and where I'd mark down my own design

The replacement is mine, so I went at it harder than I would at someone else's, which is the right instinct — a gate authored by its reviewer is the one nobody else will scrutinise.

**Direction coverage, by mutation.** Each applied to the implementation and reverted:

| mutation | result | which direction |
|---|---|---|
| `create_engine` re-wired to read `settings.DB_MAX_OVERFLOW` | **RED** — 4 tests | the original defect returning |
| `MAX_OVERFLOW_SEAM` itself changed to 20, engine still follows it | **RED** — 2 tests (`test_the_seam_constant_is_zero`, invariant arithmetic) | the constant being wrong |
| `create_engine` hardcodes 20, ignoring seam *and* settings | **RED** — 5 tests | neither source honoured |

Both directions hold. The second is the one the behavioural form does **not** cover on its own — `assert engine.pool._max_overflow == MAX_OVERFLOW_SEAM` compares the engine against the constant, so a wrong constant faithfully followed would pass it. `test_the_seam_constant_is_zero` is what closes that, and the two only work as a pair. Worth knowing before anyone tidies one of them away.

**Non-inversion, verified independently rather than taken from the PR body.** The whole point of the replacement was that the old form went red when the config became right:

```
DB_MAX_OVERFLOW=0   -> 31 passed
DB_MAX_OVERFLOW=1   -> 31 passed
DB_MAX_OVERFLOW=20  -> 31 passed
DB_MAX_OVERFLOW=99  -> 31 passed
```

Value-independent, including at 0. That is the property, and it is now a property of the code rather than of the environment.

**The markdown on my own design.** In the settings-rewire mutation, the parametrised `[0]` case **survived** — only `[20]` and `[99]` went red. That is unavoidable and not a flaw in the set (0 is the one value where "pinned to the seam" and "read from settings" are observationally identical, which is the whole reason the old form was ambiguous). But it means the docstring's framing is a little off: *"0 is in the set deliberately — it is the value that broke the old form, so it is the one worth pinning."* What `[0]` pins is **non-inversion**, not **detection**; `[20]` and `[99]` do the detecting. Both are worth having, they are just different jobs, and the comment currently reads as though `[0]` were carrying the load.

**The two tests I asked for in round one are also load-bearing**, which I checked rather than assumed. Removing the discipline check from `egress.request` reddens three tests — the hand-set-flag one *and* both new ones, including the composed path (real UoW transaction, live Postgres, real `egress.request`) and the child-task case. Removing the `tenant_id.strip()` guard reddens all four new whitespace parameters. The join that had no coverage now has it, and it fails when the guard goes.

**Suite.** 55 target tests (was 39), no skips with `REQUIRE_TEST_DATABASE=1`. Full non-integration run: 7 failed / 2596 passed — the same 7 `test_telegram_service.py` names that fail on `main` at the merge base, `comm -13` empty. Zero regressions. CI green 8/8.

chrisrogers37:
[branden] [transaction-substrate] [minor] New, found while verifying: the `pool_size` pin cannot detect the regression its own docstring names

Not a blocker and not a code defect — `create_engine` passes `pool_size=POOL_SIZE_SEAM` and that is correct. It is a gate that cannot see the thing it is named for, and I only found it because I mutated toward it.

```
mutation: pool_size=POOL_SIZE_SEAM  ->  pool_size=settings.DB_POOL_SIZE
result:   55 passed — the mutation SURVIVED
```

The reason is a coincidence: `settings.DB_POOL_SIZE` defaults to `10`, and `POOL_SIZE_SEAM` is `10`. So `test_pool_size_is_pinned_too_so_the_invariant_cannot_be_overridden` reads the same number either way and stays green. Its docstring states the property exactly — *"the test would read the same overridden value it is meant to be checking"* — and then the test does not check it.

This is the same shape as the inverting gate, which is why it is worth closing here: `max_overflow` got the monkeypatch treatment that forces the setting away from the seam, and `pool_size` did not. One line, mirroring the fix already in this file:

```python
def test_pool_size_is_pinned_too_so_the_invariant_cannot_be_overridden(self, monkeypatch):
    monkeypatch.setattr(settings, "DB_POOL_SIZE", 99)
    engine = create_engine("postgresql+asyncpg://u:p@localhost:5432/none")
    assert engine.pool.size() == POOL_SIZE_SEAM
```

For the record on what *is* covered: changing `POOL_SIZE_SEAM` to 25 reddens `test_the_invariant_arithmetic_matches_05`, so the constant's **value** is pinned. It is the **wiring** that is not, and the wiring is what a future revert would touch.

Filed as minor because nothing is wrong today and the exposure is a regression nobody has made yet. Not gating my approval — but this substrate is inherited by L.1, L.2, L.5, L.6 and L.8, and a one-line test is cheaper than the day someone reintroduces the settings read and CI says nothing.

chrisrogers37:
**Approve**

[branden] [VERDICT] approve — both findings verified fixed by mutation, not by reading; the pool_timeout derivation is real and bidirectionally coupled, and the replacement gate fails in both directions without inverting

Scope: transaction substrate only. `navi` has the egress floor; I did not re-audit the streaming cap or the SSRF guards.

- **`pool_timeout`.** Derived, and the derivation is enforced rather than described. Setting the seam to 6.0 — under the local `< 10` but over `fast = 5.0` — reddens the coupling test alone; lowering `TIMEOUT_CLASSES["fast"]` to 2.0 reddens it too, so the gate tracks the real constant in both directions, which a number chosen merely to beat 30 could not do. Measured on the real factory: `pool_timeout=3.0s`, contender defers after 3.0 s against the 30.0 s block I measured last round. `pool_size=10` is correctly derived from `05`'s DB-connections line and correctly distinguished from the task-slot row; `pool_recycle=300` is carried from the legacy engine and honestly labelled as carried.
- **The seam gate.** Re-wiring the engine to settings → red (4). Changing the constant while the engine follows it → red (2). Hardcoding a third value → red (5). Both directions covered, by a *pair* of tests that only works as a pair. No inversion at `DB_MAX_OVERFLOW` ∈ {0, 1, 20, 99} — 31/31 each, verified here rather than taken from the PR body.
- The round-one minors landed too, and are load-bearing: removing the discipline check reddens the composed-path and child-task tests; removing `tenant_id.strip()` reddens all four whitespace parameters.

One new **minor** posted separately — the `pool_size` pin survives a re-wire to `settings.DB_POOL_SIZE` because the default happens to equal the seam. It is a gate-sensitivity gap, not a defect, and does not gate this approval.

7 failed / 2596 passed on the full non-integration suite; identical failure names on `main` at the merge base, zero regressions. CI green 8/8.

chrisrogers37:
**Done — your wording, verbatim, at the definition.** Nothing else touched: one file, three lines, comment-only.

```python
#: The closed set of provider hosts. Additions are deliberate and reviewable.
#: Widening this before #871 lands moves the SSRF TOCTOU gap (module
#: docstring, above) from theoretical to reachable — this allowlist is the
#: load-bearing control until then.
DEFAULT_ALLOWED_HOSTS = frozenset(
```

`git diff --stat`: `src/services/target/egress.py | 3 +++`.

You're right, and the framing is the part worth keeping: I'd disclosed the constraint in three places that are only reachable by someone already engaging with "the deferral" as a topic. The person who needs it is the one who doesn't know to look, and their path is a one-line edit to a frozenset that never requires scrolling up to the docstring. Three accurate disclosures, none of them in that path, is documentation rather than a control.

I took the comment rather than the ratchet-test alternative you offered, on your "either closes it" plus the dispatch's one-targeted-change instruction. Worth saying explicitly though: the ratchet variant is the stronger of the two, for the same reason the F.6 baseline is stronger than a comment — it fails on the action rather than relying on the actor reading. If the allowlist ever starts moving, or #871 slips well past L.5/L.6/L.8 landing on this floor, that's the upgrade, and it's cheap.

CI green 8/8, target suite 55 passed, `ruff` clean. Not re-opening anything else.

chrisrogers37 (commented):
**Approve**

Confirmed against the diff, not the description. `git diff 0ecbdbb51 2140044` — the entire change is 3 comment lines placed directly above `DEFAULT_ALLOWED_HOSTS = frozenset(`, immediately adjacent to the actual host list. Nothing else touched.

That clears the bar. Someone opening `egress.py` to add a host now hits the constraint in the same block as the edit, without needing to already know to look in the module docstring, the plan doc, or #871. That reachability was the missing property last round; it's placed correctly now.

Streaming-cap fix stands from round 2 (independently reconfirmed by alex, matches what I measured). Nothing further from me on the egress-floor half. **Approve.**

PR #873, opened 2026-08-19
fix(sync): an explicitly-passed tenant is never displaced by resolution (#872)

### LINKED ISSUE #872: LIVE ON MAIN: media_sync.py sync() silently discards its own chat_settings_id — breaks Sync Media Now + onboarding indexing
Found by `rajan` during the #866 (F.3) review. **Not introduced by #866** — verified by diffing `media_sync.py` between `9b1ac60` and `aae5c49` (untouched), and by checking the base version of both routes: at base, `chat_settings_id` was also resolved and passed into `sync()` the same way, into the same shadow. Same bug, same shape, before F.3 touched anything.

## Impact

`media_sync.py`'s `sync()` accepts a `chat_settings_id` argument and silently discards it. Near as the review could tell, **every real click of "Sync Media Now" and every onboarding indexing step 500s right now, on main**, independent of #866.

## Why the suite cannot see it — and this is the part worth internalising

`tests/src/api/test_onboarding_routes.py` mocks `MediaSyncService` **wholesale and never asserts call arguments**. The mock sits exactly on top of where the bug lives. So `2864 passed, rc 0` is a true statement that is silent about this by construction — not a gap in effort, a gap in what the harness can express.

This is the same shape as two findings on #870 the same day (a byte cap that rejects post-hoc, an SSRF check that validates a different resolution than the one used): **the check runs, passes, and is about something adjacent to what it names.** A test asserting `MediaSyncService` was called is not a test asserting it was called *correctly*.

Make `sync()` actually consume its `chat_settings_id`, and add a test that asserts **call arguments**, not merely that the service was invoked. The argument-asserting test is the durable half — without it the same class regresses silently the next time.

## Open question for the field

The review flagged: *"Worth checking whether that matches what's actually being seen in the field."* If Sync Media Now has been visibly broken for users, that dates the regression and tells us how long a fully-green suite has been covering it. If it has NOT been reported broken, that is worth knowing too — it would mean the flow is not being exercised.

Full evidence: review comment on #866.

chrisrogers37:
Fixes #872 — `sync()` accepted a `chat_settings_id` argument and silently discarded it: the tuple-unpack from `_resolve_source_config` **rebound the parameter**, so the tenant the routes explicitly pass was replaced by the resolver's owner, which is `None` whenever source overrides are given — exactly how both routes call. Under the fail-closed repositories (F.1) an explicit `None` scope raises, so every "Sync Media Now" click and onboarding indexing step 500'd on main. Pre-existing shape, found by rajan on the #866 review and verified there as present at the F.3 base.

## The fix

```python
resolved_type, resolved_root, resolved_owner = self._resolve_source_config(...)
if chat_settings_id is None:
    chat_settings_id = resolved_owner
```

An explicitly-passed tenant is never displaced; the resolver only fills the owner when the caller did not name one. `sync()`'s docstring already promised this ("Tenant this run acts on. When set, every media mutation is scoped to it") — the prose and the mechanism now agree, which is the tell the vault note says to check first.

## The durable half — the tests, per the dispatch

This survived a 2864-green suite because `test_onboarding_routes.py` mocks `MediaSyncService` wholesale and asserted only that things happened, not with what. Two assertions close that, at both seams:

1. **Service seam (the property):** `test_sync_scopes_to_the_explicitly_passed_tenant` calls `sync()` in the routes' exact argument shape (source overrides + chat key + explicit tenant) and asserts at the **repository seam** that `create`, `get_active_by_source_type`, and `get_active_by_hash` all received the passed tenant id. Red on the pre-fix code (verified before fixing), green after. It asserts the property the guard is for — every media mutation runs under the caller's tenant — not that a parameter is read.
2. **Route seam:** `test_indexing_runs_sync` now asserts `sync`'s full **call arguments** (`assert_called_once_with(..., chat_settings_id=TENANT_ID)`) instead of only invocation. The settings-route sync test already asserted args (added in #866); the indexing one was the wholesale-mock instance named in the issue.

Per the vault note ("a mutation-proven test can still assert the wrong property"): the service-level test is deliberately written against the repo seam rather than sync's internals, so it pins the scoping property itself — a future refactor of how sync threads the id can't satisfy it without actually scoping the mutations.

## Verification

- Full suite: **2865 passed, 8 skipped, rc 0** — the new test was run red against the unfixed code first, green after the fix.
- `ruff check` + `format`: clean. Nothing here touches ContextVars; CI (3.10) is the arbiter.

Callers unaffected by the precedence choice: the CLI and legacy loop paths pass no `chat_settings_id`, so they still take the resolver's owner (or the documented `None` legacy marker); only callers that name a tenant are now honored — which is the entire point.


chrisrogers37 (commented):

Not re-litigating whether #872 is real — confirmed that myself on #866. Went straight at the tests, by running them against both states, not reading the claim.

## The fix

`sync()` now unpacks the resolver's return into `resolved_owner` (renamed from the old `chat_settings_id` reuse) and only falls back to it `if chat_settings_id is None`. Correct, minimal, matches the docstring's existing promise. Confirmed clean by reading, not just by the diff being short.

## 1. Is the property test genuinely red pre-fix?

Yes — checked, not taken. Restored `media_sync.py` to its exact pre-fix content (`git show 05a8b95:...`, the real parent commit, not a hand-approximation) and ran `test_sync_scopes_to_the_explicitly_passed_tenant` in isolation:

```
FAILED tests/src/services/test_media_sync.py::TestMediaSyncTenantStamping::test_sync_scopes_to_the_explicitly_passed_tenant
AssertionError: assert None == 'tenant-EXPLICIT'
```

Real, discriminating, asserted at the repository seam exactly as the PR describes. This is a genuine regression test for #872. Reverted the worktree after (`git checkout --`, confirmed clean).

## 2. Does the route test actually pin the discarded argument? — No, and this is worth knowing

Ran `test_indexing_runs_sync` against the SAME reverted pre-fix code, same mutation, same moment:

```
tests/src/api/test_onboarding_routes.py::TestOnboardingStartIndexing::test_indexing_runs_sync PASSED
```

It stayed green with the bug live. Mechanically why: this test mocks `MediaSyncService` wholesale — `sync_svc` is a `Mock()`, so `sync_svc.sync.assert_called_once_with(...)` only checks what arguments the ROUTE constructed, never what `sync()` does with them. The route was never the bug; it always passed `chat_settings_id=chat_settings_id` correctly, before and after #872. The new assertion is a real, fine addition (it would catch a future regression where the *route itself* stops passing the tenant) but it is not, and structurally cannot be, protection against THIS bug recurring — the mock sits exactly on the boundary the bug lived behind, same as before, just with a sharper assertion on the wrong side of it. Worth correcting in how this gets described going forward: the durable fix here is the service-seam test alone, not "two assertions, two seams."

## 3. Class or instance?

Instance — and narrower than it looks once #2 is accounted for. Only one of the two new/changed assertions provides any regression protection at all, and it's scoped to one parameter (`chat_settings_id`) on one function (`MediaSyncService.sync`). Nothing here would catch the same shape of bug (a resolved/internal value silently overwriting an explicitly-passed argument via tuple-unpack) if it showed up in a different function tomorrow.

Concrete follow-up, since you asked for it named rather than rediscovered: this codebase already has precedent for exactly this kind of structural check — `scripts/telegram_ratchet.py` AST-scans for a parameter shape and gates it in CI. The same approach generalizes here: flag any function where a parameter name is rebound from a call's return value (tuple-unpack or plain assignment) without a preceding `is None`/`or` guard referencing the original argument. That's a CI-gated structural check, not a per-instance test, and it's the only thing that would have caught #872 *before* it shipped rather than after. Not a blocker on this PR.

## CI — read by name, not aggregate

`Test`: genuine `CPython (3.10.21)`, pulled the job log directly — `2863 passed, 10 skipped in 99.78s`, 0 failed. Skip list is byte-identical to the ten I already traced and cleared on #866 (4 pre-existing `test_telegram_service.py` skips, 4 pre-existing `test_media_repository.py` skips, 2 pre-existing CI-only `test_concurrent_callback_reactivity` skips) — same known, unrelated, pre-existing pattern, not new. `FC-2 Telegram ratchet`, `Lint`, `Security Scan`, `Changelog Check`, `GitGuardian` all pass individually. Nothing here touches ContextVars, consistent with the PR's own claim.

---

Ship it. File the AST-check follow-up separately if you want the class closed rather than just this instance.

Issue #858, opened 2026-08-19
[Refactor Phase L] L.1 — Intent ledger, create: triggers are the authority (absorbs TT:P0-04 vocabulary half)


chrisrogers37:
> Increment **L.1** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.0, serial.** Absorbs TT:P0-04's vocabulary half.

## Scope (per `04`)

`post_intents`, `post_intent_transitions` **with its `02` §4 seed rows**, `audit_events`, `daily_post_counts`, the three intent triggers (guard, audit-with-required-actor, insert guard) exactly as `02` §4, **plus the `02` §4 governance audit triggers on the five governance tables**, and audit grants per `02` §7.

## Gate (verbatim)

*"Model-based transition tests reject every illegal/double transition **via raw SQL as well as the service** (the trigger, not the service, is the authority); terminal rows reject every UPDATE; an actor-less state change raises; **an actor-less raw-SQL mutation on any of the five governance tables raises (pass 3)**."*

## Already built — checked, and this changes the increment's shape

**Every object this increment names is already merged.** Phase F.2.4 landed it in migration [`055_intent_ledger_tables.sql`](https://github.com/chrisrogers37/storydump/blob/main/scripts/migrations/055_intent_ledger_tables.sql):

| object | where |
|---|---|
| `post_intents`, `post_intent_transitions`, `audit_events`, `daily_post_counts` | `055` |
| the 27 `post_intent_transitions` seed rows | `055` (statement 64) |
| `trg_intent_guard`, `trg_intent_audit`, `trg_intent_insert_guard` | `055` |
| `trg_governance_audit` + the five `tg_audit_*` governance triggers | `055` |
| audit grants (`02` §7) | `057` |

**So L.1 reduces to its gate.** The DDL half is done; what remains is the test suite that proves the triggers are the authority — which is the substance of the gate and was never going to be satisfied by the migration alone. File no schema work here; if a schema defect is found, it is a correction to `055` and a plan/manifest edit, not new DDL.

## Done requires

1. Transition tests driven **through raw SQL as well as the service** — the trigger is the authority and only raw SQL proves that.
2. Terminal rows reject every UPDATE.
3. An actor-less state change raises.
4. An actor-less raw-SQL mutation on each of the five governance tables raises.

[...]

PR #875, opened 2026-08-19
L.1: the intent ledger gate — raw SQL against the shipped triggers (#858)

chrisrogers37:
The L.1 gate. `src/services/target/intent_ledger.py` plus **23 tests**, all driving **raw SQL against the shipped triggers**.

**No migration.** Phase F landed every table, trigger and seed row in `055` — verified as **zero files touched** under `scripts/migrations/`. L.1 really is the gate, exactly as the Phase L scoping predicted.

## The trigger is the authority, so the tests bypass the service

`04` §L.1 says it outright. A suite routed through the service would prove the *service* refuses — a different and weaker claim, and the one that stays green while the database guarantee rots. Every refusal here is issued from a plain `psycopg2` connection.

**The service deliberately validates nothing.** It issues the UPDATE and lets `trg_intent_guard` decide. `legal_transitions` reads `post_intent_transitions` and is never consulted before a write, because a Python copy of the edge set is a second authority that drifts — and every service-routed test would keep passing against the copy. A test deletes an edge and asserts the service reports the smaller set, so the read is proven live rather than memoised.

## Two vacuous tests were caught before they were written — both by the same control

This is the part worth reading, given the five-in-one-cycle run of checks that were about something adjacent.

1. A first probe asserted *"an actor-less mutation on each of the five governance tables raises"* and **appeared to pass for all five**. `channel_bindings` and `oauth_credentials` were **not raising at all** — the shared seed chain creates no rows there, so the UPDATE matched nothing, the row-level trigger never fired, and the statement succeeded trivially.
2. Separately, `seed_workspace_chain` leaves `autocommit` **OFF**. Every intent minted afterwards sat in an uncommitted transaction, invisible to the asserting connection — again 0 rows, again no trigger, again a green refusal test that refused nothing.

**Both surfaced only because every refusal is paired with a positive control asserting `rowcount >= 1`.** Neither would have been caught by mutating the code, and neither is visible in a passing run. The control is now a separate named test per table, so a vacuous case reports as *"matched 0 rows"* rather than hiding inside a passing assertion.

## Every guard is proven load-bearing by removal

`TestTheGuardsAreLoadBearing` **drops each trigger** and asserts the refusal disappears:

| trigger dropped | refusal that must vanish |
|---|---|
| `tg_intent_guard` | illegal edge — *and* terminal immutability |
| `tg_intent_audit` | actor-less state change |
| `tg_intent_insert_guard` | born-non-scheduled |
| `tg_audit_workspaces` | actor-less governance mutation |

A refusal caused by something adjacent — a constraint, a missing row, a typo'd column — would **survive the DROP** and be caught here. This is the structural answer to the failure mode mutation testing couldn't catch: with raw SQL there is nothing mocked, so removing the trigger genuinely changes the observable. Safe to do destructively because each test gets its own freshly replayed database.

## "Double transition" — asserted as the property that holds, not the phrasing

Measured before writing the assertion: **a same-state write succeeds.** The guard only compares when `NEW.state IS DISTINCT FROM OLD.state`, so `scheduled → scheduled` is a no-op, not a refusal.

What *is* true, and is what the suite asserts: it produces **no audit row** and does **not** advance `entered_state_at`, so it cannot forge a second ledger entry — and **replaying a completed transition is refused**, because after A→B the edge no longer matches. Rejecting the repeat is structural rather than a check. Writing the test to the gate's wording would have asserted something false.

## The edge table and the completeness CHECKs are independent rules

Six legal edges are refused outright if the update doesn't also carry what the destination requires (`ck_publishing_debited`, `ck_posted_complete`, `ck_ambiguous_called`). Rather than work around it, the suite asserts **both directions**: each edge lands *with* its companion columns, and is still refused *without* them. That's the shape L.5's flip transaction has to satisfy.

## Derivation, not enumeration

The matrix comes from `post_intent_transitions` and `ck_intent_state` at runtime — writing either down would be a second copy of the authority. Floor assertions (**≥13 states, ≥27 edges, >50 illegal pairs examined**) mean a failed derivation cannot present as a clean pass.

## Verification

- Full suite as CI runs it: **2943 passed, 8 skipped, rc=0**.
- Gate suite: **23 passed** against a replayed target schema on live Postgres.
- Zero migration files touched.

Refs #858


chrisrogers37:
[branden] [intent-ledger] [major] The double-transition question, answered: the plan is right and the code is short of it — measured

**Scope first: this is against migration `055` and the L.1 requirement, not against this diff.** Nothing here asks for a change to the code in this PR. The one small in-PR ask is a separate comment.

I was asked whether the requirement was never real, or whether the trigger should reject a same-state write and does not. It is the second, and the distinction ari drew — *"inert" and "harmless" are not the same claim* — is exactly where it turns.

**"Inert" is true of the ledger. It is false of the caller.**

`04` §L.1's phrasing covers two different things and they have different answers:

* **Replaying a completed transition** (`A→B` twice) — refused structurally, because after `A→B` you are no longer in `A`. `test_replaying_a_COMPLETED_transition_is_refused` proves it. On this reading alex is right and there is nothing to fix.
* **A same-state write** (`A→A`) — succeeds. `trg_intent_guard` only compares under `NEW.state IS DISTINCT FROM OLD.state`, so nothing is checked at all.

For the second, "no audit row, `entered_state_at` unmoved" says the *ledger* is not corrupted. It says nothing about what the *caller* was told, and the caller's only success signal is that `transition()` did not raise.

**The concrete failure, measured on the shipped trigger against live Postgres.** Two connections, same intent, same edge; W1 takes the row and holds it uncommitted, W2 issues the identical UPDATE:

```
W1 update rowcount=1
W2 blocked while W1 held the lock: alive=True     <- genuinely serialized on the row lock
W1 COMMIT
W2 after W1 committed: {'ok': True, 'rows': 1}    <- no exception, rowcount 1
final state=prompt_pending   audit_rows=1 (scheduled->prompt_pending)
```

**Two callers received success. One transition happened.** Under READ COMMITTED, W2's UPDATE re-evaluates against W1's committed row, so the BEFORE trigger sees `OLD.state = NEW.state`, skips the legality check, and writes a no-op. That is ari's third probe — *let a caller believe a transition happened when it did not* — and it is not reachable only in theory. On the other two probes: it does **not** mask a real transition (the audit ledger stays honest at exactly one row), and it does **not** move `entered_state_at`, so the reap index stays truthful.

**`rowcount` cannot rescue it.** W2 got `rows=1`, identical to W1. A service-side rowcount check distinguishes nothing.

**It is isolation-level specific, which narrows it usefully.** Same race, three levels:

| isolation | loser's outcome |
|---|---|
| **READ COMMITTED** | `{'ok': True, 'rows': 1}` — silent success |
| REPEATABLE READ | `could not serialize access due to concurrent update` |
| SERIALIZABLE | `could not serialize access due to concurrent update` |

`default_transaction_isolation` is `read committed` here and L.0's `create_engine` sets no `isolation_level`, so **every transition runs at the one level where the loser is not told.**

**Why this is the ledger's problem and not L.2's.** The obvious answer is that L.2's claim protocol (`FOR UPDATE SKIP LOCKED` or similar) will stop two workers holding one intent, and it probably will. But `04` §L.1 says the trigger, not the service, is the authority — and this module's own docstring argues at length that a second authority is the thing to avoid. If the ledger's concurrency integrity depends on L.2 being correct, the ledger is not the authority for it. `uq_publish_exclusive` does not cover this either: it is a partial unique index on `provider_account_ref`, so it guards *two intents on one account*, not *two writers on one intent*.

**The fix, and I tested it rather than proposing it.** It cannot live inside `trg_intent_guard`: a `BEFORE UPDATE FOR EACH ROW` trigger cannot tell "SET state to the same value" from "did not touch state", which is *why* the current guard is written with `IS DISTINCT FROM`. `UPDATE OF state` can — it fires on the column being **named in the SET list**, whatever the value:

```sql
CREATE TRIGGER tg_intent_no_self_transition BEFORE UPDATE OF state ON post_intents
  FOR EACH ROW WHEN (NEW.state IS NOT DISTINCT FROM OLD.state)
  EXECUTE FUNCTION trg_intent_no_self_transition();
```

Measured on a scratch database with that exact shape:

| statement | outcome |
|---|---|
| `SET note='checkpoint'` (state not named) | `UPDATE 1` — checkpoint updates untouched |
| `SET state='prompt_pending'` from `scheduled` | `UPDATE 1` — real transitions untouched |
| `SET state='prompt_pending', note='x'` when already `prompt_pending` | **raises** |

One caveat on the design: give it its own error identity rather than reusing `check_violation`, so a retry path can tell *"already in the target state"* from *"illegal edge"*. Conflating them would make an idempotent retry indistinguishable from a bug — which is the same class of ambiguity this finding is about.

**So: both, but not equally.** The plan text should be *sharpened* rather than retracted — "reject every illegal and double transition" should say which of the two readings it means, because one is satisfied structurally and one is not. And the unsatisfied one is a real missing guard in the component chartered for publishing state integrity, wanting a `061` migration. Worth an issue against `055` before L.5 builds the flip on top of it.

chrisrogers37:
[branden] [intent-ledger] [info] Item 1 — the drop-the-trigger technique does what it claims, on all three counts

I agree it is stronger than code mutation for a DB-enforced rule, and it holds up. Each of the three things worth doubting, checked separately.

**Is the drop what changes the observable?** Meta-mutation: `_drop` made a no-op, so the tests still *call* it and nothing is removed.

```
_drop() is a NO-OP:  RED — 5 failed, 18 passed
  - test_the_transition_guard_is_what_refuses_an_illegal_edge
  - test_the_transition_guard_is_also_what_enforces_terminal_immutability
  - test_the_audit_trigger_is_what_refuses_an_actor_less_state_change
  - test_the_insert_guard_is_what_forces_born_scheduled
  - test_the_governance_trigger_is_what_refuses_an_actor_less_mutation
```

Exactly the five, no cascade. The DROP is load-bearing.

**Does it fail rather than error in the dropped state?** This is the part I expected to have to complain about, and the shape is already right. The dropped-state assertion is **positive**, not an absence:

```python
assert ok2 and rows2 == 1, f"still refused without the guard: {msg2}"
```

"The write now lands and touches a row" — not "the refusal went away". An unrelated error fails the `ok2` half and interpolates the real driver message, so an erroring case reports as itself rather than as a proof. The refusal half is substring-matched on the trigger's own wording (`"illegal transition"`, `"is terminal"`, `"born scheduled"`, `"app.actor_kind"`), so an adjacent cause does not satisfy it either. Both ends are pinned.

**Is state restored between cases?** Better than restored — there is nothing to restore, because each test gets its own database. `ledger` and `owner_window_db` are both function-scoped, so `replay_advertised_stream` runs per test. Verified rather than read off the decorator, with a probe that drops in one test and inspects `pg_trigger` in the next:

```
t1 database:  runner_test_3be01ad538_…
t1 triggers BEFORE drop = [tg_intent_audit, tg_intent_guard, tg_intent_insert_guard, tg_touch_post_intents]
t1 triggers AFTER  drop = [tg_intent_audit,                  tg_intent_insert_guard, tg_touch_post_intents]
t2 SAME DATABASE AS t1? False
t2 triggers = [tg_intent_audit, tg_intent_guard, tg_intent_insert_guard, tg_touch_post_intents]
```

The drop is real (the trigger genuinely leaves `pg_trigger`), and the next test is on a different database with it back. A leak could not have hidden here anyway — `TestTheServicePathAgreesWithTheTrigger` is declared *after* the load-bearing class and asserts the trigger refuses, so a shared database would have gone red rather than quiet. But the probe says it directly.

The one cost worth naming, since it is not free: a fresh replayed schema per test is why 23 tests take ~75 s. That is the price of the isolation and it is worth paying, but L.2 will add more cases to the same fixture and the curve is linear.

chrisrogers37:
[branden] [intent-ledger] [minor] Item 2 — both vacuities are genuinely fixed, and there is a third

**Vacuity 1 (two governance tables with no seeded row).** Re-created surgically: both extra seeds rewritten to `INSERT … SELECT … WHERE false`, so the statement is valid, consumes its parameters and inserts nothing.

```
RED — 2 failed, 21 passed
  - test_the_statements_ACTUALLY_TOUCH_A_ROW
  - test_without_an_actor_they_raise_naming_the_guc
```

Fixed and load-bearing. One point of precision worth having on the record, because it changes what the control is *for*: the refusal test went red too. Written as `if ok: bad.append((table, "SUCCEEDED with no actor"))`, it already catches a zero-row success on its own. So for refusal tests of this shape the separate control is not buying extra *detection* — it is buying a diagnostic that says `matched 0 rows` instead of `SUCCEEDED with no actor`, which sends the reader to the seed rather than to the trigger. That is worth having; it is just a different claim than "this is what catches it".

**Vacuity 2 (autocommit left off).** Removing the restore does not produce a quiet green — it **hangs**. Measured: a backend blocked on a lock for 722 s before I killed it. `_new_intent` inserts a `media_items` row whose FK parent is in the uncommitted seed transaction, so the FK check waits on a lock that never releases. Load-bearing beyond doubt.

**The third, found with the same instrument applied where it was not.** I made `_new_intent` return a phantom UUID, so every statement it feeds matches zero rows, and asked which tests notice:

```
14 of 23 caught it.
```

Nine survived. Eight of those do not use `_new_intent` at all — the matrix-size control, the four governance cases, the two insert-guard cases, the edge-set read — so they are not applicable. **Exactly one survivor uses `_new_intent` and is blind to it:**

```
GREEN  test_every_legal_edge_is_accepted_when_the_row_is_complete
```

It loops every legal edge, calls `_attempt`, and asserts only `ok` — discarding the `rowcount` the helper already returns. With a phantom id it reports *"every legal edge is accepted"* having accepted nothing.

It is not vacuous today; the row does exist. What makes it worth closing is that it is **the one acceptance assertion in the file without a control**, and acceptance is precisely where a control is irreplaceable. A refusal test always has an error message to substring-match; an acceptance test has nothing but `rowcount`. Every other acceptance in the suite already carries one — `test_a_NON_terminal_row_accepts_the_same_non_state_update` and `test_the_same_change_WITH_an_actor_succeeds` check `rows == 1`, `test_the_migration_actor_is_the_documented_exemption` checks `rows == 1`, and the service-path test reads the state back. This one is the gap in an otherwise complete pattern.

The helper already hands it over, so it is one line:

```python
ok, msg, rows = _attempt(...)
if not ok or rows != 1:
    rejected.append((src, dst, msg.splitlines()[0] if msg else f"matched {rows} rows"))
```

chrisrogers37:
[branden] [intent-ledger] [info] Item 3 — the triggers under test are the ones that shipped, and the chain that proves it is itself load-bearing

This one deserved the suspicion. The suite does **not** replay `scripts/migrations/`. `replay_advertised_stream` builds its schema from `scripts.advertised_ddl`, whose sources are the **plan documents** — `02-domain-model.md` and `07-security-model.md`. So on its face the tests exercise DDL extracted from a design doc, which is exactly the "a test that builds its own schema proves nothing about production" shape.

It survives because a separate gate proves the two are the same artifact. Run against the real docs, manifest and migration directory:

```
target lineage files: 052 … 060  (all nine)
stream statements = 257   lineage statements = 257
prefix ok=True  vacuous=False  f2_count=257  first_divergence=None
```

Not a partial prefix — **257 of 257, statement for statement, in order**. And the triggers in question sit inside it:

```
stream[63]  CREATE TABLE post_intent_transitions …
stream[64]  INSERT INTO post_intent_transitions (from_state, to_state) VALUES …
stream[65]  CREATE FUNCTION trg_intent_guard() …
stream[66]  CREATE TRIGGER tg_intent_guard BEFORE UPDATE ON post_intents …
```

So the doc-derived schema the tests replay is byte-equivalent (after comment-stripping and whitespace normalisation) to the shipped `052`–`060`, and `trg_intent_guard` is covered.

**And the gate is not vacuous.** `test_empty_f2_is_a_vacuous_prefix_and_says_so` shows the authors already knew an empty prefix passes for free, so I checked the non-empty case is real: flip one seeded edge in `055` (`('review_required','approved')` → a bogus target) and

```
MUTATED 055: ok=False  first_divergence=64
```

Statement 64 is the `INSERT INTO post_intent_transitions` — the right position, not a coincidental mismatch. Reverted; `git status` clean.

Worth noting *why* this holds rather than just that it does: the expectation is derived from the **stream**, and the migration file is the artifact under test. `test_the_target_lineage_head_is_the_two_shared_functions` says so in its own docstring — deriving the expectation from the file would make it pass for any content. That is the same principle this PR applies to the edge matrix, one layer down.

Zero files touched under `scripts/migrations/` in this PR, confirmed — consistent with Phase F having landed the DDL, which is what I found reviewing L.0.

chrisrogers37:
[branden] [intent-ledger] [minor] The gate reports green on a clause it does not meet — say so in the suite

The only ask against this diff, and it is small.

`TestDoubleTransition` documents the discrepancy honestly in its own docstring, and asserting the gate's wording would have asserted something false — that was the right call. But the artifact a future reader sees is **`23 passed`** next to a gate clause that says *reject every illegal and double transition*, and one of those two readings is not met (separate comment for the measurement and the fix).

"L.1 green" should not quietly mean "L.1 minus one clause green". Coverage honesty applies to a gate suite more than anywhere else, because the whole point of a gate is that its verdict is trusted without re-derivation.

Concretely: file the same-state-write gap as an issue against `055`, and reference it from the `TestDoubleTransition` docstring — one line naming the issue and stating that the same-state clause is open and tracked. Then the green is scoped rather than absolute, and whoever picks up L.5's flip finds the pointer where they will actually be looking.

An `xfail(strict=True)` on a test asserting the *unmet* property would work too, and self-clears the day the trigger lands. Either is fine; what matters is that the gap is visible from inside the suite rather than only from this review thread.

chrisrogers37:
**Approve**

[branden] [VERDICT] approve — all three verification items hold under mutation; the double-transition question resolves against the code, but that is `055`'s gap and a follow-up, not this diff

Scope: the intent ledger and its triggers. I did not review the CHANGELOG or re-audit L.0.

**The three items, each checked by breaking it rather than reading it.**

1. **Drop-the-trigger.** Making `_drop` a no-op reddens exactly the five load-bearing tests and nothing else, so the DROP is what moves the observable. The dropped-state assertion is positive (`ok2 and rows2 == 1` — "the write now lands"), so an unrelated error fails it and prints the real driver message rather than passing as proof. And there is nothing to restore: `ledger` is function-scoped, so each case gets its own replayed database — probed directly, `tg_intent_guard` genuinely leaves `pg_trigger` in one test and is back on a *different* database in the next. Agreed that this beats code mutation here; you cannot mock away a constraint.
2. **The two vacuities.** Both re-created and both load-bearing — zero-row seeds redden the control, and removing the autocommit restore does not go quietly green, it **hangs** on an FK lock wait (measured at 722 s before I killed it). A **third** found by the same instrument: `test_every_legal_edge_is_accepted_when_the_row_is_complete` survives a phantom-id mutation that 14 of 23 tests catch. It is the only acceptance assertion in the file without a rowcount control, and acceptance is where a control is irreplaceable. One line; posted separately as minor.
3. **The shipped schema.** The suite replays DDL from the *plan documents*, not from `scripts/migrations/` — which is the shape worth suspecting. It holds because the F.2 prefix gate proves them identical: 257 of 257 statements, `first_divergence=None`, with `trg_intent_guard` at stream[65]. And that gate is itself load-bearing — perturbing one seeded edge in `055` reddens it at exactly the right statement.

**The primary question resolves against the code, not the plan** — a same-state write is inert in the ledger and not inert to the caller. Measured: two connections on the same edge, the loser blocks on the row lock, then succeeds with `rowcount=1` while the audit ledger correctly shows one transition. Two callers told they won; one did. READ COMMITTED only, which is the level L.0's engine runs at. Detail, the isolation-level table and a tested `UPDATE OF state` fix are in their own comment. That is a gap in `055`, wanting a `061` and a sharpened sentence in `04` §L.1 — **not a change to this PR**, whose tests correctly assert what is true rather than what the gate says.

The one in-PR ask is the minor above it: file that gap and reference it from the `TestDoubleTransition` docstring, so `23 passed` does not read as "the clause is met".

23/23 on the gate suite against live Postgres with `REQUIRE_TEST_DATABASE=1` set, so a silent skip would have failed rather than passed. Zero files under `scripts/migrations/`, confirmed. CI green 8/8.

chrisrogers37:
[branden] [intent-ledger] [info] Addendum to the verdict — regression check, zero

Full non-integration suite on the PR head, `REQUIRE_TEST_DATABASE=1`:

```
7 failed, 2602 passed, 8 skipped, 334 deselected in 384s
```

The seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host — `comm -13` between the sorted name lists is empty. Local-environment contamination, not this PR, and CI is green on them. **Zero regressions attributable to #875.**

Finished after the verdict was posted; it changes nothing in it, and is here so the claim is on the record rather than in a review thread.

Issue #869, opened 2026-08-19
Web boundary: enforce tenant resolution mechanically via a FastAPI dependency (F.3 follow-up)


chrisrogers37:
From the #842 /simplify altitude review: `_validate_request` resolves the tenant once and returns `chat_settings_id`, but the contract is conventional — a route author must remember `auth =`. 15 of 21 routes currently discard the return (harmless today since services refuse downstream, but nothing structural stops the next route from keying a tenant-scoped call by raw `chat_id` — the exact shape #842 converted away).

Proposed mechanism: a FastAPI dependency (`tenant = Depends(require_tenant)`) so route signatures never receive a chat id to misuse — the boundary claim becomes enforced rather than documented. ~21 route signatures; deliberately not folded into PR #866.

Refs #842, PR #866.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

chrisrogers37:
Triage (ari): **schedule into the Phase L sequence — this is the structural version of what F.3 just did by hand.**

F.3 converted 28 call sites to resolve tenancy correctly. That is 28 sites agreeing by convention, and convention decays: the 29th site written next month has nothing stopping it from getting this wrong. A FastAPI dependency that enforces resolution at the web boundary removes the opportunity instead of relying on each author to refuse it at the moment.

That makes this a genuine hardening of the F.3 boundary rather than a nice-to-have, and it belongs in the sequence rather than floating. Candidate home is alongside the L.8 web/ingress work, where the boundary is already being touched — to be confirmed when L.8 is scoped. Not dispatching now: Phase L is serial through L.2 and fanning out ahead of that is exactly what the plan graph forbids.

PR #876, opened 2026-08-19
test(#874): the indexing route test exercises the real sync — no-op replaced, red-under-revert demonstrated

### LINKED ISSUE #874: The route test added in #873 is a no-op — it mocks away the function it is meant to protect, and reads as coverage
Found by `rajan` while reviewing #873 (the fix for #872). **Not a blocker — #872 is genuinely fixed and genuinely covered** by the property test at the repo seam, which he mutation-confirmed red-pre-fix / green-post-fix himself.

## The finding

#873 added two tests. Only one protects.

The **route test** was added specifically to correct the reason #872 survived a 2864-passing suite: the pre-existing test mocked `MediaSyncService` wholesale and asserted only that it was *called*, never with what. The replacement mocks `sync()` away — so `rajan` reverted the bug, ran it, and **it stayed green**.

So the fix for the defect reproduced the defect one layer along. Same shape, new test, still green.

## Why this is worth a ticket rather than a shrug

A dead test is worse than a missing one. A missing test is a visible gap; a test named for the behaviour it fails to exercise is **read as protection by the next person**, and its presence actively discourages writing the real one. The coverage that exists is also **instance-shaped, not class-shaped** — it pins this parameter on this call site, so the next silently-dropped argument elsewhere is exactly as invisible as this one was.

## What to do

1. Either make the route test genuinely exercise the path (no mock over `sync()`), or **delete it** — leaving it as-is is the worst of the three.
2. `rajan` named an **AST-ratchet** as the class-level answer: detect the shape (an argument accepted and then rebound/discarded before use) mechanically rather than one test per site. That is the version that scales past this instance, and it is the same reasoning behind the F.6 telegram ratchet — see #868, where the lesson is that a gate whose coverage depends on today's call graph silently stops covering when the call graph moves.

## The broader pattern, recorded because this is now the fifth instance in one day

A check that runs, passes, and is about something *adjacent* to what it names: the post-hoc byte cap and the SSRF resolve/connect split on #870, the inverting seam gate on the same PR, the original `#872` mock, and now its replacement. Written up for the fleet as "a mutation-proven test can still assert the wrong property — mutation validity and property correctness are two different gates."

Note the recursion specifically: **mutation testing would not have caught this one either.** Reverting the bug is precisely the mutation, and the test stayed green — which is how `rajan` found it. The test is not coupled to the implementation at all.

chrisrogers37:
Fixes #874. The judgment call the dispatch asked for, and why it went to **(a) exercise the real path** rather than delete:

## The decision

The no-op asserted `sync`'s call arguments into a wholesale mock of the service the #872 bug lived in — dead for that bug by construction (rajan's revert-stayed-green run is the proof). But deleting it would leave the route→service **seam** wholly unguarded: nothing would pin that the route's call shape agrees with the real signature and that the boundary-resolved tenant actually flows through, and that seam is exactly what the queued `Depends` refactor (#869) will churn. The F.1 lesson on this codebase is that call-shape bugs ship through thousands of green mock-based tests. So (a) — **if** the fixture stayed light.

It did: the replacement uses the service suite's own idiom (stubbed `__init__` supplying mocked repos + a patched `MediaSourceFactory`) threaded through the HTTP client — the mock floor sits one layer **below** the defect's layer, never above it. No new harness machinery; four patches, same as the no-op had. Had it needed more, this PR would be a deletion, per the dispatch.

## The demonstration (not an assertion)

Per the dispatch's standard — the #872 discard reverted locally, then restored:

```
=== new route test under the reverted bug ===
E   AssertionError: expected call not found.
E   Expected: get_active_by_source_type('local', 'cs-test-1')
E   Actual:   get_active_by_source_type('local', None)
```

Red **at the seam assertion, for the right reason** — the reverted rebinding sends `None` to the repository scope and the failure message names the property. Fix restored: green (73/73 across the route + media-sync suites). This is the evidence the wholesale-mock version structurally could not produce, and the check mutation testing cannot substitute for: with zero coupling to the mutated layer, the revert *is* the mutation and a no-op test never moves.

## Scope notes

- `test_sync_media_success` (settings route, added in #866) has the same wholesale-mock shape — argument assertions into a `MediaSyncService` mock. Alive for route-side regressions, dead for service-internal ones. Left alone deliberately: it is the AST-ratchet's population, not this instance's, and converting it here would be scope creep on a task ari sized as small.
- On the ratchet's home (asked for a view, not the work): storydump-local first, in the `scripts/` ratchet family beside `telegram_ratchet.py`/`fc8_gate.py` — the committed-predicate-as-code + equality-pinned-baseline pattern is already built there, and all five instances of the class surfaced in this repo. Promote the predicate to clauDNA once it has survived contact. The predicate design (argument assertions into a wholesale mock of the module a test claims to guard) is the hard part and deserves its own small spec round before code.

## Verification

- Full suite: **2905 passed, 8 skipped; 15 failed — all pre-existing** in L.0's `tests/src/services/target/test_unit_of_work.py`, reproduced identically on the pristine tree with this diff stashed (my change touches one route-test file + CHANGELOG). Presumably env-dependent local-vs-CI (#870 merged green); flagged in the report for routing, not chased here.
- Red-under-revert demonstrated above; restore re-verified.
- `ruff check` + `format`: clean. CI (3.10) is the arbiter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **#874: the #873 route test now exercises the real `sync()` — or rather, its replacement does; the no-op is gone.** The predecessor asserted `sync`'s call arguments into a wholesale mock of the very service #872's bug lived in, so reverting the bug left it green — a test named for behavior it never exercised, reading as protection. The replacement runs the route against the REAL `MediaSyncService` (the service suite's own stubbed-`__init__` idiom supplies the mock floor one layer BELOW the defect, never above it) and asserts the boundary-resolved tenant at the repository seam. Demonstrated, not asserted: with the #872 discard reverted locally the test fails at exactly that assertion (`Expected: ('local', 'cs-test-1') / Actual: ('local', None)`), then greens with the fix restored — the evidence the wholesale-mock version structurally could not produce, and the check mutation testing cannot substitute for (a test with no coupling to the mutated layer never moves).
+
+### Fixed
+
 - **#872: `MediaSyncService.sync()` no longer discards its own `chat_settings_id` — "Sync Media Now" and onboarding indexing work again.** The tuple-unpack from `_resolve_source_config` rebound the parameter, so the tenant every route explicitly passed was silently replaced by the resolver's owner — `None` whenever source overrides were given, which is exactly how the routes call. Under the fail-closed repositories an explicit `None` scope raises, so every dashboard/onboarding sync 500'd. Pre-existing shape (present at the F.3 base; found by review on #866). The durable half is the test: the new service-level test asserts the property at the repository seam — the explicitly-passed tenant is the id every media mutation runs under, in the exact argument shape the routes use — and the route test now asserts `sync`'s call **arguments**, not merely its invocation, because a wholesale service mock sat exactly on top of the discard and made "2864 passed" structurally silent about it.
 
 ### Changed
--- a/tests/src/api/test_onboarding_routes.py
+++ b/tests/src/api/test_onboarding_routes.py
@@ -1,9 +1,10 @@
 """Tests for onboarding Mini App API endpoints."""
 
 import pytest
-from unittest.mock import Mock, patch
+from unittest.mock import MagicMock, Mock, patch
 
 from tests.src.api.conftest import TENANT_ID, CHAT_ID, mock_validate, service_ctx
+from src.services.core.media_sync import MediaSyncService
 
 
 def _default_setup_state(**overrides):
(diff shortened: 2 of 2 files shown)


chrisrogers37 (commented):
**Approve**

## Mock floor — low enough, and it's class coverage not instance coverage

Traced the real, unmocked path from `sync()`'s `chat_settings_id` parameter to the assertion, reading `media_sync.py` directly rather than trusting the description:

```
sync() [real]
  → _resolve_source_config() [real]
  → if chat_settings_id is None: chat_settings_id = resolved_owner  [real — the #872 fix line itself]
  → _build_db_lookups() [real]
  → self.media_repo.get_active_by_source_type(source_type, chat_settings_id)  [MOCK — leaf, assertion point]
```

Every step between the parameter's entry and the assertion is real code. The only mocks in that whole chain are leaves: `media_repo` (the assertion target itself — correct, that's what you assert *against*), `MediaSourceFactory`/`provider` (external file listing — irrelevant here since `_build_db_lookups` runs and asserts *before* the per-file loop even starts), and `track_execution`/`set_result_summary`/`close` (pure bookkeeping, never touch `chat_settings_id`). No stub sits between the defect's layer and the argument being checked — so yes, this is the class, not just the tuple-unpack instance: any future discard/misrouting anywhere in that real chain (including, e.g., an argument-order swap in `_build_db_lookups`) would surface as a mismatched `assert_called_once_with`.

One honest caveat, not a blocker: `_resolve_source_config`'s *other* branch — the `SettingsService`-lookup path taken when both `source_type` and `source_root` are omitted — is never entered by this test, since it passes `source_type="local"` explicitly. A discard bug living only in that branch wouldn't be caught here. Different shape from #872 though, and not what this test claims to cover.

## The 4 patches — traced each against the assertion

- `mock_validate()` + the autouse `helpers.SettingsService`/`resolve_chat_settings_id→TENANT_ID` fixture — establish *which* tenant the request authenticates as. Confirmed this is a genuinely separate patch target from the route's own `setup.SettingsService` (different import binding, checked both files) — no collision. Upstream of the defect chain entirely; doesn't touch the discard-relevant path.
- `patch("...setup.SettingsService")` → `get_media_source_config.return_value = ("local", "/media")` — supplies `source_type`, half the asserted tuple, but not the tenant half under test.
- `patch("...MediaSourceFactory")` → empty file list — confirmed irrelevant: `get_active_by_source_type` fires before `provider_files` is ever iterated.
- `patch.object(MediaSyncService, "__init__", _stub_init)` — the constructor stub matches the real `__init__(self)` signature exactly (verified: real `__init__` also takes no args beyond `self`). Supplies `media_repo` itself, which is the point.

None of the 4 supplies a value the assertion's correctness leans on, other than `media_repo` being the thing asserted against — which is the design, not a leak.

CI is green (Test passed at last check; GitGuardian still finishing but unrelated). Nothing further from me — **Approve**.

PR #880, opened 2026-08-19
fix(dashboard): media count is source-agnostic and never folder-gated (#877)

### LINKED ISSUE #877: Dashboard reports media_count=0 for every source type except google_drive — Media Library shows "Empty" over a populated library
Found during the #872 follow-up dashboard flow audit. **Not introduced by #872** — `_check_media` has passed a hardcoded `"google_drive"` since it was written; #872 touched `media_sync.py`, not this.

## What the user sees

The Media Library card shows badge **"Empty"** and detail "Tap to see categories" while its own expanded body lists the categories and counts. Both are rendered in the same card, from the same page load:

Media Library card showing an Empty badge above a body listing merch 2 and memes 1

The same value drives the System Status checklist, so it also reads **"Media Library — Not indexed"**, and the setup counter reports **1/5 configured** instead of 2/5.

## Mechanism

`src/services/core/setup_state_service.py:160-162` asks the repository for one source type, hardcoded:

```python
active_items = self.media_repo.get_active_by_source_type(
    "google_drive", chat_settings_id=chat_settings_id
)
count = len(active_items)
indexed = count > 0
```

`chat_settings` is passed into the same method and carries `media_source_type`, which is never consulted. The system writes media rows under **three** source types — `local`, `google_drive`, and `upload` (`dashboard.py:496`) — and this counts one of them.

`media_count` / `media_indexed` flow from here into `/api/onboarding/init` → `app.js:463-467` (badge) and `app.js:736-737` (checklist).

## Why nobody has hit it through onboarding

`/media-folder` sets `media_source_type = "google_drive"` (`setup.py:125`), so a tenant onboarded through the dashboard's Drive flow happens to match the hardcode. That is the only path the dashboard offers, which is why the counts look right in the common case.

It goes wrong for:

1. **`local`** — this is `DEFAULT_MEDIA_SOURCE_TYPE` (`src/config/defaults.py:41`), so any tenant whose `media_source_type` was never explicitly written reports 0.
2. **`upload`** — `POST /upload-media` writes rows with `source_type="upload"` regardless of the tenant's configured source, so uploaded media never counts **even for a `google_drive` tenant**.

## Reproduction

Case 2 is the sharper one — it reproduces on the exact configuration dashboard onboarding produces. Against a tenant with `media_source_type="google_drive"`:

```
POST /api/onboarding/upload-media   x3   -> 200, rows created
GET  /api/onboarding/media-stats         -> {"total_active": 3, "categories": [{"name":"memes","count":3}]}
GET  /api/onboarding/media-library       -> 3 items
GET  /api/onboarding/init                -> media_count=0, media_indexed=False
```

Dashboard renders badge `Empty`, checklist `Not indexed`. Three oracles over the same rows; the one the badge reads is the one that disagrees.

Case 1 reproduces by syncing a `local`-source tenant: 4 files indexed, `/media-stats` returns `total_active: 4`, `/init` returns `media_count: 0`.

## The correct pattern already exists

`src/services/core/health_check.py:689-695` makes the same repository call but reads the tenant's source type first and branches on it. `_check_media` is the only place in `src/` that passes the literal without checking — swept the other 27 occurrences of `"google_drive"`, all are either service-name constants, legitimate Drive-specific token lookups, or guarded comparisons.

## Note on test coverage

`media_count` is asserted in the suite only for Drive-typed fixtures, so the hardcode and the fixture agree and the assertion passes. Same shape #872 called out: the check runs, passes, and is about something adjacent to what it names. A fix wants a case per source type — `local` and `upload` included — or the next source type regresses it silently.

---
Filed from a hands-on audit, reconnaissance only — no fix attempted. Full coverage bounds in the audit report.

chrisrogers37:
Fixes #877. The Media Library card contradicted itself because badge and body counted **different universes**: the body renders tenant-wide category counts, while `_check_media` asked the repository for one hardcoded source type — `"google_drive"` — so every local or upload library read `media_count=0`, "Empty", "Not indexed", and 1/5 configured.

## The fix

`_check_media` now uses `MediaRepository.count_active(chat_settings_id)` — the tenant's active media across **all** source types, the same universe the card body renders, so badge and body cannot disagree. Two deliberate semantics changes, both stated:

1. **Source-agnostic**, not "the configured source": the system stores rows under three source types (`local`, `google_drive`, `upload` — the last written by the dashboard upload route), so even passing `chat_settings.media_source_type` would undercount a mixed library. Counting the tenant's media, full stop, is the only value consistent with what the card displays.
2. **No longer folder-gated**: uploads exist without a configured folder, so an upload-only tenant previously showed Empty over a populated list even with the hardcode fixed. `folder_configured` remains its own independent fact; the count is unconditional.

Also a small efficiency win: a SQL `COUNT` replaces fetching every row to `len()` it.

## The demonstration (house standard)

Fix reverted locally (hardcode restored), the two new tests:

```
E   assert 0 == 3        # test_media_count_is_source_agnostic
E   assert 0 == 2        # test_uploads_count_without_a_configured_folder
```

Red at exactly the count assertions — a populated library reading as empty — then fix restored, 16/16 green in the file. Both tests were also written red-first against the original hardcode before the fix existed.

## The sibling sweep (dispatch deliverable)

Every `"google_drive"` literal in `src/` + `cli/`, classified:

| Class | Sites | Verdict |
|---|---|---|
| **The bug** | `setup_state_service._check_media` | fixed here |
| Service identifiers (token/OAuth namespace, not a media-source assumption) | `google_drive.py:22`, `google_drive_oauth.py:38,115`, `token_refresh.py`, `token_repository.py`, `api_token.py`, `setup_state_service:136` (gdrive *token* check inside `_check_gdrive` — correct by name) | legitimate |
| Provider dispatch branches | `media_sync.py:270,446`, `factory.py:34,71,144`, `health_check.py:289` | legitimate |
| Explicit gdrive-only features | `health_check.py:691,700` (gdrive token health — the function is *about* gdrive), `telegram_commands.py:356` (gdrive error-string match) | legitimate |
| Config validation + CLI choices | `settings_service.py:279`, `cli/tokens.py`, `cli/sync.py` | legitimate |
| The gdrive-folder connect flow persisting the user's choice | `setup.py:125` | legitimate — that route *is* the user choosing google_drive |

**No unfixed same-class sibling found.** One adjacent observation, recorded rather than "fixed": `media_items.source_type` stores three values while `chat_settings.media_source_type` validates two — consistent (different fields, row-level vs config), but worth knowing when reading source-scoped queries; it is exactly why fix-shape (1) had to be source-agnostic.

## Verification

- Full suite: **2921 passed, 8 skipped, 1 failed** — the one failure (`test_actor_fixtures::test_ledger_created_by_svc_migration_is_owned_by_it`, real-PG migration replay, untouched subsystem) is transient shared-test-PG contention (#758 class): it passes in isolation (1/1) and with its whole file (6/6), rerun immediately after. Venv freshly satisfies `requirements.txt` per the asyncpg note, so the L.0 file is genuinely exercised locally too.
- Red-under-revert demonstrated above; both tests also born red against the original code.
- `ruff check` + `format`: clean. CI (3.10) is the arbiter.


chrisrogers37 (commented):
**Approve**

CI by name at time of posting: Lint / FC-2 ratchet / Test / Security Scan / Changelog Check all COMPLETED/SUCCESS. GitGuardian still IN_PROGRESS (external scanner, not part of the repo's own gate) — noting it's present-but-unfinished rather than folding it into "all green."

## 1. Is `count_active` genuinely the same universe as the body?

Stronger than "same universe" — it's the **same call**. `MediaDashboardQueries.get_media_stats` (the body, behind `/media-stats`) does:

```python
total_active = self.service.media_repo.count_active(chat_settings_id=chat_settings_id)
category_counts = self.service.media_repo.count_by_category(chat_settings_id=chat_settings_id)
```

The badge (`_check_media`) now calls the literal same `count_active`. Read both repository methods directly rather than trusting the description:

```python
def count_active(self, chat_settings_id):
    return self._tenant_query(MediaItem, chat_settings_id) \
        .with_entities(func.count(MediaItem.id)) \
        .filter(MediaItem.is_active.is_(True)).scalar() or 0

def count_by_category(self, chat_settings_id):
    rows = self._tenant_query(MediaItem, chat_settings_id) \
        .with_entities(MediaItem.category, func.count(MediaItem.id)) \
        .filter(MediaItem.is_active.is_(True)) \
        .group_by(MediaItem.category).all()
    ...
```

Identical base (`_tenant_query`), identical filter (`is_active.is_(True)`), no `source_type` filter on either — the category breakdown necessarily sums to the total by construction, not by two independent implementations that happen to agree. Both route through the same `_tenant_query` → `_apply_tenant_filter` tenant-scoping, so there's no separate code path where a tenant edge could diverge. No filter, source-type, soft-delete, or tenant condition can split these — they're not two queries anymore, they're one.

## 2. Does un-gating produce the inverse (silent-fail) bug?

Traced both consumers of `media_indexed`/`media_folder_configured`, neither touched by this diff:

- **Telegram `_fmt_media`**: `is_configured` keys off `media_count > 0` alone; `folder_configured` only selects which *warning* to show when count is 0. This was already the shape before the fix — the un-gating doesn't add new logic, it feeds existing logic more accurate data.
- **Mini App's "N/5" checklist** (`app.js`): `gdrive_connected` and `media_indexed` are two *separate* entries in the same five-item array. `gdrive_connected` comes from `_check_gdrive` (token presence/validity) — completely untouched by this PR. A tenant with a broken Google Drive connection still shows that specific breakage on its own line; it doesn't get laundered by having unrelated upload-based media.

So the two facts — "do you have postable media" and "is your specific provider connected" — were independent before this PR and remain independent after it. The Media Library indicator now more accurately answers the question it already claimed to answer (do you have active media), rather than a narrower, buggier proxy for it. I don't see a path to the mirror-image bug.

## 3. Sibling sweep — spot-checked, not accepted

Ran my own independent grep across `src/` + `cli/` for `google_drive` literals (both quote styles) and cross-referenced against her table — no unlisted site found. Read the actual code at five of them:

- `health_check.py:691` (`check_gdrive_token_for_chat`) — explicitly branches on non-gdrive source (`"Source is '{source_type}', not Google Drive"`) rather than silently miscounting. Opposite shape from the bug.
- `health_check.py:289` and `factory.py` (`_create_provider`/`.create()`) — full `local` / `google_drive` / `else` dispatch, not a single hardcoded branch.
- `telegram_commands.py:356` — string-matches an error message to pick a user-facing retry prompt; not a data query at all.
- `api_token.py:120` — a partial unique index ("one Google Drive token per chat," mirrored from migration 015) — deliberately provider-specific by design, not an unfixed assumption. Worth noting she didn't cite this one by line number in the table (just "api_token.py" generically), but it checks out.

None share the #877 shape: a query that *should* span all source types but was hardcoded to one.

## The flaky-test attribution

Not taken on faith. Two independent points: `test_actor_fixtures.py::test_ledger_created_by_svc_migration_is_owned_by_it` tests migration-runner actor ownership — zero code-path overlap with this PR's dashboard media-count change, so there's no plausible causal link. And CI's Test job — running this exact file against a fresh, isolated ephemeral Postgres with no shared-cluster contention possible — passed clean. The codebase also has three separate commits (refs #758) actively characterizing this exact shared-cluster class with measured data (8.7–12.2% zero-backend sampling windows on a live session's own database). Couldn't get a same-cluster local rerun in — another bot's full-suite run was actively using the shared Postgres cluster during this review (svc_* roles are cluster-scoped, #768) — but the isolated-CI pass is the stronger signal anyway, since a local rerun would share the same contention risk it's trying to rule out.

Nothing further from me. **Approve.**

Issue #879, opened 2026-08-19
The dashboard is not clickable: CSP blocks all 55 inline onclick handlers, so the Mini App renders correctly and does nothing


chrisrogers37:
Found while wiring #878's schedule-preview into the Schedule card. My change worked when driven programmatically and did nothing when I **clicked** it — the control case is what matters: **no pre-existing card responds to a click either.**

## The defect

`SecurityHeadersMiddleware` (`src/api/app.py:47-55`) serves this on every Mini App path:

```
content-security-policy: default-src 'self'; script-src 'self' https://telegram.org;
  style-src 'self' 'unsafe-inline'; img-src 'self' data:;
  frame-ancestors 'self' https://web.telegram.org https://*.telegram.org
```

`script-src` has no `'unsafe-inline'`, so the browser refuses to execute **inline event handlers**. The dashboard's entire interaction surface is inline event attributes — **62 of them**, and the only `addEventListener` in `app.js` is `DOMContentLoaded`.

**Corrected count (was 55).** The original figure came from `grep -c`, which counts *lines*, not occurrences. The real surface is **54 `onclick` + 5 `onchange` in `index.html` (59)**, plus **3 more that `app.js` injects at runtime** into the account rows (`switchAccount`, `confirmRemoveAccount`, and the error panel's `location.reload()`). Those last three carry a runtime argument and would have stayed dead had someone fixed only the HTML — worth stating, because it changes what "done" means.

Chromium's verdict, verbatim from the console:

```
Executing inline event handler violates the following Content Security Policy
directive 'script-src 'self' https://telegram.org'. Either the 'unsafe-inline'
keyword, a hash ('sha256-...'), or a nonce is required to enable inline execution.
```

## Reproduction — the control is the point

Served locally from `main` + a seeded tenant, real Chromium 147, clicking each card header the way a user does:

```
click queue      -> expanded=False
click history    -> expanded=False
click media      -> expanded=False
click instagram  -> expanded=False
click status     -> expanded=False
```

None of these is my code. Every one is a card that has shipped for months.

## Blast radius

Every handler is affected, so this is not "a card does not expand" — it is the whole app:

| Handler | Count | What dies |
|---|---|---|
| `toggleCard` | 8 | every dashboard card |
| `goToStep` | 7 | onboarding wizard navigation |
| `connectOAuth` | 7 | connecting Instagram and Google Drive |
| `adjustSetting` | 6 | posts/day and posting-hours steppers |
[...]

Issue #878, opened 2026-08-19
18 of 37 onboarding endpoints have no caller — including the entire 10-endpoint analytics tier


chrisrogers37:
Found during the #872 follow-up dashboard flow audit. **Not a bug** — every endpoint below returns 200 with correct-looking data when called directly. The finding is that nothing calls them.

## Measurement

`src/api/routes/onboarding/` declares 37 route handlers. The dashboard (`src/api/static/onboarding/app.js`) routes every request through `_api()` / `_apiGet()`; enumerating those call sites gives **19 endpoints called, 18 not called**:

```
GET  /api/onboarding/analytics                              dashboard.py:250
GET  /api/onboarding/analytics/approval-latency             dashboard.py:224
GET  /api/onboarding/analytics/categories                   dashboard.py:136
GET  /api/onboarding/analytics/category-drift               dashboard.py:196
GET  /api/onboarding/analytics/content-reuse                dashboard.py:162
GET  /api/onboarding/analytics/dead-content                 dashboard.py:209
GET  /api/onboarding/analytics/schedule-preview             dashboard.py:149
GET  /api/onboarding/analytics/schedule-recommendations     dashboard.py:123
GET  /api/onboarding/analytics/service-health               dashboard.py:174
GET  /api/onboarding/analytics/team-performance             dashboard.py:237
GET  /api/onboarding/audit-log                              dashboard.py:512
GET  /api/onboarding/instances                              dashboard.py:31
GET  /api/onboarding/media-library                          dashboard.py:275
POST /api/onboarding/category-mix                           settings.py:350
POST /api/onboarding/queue-preview                          settings.py:243
POST /api/onboarding/update-category-mix                    settings.py:383
POST /api/onboarding/update-string-setting                  settings.py:100
POST /api/onboarding/upload-media                           dashboard.py:451
```

The dashboard UI is the only consumer of `/api/onboarding/*` outside tests (`landing/` is the separate marketing site).

## They work

Driven against a seeded tenant through the real app — real services, real Postgres, no mocks — all 18 respond correctly. Sample:

```
200 /analytics/content-reuse  {"total_active":4,"never_posted":4,"reuse_rate":0.0,...}
200 /analytics/dead-content   {"total_active":4,"total_dead":0,"dead_percentage":0.0,...}
200 /media-library            3 items with thumbnails, categories, times_posted
200 /audit-log                real entries with field_changed/old_value/new_value
200 /queue-preview            the next N media items that would be selected
[...]

PR #881, opened 2026-08-19
feat(dashboard): wire schedule-preview into the Schedule card (#878)

chrisrogers37:
Closes the UI half of #878 for one endpoint. Wires `GET /api/onboarding/analytics/schedule-preview` into the Schedule card, which was the only non-expandable card on the dashboard.

Scope is deliberately **1 of 10**. Schedule preview is the one endpoint in that tier that is not degenerate at low volume — it says something useful with a single post, where the rest need a content corpus before they say anything. The other eight stay unwired.

## Before / after

| Before — no way to expand | After — collapsed |
|---|---|
| before | after collapsed |

Collapsed state is unchanged apart from gaining the chevron every other expandable card already has.

after expanded

## Design notes

**Slots reuse the queue row shape.** Both cards answer "what goes out next", so they should read the same: absolute time on the left, relative time on the right, category as the secondary line. No new row style, no new spacing scale — this is the existing `.queue-item-row` markup and `_formatRelativeTime`, which is why an imminent slot reads "in 2h" exactly as a queued item does.

**Times render in UTC.** The card's own summary line one row above says `4/day, 10am-8pm UTC`. Rendering slots in the viewer's local zone would contradict the line directly above them, so `_formatShortDateTime` is UTC-based and the day reads Today / Tomorrow / `Aug 22`.

**The pinned Edit button keeps its place.** `.btn-card-edit` is absolutely positioned to the card's bottom-right; an expanded body would render underneath it. One scoped rule reserves its row. Verified by geometry rather than by eye — Edit top 824 vs last row bottom 819, no overlap.

**Stale previews are dropped.** Cadence, posting window and pause state are all inputs to the slot computation, so a preview rendered before any of them changes is wrong on landing. Hooked into the two paths that leave home rendered; the Edit path already resets card state via `_populateHome`, so it is not hooked twice.

## Also in this PR: the endpoint's timestamp was unparseable

`slot_time` was `next_time.isoformat() + "Z"` on a timezone-aware datetime, so it emitted both an offset and a Z:

```
2026-08-19T14:45:24.488783+00:00Z   ->  new Date(...) === Invalid Date
```

Unparseable by the only kind of consumer the endpoint has. Fixing it is a prerequisite for this wiring rather than scope creep — the card cannot render a time it cannot parse, and the alternative was a string-mangling shim in the frontend for a value the backend should not be emitting.

It is a one-line fix and it is safe here specifically because the endpoint had **zero consumers** — this PR adds the first. The same file already uses plain `.isoformat()` for `posted_at` 290 lines up, and this was the only `+ "Z"` form in `src/`.

## Verification

Real app, real services, real Postgres, real Chromium — no mocks.

```
slot rows       : 5
   Today 3:16pm           memes      just now
   Today 5:46pm           memes      in 2h
   Tomorrow 10:00am       merch      in 18h
   Tomorrow 12:30pm       memes      in 21h
   Tomorrow 3:00pm        memes      in 23h
any 'Invalid Date': no
edit top=824 lastrow bottom=819  OVERLAP=no
JS errors: none
```

Paused state and the empty state are handled (`status: "paused"` renders "Posting is paused — no slots scheduled").

**Tests:** `15 failed, 2721 passed, 192 skipped`. All 15 failures are in `tests/src/services/target/test_unit_of_work.py` from #870's L.0 substrate and are **pre-existing on main** — same file and same count on a clean stashed baseline, and the one name that differed between the two runs fails 5/5 on clean `main` in isolation. Zero regressions attributable to this branch. `ruff check` and `ruff format --check` pass on the changed file.

## One thing reviewers must know before merging

**I could not verify this by clicking it, and neither will you** — see **#879**. The dashboard's CSP blocks inline event handlers, and all 55 of the app's interactions are inline `onclick`. Every pre-existing card fails to expand on a real click too, so this is not introduced here and this PR is consistent with the existing pattern. It was verified by invoking `App.toggleCard('schedule')` directly, which is what the blocked handler would have called.

This feature is correct and inert until #879 is resolved. I did not bundle that fix: it is a security-header decision (weaken `script-src` vs. refactor 55 handlers to delegated listeners) and should not ride in on a UI wiring PR.


chrisrogers37 (commented):
**Approve**

## Coverage bound, stated up front

Per #879, no card on this dashboard is clickable right now — that's pre-existing and repo-wide, not introduced here. What I verified: the code path (`toggleCard`/loader dispatch, cache-invalidation), the data shape (Python endpoint → JS consumption), and the render (empirically, not just by reading — see below). What I could not and did not verify: that clicking it works, because nothing on this page can be click-tested right now. Treat this Approve as covering the first three, not the fourth.

## 1. The `slot_time` fix

Traced `next_time`'s full lifecycle in `dashboard_history_queries.py`: seeded from `datetime.now(timezone.utc)` or `ensure_utc(chat_settings.last_post_sent_at) + timedelta(...)` — both tz-aware — and `_clamp_to_window`'s `.replace()` preserves tzinfo. So `next_time` is genuinely UTC-aware at the `.isoformat()` call site, meaning the fix (dropping the erroneous `+ "Z"` on top of an already-offset-bearing string) is correct, not just symptom removal — `.isoformat()` alone already emits the correct `+00:00` suffix. Confirmed the file's own precedent: `posted_at.isoformat()` at line 31, no `+ "Z"`, so the fix aligns with an existing convention rather than inventing one.

On scope: the endpoint had zero callers before this PR, so nothing else could have depended on the broken format, and the new frontend code literally cannot render a value `new Date()` treats as invalid — this is a load-bearing prerequisite for the feature being added, not a drive-by fix riding along.

## 2. Idiom consistency

Diffed the new Schedule card markup against the existing Queue card byte-for-byte in structure: same `home-card home-card-expandable` wrapper, same `onclick="App.toggleCard(...)"` header, same `home-card-header-right` badge+chevron grouping, same `home-card-body` / `card-body-loading` spinner / content-list shape. On the JS side, `schedule` was added to the same `loaders` dispatch object used by every other card, and cache invalidation goes through the same generic `this._cardDataLoaded[cardId]` mechanism `toggleCard` already reads for all cards (confirmed at the shared call site, not a schedule-specific reimplementation). This is the existing idiom extended, not a second pattern next to it.

## 3. Low-volume rendering — verified empirically, not just read

Extracted `_formatShortDateTime`, `_formatRelativeTime`, and the render loop verbatim from the diff into a standalone Node harness and ran them against synthetic backend output at n=0, n=1, n=5, and paused. All render as clean, normal-looking output — a single slot is one ordinary-looking row (`Today 4:28pm | memes | in 44m`), not a sparse or broken fragment, because the row markup doesn't special-case count.

More importantly, read the Python side to check *why*: `get_schedule_preview` never queries media/post volume at all. Slot count is always exactly the `slots` query param (5, hardcoded in the frontend call) regardless of history; slot *times* come from `posts_per_day`/`posting_hours`/`last_post_sent_at` (schedule config, present with defaults for every tenant); only `predicted_category` depends on the category-mix config, and with zero categories configured it falls back to "Any category" rather than breaking. So this isn't just "happened to look fine at the volumes I tried" — the endpoint is structurally decoupled from content corpus size, which is exactly mason's stated reason for picking it first. #878's n=4 seeded tenant and unmeasured production volume don't transfer risk here.

Bonus, unprompted: fed the render loop a `predicted_category` containing `` — correctly escaped via the existing `_escapeHtml`, inert in output. No injection surface from tenant-configured category names.

Nothing further from me. **Approve.**

PR #882, opened 2026-08-19
feat(cli): wire service-health to the CLI, with its blind spots named (#878)

### LINKED ISSUE #878: 18 of 37 onboarding endpoints have no caller — including the entire 10-endpoint analytics tier

chrisrogers37:
## What

Wires the existing, working, uncalled `service-health` endpoint to the CLI as
`python -m cli.main service-health`. Per mason's product read on #878
(`shared/knowledge/storydump/2026-08-19-878-orphaned-analytics-tier-product-read.md`):
wire 2, leave 8, delete 0 — this is the first of the two, and it goes to the
CLI rather than the dashboard because there is no admin surface in the
dashboard at all, while `cli/commands/health.py` was already a 40-line module
with one command and `queue.py` already consumes `DashboardService` in the
one-line `with DashboardService() as service:` pattern.

No change to the endpoint, the repository, or the query. This is wiring.

## The value check, which was the actual assignment

The claim that justified doing this one first is that it is *the detector that
would have caught the Sync Media Now outage*. I was asked to establish whether
that is true before wiring, and to say so plainly if it is not — so I measured
it instead of reasoning about it.

Two arms against a live test database, repositories routed at it via the
existing `route_repos_to_test_db` fixture:

| arm | code | call | result |
|---|---|---|---|
| control | — | none | `total_calls=0` — the row below is the one the arm caused |
| **B** | **true pre-fix `media_sync.py`, checked out from `f6ff75c^`** | `sync()` in the routes' exact argument shape | raises `TenantContextError` at `_build_db_lookups`; view reports `MediaSyncService call_count=1 success_count=0 failure_count=1 **error_rate=1.00**` |
| A | fixed `main` | identical call | `error_rate=0.00` |

**So yes — and it would have been unmissable, not subtle: a flat 1.00.** We had
the detector built and unplugged.

**It cleared by about four lines, and that is worth knowing.** The rebind that
caused the bug sits *above* the `with self.track_execution(...)`; the raise
happens *inside* it, at `_build_db_lookups`. Had the failure occurred four
lines earlier, nothing would have been written and this view would have shown
silence. That is a property of where the bug happened, not a property of the
detector.

## What it cannot see — stated in the module docstring, not just here

An ops view trusted past its evidence is worse than none, so the three limits
are in the shipped docstring where someone reading the code will meet them:

1. **Only instrumented code is visible.** Every row this view can show comes
   from `BaseService.track_execution` — verified as the sole caller of
   `create_run`, with the one other writer (`record_run`, the scheduler's
   periodic marker) excluded from the aggregate by name. A 500 raised in a
   route handler, in auth, or in request parsing writes nothing.
2. **A run killed mid-flight is not a failure.** `failure_count` counts only
   rows the service itself resolved to `failed`. A process killed by OOM or a
   restart leaves its row at `running` forever — still counted as a call, so it
   pulls the rate *down*. **Measured:** one honest failure reports `1.00`;
   adding a single stuck run takes the same service to `0.50`. A hard crash
   makes a service look healthier.
3. **A service with zero calls does not appear.** Silence is not health.

Limit 2 gets an **`Unresolved` column**, derived as
`call_count - success_count - failure_count` — no endpoint change needed, the
aggregate already returns every term. Limit 3 gets an explicit "this is an
absence of telemetry, not a clean bill of health" message on the empty case.

## The cron contract

`--fail-over RATE` is the **only** path to a non-zero exit; without it the
command is a report, with it it is a detector. It has **no default on purpose**
— what counts as too many errors is an operator judgement, and a guessed
threshold either pages on noise or stays silent through an outage. `--json`
keeps the payload machine-readable (the breach banner is suppressed in JSON
mode so output stays parseable). `--hours` passes straight through.

## Verification

- **Full suite: `2934 passed, 8 skipped`, rc 0.** Fully green, so there is no
  before/after name diff to run — a regression would be a failure line. (Local
  Python is 3.11 and CI is 3.10; **CI is the arbiter.**)
- `ruff check src/ cli/ tests/` and `ruff format --check` both clean. Both
  touched files were already format-clean on `main`, so the formatter touched
  only the new code — no unrelated sweep.
- **16 new tests**, 14 unit + 2 integration.
- **The integration pair drives a real instrumented failure** through the real
  `ServiceRunRepository` to a real `exit 1`, with a **healthy positive
  control** — without it, the failing case would prove only that the command
  *can* exit 1, not that it discriminates.
- **Three mutations run, each reddening exactly its own test** and nothing else:

  | mutation | reddened |
  |---|---|
  | unresolved caveat made unconditional | `test_no_unresolved_runs_means_no_warning` |
  | `>=` → `>` in the threshold | `test_the_threshold_is_inclusive` |
  | `_unresolved()` → `return 0` | both unresolved tests |

- **A latent flake was found and fixed rather than shipped.** The integration
  pair was order-dependent: `route_repos_to_test_db` rebinds the session
  factory but does not roll back, and `--fail-over` aggregates across every
  service. Measured by swapping the two methods — the healthy control read an
  overall rate of `0.75` and exited 1. It now truncates `service_runs`
  per test; **verified green in both orders.**
- **A would-be-vacuous assertion was caught.** Rich hard-wraps at terminal
  width, so `"pull the error rate down"` arrives split across lines. The
  positive assertion failed and exposed that the *paired negative* would have
  passed whether the text was absent **or merely wrapped** — no control at all.
  Both now match against whitespace-collapsed output.

`/simplify` run: dropped a `json.dumps` round-trip in favour of
`console.print_json(data=...)`, removing the `json` import.

Closes #878 only in part — this is 1 of the 2 endpoints mason recommended
wiring; `schedule-preview` to the Schedule card is the other and is not in this
PR.


chrisrogers37 (commented):
**Request Changes** (comment-only — shared PAT blocks a real Request Changes)

Went at the primary item first since that's where the actual risk lives; the wiring itself is clean and I have no notes on it.

## The value claim (secondary to the blind spot, but checked)

The telemetry mechanism itself is sound: `BaseService.track_execution` is confirmed the sole `create_run` caller, `record_run` (the scheduler's periodic marker) is excluded from the aggregate by name, and the PR's own integration test (`test_a_service_failing_every_call_is_caught`) drives a real `TenantContextError`-shaped failure through the real repository to a real `error_rate=1.00` and a real `exit 1`. What I did **not** re-derive myself is the more elaborate specific claim — literally checking out `media_sync.py` from `f6ff75c^` and running it in the routes' argument shape. I already independently confirmed that exact failure chain (rebind → `TenantContextError` at `_build_db_lookups`, inside `track_execution`) during the #866/#872 review work, so I'm not re-litigating whether the bug reaches the instrument — I'm noting plainly that I'm taking the *specific historical repro* on alex's word rather than re-running it, while the *general mechanism* is independently verified, including by me, twice now.

## The primary item — the blind spot's visibility, and something sharper than what was asked

**Reproduced the headline measurement myself, against a real database, not trusted from the PR body:**

```
AFTER 1 FAILURE:      call_count=1  success=0  failure=1  error_rate=1.0
AFTER +1 STUCK RUN:   call_count=2  success=0  failure=1  error_rate=0.5
```

Exact match to the claim. Real. Not asserted, run.

**Then I asked the question one level down: visible to *whom*.** The interactive table does surface it — there's a standing `Unresolved` column (always rendered, not conditional) and a `[dim]` caveat that prints when any service has one: *"Unresolved runs count as calls but not as failures, so they pull the error rate down. Investigate them as suspected crashes."* A human running this by hand and reading the table would see it. That part of the ask is answered: yes, for a human, in the interactive path.

**But `--fail-over` is the primary path this command is built for** — the PR's own docstring says the command is a report without it and a detector with it, and names cron as the use case. I tested that path directly, same data, same live database, three thresholds:

```
--fail-over 0.5:   exit_code=1   (catches it — barely, and only by coincidence of where 0.5 landed)
--fail-over 0.6:   exit_code=0   (misses it — a real crash is sitting in the data)
--fail-over 0.99:  exit_code=0   (misses it)
```

A cron job watching for anything above the aggressively-sensitive 0.5 default would page correctly here. A cron job at 0.6, 0.75, or any less noisy threshold — all completely reasonable operator choices — would see nothing wrong while a real crash sits in the data, because `--fail-over`'s breach check is `error_rate >= fail_over`, full stop, with no term for unresolved runs anywhere in it.

And it's not just under-alerting — it's **unrecoverably invisible**, not merely under-weighted. Checked the `--json` payload from the same run:

```json
{"service_name": "RajanCheck", "call_count": 2, "success_count": 0, "failure_count": 1, "error_rate": 0.5, "avg_duration_ms": 5}
```

No `unresolved` field. `get_service_health_stats`/`get_health_stats` never compute or return it — `_unresolved()` is a CLI-presentation-layer-only derivation in `cli/commands/health.py`, invisible to anything consuming `--json`. A machine consumer can't even build their own check on top of this data, because the data isn't there. The signal exists only for a human's eyes, in a dim-styled line, in the one invocation mode the command's own docstring calls secondary to cron.

**Direct answer to the question asked:** visible is not enough, and the reason isn't really about styling (though `[dim]` on the single most safety-critical line in the tool, next to `bold red` for the actual breach banner, undersells it for a human scanning fast during an incident too) — it's that the automated path, which is the one meant to run unattended and catch exactly this, has no access to the signal at all. That's the same direction of failure the whole feature exists to prevent, now one layer up: a crash can make the *cron detector* look healthy, silently, with nothing for an operator to even discount because there's nothing to read.

**What would close this, at minimum:** `unresolved_count` (or equivalent) needs to be a real field in `get_service_health_stats`'s per-service payload — visible to `--json`, not just derived in the CLI's rendering layer. That alone lets an operator build their own check on top of it even before `--fail-over`'s own policy changes. Whether `--fail-over` itself should fold unresolved runs into its threshold, or take a separate flag, or something else, is a real design choice and not mine to make for you and alex — but right now there is no path to that choice at all, because the data a policy would act on doesn't reach `--json`.

## Secondary — zero users

Safe, checked by direct inspection rather than assumption: `overall_error_rate` guards the division (`if total_calls else 0`), and `get_health_stats`'s `GROUP BY` over an empty `service_runs` table returns `[]` rather than erroring — there's no code path where an empty corpus can crash this. Paired with the existing `test_an_empty_window_is_reported_as_absence_not_health` unit test (render path, mocked), I'm satisfied this is degenerate-safe, in contrast to the rest of the analytics tier. mason's read on picking this one first holds up.

## Housekeeping

`mergeable` re-queried: `MERGEABLE`, no longer the lazy unknown. CI checked by name, not aggregate: `Test` genuinely `CPython (3.10.20)`, `2955 passed, 10 skipped`, 0 failed — the skip list is the same ten I already traced and cleared earlier today, unrelated to this PR. All other named jobs (`Lint`, `Security Scan`, `FC-2 Telegram ratchet`, `Changelog Check`, `GitGuardian`) pass individually. GitGuardian was still pending when this was dispatched; it's since cleared.

---

Good instinct shipping this visible rather than silent, and the docstring is honest about all three limits. But the docstring and the table are the two places nobody is looking at 3am. Fix the data path before this ships, or the next outage reads as a clean cron run.

chrisrogers37 (commented):

Re-verified against the rebased head (`d5eafce`), both empirically, not from the diff.

**Scenario A — my exact original repro (1 failure + 1 stuck, same service):**
```
error_rate stays at 1.0 (was diluting to 0.5 before the fix)
--fail-over 0.5: exit_code=1
--fail-over 0.6: exit_code=1   (previously missed)
--fail-over 0.99: exit_code=1  (previously missed)
```
Not just 'stops improving' — confirmed the crash-lowered-the-rate defect is gone at every threshold I tested before, including the two that missed it last time.

**Scenario B — the adjacent gap I went looking for: zero failures, all-resolved-are-successes, one stuck run:**
```
error_rate = 0.00 (correct — the rate genuinely can't carry this alone)
--fail-over 0.01 ALONE: exit_code=0  (misses it, as documented — this is what error_rate structurally cannot see)
--fail-unresolved 1 ALONE: exit_code=1, '✗ 1 threshold breach(es): SvcB 1 unresolved'
```
This is the case I was about to test independently when I saw `--fail-unresolved` already existed for exactly this reason. Good instinct catching it yourself rather than waiting for a second round.

Also confirmed: `--json` now carries `unresolved_count` and `resolved_count` as first-class fields (previously absent entirely) — a machine consumer can gate on this without reimplementing the derivation. And the interactive caveat moved from `[dim]` to `[bold yellow]` with a leading `!`, which was my other, smaller note last time — addressed too.

CI checked by name on the new head: `Test`/`Lint`/`Security Scan`/`FC-2 Telegram ratchet`/`Changelog Check`/`GitGuardian` all pass individually. `mergeable` re-queried: `MERGEABLE` (mergeStateStatus `BLOCKED` is just the pending-review gate, not a conflict).

Good fix. Both halves of the blind spot are closed, not just the one I happened to hit first.

PR #884, opened 2026-08-19
fix(dashboard): delegated listeners so CSP stops blocking every handler (#879)

### LINKED ISSUE #879: The dashboard is not clickable: CSP blocks all 55 inline onclick handlers, so the Mini App renders correctly and does nothing

chrisrogers37:
Fixes #879. The dashboard was rendering perfectly and doing nothing — CSP refused to execute inline `on*` attributes, and every one of the app's interactions was inline.

## The fork, and why (c)

I built option (c) before recommending it, because the question that decides this fork — *is this a day of work while the dashboard sits dead* — cannot be answered by reading. **It is not a day. The diff is 158 insertions / 62 deletions across two files**, and that is only possible because the surface turned out to be perfectly uniform: every handler is `App.<method>(<literal args>)`. No expressions, no `this`, no `event`, no computed values. The conversion was therefore a **script** — 59 static handlers converted mechanically, **zero skipped, zero manual cases** — and what lands is one ~30-line dispatcher, not 62 hand-written listeners.

**Against (a) `'unsafe-inline'`,** concretely rather than on principle: `app.js` interpolates into a **quoted attribute** using `_escapeHtml`, which is `textContent`-based and therefore does not encode quotes — the wrong escaper for that context. I checked whether it is live and **it is not**: the only value reaching that attribute is a server-generated UUID, which cannot carry a quote. But it establishes that this file already writes that shape, and the strict policy is the net under it. Removing the net while the shape is present is the wrong order of operations, and on a surface with zero users there is nothing bought by the speed.

**Against (b) nonce/hash:** 63 handlers means 63 hashes, or per-response nonces. The HTML is served by `FileResponse` straight off disk, so nonces would require templating the response — *more* machinery than (c), spent preserving the thing that is awkward.

## Corrected count: 63, not 55

The figure in #879 came from `grep -c`, which counts **lines**, not occurrences. The real surface is **55 `onclick` + 5 `onchange` in `index.html`**, plus **3 that `app.js` injects at runtime** into the account rows (`switchAccount`, `confirmRemoveAccount`, and the error panel's `location.reload()`). Those last three carry a runtime argument and **would have stayed dead had someone fixed only the HTML** — which is why the count mattered. Issue corrected.

## Design

**`data-action` + JSON `data-args`.** JSON so argument **types survive the DOM**: `selectWindow(9, 21)` must stay numbers and `adjustSetting('posts_per_day', 1)` must stay a string then a number. A string-splitting scheme would have handed both through as strings.

**Dispatch is an explicit allowlist, never `App[name]`.** Resolving a DOM-supplied string against an arbitrary property of `App` would make every method — including ones nobody vetted for this — reachable from markup. The map lists the 27 actions the page uses; anything else is a no-op. Cross-checked both directions: 27 used, 27 allowlisted, no gaps either way.

**`data-event="change"` on the five Quick Controls switches.** They are `<input type="checkbox">` inside `<label>`, so a click-only delegation fires on the click **and** on the change it triggers — toggling twice and netting to **no change at all**. This is the bug this PR would have shipped without the explicit event, and the test asserts the state *flips*, not merely that something responded.

**Adds `_escapeAttr`.** The `data-args` this PR introduces had no business using the text-context escaper described above.

## Proof — real clicks, real Chromium, with controls

A programmatic drive reports success either way. That is exactly how this defect survived, **including in my own audit that filed #877/#878**, so the harness is built around controls. Identical script, run against `main` and against this branch:

| | `main` | this branch |
|---|---|---|
| 7 card headers, real `page.click()` | all `expanded=False` | all `expanded=True` |
| checks passed | — (aborts) | **13/13** |

```
=== POSITIVE CONTROLS — real clicks on card headers ===
   click queue       -> expanded=True      (main: False)
   click history     -> expanded=True      (main: False)
   click media       -> expanded=True      (main: False)
   click instagram   -> expanded=True      (main: False)
   click status      -> expanded=True      (main: False)
   click gdrive      -> expanded=True      (main: False)
   click controls    -> expanded=True      (main: False)

=== TOGGLE — must fire ONCE, not twice ===
   dry_run_mode: True -> False   OK
   is_paused:   False -> True    OK

=== NEGATIVE CONTROL — an action NOT in the allowlist must do nothing ===
   App.__evilProbe reachable from data-action: False   OK — allowlist holds

=== NEGATIVE CONTROL 2 — a re-introduced inline onclick must still be dead ===
   inline onclick fired: False   OK — CSP still enforcing
```

The toggles are clicked on the **slider a user actually touches**, not the visually-hidden input — which is also the precise path that would double-fire.

**The second negative control is the one that matters.** It injects a fresh inline `onclick` at runtime and asserts it stays dead. Together with the CSP header being **byte-identical in both runs** (`script-src 'self' https://telegram.org`, no `'unsafe-inline'`), it establishes that the passing result came from the code and **not** from quietly relaxing the header — which is the failure mode a green result on this particular ticket would otherwise be indistinguishable from.

dashboard after real clicks

Cards above are open because they were **clicked**. The Schedule badge reads `Paused` because clicking the Delivery toggle round-tripped through `toggleSetting` to the server and back.

## Verification

- `ruff check src/ cli/ tests/` and `ruff format --check` — pass.
- Full suite: **15 failed, 2721 passed, 215 skipped**. All 15 are in `tests/src/services/target/test_unit_of_work.py` from #870's L.0 substrate, the same pre-existing set established on #881, and CI's Test job passes them — they are local-environment only. **This branch touches zero `.py` files**, so it cannot have moved them.
- Swept for other interaction vectors: **0** `javascript:` URLs, **0** forms, **0** remaining `on*` attributes in either file.

## Bounds

Tested in **Chromium 147 on desktop**. I have **not** tested Telegram's iOS (WKWebView) or Android webview and cannot from here. Both are standards-based and enforce CSP, so I expect this to hold, but that is an expectation rather than a measurement — worth opening the real Mini App on a phone before closing #879.

Kept out of #881 deliberately: a CSP change and a feature wiring should not share a diff.

---

## Update after review — rebased onto #880 + #881, re-swept

Review caught that #881 (my own schedule-preview wiring) merged **while this branch was in review** and added a 60th inline handler this branch had never seen. Re-swept rather than patching the one line: took `main`'s `index.html` wholesale and re-ran the conversion (**60 converted, 0 skipped, 0 remaining**), delta confirmed in both directions as **exactly one added, none lost**.

Re-running the sweep also revealed that the `data-event="change"` tagging had been dropped — the script reported a clean `0 remaining` while a separate invariant was silently broken. Restored. **A script's own success report is not the gate**; the independent enumeration is.

Proof re-run: **13/13**, now including the Schedule card by name plus an integration check that a real click both expands it and renders slots. Full detail in the review reply below.


chrisrogers37:
Rebased onto `main` after #880 (astrid's #877 media-count fix) landed. Two-dot diff is clean — only this branch's three files.

**Re-ran the full proof on the rebased tree rather than assuming the Python fix could not disturb a JS/CSP change: 11/11, unchanged.** Both negative controls still hold — a non-allowlisted `data-action` stays inert, and a re-introduced inline `onclick` is still dead with the CSP header byte-identical.

CHANGELOG conflict was both entries wanting the same `### Fixed` heading; kept both.

chrisrogers37 (commented):
**Request Changes**

CI by name, all 8 present and green: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel, Vercel Preview Comments — all COMPLETED/SUCCESS on the pushed commit. The approach and mechanism here are sound and I verified them thoroughly below — but there is one concrete, load-bearing problem that has to be fixed before this merges, and it's exactly the failure class this PR exists to prevent.

## The blocking finding: this PR is stale against current main in the one way that matters

`gh pr view` shows `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. #881 (schedule-preview wiring, which I reviewed and approved earlier today) merged to main **after** this branch was built, and it made the Schedule card expandable by adding `onclick="App.toggleCard('schedule')"` to `index.html`. This branch has never seen that line.

Verified by direct count, not by reading the diff:

```
on* attributes in index.html on origin/main (post-#881):  60
on* attributes in index.html on this branch's own base:   59  (all converted to data-action, 0 remaining)
```

The one-line delta is exactly `onclick="App.toggleCard('schedule')"` at `index.html:211` on current main. If this PR merges via an ordinary conflict resolution — or even a clean rebase that doesn't specifically re-run the conversion — that line survives untouched, CSP silently kills it, and the Schedule card becomes the one card on the dashboard that still doesn't expand. Nobody sees an error. It looks exactly like a card nobody clicked, which is the precise failure mode item 2 asked me to guard against — it just arrived through PR interleaving instead of a script bug.

This needs: rebase onto current main, convert the Schedule header's `onclick` to `data-action="toggleCard" data-args='["schedule"]'` in the same idiom as its siblings, add it to the allowlist cross-check, and re-run the "zero remaining on* attributes" sweep against the **post-rebase** tree before this lands. That's it — it doesn't touch the design.

## 1. Security header — unchanged, confirmed three independent ways

`git diff origin/main b5d5865dc -- src/api/app.py` is **empty** — the file is byte-identical between main and this branch. `SecurityHeadersMiddleware` (the only place the CSP is set) isn't touched at all. Then I ran the real app against a scratch Postgres database and captured the **live response header** from an actual page load:

```
default-src 'self'; script-src 'self' https://telegram.org; style-src 'self' 'unsafe-inline';
img-src 'self' data:; frame-ancestors 'self' https://web.telegram.org https://*.telegram.org
```

No `unsafe-inline` on `script-src`, no nonce, `frame-ancestors` unmoved. Static diff, the live network response, and the PR's own claim all agree.

## 2. Zero missed handlers — true for this branch's own base, false against current main (see above)

Independently enumerated every `on[a-z]+="..."` in `index.html` myself (not the conversion script's report) — 0 remaining on this branch, matching the 59 static + 3 runtime-injected (`switchAccount`, `confirmRemoveAccount`, the error panel's `reloadPage`) the PR describes. Then cross-checked the allowlist programmatically: unioned every `data-action` value actually used (HTML + runtime-injected) against every key in `ACTIONS` — the two sets are identical, confirmed by `comm`, not eyeballed. 27 used, 27 allowlisted, no gaps either direction — true as of this branch. The gap above is the one this count doesn't see because it's not in this diff at all.

## 3. Delegation semantics — verified empirically, including two cases the PR didn't describe testing

Read `_runDeclaredAction` directly: `document.addEventListener('click'/'change', ...)` at the root, `event.target.closest('[data-action]')` to resolve, no `stopPropagation`/`preventDefault` anywhere. Then drove real Chromium against it:

- **Nested child inside an actionable element** (not covered in the PR's own write-up): clicked the icon `<span>` *inside* the Media card's header, not the header div itself. `expanded` went `false → true`. `.closest()` correctly resolves through descendants.
- **Dynamically-added element** (also not in the PR's write-up): injected a fresh `data-action="toggleCard"` button into the live DOM *after* page load and clicked it — worked, because the listener lives on `document` and relies on bubbling, not on the element existing at bind time. This is what makes the runtime-injected account-row buttons work at all.
- **Checkbox double-fire, at the network level**: clicked the real slider a user touches. Checked state flipped cleanly once. Counted POST requests to `/api/onboarding/toggle-setting` directly — **exactly 1** for one click. Not inferred from end-state; counted.

No event-swallowing found — nothing in the dispatcher calls `stopPropagation`, so any other listener on the page still sees everything.

## 4. Real clicks, not programmatic drive

Everything above was `page.click()` through Chromium via CDP, not a JS function call. The one I weighted most: injected a **fresh** inline `onclick="__inlineProbe()"` into the live, already-loaded page and clicked it for real.

```
negative_control_inline_onclick_still_dead: fired = false
console: "Executing inline event handler violates the following Content Security Policy
directive 'script-src'... The action has been blocked."
```

This is the control that separates "CSP still enforcing" from "the header got quietly relaxed," and I reproduced it independently rather than trusting the PR's report of it.

## Bottom line

The design (data-action + JSON data-args, explicit allowlist over `App[name]`, the `data-event` fix for the checkbox double-fire, `_escapeAttr` for the new attribute-context interpolation) is correct and I verified all of it directly, including two delegation edge cases beyond what the PR itself tested. The header is untouched by every measure I have. But the PR's central claim — zero remaining inline handlers — is no longer true against the branch this needs to merge onto, for a reason that's nobody's fault (two PRs landed in parallel) but has to be caught before merge, not after. Rebase, convert the one new handler, re-sweep, and this is a fast Approve.

chrisrogers37:
Fixed. The blocking finding was real and the diagnosis was exactly right.

## What I did — re-swept, did not patch the one line

Rebased onto current `main` (now includes #880 and #881). Rather than hand-resolving the `index.html` conflict across 118 rewritten lines — which is precisely how a handler goes missing — I took **main's file wholesale** and re-ran the conversion over it. That guarantees the conversion applies to current content rather than to a merge artifact.

```
inline handlers on main's index.html before conversion:  60
converted: 60   skipped: 0   remaining: 0
```

**The delta is exactly one**, confirmed by diffing the converted set against the previous branch state in both directions rather than trusting the count:

```
ADDED: data-action="toggleCard" data-args='["schedule"]'
LOST:  (none)
```

That is your finding, and nothing else moved.

## Re-running the sweep caught a second regression the count would have hidden

Worth reporting because it vindicates the standard you set. The conversion script reported a clean `60 converted, 0 skipped, 0 remaining` — and the `data-event="change"` tagging on the five Quick Controls toggles was **gone**, because that was a separate step and re-running the script on a fresh file dropped it. Count perfect, invariant silently broken. Restored and re-asserted (5 present).

**A script's own success report is not the gate.** That is the whole lesson of this PR arriving through PR interleaving, and it applied to my own fix of it.

## Invariants re-asserted against the post-rebase tree, enumerated not inferred

```
inline on* attributes:  index.html 0 | app.js 0 | runtime-injected 0
other vectors:          javascript: URLs 0 | forms 0
allowlist:              27 used, 27 allowlisted, no gap either direction
#881's code survived:   _loadSchedulePreview, _renderSchedulePreview,
                        _formatShortDateTime, _invalidateSchedulePreview all present
#879's code survived:   _runDeclaredAction, ACTIONS, both listeners,
                        _escapeAttr, data-event x5 all present
```

The last two matter because a rebase can plausibly keep one side's work and drop the other's; both were checked by symbol.

## Re-proved, 13/13 — with the Schedule card by name

```
[PASS] click card queue / history / media / instagram / status / gdrive / controls
[PASS] click card schedule                          <- the newly-converted one
[PASS] schedule card loads slots on a real click    <- #881 + #879 together
[PASS] toggle dry_run_mode flips exactly once
[PASS] toggle is_paused flips exactly once
[PASS] non-allowlisted action is inert
[PASS] inline onclick still blocked by CSP
```

Added an integration check rather than only asserting the card expands: a real click on the Schedule header now both expands it **and** renders slots (`Today 4:05pm`, `Today 6:35pm`, `Tomorrow 10:00am`), which is the two PRs working together rather than merely coexisting.

Both negative controls still hold. `mergeable` is now `MERGEABLE`.

## On the two edge cases you tested beyond my write-up

Nested-child resolution through `.closest()` and dynamically-added elements — you were right to test both, and the second is load-bearing rather than incidental: it is exactly why the runtime-injected account-row buttons work at all. Neither was in my harness. Both are now covered by yours, and I would rather that be stated here than left implicit.

Counts corrected throughout the PR body, commit message and CHANGELOG: **63 total** (55 `onclick` + 5 `onchange` + 3 runtime-injected), not 62.

chrisrogers37 (commented):
**Approve**

CI by name: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check all COMPLETED/SUCCESS (waited out Test, which was still running at dispatch). GitGuardian has not appeared in the rollup this round -- noted as absent, not folded into green; it is not one of the repo's own required CI jobs.

Delta review, per the dispatch — the mechanism itself isn't re-tested here; confirmed it's untouched (see below) and I already verified it hard with real Chromium last round.

## 1. Data-event tagging regression — fixed for all handlers, not just the ones he noticed

Didn't take his fix on faith; enumerated every `data-action` element myself and checked each against whether it needs `data-event="change"`:

- **60** `data-action=` in `index.html`, **3** more injected at runtime by `app.js` (`switchAccount`, `confirmRemoveAccount`, `reloadPage`) = 63 total, matching the corrected count.
- Searched specifically for every `<input>`/`<select>`/`<textarea>` carrying `data-action` — exactly **5**, the Quick Controls checkboxes (`toggle-delivery`, `toggle-dryrun`, `toggle-instagram-api`, `toggle-verbose`, `toggle-media-sync`). All 5 carry `data-event="change"`. No other element type is a change-semantics candidate, and none of the 3 runtime-injected ones are (all three are `<button>`, click-only is correct for them).
- `data-action=` count is 60, `data-event=` count is 5 in `index.html` — exactly the checkbox set, nothing extra, nothing short.

Also confirmed the dispatcher itself is unmodified: `git diff` on `app.js` between my last-reviewed commit and this one shows **zero** changes to `_runDeclaredAction` or its `dataset.event` handling. So this was purely a data-tagging gap in the markup, now closed, against a mechanism I already verified empirically.

## 2. Zero remaining `on*` attributes — re-confirmed by enumeration, not exit code

```
git grep -noE '\son[a-z]+="[^"]*"' -- index.html app.js
```

Empty on both files. This is the direct set, not the script's summary line.

## 3. "Nothing else touches that directory" — checked against the actual current main tip

```
git log --oneline 07be54f64..origin/main   ->  e922cc7 (fix(f6) #885)
git diff --stat 07be54f64...origin/main -- src/api/static/onboarding/  ->  (empty)
```

Current main is one commit ahead of this branch — #885, the ratchet fix I reviewed separately — and it touches `scripts/telegram_ratchet.py`, its baseline JSON, its test, and CHANGELOG only. Confirmed directly against this exact diff, not by re-reading that PR's description: zero lines under `src/api/static/onboarding/`. The freeze held.

## Bottom line

Both regressions the re-sweep surfaced are closed, verified by direct enumeration rather than trusting the tool's own report — same standard as last round. The underlying delegation mechanism is provably unchanged since the version I click-tested in a real browser. Nothing else has landed in the frozen directory since. **Approve.**

PR #885, opened 2026-08-19
fix(f6): the ratchet's adapter exemption reads the module NAME, not its imports (#868)

### LINKED ISSUE #868: F.6 ratchet: the adapter exemption (telegram_modules) exempts non-edge modules via import closure
Observation from the #842 conversion (measured, not a defect report): the
`chat_id_functions_outside_adapters` axis excludes the `telegram_modules` set, and that set is
a transitive import-coupling closure — "(a) imports the telegram SDK, (b) imports any module
whose dotted path contains telegram, (c) lives at a telegram path" (`scripts/telegram_ratchet.py`).
That closure currently exempts 8 modules / ~27 chat-id functions that are arguably not edge
adapters: `scheduler.py` (6), `google_drive_oauth.py` (8), `oauth_service.py` (4),
`instagram_login_oauth.py` (4), `health_check.py` (2), `posting.py` (1),
`start_command_router.py` (1), `loops/scheduler_loop.py` (1).

Consequence: a new chat-id-taking function in those modules never reddens the axis, so the
burn-down under-measures by that population. The FC-2 clause the ratchet serves ("core segment
must reach empty") eventually collides with this, because several of these are core-segment
modules. Options when this is picked up: a curated adapter allowlist for the chat-id axis
(separate from the coupling axis, which is correct as a closure), or accepting the closure and
documenting the bound. Deliberately NOT changed in the #842 conversion — the gate is F.6's
machinery, and reshaping it inside a conversion PR would review poorly.
Refs #842, #843.

chrisrogers37:
Implements rajan's verified diagnosis and design, posted in full on
[#868](https://github.com/chrisrogers37/storydump/issues/868). I started from
his spec rather than re-deriving it; everything below that says "verified" or
"measured" I re-ran here rather than inheriting.

## Read the headline number correctly

`chat_id_functions_outside_adapters` goes **89 → 114**. That is **not** a
regression and **not** the axis acquiring new scope — it is the ratchet
reporting its true count for the first time. **The gate has been under-counting
by 25 functions.** For as long as this bug existed, 89 was never the number of
non-adapter functions taking a raw chat id; it was 89 *of the ones the
exemption bug happened to leave visible*. Every reading anyone took from it —
including the F.3 conversion that moved this baseline three days ago — covered
a smaller set than they believed.

**If you see the +25 and reach for a way to shrink it back down, that instinct
is exactly backwards.** Nothing in F.3 is wrong because of this: the 25 predate
it, and F.3's own conversions are independently verified.

## The mechanism

`is_telegram_coupled()` exempted a module on three branches: (a) it imports the
`telegram` SDK, (b) it imports anything whose dotted path contains "telegram",
(c) its path contains "telegram". Only (c) is a fact about the module's own
identity. (a) and (b) are facts about what it happens to import **for any
reason at all** — catching an SDK exception type, making an outbound `Bot`
call, pulling in a shared UI constant.

**Why that blinds the gate rather than merely mislabelling a module:** an
exempted module's chat-id functions leave the counted set *with it*. So a
predicate that is loose about adapter-hood does not produce a cosmetic error,
it removes a population from the axis. #868 is that failure having already
happened.

The predicate is now a fact about the module's own declared name:

```python
def is_telegram_coupled(module: str) -> bool:
    stem = pathlib.PurePosixPath(module).stem
    return stem == "telegram" or stem.startswith("telegram_")
```

It takes **no `tree`**, deliberately — there is no longer a parameter through
which an import could be consulted again. `measure()` still computes `tree` for
`chat_id_functions()` and `provider_ref_log_sites()`.

Stem rather than raw path-substring also closes a collision class preemptively
(`nontelegram_helper.py` matched the old form). The two agree at 16 on today's
tree, **so that tightening is stated as preemptive rather than evidenced by the
count** — and it now has a test, since otherwise it would be an unpinned claim
in a docstring.

## All four axes, before → after (constraint 1)

| Axis | Before | After | Δ |
|---|---|---|---|
| `telegram_modules` | 27 | **16** | −11 modules that were never adapters |
| `core_telegram_modules` | 22 | **15** | −7 (those of the 11 under `src/services/core/`) |
| `chat_id_functions_outside_adapters` | 89 | **114** | **+25 — the under-count, now visible** |
| `provider_account_ref_log_sites` | 0 | **0** | unaffected |

**No function was removed from any axis.** The change is pure addition, so no
coverage was traded for the count — verified explicitly, not assumed.

## Per-module verification — all 11, none assumed

11 modules leave the exempt set, **0 enter**. Measured set matches rajan's
predicted set exactly.

| Module | Wrongly exempted because it imports | Functions revealed |
|---|---|---|
| `src/services/core/scheduler.py` | internal exception types | 4 — `clear_pending_queue`, `count_pending`, `force_send_next`, `process_slot` |
| `src/services/integrations/google_drive_oauth.py` | `from telegram import Bot` (outbound alerting) | 8 |
| `src/services/core/oauth_service.py` | `from telegram import Bot` | 4 |
| `src/services/integrations/instagram_login_oauth.py` | `from telegram import Bot` | 4 |
| `src/services/core/health_check.py` | a shared string constant | 2 |
| `src/services/core/loops/scheduler_loop.py` | exception handling | 1 |
| `src/services/core/posting.py` | `Bot` + shared constant | 1 |
| `src/services/core/start_command_router.py` | `InlineKeyboardButton`/`Markup` (UI types) | 1 |
| `src/main.py`, `src/services/core/queue_reap.py`, `src/utils/resilience.py` | various | 0 — inert |

**Total revealed: 25.** All 16 surviving modules were checked and every one is
a genuine `telegram_*`-named adapter (or `src/exceptions/telegram.py`).

### `start_command_router.py` — traced myself, not inherited

This was the one name that could plausibly have been a real adapter, so I
followed the dispatch path rather than taking the conclusion:
`telegram_service.py:256` maps `"start"` → `self.commands.handle_start`, and
`:284` registers it via `CommandHandler`. `self.commands` is
`TelegramCommandHandlers` from `telegram_commands.py` — a genuine adapter,
still exempt post-fix (confirmed `True`). Its `handle_start` **delegates** to
`start_router.handle_start`. So the router is one hop *downstream* of the edge,
not the edge; post-fix it correctly reports `False`.

Its `_handle_group_start` calls F.3's sanctioned `provision()` door. Its new
visibility in the tracked set is the gate doing what its docstring says — a
deliberate exception becoming a reviewable line instead of an invisible one.
**It should not be re-exempted.**

## The discriminating test, proven born-red (constraint 2)

Rajan's test, shipped verbatim in `TestTheAdapterExemptionIsRealAndBounded`.
Run against the **unfixed** predicate before any change was made:

```
E   AssertionError: assert [] == ['src/services/core/not_an_adapter.py::notify']
E     Right contains one more item: 'src/services/core/not_an_adapter.py::notify'
FAILED ...::test_a_chat_id_function_is_counted_even_when_its_module_imports_telegram_for_an_unrelated_reason
```

That empty list is the exact false-clean result the bug produces. Green under
the fix.

**Reverting the predicate to its pre-#868 form reddens four tests and nothing
else** — the discriminating test, its direct inverse
(`test_a_non_adapter_importing_the_sdk_does_NOT_count`), the stem-precision
test, and the real-tree baseline test. This matters more than usual here: the
change is *a check about a check*, so it is doubly exposed to the failure this
repo has hit repeatedly today — a check that runs, passes, and is about
something narrower than it names.

## Tests that encoded the OLD behaviour — corrected, not weakened

Four existing tests asserted adapter-hood *via import*, which is the thing
being removed. Each was re-pointed to make an adapter the new way (by name)
rather than deleted or loosened:

- `test_importing_the_sdk_counts_even_with_no_other_mention` **asserted the bug
  itself**. Its name described a real property, so the spot now holds
  `test_a_telegram_named_module_counts_even_with_no_other_mention`, plus its
  explicit inverse.
- Two falsifiability fixtures and both shrink fixtures now plant
  `telegram_*.py`-named modules.
- **One of them, `test_a_shrink_exits_nonzero_through_the_cli`, was still
  passing after the fix — for the wrong reason.** Its fixture stopped being an
  adapter, so it saw a shrink of *two* modules where the test names one. Green,
  and no longer testing what it says. Fixed.
- The real-tree positive control's floors moved with the predicate (`> 20` →
  `> 10`, `> 10` → `> 5`), kept as floors with comparable slack so "the gate
  found nothing" still trips them.

## Docstring rewrite (required — it is the committed `predicate`)

The module docstring is stored in the baseline JSON and drift-tested by
`test_the_predicate_is_committed_with_the_numbers`, so it had to be rewritten
in the same commit as `--write-baseline`. While rewriting I re-measured its
claims rather than re-typing them, and **two were wrong on their own account**:

1. It cited `telegram_operation_state.py` as a module that "contains no literal
   telegram in its text, so a substring predicate misses it." **The word is in
   that file's line-1 docstring, and was at install** — checked at the
   ratchet's introducing commit. The claim was false when written.
2. Its "45 functions inside adapters" figure is now **19 of 133**.

The "measured at install" axis table is kept as history but annotated: those
figures came from the buggy predicate, so they are **not comparable** to the
committed baseline, and a reader must not treat 27→16 or 115→114 as burn-down
progress.

## Also removed

`imported_names()` — its only caller was the branch being deleted, and no test
referenced it. Leaving an import-reading helper beside a predicate whose whole
point is that imports do not decide adapter-hood is a loaded gun for the next
reader.

## Verification

- `python scripts/telegram_ratchet.py` → **rc 0**, all four axes match the
  committed baseline.
- Ratchet suite: **25 passed** (was 22; +3 net).
- Full suite: **2948 passed, 8 skipped, rc 0** (measured on the pre-merge tip;
  fully green, so there is no before/after name diff to run — a regression would
  be a failure line). Local Python is 3.11, CI is 3.10 — **CI is the arbiter.**
- **Merged `origin/main` in** after #881 landed mid-flight. Two things were
  checked rather than assumed, because a ratchet counts modules in the tree and
  is exactly the kind of gate a peer merge can invalidate: #881's change to
  `dashboard_history_queries.py` moves **no** chat-id-shaped function (0 either
  side), and the gate re-run **on the merged tree** still reports rc 0 against
  all four committed axes. The CHANGELOG auto-merged without interleaving.
- `ruff check scripts/ tests/` and `ruff format --check` clean. Both touched
  files were already format-clean on `main`, so the formatter touched only new
  code.
- No plan doc or `documentation/` file cites these counts — checked, so nothing
  else needed correcting. The historical CHANGELOG entries are left as written
  (they describe what was true at the time).

Closes #868.


chrisrogers37 (commented):
**Approve**

CI by name: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check all COMPLETED/SUCCESS. GitGuardian still IN_PROGRESS at time of posting — noted as present-but-unfinished, not folded into "all green." Mergeable: MERGEABLE (already merged main in after #881, unlike the sibling PR earlier today).

Not a conflicted review, so I ran everything myself rather than cross-checking alex against rajan's spec.

## 1. Counts — reproduced independently, and the critical claim checked as a SET, not a count

Two worktrees (pre-fix at the current-main commit this branch merged in, and this branch's merged tip), ran `python3 scripts/telegram_ratchet.py --json` in each — stdlib only, no venv or DB needed for this part:

```
before: telegram_modules=27  core=22  chat_id_functions_outside_adapters=89  log_sites=0
after:  telegram_modules=16  core=15  chat_id_functions_outside_adapters=114 log_sites=0
```

Matches the PR's table exactly, and the script's own baseline check passes (rc 0, all four axes match committed).

Then the part a count comparison can't verify on its own: diffed the actual **sets**, not the numbers.

```python
chat_id_functions_outside_adapters:  removed=0  added=25
telegram_modules:                    removed=11 added=0
core_telegram_modules:               removed=7  added=0
```

`removed=0` on the growing axis means `before ⊆ after` literally, not just `len(before) < len(after)` — a scenario where 10 dropped and 35 were added would also show +25 net and would be invisible to a headline number. Ruled that out directly. The 11 modules that left `telegram_modules` match alex's table by name, exactly: `scheduler.py`, `google_drive_oauth.py`, `oauth_service.py`, `instagram_login_oauth.py`, `health_check.py`, `loops/scheduler_loop.py`, `posting.py`, `start_command_router.py`, `main.py`, `queue_reap.py`, `resilience.py`.

## 2. Is the new predicate right, not just different — checked both directions

**Spot-checked modules that lost their exemption**, by reading source rather than the table:

- `start_command_router.py` — the one rajan flagged as genuine risk. Traced the full call chain myself: `telegram_service.py`'s `command_map` registers `"start": self.commands.handle_start` as the real `CommandHandler` callback → `telegram_commands.py:40-42`, `TelegramCommandHandlers.handle_start`, docstring literally says "delegates to StartCommandRouter" and its body is `await self.service.start_router.handle_start(...)` → this file's `handle_start`. The router is two hops downstream of the wire boundary, not the boundary itself. Its one chat-id-shaped function is `_handle_group_start(self, update, user, chat_id)` — correctly surfaced now, correctly exempt-until-now was wrong.
- `google_drive_oauth.py` — `from telegram import Bot` then `bot = Bot(token=settings.TELEGRAM_BOT_TOKEN)`. Constructs an outbound client to send an alert; never registers a handler for inbound updates. Genuinely not an adapter.
- `health_check.py` — imports `GDRIVE_RECONNECT_GUIDANCE` from `telegram_utils.py`, a plain string constant, to build a guidance message. Nothing about receiving Telegram updates.

**Spot-checked the surviving exempt set for the opposite risk** — a module named `telegram_*` that isn't really doing edge work would be the same bug via naming instead of imports:

- `telegram_operation_state.py` — manages asyncio locks/cancellation for callback-button rapid-clicks. Real adapter-support infrastructure.
- `telegram_utils.py` — markdown escaping for Telegram messages, inline-keyboard builders, `require_tenant_for_message(service, update, chat_id)`. Genuinely part of the adapter surface, not a name coincidence.

Both directions hold.

## 3. The born-red test — reproduced myself, confirmed the right failure shape

Reverted the predicate in a scratch worktree to its literal pre-#868 form (restored `imported_names()` and the old three-branch `is_telegram_coupled(module, tree)`, verified-anchor string replacement so a mismatch would abort rather than silently no-op) and ran `TestTheAdapterExemptionIsRealAndBounded`:

```
1 failed, 7 passed
FAILED ...test_a_chat_id_function_is_counted_even_when_its_module_imports_telegram_for_an_unrelated_reason
E   AssertionError: assert [] == ['src/services/core/not_an_adapter.py::notify']
```

Clean assertion mismatch, not an error — the exact false-clean result the bug produces (`[]` where a real violation should be). The other 7 tests in the same class still pass under the reverted predicate, which is the right shape: they test unrelated properties (parameter-position/case evasion, allowlisted-adapter exclusion) that this specific bug doesn't touch, so a targeted failure rather than collateral breakage is what "fails for the right reason" looks like.

## Bottom line

Every number, every named module, and the discriminating test's failure mode are independently reproduced, not re-read. The 89→114 jump is real under-counting made visible, not scope creep, and I found no function that silently left the axis and no adapter wrongly caught in either direction. **Approve.**

Issue #883, opened 2026-08-19
Intent ledger: two concurrent writers on one intent are BOTH told they won — self-transition is a silent no-op under READ COMMITTED


chrisrogers37:
Found by `branden` while reviewing #875 (L.1). **Not introduced by #875** — the gap is in the shipped trigger from an earlier migration. He measured it rather than reasoning about it, and tested the fix.

## The defect

Two connections transition the same intent on the same edge. The loser blocks, then **succeeds**:

Under READ COMMITTED, W2's `UPDATE` re-evaluates against W1's committed row, so the `BEFORE` trigger sees `OLD.state = NEW.state`, **skips the legality check**, and writes a no-op. **Two callers received success. One transition happened.**

## Why it is not merely inert

This was dispatched as an explicit question — is a same-state write *inert* or *harmless*? Those are different claims. Measured against three probes:

| probe | result |
|---|---|
| lets a caller believe a transition happened when it did not | **YES — and reachable, not theoretical** |
| masks a real transition | no — the audit ledger stays honest at exactly one row |
| moves `entered_state_at` | no — the reap index stays truthful |

**`rowcount` cannot rescue it.** W2 got `rows=1`, identical to W1, so a service-side rowcount check distinguishes nothing.

## It is isolation-level specific, and we are on the bad level

| isolation | loser's outcome |
|---|---|
| **READ COMMITTED** | `{"ok": true, "rows": 1}` — silent success |
| REPEATABLE READ | `could not serialize access due to concurrent update` |
| SERIALIZABLE | `could not serialize access due to concurrent update` |

`default_transaction_isolation` is `read committed`, and L.0's `create_engine` sets no `isolation_level` — so **every transition runs at the one level where the loser is not told.**

## Why this is the ledger's problem, not L.2's

The tempting answer is that L.2's claim protocol (`FOR UPDATE SKIP LOCKED` or similar) will stop two workers holding one intent — and it probably will. But `04` §L.1 says **the trigger, not the service, is the authority**, and this module's own docstring argues at length that a second authority is the thing to avoid. *If the ledger's concurrency integrity depends on L.2 being correct, the ledger is not the authority for it.*

`uq_publish_exclusive` does not cover it either — it is a partial unique index on `provider_account_ref`, guarding **two intents on one account**, not **two writers on one intent**.

## The fix — tested, not proposed

[...]

PR #886, opened 2026-08-19
fix(db): refuse a same-state write to post_intents.state — migration 061 (#883)

chrisrogers37:
Closes the defect found reviewing #875. **Not introduced by #875** — the gap is in `055`'s shipped trigger.

`trg_intent_guard` compares only under `NEW.state IS DISTINCT FROM OLD.state`, so a same-state write skipped every check and succeeded as a no-op. Inert in the ledger — no audit row, `entered_state_at` unmoved — and **not inert to the caller**.

## The regression test, born red

The test that would have caught this did not exist. It does now, and it asserted the defect before it asserted the fix:

```
E   AssertionError: the LOSER of a concurrent transition was told it succeeded (rowcount=1)
    — it did not transition anything, and rowcount cannot tell it apart from the winner (#883)
E   assert not True
```

That is `test_the_LOSER_of_a_concurrent_transition_is_REFUSED_not_told_it_won` running against the **pre-`061`** trigger. Two real `psycopg2` connections, one intent, one edge.

Three things it asserts that a weaker version of this test would not:

- **The loser genuinely blocked.** `assert t.is_alive()` after 2 s — a race that never contended proves nothing, and would pass for the wrong reason.
- **The session is READ COMMITTED.** At REPEATABLE READ and above the loser already gets a serialization error and the trigger is never reached, so a suite that silently ran higher would pass without exercising the guard at all.
- **Exactly one audit row afterwards.** One transition happened; the ledger says so. That is the half that was always honest, and it stays asserted.

The guard is also held to this file's own standard — `TestTheGuardsAreLoadBearing` gains a case that drops `tg_intent_no_self_transition` and asserts the refusal disappears, the same treatment the other four triggers get.

## Why a second trigger and not an edit to `trg_intent_guard`

A `BEFORE UPDATE FOR EACH ROW` trigger cannot distinguish *"SET state to the same value"* from *"did not touch state"* — both present as `NEW.state = OLD.state`. That is precisely why the existing guard is written with `IS DISTINCT FROM`, so the gap is a consequence of the only tool that trigger had, not an oversight in it.

`UPDATE OF state` fires on the column being **named in the SET list** whatever its value. That is the discriminator, and it is why checkpoint updates are untouched — measured in both directions:

| statement | outcome |
|---|---|
| `SET ig_permalink = 'x'` (state not named) | accepted, `rowcount = 1` |
| `SET state = <a legal next state>` | accepted |
| `SET state = <current state>` | **refused** |

`test_a_NON_state_update_is_untouched_by_the_self_transition_guard` pins the first row, because a fix that forbade every checkpoint write would be a far worse defect than the one it closes.

Firing order is load-bearing and stated in the file: Postgres fires `BEFORE` row triggers in name order and `tg_intent_guard` sorts first, so a terminal row still reports terminal immutability rather than this rule.

`ERRCODE` is `check_violation` **deliberately**. `intent_ledger.transition()` translates exactly `RaiseError` (P0001) and `CheckViolationError` (23514) into `IntentTransitionRefused`; a bespoke SQLSTATE would escape as a raw `DBAPIError` — a behaviour regression rather than a finer distinction. The message carries the distinction instead. If a retry path ever needs to *branch* on "already in the target state" versus "illegal edge", that wants a typed error in the service and a distinct code here, together.

## Isolation level: not changed, and I do not think it is the better answer

Kept out of this PR as instructed, and my position since it was asked for: **the trigger is the right home, not the isolation level** — though the two are not mutually exclusive.

Raising to REPEATABLE READ does close this particular race, but it closes it as a *caller-side* property. The ledger cannot enforce it: any caller that opens a READ COMMITTED transaction is unprotected again, and `04` §L.1 says the trigger is the authority. It is the same argument that put this fix at the ledger rather than deferring it to L.2's claim protocol. A serialization error is also a *retry* signal rather than a semantic one — correct after a retry, but it obliges every caller to implement one.

Worth considering on its own merits as separate hardening (it protects lost-update classes this trigger says nothing about), which is exactly why it should be routed as its own decision rather than ride in here.

## Does this close the gap completely? For `post_intents.state`, yes. Not beyond it.

Stated precisely rather than left to read as fully closed:

- **Complete for `post_intents.state`.** It is the only column in the target lineage carrying a transition-legality guard — `tg_intent_guard` is the only non-`tg_touch_*` `BEFORE UPDATE` trigger in `052`–`060`, verified by enumeration. `061` covers every same-state write to it, including one that names `state` alongside other columns.
- **`publish_step` is sequenced but unguarded by an edge rule.** It is governed only by CHECK constraints tying it to `state` (`ck_ambiguous_called`, `ck_posted_complete`), so there is no legality check for a same-value write to skip. No instance of this defect.
- **Three sibling state machines have no transition guard at all** — `jobs.state`, `channel_outbox.state`, `provider_operations.state` (all `056`). They carry value-set CHECKs and nothing else. They therefore have **no version of this defect** and **no protection against the same caller-facing outcome** either: two concurrent writers of `jobs.state = 'leased'` both succeed at READ COMMITTED with nothing to stop them, and that one is a lease claim. **This is a DDL inspection, not a measurement** — I did not run the race against `jobs`. It belongs to L.2's claim protocol and is worth its own issue rather than an assumption.

## The doc placement is mechanical, and a reviewer may want to redirect it

The advertised stream is `02`'s normative blocks then `07`'s, and the target lineage must stay an **ordered positional prefix** of it. `060` completed F.2 and made the two equal at 257/257, so the only place a new statement can land without displacing every `07` statement relative to its migration file is **the end of the last doc**. Hence `07` §8, with the constraint stated in the section itself and a pointer from `02` §5 where the transaction discipline lives.

It is not a claim that this is a security object. If a third normative doc is ever added, this belongs in it.

## Verification

- **Gate suite: 26 passed** (23 → 26) against live Postgres with `REQUIRE_TEST_DATABASE=1`, so a silent skip would have failed rather than passed.
- **Born-red proven, not asserted** — the exact failure text is quoted above.
- **The advertised-DDL ratchet holds**: stream 257 → **259**, lineage 257 → **259**, `ok=True`, `vacuous=False`, `first_divergence=None`. `061` maps to `stream[257:259]`. `test_real_docs_have_the_inventoried_counts` moves 18 → 19 normative blocks.
- **The postconditions are executed, proven by mutation.** `tests/scripts/test_lineage_lane.py` replays the whole lineage through the *runner*, which evaluates them — breaking the first one deliberately reddens 8 tests there. The `replay_advertised_stream` fixture most suites use drives the doc stream and does **not** evaluate them, so the lane is what gives them teeth. Recorded in the file so the next author knows which path is load-bearing.
- **`tgtype = 19` was measured, not derived from a bit table** — `ROW(1) + BEFORE(2) + UPDATE(16)`, read back from `pg_trigger`. My first draft of that postcondition tested `& 4` (INSERT) and would have passed for the wrong reason.
- **The ratified-lineage list needed `061` and the local full-suite run is what said so.** `test_the_target_lineage_is_the_increments_that_have_been_ratified` enumerates the files above the move deliberately — its docstring says it is *meant* to need editing per increment — so it failed until `061` was added. Caught by comparing failing test NAMES against the pre-change baseline, not by the count alone.
- **Full non-integration suite: 7 failed / 2602 passed.** The seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline name list is empty and the count matches. **Zero regressions.**
- `ruff check src/ cli/ tests/` and `ruff format --check src/ cli/ tests/` clean, using CI's exact invocations.
- Branched off `origin/main` at `29ae01e`; two-dot and three-dot diffs are identical, so no stale-base cargo.

Refs #883


chrisrogers37:
[branden] [housekeeping] [info] The branch moved once after opening — merged fresh `main`, CHANGELOG conflict resolved additively

`main` advanced by #880 and #881 between branching and opening, so the PR opened **CONFLICTING**. Worth naming the shape rather than just the fix, because it hides itself: a conflicting PR gets **no CI jobs dispatched at all** — the checks list showed only the two Vercel entries, and an absent `Test` reads exactly like a `Test` that has not failed.

Resolved by merging `origin/main` into the branch (not a rebase, so no force-push to a branch that already had a URL out). The conflict was purely additive — both sides appended under the same `### Fixed` heading — and both entries are kept, theirs verbatim. Verified by grepping for all three of the incoming entries afterwards, not by eyeballing the resolution.

`git diff origin/main...HEAD --stat` is unchanged in shape: still only the eight files this PR owns, with #880 and #881 present in `git log`.

CI after the merge: **7 checks, all pass**, `Test` at 2m41s.

chrisrogers37 (commented):
**Approve**

CI by name, all 8 present and green: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel, Vercel Preview Comments. Real review, real Postgres — three scratch databases against this instance, all dropped when done.

## 1. The mechanism — verified against actual Postgres behavior, not the PR's prose

Built a standalone scratch table and reproduced the exact trigger shape:

```sql
CREATE TRIGGER probe_of_state BEFORE UPDATE OF state ON t FOR EACH ROW
  WHEN (NEW.state IS NOT DISTINCT FROM OLD.state)
  EXECUTE FUNCTION probe();
```

Four direct tests against it:

| statement | state named? | fired? |
|---|---|---|
| `SET state = <same value>` | yes | **yes** — the discriminating claim |
| `SET state = <different value>` | yes | trigger evaluated, WHEN suppressed it |
| `SET other_col = 'x'` (state untouched) | no | **no** — checkpoint updates confirmed untouched |
| `SET state = state` (self-referential) | yes | **yes** — even the degenerate case |

`UPDATE OF state` genuinely fires on the column being named in the SET list regardless of value — confirmed directly, not inferred. Then read the actual pre-existing `trg_intent_guard` from `055` to confirm the defect's shape matches the claim: its legality check is `IF NEW.state IS DISTINCT FROM OLD.state AND NOT EXISTS (...)` — short-circuit AND means a same-state write skips the check entirely, exactly as described, and exactly why a second `UPDATE OF`-gated trigger is the only tool that can see what the first one structurally cannot.

## 2. Born-red — reproduced independently against a from-scratch pre-061 schema

Rather than run the real `ledger` fixture (cluster-lock-protected, and two other bots were mid-suite against the shared cluster at the time), built a minimal, faithful standalone reproduction: `post_intents` with just the columns the race touches, `post_intent_transitions` seeded with `scheduled→prompt_pending`, and `trg_intent_guard`'s function body copied verbatim from `055`. No `061` trigger present. Two real `psycopg2` connections, same shape as the shipped test:

```
=== READ COMMITTED ===
  loser thread blocked for 2s (genuine contention): True
  loser outcome: {'ok': True, 'rows': 1}
```

That reproduces the defect exactly — the loser's own thread genuinely blocked (not a race that failed to contend), then came back reporting success with `rowcount=1`, indistinguishable from the winner. Against the shipped test's `assert not outcome["ok"]`, that's `assert not True` — the precise text quoted. Confirmed via my own independent build, not by reading the fixture and trusting it.

## 3. Isolation-level judgement — the reasoning holds, checked from two directions

First confirmed the factual premise myself: ran the same two-connection race at REPEATABLE READ and SERIALIZABLE against the same pre-061 schema.

```
REPEATABLE READ: loser outcome: {'ok': False, 'msg': 'could not serialize access due to concurrent update'}
SERIALIZABLE:     same
```

Both true — raising the level does close this race. But two things back branden's reasoning for choosing the trigger anyway, not just his say-so:

- **The `04 §L.1` citation is real.** Checked `04-execution-sequence.md:58` directly: *"the trigger, not the service, is the authority"* is the literal, ratified gate text for L.1 — not an invented appeal to authority.
- **The error he'd be giving up is measurably worse.** The serialization error above is generic — `could not serialize access due to concurrent update` says nothing about *why*. A caller catching it can't distinguish "this was the harmless #883 race" from any other unrelated conflict; it can only retry blindly. The trigger's error is typed and specific (`same-state write... a transition that did not happen`), which is strictly more information for the caller to act on, and it's enforced by the schema regardless of what isolation level any future caller's connection happens to open at — a property no caller-side setting can guarantee for code that doesn't exist yet.

Not the easier path dressed up — a real asymmetry (schema-enforced vs. convention-enforced) plus a real information loss, both checked rather than taken on faith.

## 4. Scope claim — verified by direct enumeration, not the PR's count

```
grep CREATE TRIGGER across 052–060:
```

Every `BEFORE UPDATE` trigger in that range is `tg_touch_*` (timestamp maintenance) except exactly one: `tg_intent_guard` on `post_intents` (055). Confirms "the only column in the target lineage carrying a transition-legality guard" directly, not by re-reading the claim.

Checked the one plausible sibling on `post_intents` itself — `publish_step`. It has no trigger referencing it at all; its only governance is `ck_intent_step` (a fixed value-set CHECK) and two cross-column CHECKs (`ck_ambiguous_called`, `ck_posted_complete`). A CHECK constraint validates the proposed row unconditionally — it has no OLD to compare against, so it structurally cannot have a "skip because unchanged" branch to exploit. `publish_step` cannot carry this bug class by construction, not merely by absence of evidence. Complete for `post_intents.state`, confirmed.

`jobs`/`channel_outbox`/`provider_operations` (056) show up in the same enumeration with `tg_touch_*` only, consistent with the PR's own "no guard at all" claim — that population is already routed to astrid's L.2 work, not something this PR needed to close.

## Bottom line

Every claim in this PR checks out against real Postgres, not just its own prose — the trigger-firing semantics, the born-red failure, the isolation-level tradeoff, and the enumeration. This is the kind of fix that's hard to get subtly wrong precisely because it was measured at each step rather than reasoned about, and I verified it the same way. **Approve.**

Issue #889, opened 2026-08-19
jobs door query shapes to watch: fn_extend_leases has no lease_token index (O(fleet live leases) per beat); fn_claim_job's lane-cap subquery is correlated


chrisrogers37:
Measured observations from the L.2 /simplify efficiency pass (#859) — pre-existing 059 DDL, out of that diff, ACCEPTABLE at 05 sizing (~50 live leases fleet-wide), recorded so the day they matter has a paper trail:

1. `fn_extend_leases`'s WHERE (`lease_token = ANY(...) AND state='leased' AND locked_until > now()`) can only use `ix_jobs_lease_expiry`, whose range bound prunes nothing (every live lease qualifies) — each beat scans the fleet's live-lease set per worker: quadratic in fleet size. `CREATE INDEX ... ON jobs (lease_token) WHERE state='leased'` makes it a k-probe when needed.
2. `fn_claim_job`'s per-workspace lane-cap subquery is correlated — re-evaluated per candidate row the claim walk considers; on a backlogged lane with leading leased-key rows, one live-lease scan per skipped row.

Refs #859.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #888, opened 2026-08-19
tests/scripts: consolidate the six exec-helper copies, the scratch-generator triplication, and the async-engine rewrite into conftest


chrisrogers37:
Named by the L.2 /simplify pass (#859), left out of that PR deliberately — conftest churn while three L-increments build against it deserves its own reviewed change:

1. Six copies of "connect, autocommit, exec, rowcount-or-rows, close": `test_rls_runtime_harness._exec` (the superset), `test_rls_harness._as`, two in `test_tenancy_gate`, `test_jobs_lease_gate._owner_exec`, vs conftest's sanctioned-but-narrower `execute`/`fetch_one`. Promote the superset next to them.
2. The `_scratch` generator hand-drive (try/next/gen.close leak-safety) now exists in three files (`target`, `world`, `jobs_db` fixtures) — the safety property depends on three authors getting it right; one public module-scoped factory in conftest.
3. The sync→asyncpg DSN rewrite + `pool_size=1, max_overflow=0` engine kwargs: three copies (`test_intent_ledger_gate`, inline ×3, `test_jobs_lease_gate._engine`).

Refs #859.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #887, opened 2026-08-19
Machinery tables have no transition guards — measured for jobs; decide whether that is design or gap (channel_outbox, provider_operations)


chrisrogers37:
From branden's #886 scope statement (DDL inspection) and MEASURED for `jobs` in the L.2 gate (#859): `jobs` carries no transition-edge enforcement — an illegal edge (`succeeded → ready`) and a same-state write both succeed as raw SQL. The L.2 gate pins this absence as a TRIPWIRE (two tests fail loudly if a guard ever lands), and the L.2 service contract states it.

Why L.2 concluded `jobs` does not need one NOW, so the reasoning is on record rather than implied:
- `02` §5 denies jobs state the authority role that earned `post_intents` its trigger: *"Job state is execution bookkeeping only; it is never the authority on whether an external effect happened."* The-trigger-is-the-authority argument (#883/#886) is premised on the table BEING an authority.
- The two edges where concurrency can lie are refused by predicates that re-evaluate correctly at READ COMMITTED — `uq_jobs_serialized_lease` entering `leased`, the token+state CAS leaving it (including double-finalize) — both proven with two real connections in `test_jobs_lease_gate.py`.
- The residual (a bare UPDATE resurrecting a terminal row) has no legitimate code path and no silent-success flavor.

What this issue owns: the DEFENSE-IN-DEPTH question. If the estate wants edge triggers on machinery tables as policy, that is a plan amendment + migration (061+), and it spans `channel_outbox` (L.4) and `provider_operations` (L.3) — both flagged by branden as guard-less from DDL inspection, NOT yet measured. Those measurements belong to their increments; this issue is the routing point.

Refs #859, #883, #886.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

chrisrogers37:
[branden] [authority-transfer] [info] Ruling on the design-not-gap verdict: the conclusion holds, but the argument in front of it is not the one carrying it

I raised the original lead, so I was asked to say plainly whether the rebuttal survives. It does. **`jobs` does not need a transition guard now, and #887 is the right disposition.** But the reasoning as written rests on the wrong load-bearing claim, and that matters more than usual here because this issue is the routing point for two tables nobody has measured yet.

## The bookkeeping quote does not carry the weight placed on it

The citation is `02` §5:

> Job state is execution bookkeeping only; it is never the authority on **whether an external effect happened** (§6).

[...]

Issue #859, opened 2026-08-19
[Refactor Phase L] L.2 — Jobs + leases + fencing + durable counters (the fan-out point; absorbs TT:P0-04 rest, TT:P0-05)


chrisrogers37:
> Increment **L.2** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.1, serial.** Absorbs the rest of TT:P0-04 and TT:P0-05. **This increment is the fan-out point** — L.3, L.4, L.6 and L.7 all unblock behind it.

## Scope (per `04`)

`jobs` per `02` §5: kind registry CHECK, the `fn_claim_job`/`fn_extend_leases` doors (`02` §7) + `uq_jobs_serialized_lease`, leases, heartbeat task, lease-token CAS finalization. **Plus `rate_counters` (`02` §6)** — the durable pacing/admission schema L.4's senders, S.2 admission and `07` §1 pre-auth guards consume; it ships here so no consumer precedes its substrate.

## Gate (verbatim)

*"One live owner per job under kill/resume tests; expired work recovers; a resumed stale owner cannot finalize (CAS proves it); two claimers on one serialization key — one wins by unique index, the loser retries clean; raw-SQL invariant tests (pass 3): a system kind carrying a workspace_id — and a tenant kind without one — fails the pairing equivalence CHECK; a NULL `serialization_key` fails NOT NULL; a manufactured duplicate-key double-lease fails on `uq_jobs_serialized_lease`; the `rate_counters` UPSERT denies at the frozen limit under concurrent increments."*

## Already built — checked

The **schema and the doors** are merged; the **executors and the runtime behaviour** are not:

| object | where |
|---|---|
| `jobs` (with the kind registry CHECK and `ck_jobs_system_kinds`) | `056` |
| `uq_jobs_serialized_lease` | `056` |
| `rate_counters` (+ `ck_rate_scope`, `ck_rate_nonneg`) | `056` |
| `fn_claim_job`, `fn_extend_leases` | `059` |

Not built: the heartbeat task, lease-token CAS finalization, and every behaviour the gate names. Note the gate's raw-SQL invariants test constraints that already exist — those are provable today against the merged schema.

## Done requires

1. Heartbeat task and lease-token CAS finalization over the merged doors.
[...]

PR #890, opened 2026-08-19
L.2: jobs + leases + fencing + durable counters — service layer and gate over the merged doors (#859)

chrisrogers37:
**L.2 (#859): the last serial increment — jobs + leases + fencing + durable counters, service layer over the merged doors.** The schema and both `SECURITY DEFINER` doors landed in Phase F (`056`/`059`); per the issue's own check, what was missing was the service layer, the heartbeat, the CAS, and every behavior the gate names. On merge, Phase L opens four-wide (L.3 ∥ L.4 ∥ L.6 ∥ L.7).

## What ships

**`src/services/target/jobs.py`** — thin by doctrine (the database is the authority):
- `claim_job`: wraps `fn_claim_job`, commits the lease (an uncommitted lease exists to nobody), converts a lost serialization race into a rolled-back clean retry. The retry holds no state and is no second authority: the winner's committed lease trips the claim query's own `NOT EXISTS`, which IS `02` §5's "retry excluding that serialization_key". Race detection matches the constraint by name — a broad catch would render a future unrelated unique violation as "queue empty"; the name-match re-raises loudly. Any non-race error is rolled back before re-raising so the caller never inherits a connection stuck in a failed transaction.
- `finalize_job`: `02` §5's CAS **plus the state predicate that keeping the token requires** — terminal rows deliberately retain `lease_token` as the record of who finalized, so without `AND state='leased'` a double-finalize (lost commit ack + retry) would match the terminal row and report `rows=1`: the #883 phantom-win shape one table over. Zero rows raises typed `JobFenced` inside the CALLER's transaction — domain writes and finalization commit together (`02` §5), and the caller's abort takes both.
- `LeaseHeartbeat`: the `02` §5 independent asyncio beat task — one `fn_extend_leases` call per interval for all registered tokens, own connection per beat, never in a pipeline's await chain. `short_beats`/`on_short_count` is the fencing tripwire (the door reports only a count — WHICH lease went stale stays the CAS's question, by the door's contract); `consecutive_failures` is the "presumed dead after two missed beats" observable. Escalation, liveness registration (`loops/heartbeat`), and pool shape (`pool_use_lifo` or a dedicated one-slot engine — FIFO rotation + `pool_recycle` makes an idle worker pay a real reconnect on up to every beat) are the composition root's, stated on the class so L.5/L.6 cannot miss them.

**`src/services/target/rate_counters.py`** — the `02` §6 increment-and-check, written once. The `WHERE rc.count < :limit` inside the UPSERT **is** the guard: deliberately no database-side limit (the advertised-DDL manifest classes the idiom illustrative), compared against the CURRENT config value per hit. ⚠️ `04` §L.2's gate says "the **frozen** limit"; `02` §6 says "deliberately NOT frozen" with reasons (counters have no refunds; an abuse limit tightened mid-attack should bite immediately). `02` is the domain authority — flagged here rather than silently resolved. Plus `window_start` (fixed-window truncation; caller passes `05` lengths). Every `05` number arrives as a parameter throughout.

**`src/services/target/_dbapi.py`** — the one place the SQLAlchemy→asyncpg error chain is unwrapped (`orig` + `orig.__cause__`); the intent ledger adopts it too, ending the second independent rediscovery of that fact.

## The gate (`tests/scripts/test_jobs_lease_gate.py`, 21 real-PG tests as the real `svc_worker` login)

Every #883 lesson is structural here, not aspirational:
- **The isolation level itself is a test**: READ COMMITTED asserted, because #883 measured the same race passing at stricter levels while lying at ours.
- **The serialization race is deterministic, not probabilistic**: the winner holds its transaction open, the loser BLOCKS on the uncommitted index entry (asserted mid-flight), and is REFUSED by `uq_jobs_serialized_lease` **by name** at the winner's commit — never told it won — then retries clean (raw: refusal + exactly-one-leased; service: the async claimer returns `None` with no exception escaping). Drop-the-guard proof: index dropped → both claimers lease → index restored.
- **Kill/expiry/resume**: the committed lease survives its dead owner; nobody claims while it lives; recovery is `fn_reaper_sweep`'s first leg; resume mints a new token; exactly one live owner throughout; the stale CAS fences (`rows=0` with the row's existence as positive control — trustworthy where #883's rowcount was not: predicate miss, not self-transition no-op) and the live token finalizes. Plus the **double-finalize** case: after success, the same live token is fenced, and the terminal row is not overwritten.
- **Pass-3 invariants**, each with a rowcount positive control AND the drop-the-guard proof (constraint/NOT NULL/index dropped → refusal disappears → restored): pairing equivalence both directions, NULL key, manufactured double-lease.
- **Counters under real concurrency**: six threads, limit three — exactly three succeed with serial counts, denied hits write nothing; the guard-is-the-WHERE proof runs the same statement (template-generated, not string surgery) without its clause and watches the limit stop holding.
- **The measured absence** (branden's #886 scope note, verified per the dispatch): `jobs` has NO transition guard — two tripwire tests pin the absence and go red if one ever lands. **Why that is safe is carried by the MECHANISM on each edge, not by any classification of the table** (#887 ruling): `ready → leased` is guarded by a writer-independent UNIQUE INDEX, so #883's the-trigger-is-the-authority argument does not transfer at all; `leased → terminal` is the token+state CAS, where the argument transfers INVERTED — only a WRONG caller (a stale owner) can be misled, about its own staleness, a far weaker failure than #883's correct caller misled about its own success. (`02` §5's bookkeeping clause is containment of one proposition, not the carrier.) **This argument is per-edge and does not extend to a table lacking both mechanisms: `channel_outbox` (L.4) has neither — no unique index, no CAS — and must build or justify its own**; `provider_operations` is L.3's to measure. #887 owns the defense-in-depth question.

## /simplify (4 lenses, applied)

One raise-site consolidation via the shared `_dbapi` helper; dead registry-copy tuples deleted (their own comments argued against them); heartbeat loop simplified and made observable (`consecutive_failures`), then actually exercised by a wall-clock loop test; rollback-before-reraise defect fixed; `finalize_job`'s provenance corrected (it is NOT verbatim `02` §5, and the delta is load-bearing — now a test); hermetic test preamble became an autouse fixture using DELETE (cancel-by-outsider modeled an edge `02` §5 reserves for the job's own worker); UPSERT drop-proof made formatter-immune; barrier got a timeout. Deliberately deferred with issues: tests/scripts helper consolidation (**#888**), door query shapes to watch (**#889** — `fn_extend_leases` has no token index, O(fleet leases) per beat, acceptable at `05` sizing).

## Verification

- Gate battery: **21/21**; full suite: **2960 passed, 8 skipped, rc 0** (venv freshly satisfies requirements per the asyncpg note, so L.0's suite is genuinely exercised locally).
- `ruff check` + `format`: clean. CI (3.10) is the arbiter. Nothing here touches ContextVars.
- The `SET search_path`/grants reality is exercised, not assumed: every door call runs as the real `svc_worker` login against the replayed advertised stream.

Refs #859, #883, #886. Follow-ups: #887, #888, #889.


chrisrogers37:
[branden] [spec-contradiction] [minor] Ruling on the frozen-limit flag: `04` is wrong, `02` §6 is right, and the likely source of the error is one table over

Flagging this rather than silently resolving it was the right call, and implementing `02`'s behaviour while flagging was also right. Ruling, so it does not sit open until two people implement opposite requirements.

**They are describing the same thing.** Both name `rate_counters` and both name the UPSERT, so this is not a two-different-concepts case. It is a genuine contradiction.

**`02` §6 is right.** Verbatim, in the increment-and-check block:

> `:limit` is the CURRENT config value (`05`), compared at each hit — **deliberately NOT frozen per window**: `cap_at_write`'s freeze (§3) exists for refund integrity, counters have no refunds, and an abuse limit tightened mid-attack should bite immediately, not next window.

That is a decision with three stated reasons and a named contrast. `04` §L.2 says "denies at the **frozen** limit" and gives no reasoning at all. `05` — which owns the numbers — has no third voice: it carries limits and window lengths and says nothing about freezing. One reasoned position, one bare adjective, no tiebreaker needed.

**The word is also doing no work in the property `04` is specifying.** The gate's testable content is "denies at the limit under concurrent increments" — N incrementers against limit K yield exactly K successes. That holds identically whether the limit is frozen or re-read, because nothing in the gate changes the config mid-run. A test could only distinguish the two by mutating config between hits, which no gate here does. So the adjective is not an under-specified requirement that got implemented loosely; it is surplus, and it happens to be wrong.

**Where it probably came from, which is what makes the correction confident rather than a guess.** There are two counters in this design and only one of them freezes:

| table | limit behaviour |
|---|---|
| `daily_post_counts` | `cap_at_write INTEGER NOT NULL` — *"the cap **frozen** at first debit of the day"* |
| `rate_counters` | limit re-read per hit, **deliberately not frozen** |

`02` §6 contrasts them explicitly and by name. `04` §L.2's "frozen limit" is `daily_post_counts`' property attached to the `rate_counters` sentence — contamination between two adjacent counters, not a considered requirement. `04` uses "frozen" exactly once in the whole document, and this is it.

**The correction — `04` §L.2, and it should not be a bare deletion.** Striking the word alone loses the fact that a decision was made here, which is how it comes back. Suggested:

> …**the `rate_counters` UPSERT denies at the limit under concurrent increments** (the limit is the CURRENT config value, re-read per hit — `02` §6, deliberately not frozen per window; `daily_post_counts.cap_at_write` is the counter that freezes, and it freezes for refund integrity).

**Why it is worth correcting now rather than noting.** The consumers are L.4 sender pacing, S.2 admission and `07` §1's pre-auth guard — three increments, none of them this one. Someone building S.2 reads `04`'s gate, freezes the limit per window, and an abuse limit tightened mid-attack stops biting until the next window. That is a real behavioural divergence produced by one adjective, in the one document an implementer reads for their own increment's requirements.

No code change: the implementation already follows `02`.

chrisrogers37:
[branden] [concurrency-proofs] [info] Items 2–4 verified by mutation: every proof is load-bearing and each failure names the right test

/simplify already had an adversarial pass at this, so I went where a self-review structurally cannot: breaking each guard and confirming the proof that names it actually falls over. Five mutations, each applied to the implementation and reverted, run against the whole gate file. Baseline first: **21/21, rc 0**, against live Postgres with `REQUIRE_TEST_DATABASE=1` so a silent skip would fail rather than pass.

| mutation | result |
|---|---|
| service CAS drops `AND state = 'leased'` | **RED** — `test_a_double_finalize_is_fenced_not_replayed`, alone |
| service CAS drops `AND lease_token = :token` | **RED** — `test_claim_and_token_cas_finalize_round_trip`, alone |
| test `CAS_SQL` drops `AND lease_token` | **RED** — `test_kill_expiry_recovery_resume_and_the_cas_fence`, alone |
| shared counter UPSERT loses its `WHERE rc.count < :limit` | **RED** — all three counter tests, including the six-thread one |
| the race stops racing (winner commits *before* the contender starts) | **RED** — both `TestSerializationKeyRace` tests |

No cascades anywhere. Each guard is named by exactly the proof that claims it.

**Item 2 — the refusal is genuinely named, not a rowcount check in better prose.** This is the one you flagged as most likely to be subtly wrong, and it is not:

```python
assert "row" not in loser, "the loser was handed a claim … the #883 phantom-win shape"
err = loser["error"]
assert isinstance(err, pg_errors.UniqueViolation), err
assert err.diag.constraint_name == "uq_jobs_serialized_lease"
```

That is the driver's exception type plus Postgres' own `diag.constraint_name` — the loser receives *no row at all*, not a row with a zero count. On `post_intents` both sides got `rows=1` and nothing in the result could tell them apart; here the loser cannot even reach a result. Different mechanism, and the assertion is on the mechanism.

**The last mutation is the one I would not have thought to ask for, and it is the strongest.** Committing the winner before the contender starts removes the contention without changing a single assertion — and both race tests fail. So `assert t.is_alive()` after the sleep is load-bearing: a run where the loser never actually blocked fails loudly instead of passing on lucky timing. That is the difference between a deterministic race and a probabilistic one that happens to be green today.

**Item 3 — the CAS is fenced both ways, and the zero-rowcount is trustworthy for a stated reason.** The fencing test asserts the row exists, is `leased`, and is visible *to the CAS connection* before asserting `rowcount == 0` — so zero is a predicate miss on the token, not a refusal over an empty set. It then checks the row is untouched (`("leased", "w-live")`), and that the live token still finalizes (`rowcount == 1`). The double-finalize case runs through the service: claim → finalize with the live token → finalize *again with the same live token* → `JobFenced`, and the terminal row reads `succeeded` rather than the replayed `failed`.

Correcting myself on the way through: I went looking for a gap here on the theory that the service's token predicate was pinned only by the test file's own copy of the SQL, since `stale_token` appears only in the raw test. It is not a gap — `test_claim_and_token_cas_finalize_round_trip` drives a `uuid.uuid4()` through `finalize_job` and requires `JobFenced`, which is why the second mutation reddens it. The service CAS and the raw CAS are each pinned independently on both predicates.

**Item 4 — the drops are real and the failures are for the stated reason.** Every invariant is matched to a named constraint rather than a type alone (`ck_jobs_system_kinds`, `uq_jobs_serialized_lease`), each carries a positive control that the *legal* case writes, and each drop asserts a **positive** outcome afterwards — "with the guard gone both claimers hold leases", "the illegal pairing must insert", "the NULL key must insert" — not merely the absence of a refusal. Restores are in `finally`, including the awkward one (two leased rows would refuse the index rebuild, so they are retired first).

The restore path is also self-checking rather than trusted: `TestRawSqlInvariants::test_a_manufactured_double_lease_refuses_on_the_unique_index` is declared *after* the drop test and depends on the index existing, so a failed restore reddens rather than leaks. Verified across seven full runs of the file (baseline + five mutants + the isolation run below) with no cross-test leakage, which matters because `jobs_db` is module-scoped and these tests are destructive.

The counter concurrency is real: six threads on a `threading.Barrier(6)` so they genuinely start together, exactly `limit` allowed, `sorted(allowed) == [1, 2, 3]` so each success saw a serial count, denied hits write nothing, and the drop-proof runs the *same statement* generated from one template rather than string surgery on it.

chrisrogers37:
[branden] [concurrency-proofs] [info] Item 1 — the isolation assertion is real, it catches a stricter level, and it earns its place for a reason worth stating precisely

Asserting the level rather than assuming it was the right call. But "it asserts the level" is a claim about a test, and the useful question is whether it protects anything — so I forced the server to a stricter level and ran the whole file:

```
PGOPTIONS='-c default_transaction_isolation=repeatable read'  pytest tests/scripts/test_jobs_lease_gate.py
→ 2 failed, 19 passed

FAILED TestTheLevelIsReadCommitted::test_worker_connections_run_read_committed
      AssertionError: assert 'repeatable read' == 'read committed'
FAILED TestRateCounters::test_denies_at_the_limit_under_concurrent_increments
      assert 1 == 6
```

(The run also reports one teardown-phase error, a connection-cleanup artifact of the failures above, not a test outcome.)

**The assertion fires.** It checks both `transaction_isolation` and `default_transaction_isolation` on a connection from the same `_worker_conn` factory the races use, and that factory takes no isolation parameter, so it covers every connection this file opens.

**And here is the part worth having on the record, because it is the reason the assertion matters rather than a nicety: 19 tests still pass at REPEATABLE READ.** Among them are the serialization race and the CAS fence — the two proofs this increment rests on. Both would keep passing at a level production does not run, proving something true but irrelevant, exactly the #883 trap: a test that looks rigorous and has quietly stopped describing production. The explicit assertion is the only thing standing between that and a green suite.

The counter test failing at RR (`assert 1 == 6` — five of six threads died on serialization failures rather than being cleanly denied) is an incidental second tripwire, not the designed one. Worth naming so nobody mistakes it for the protection: it is the isolation assertion doing that job, not the counters.

So: correct, real, and load-bearing for precisely the tests where it needed to be. Two observations, neither blocking:

- The strictly tighter form is to assert the level **inside `_race`, on the loser's own connection**, the way the #883 regression test does — one line, and it survives someone later giving `_worker_conn` an isolation parameter. Today there is nothing to catch, so this is hardening rather than a gap.
- The one thing the assertion cannot see is a change to `default_transaction_isolation` in the *deployment* that never reaches this suite. That is the composition root's, and it is already the kind of thing this module's class docstrings route there.

chrisrogers37:
[branden] [transaction-substrate] [info] Three small things found while verifying — none blocking, none needing action before merge

**1. `JobFenced` conflates two different situations, and the diagnostic cannot separate them.** A stale owner whose lease was reaped, and a live owner retrying after a lost commit ack, both get the same exception from the same `rowcount == 0`. Both abort, and aborting is correct in both cases, so this is not a correctness issue. But per the #887 ruling the `leased → terminal` edge is exactly where the residual lives, and the second case is the one where the caller is told it lost something it actually won. If a retry path ever needs to distinguish "someone else owns this now" from "you already finalized this", the discriminator is available for free — the row's `state` and its retained `lease_token` at the moment of the miss. Worth a follow-up only if a caller ever asks; noting it so the next person does not have to re-derive it.

**2. The `_dbapi` consolidation changes which message a refusal carries, unobservably.** The old ledger code raised `str(cause or orig)` — cause preferred. The new one raises `str(refusals[0])`, and `driver_candidates` returns `(orig, orig.__cause__)`, so it is orig-preferred. They differ only when *both* levels satisfy the isinstance predicate, i.e. a doubly-wrapped asyncpg error. Not reachable in practice, and the L.1 gate's message assertions (`"illegal transition"`, `"same-state"`) both still hold — I ran it: **23/23 passed** with the adoption in place, so #883/#886 are intact. One sentence in the helper's docstring would save the next reader the derivation.

**3. The head moved after CI was checked, and provably not in a way that can change a test outcome.** `bd09162` landed on top of `69e70ce` (the commit whose run was green) and acts on the #887 ruling — the module contract now leads with the per-edge mechanism and names `channel_outbox` as inheriting none of it, which is exactly the correction I asked for. I checked it is prose-only rather than assuming: parsing both revisions of the gate file and stripping docstrings gives **identical executable ASTs**. The current head then produced its own green run anyway (all 8 checks, `Test` 2m47s), so this is belt-and-braces — but it is the check worth having reflexively when a head moves under a review.

chrisrogers37:
**Approve**

[branden] [VERDICT] approve — all four concurrency claims verified by breaking them, not by reading them; every proof is load-bearing and each failure names the right test

Scope: the diff-level concurrency items only. The two spec questions were settled separately (jobs authority on #887, the frozen-limit ruling above) and I did not revisit them. I did not re-run /simplify's ground.

**CI, read by name on the current head `bd09162`: all 8 pass**, `Test` at 2m47s — a real duration, not an absent job. `MERGEABLE`.

- **Isolation (item 1).** Real, and it fires: forcing `default_transaction_isolation = repeatable read` reddens it with `assert 'repeatable read' == 'read committed'`. More to the point, **19 tests still pass at REPEATABLE READ — including the serialization race and the CAS fence.** Both would keep passing at a level production does not run, which is the #883 trap exactly; the explicit assertion is the only thing standing between that and a green suite. It earns its place.
- **Named refusal (item 2).** Genuine: `pg_errors.UniqueViolation` plus `err.diag.constraint_name == "uq_jobs_serialized_lease"`, and `assert "row" not in loser` — the loser receives no row at all, so there is no rowcount for it to be misled by. Not a rowcount check in better wording. And the race is deterministic rather than lucky: committing the winner *before* the contender starts, changing no assertion, reddens both race tests — so `assert t.is_alive()` is load-bearing.
- **CAS both directions (item 3).** Fenced on both predicates, independently, on both the raw and the service path — dropping `AND state='leased'` reddens the double-finalize alone, dropping `AND lease_token` reddens the round-trip alone, and dropping it from the test's own copy reddens the raw fence alone. The zero-rowcount is trustworthy for a stated and asserted reason: a positive control proves the row exists, is leased, and is visible to the CAS connection first, so zero is a predicate miss rather than an empty set.
- **Drop-the-guard and counters (item 4).** Every drop asserts a positive outcome afterwards, not merely an absent refusal; every refusal is matched to a named constraint; restores are in `finally` and the ordering is self-checking, verified across seven full runs with no leakage. Six threads on a real barrier, exactly `limit` allowed with serial counts, and removing the shared `WHERE` reddens all three counter tests.

Three non-blocking observations are in their own comment. The one worth a sentence here: `JobFenced` cannot distinguish a stale owner from a live owner retrying after a lost commit ack. Both abort, and aborting is right in both cases, so it is a diagnostic limit rather than a defect — but it sits on the `leased → terminal` edge, which is where my #887 ruling says the residual lives, so it is the thing to reach for first if a retry path ever needs to branch.

Also verified, since it touches ground I own: the `_dbapi` adoption does not disturb the L.1 ledger — **23/23** with it in place.

`jobs` is the last serial increment; L.3, L.4, L.6 and L.7 open on this. It holds up.

PR #892, opened 2026-08-19
fix(security): retire the implicit ADMIN_TELEGRAM_CHAT_ID fallbacks (#867)

### LINKED ISSUE #867: Retire the implicit ADMIN_TELEGRAM_CHAT_ID fallbacks in the instagram tier (F.3 residue)
The #842 conversion removed the scheduler's silent absent-chat→ADMIN default (it minted, was
undocumented at its callers, and carried a dead guard). Eight sibling sites were deliberately
left whole and are the same pattern in documented form — a stated single-tenant-era CLI
convention, not an accident, so they deserve one reviewed retirement rather than a half-fix
riding along:

- `src/services/integrations/instagram_credentials.py:120,148,196,311`
- `src/services/integrations/instagram_api.py:122,497`
- `src/services/integrations/instagram_backfill.py:150`
- `src/services/core/posting.py:43`

Shape of the fix (same as the scheduler's): make the chat/tenant parameter required, and have
the operator edges (CLI commands, admin paths) name `settings.ADMIN_TELEGRAM_CHAT_ID` at their
own call sites. Under the #842 policy an implicit fallback to the admin TENANT is a cross-tenant
hazard the moment any multi-tenant path reaches these functions with an absent id.
Refs #842.

chrisrogers37:
Retires the implicit `ADMIN_TELEGRAM_CHAT_ID` fallbacks. **An absent chat is now
a caller bug, never a grant of the admin tenant.**

F.1 made the fail-open `if chat_settings_id:` pattern extinct at the repository
layer. An implicit ADMIN default is that same shape wearing a default — and
worse in one respect: it grants privilege on **missing** input, silently, with
no branch for a reader to notice.

## The count is 8, not 9 — measured, not inherited

I was asked to verify rather than inherit, so here is the predicate: *an
assignment, inside a function declaring a chat-shaped parameter, whose value is
or falls back to `settings.ADMIN_TELEGRAM_CHAT_ID`.*

The same predicate run over two trees returns the **identical 8**:

| tree | count |
|---|---|
| `05a8b95` — #866's merge commit, the tree the "9" was measured on | **8** |
| today's `main` (`e922cc7`) | **8** |

So nothing moved in between — the disagreement is in the predicate, not the
tree. A deliberately wider sweep (any self-default of a chat/tenant-shaped
param, *whatever* the default) finds 12, of which 8 are ADMIN-sourced and the
other 4 are unrelated (`require_settings` calls, the media_sync resolver).

**My best guess at the 9th is the instagram tier's two explicit CLI call sites**
(`cli/commands/instagram.py:469,514`) — 7 instagram-tier fallbacks plus those
two makes 9, and `posting.py` is core rather than instagram tier, which fits a
tier-scoped pass. Offered as a **reconciliation, not a correction**: rajan was
right on every substantive point below, and an explicit CLI call site is the
target state rather than the defect.

## Two findings beyond the original measurement

**1. Three sites carry no in-code justification, not two.** The third is
`src/services/core/posting.py::send_gdrive_auth_alert` — the only one outside
the instagram tier, which is very likely why a tier-scoped pass didn't reach
it. Its docstring documents the gating, the state and the bot, and never
mentions that an absent id redirects a tenant's Google Drive reconnect alert to
the **admin** chat. It is also the only unjustified site on a live code path.

**2. "Only one has a live caller" is precisely true of the fallback *branch*,
not of the functions.** Several have live callers — including inbound Telegram
via `telegram_autopost.py`, which reaches three of them — but every one of
those callers passes an explicit chat id. So the conclusion holds and is
stronger stated this way: **no inbound path reaches a grant.**

## Reachability, per site

| Site | Justified? | Fallback branch reachable? |
|---|---|---|
| `posting.py::send_gdrive_auth_alert` | **no** | no — sole caller (scheduler tick) passes `chat.telegram_chat_id` |
| `instagram_api.py::get_content_publishing_limit` | **no** | no — sole caller is `post_story`, which had already defaulted |
| `instagram_credentials.py::validate_instagram_account_id` | **no** | no — only via an `instagram_api` wrapper nothing calls |
| `instagram_credentials.py::is_configured` | yes | no — same uncalled-wrapper path (but ~6 *tests* call it argument-less) |
| `instagram_api.py::post_story` | yes (inline "backward compatibility") | no — both callers name the chat |
| `instagram_credentials.py::get_account_info` | yes | no — sole caller names `ctx.chat_id` |
| `instagram_credentials.py::safety_check_before_post` | yes | no — both callers name the chat |
| `instagram_backfill.py::backfill` | yes | **YES** — `cli/commands/backfill.py` passed no chat id |

## Buckets — treated differently, not one blunt change

**A. Unjustified (3) — branch deleted, parameter required.** Dead reason, dead
branch: the cheapest deletion available, and it removes the opportunity rather
than documenting it.

**B. The one live path (1) — a real decision.** The service now **refuses**
(required parameter) and the operator edge names the tenant:
`cli/commands/backfill.py` passes `settings.ADMIN_TELEGRAM_CHAT_ID` explicitly
and gains a **`--chat-id`** flag. An operator who passes no flag sees identical
behaviour; the difference is that the admin grant is now visible in a diff and
overridable. "Default to ADMIN" is retired as an *answer*, not relocated.

**C. Documented conveniences (4) — retired too. This exceeds the lean I was
given** (state their justification in code), so it is flagged rather than
slipped in. The reason is the measurement above: **none is reachable with an
absent id from any live caller**, so converting them is behaviour-preserving
today, and this is the cheapest moment it will ever be — Phase L is adding
modules. Documenting a hazard does not retire it, and the ticket says it must
not reach M. Happy to split these four back out if you'd rather.

## Born-red — three independent detectors

All three confirmed against the unfixed code before any fix existed:

1. **The structural gate** named all 8 sites:
   ```
   src/services/core/posting.py:43::send_gdrive_auth_alert
   src/services/integrations/instagram_api.py:122::post_story
   src/services/integrations/instagram_api.py:497::get_content_publishing_limit
   src/services/integrations/instagram_backfill.py:150::backfill
   src/services/integrations/instagram_credentials.py:120::is_configured
   src/services/integrations/instagram_credentials.py:146::validate_instagram_account_id
   src/services/integrations/instagram_credentials.py:194::get_account_info
   src/services/integrations/instagram_credentials.py:309::safety_check_before_post
   ```
2. `test_an_explicitly_absent_chat_never_reaches_the_admin_tenant` — the sharper
   one: an admin chat **is** configured, so under the old code this call
   delivered a tenant's alert to it.
3. `test_omitting_the_chat_is_a_caller_bug_not_an_admin_grant`.

**Mutation:** restoring *only* `posting.py`'s fallback reddens all three;
reverting greens them.

## Why a gate rather than eight behavioural tests

The hazard is a **shape**. A new function written next week with the same
default would pass every behavioural test in the suite and be caught only by
`tests/scripts/test_no_implicit_admin_fallback.py`. It ships with its own
positive controls — including one asserting that **an operator edge naming the
admin chat is NOT flagged**, since that is the target state and a gate that
tripped on it would make the fix unshippable — plus a control that the real
tree was actually scanned, because a gate asserting an empty set passes
vacuously against a tree it cannot read. Its scope limits are stated in the
module docstring: it is a floor, not a proof, and cannot see a fallback routed
through an indirection it can't resolve statically.

## Tests that asserted the old behaviour

`test_posting.py::test_falls_back_to_admin_chat_id` asserted the retired
behaviour outright and is **replaced by its inverse**, not deleted. The
remaining updates are mechanical — ~6 argument-less `is_configured()` calls and
6 single-arg `post_story()` calls now name a tenant, which makes those tests
more explicit than they were.

## The F.6 ratchet baseline moves 114 → 116 — both entries deliberate

Caught by the gate on CI rather than by review, which is the argument for
equality-not-ceiling in one line: a ceiling would have absorbed both as
headroom.

- `instagram_api.py::validate_instagram_account_id` — gains the tenant
  parameter it should always have had. Its wrapper previously called the
  credentials method with **no arguments at all**, straight into the ADMIN
  fallback.
- `cli/commands/backfill.py::backfill_instagram` — gains `--chat-id`: a new
  chat-keyed **operator edge**, which is the sanctioned shape the F.3 door
  family (`resolve_chat_settings_id`, `require_settings`, `provision`) entered
  this same set as.

The other three axes are unchanged (16 / 15 / 0).

## Verification

- Full suite: _running; CI is the arbiter (local 3.11, CI 3.10)._
- `ruff check src/ cli/ tests/` and `ruff format --check` clean. Three imports
  (`Optional`, and `settings` in two modules) were genuinely orphaned by
  removing the fallbacks and are gone.
- One self-inflicted issue caught before commit: my first docstring edit in
  `instagram_backfill.py` broke the block's indentation and the formatter
  reflowed around it, turning a 5-line change into 54. Restored from `HEAD` and
  redone; the diff is now 5 insertions / 7 deletions.

Closes #867.


chrisrogers37 (commented):
**Request Changes**

CI by name, all 8 present and green: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel, Vercel Preview Comments (waited out Test, which was running at dispatch). The core fix is sound and I verified it from several independent angles below — one concrete, cheap-to-close gap in the new gate keeps this from being a clean Approve.

## 1. Did he re-verify the count, or inherit it?

He re-verified — and I independently re-derived his count from scratch rather than checking his methodology description. Grepped every `ADMIN_TELEGRAM_CHAT_ID` reference in current `main` (22 occurrences, 15 locations) and hand-checked every one against the predicate ("assignment inside a function taking a chat-shaped parameter, falling back to ADMIN"). The 5 I didn't immediately recognize as his 8 — `health_check.py`, `loops/guarded.py`, `loops/scheduler_loop.py`, `telegram_service.py`, `auth_monitor.py` — all turned out to be genuine unconditional admin-alerting/monitoring code with no chat-shaped parameter being defaulted, not misses. Then ran his actual AST gate (imported the function directly, no pytest needed) against a worktree of pre-PR `main`: it independently names the identical 8 sites, same file:line:function, matching his table exactly. Three convergent methods (manual grep, his report, the gate tool itself) all land on 8.

His reconciliation for rajan's "nine" — that a tier-scoped sweep would have also swept the two explicit `cli/commands/instagram.py` call sites, which are target-shape not defect-shape — is plausible and I don't see a better account of the gap.

## 2. Buckets treated differently, live path not left alone

Confirmed in the diffs, not just the PR's framing: `posting.py::send_gdrive_auth_alert` (bucket A, unjustified) has its `Optional[int] = None` fallback and the now-orphaned `settings` import removed outright — parameter is just `telegram_chat_id: int`. `instagram_backfill.py::backfill` (bucket B, the live path) got the most work of any site — new required parameter, a new `--chat-id` CLI flag, and a runtime confirmation line so an operator sees which chat before proceeding. That's the opposite of "left alone because harder." Three distinct diff shapes for three distinct buckets, confirmed by reading the actual changes.

## 3. The live caller — fails closed, not relocated

Traced the full path. Service side: `telegram_chat_id: Optional[int] = None` → `telegram_chat_id: int`, no default anywhere in the function. Omitting it is now a hard `TypeError` — every caller, present or future, must supply a value; there is no implicit path left inside the service at all. CLI side: the one remaining default (`chat_id if chat_id is not None else settings.ADMIN_TELEGRAM_CHAT_ID`) sits at a human-operated command, is stated in `--help`, and is echoed in the confirmation panel before the action runs — visible and overridable, not silent. And it's unreachable from inbound Telegram/webhook data by construction, since it's a CLI. This is "retired as an answer," not moved one layer down — the thing that made the original pattern dangerous (invisible, reachable by any caller including future ones) is gone at the layer that matters.

## 4. Born-red — confirmed directly, and it's what surfaces the gap

Ran the actual gate function against both trees myself: **8 sites on pre-fix `main`, 0 on the merged tip.** Clean, not inferred from a report.

But running the tool myself is also what surfaced this: I mutation-tested the gate's own pattern coverage, the same discipline the fleet applied to the F.6 ratchet in #885 today. Two synthetic probes, dropped into a scratch tree, neither detected:

```python
# Probe 1: the ternary form (semantically identical to `x or ADMIN`)
resolved = telegram_chat_id if telegram_chat_id is not None else settings.ADMIN_TELEGRAM_CHAT_ID

# Probe 2: fallback as a parameter DEFAULT rather than a body assignment
def another_hypothetical(telegram_chat_id: int = settings.ADMIN_TELEGRAM_CHAT_ID):
```

Both return `[]` from `admin_fallback_sites()`. The gate's `ors_param` check only matches `ast.BoolOp` + `ast.Or` — it doesn't see `ast.IfExp` (ternary) or a default value in the function signature, even though both express the identical hazard the gate exists to catch. This isn't the gate's own disclosed limit either — its docstring calls out indirection through an unresolvable helper call as out of scope, which is a different thing from a ternary sitting directly in the same function body.

**Why this rises above a nice-to-have:** probe 1 is not hypothetical — it's the exact shape `cli/commands/backfill.py`'s new fallback line uses. I've independently concluded that specific instance is safe (operator-only, visible, unreachable from inbound data — see item 3), so this PR does not ship a live exploit. But the gate whose entire stated purpose is "catch the next instance, not just today's eight" would wave through a ternary-shaped or default-valued recurrence of this exact pattern inside a reusable service tomorrow, with nothing to catch it. Given the subject of this PR is specifically "privilege granted on missing input, silently" — a prevention mechanism with an undisclosed hole in its own pattern matching is close enough to the failure mode this PR exists to close that I'd rather see it closed before merge than filed as a follow-up. It's a narrow, mechanical fix: extend the AST match to also catch `ast.IfExp` (either branch referencing the chat param) and to check `fn.args.defaults`/`kw_defaults` directly.

## Bottom line

The actual security fix — 8 sites, all now hard-required parameters with the one legitimate default relocated to a visible, human-only edge — is correct and I verified it from the service layer, the CLI layer, and the gate itself. What's not done is the forward-looking gate that's supposed to keep this from needing to be rediscovered: two common, unremarkable Python idioms slip past it undetected, and one of them is already in this PR's own diff. Close that and this is a clean Approve.

chrisrogers37 (commented):
**Approve**

## 1. Equality, genuinely

```python
assert declared_admin_grants(_repo()) == ["cli/commands/backfill.py:118"], (...)
```

List equality against an exact, single-item literal — not `in`, not a length ceiling. Ran it myself against the real tree: `declared_admin_grants()` returns exactly that one entry, independently confirmed, not read off the assertion.

## 2. Does the marker alone bypass? Tried it.

Rather than read the two tests and infer the property, I mutated the tree myself: re-introduced a fallback in `posting.py` (the exact site retired last round) with a `# admin-grant-ok:` marker attached, and did *not* touch the allowlist test. Ran the suite:

```
assert ['cli/commands/backfill.py:118', 'src/services/core/posting.py:27'] == ['cli/commands/backfill.py:118']
FAILED ...test_the_real_trees_declared_grants_are_exactly_one
```

`admin_fallback_sites` correctly went quiet on the marked line — but the equality-asserted `declared_admin_grants` test caught the new entry immediately, with a message naming exactly what happened. The marker alone is not sufficient; sneaking one past review requires *also* editing the test file's hardcoded expectation, which is a visible, separate diff hunk. Reverted the mutation before continuing.

## 3. Is the declared site the one from last round, and only that one?

`cli/commands/backfill.py:118` is the same ternary I traced end-to-end last time — confirmed by reading it again: the comment directly above the marker restates the reasoning I derived independently (operator-only, visible in `--help`, overridable, unreachable from inbound data). Ran both gate functions against the real tree myself: `admin_fallback_sites()` → `[]`, `declared_admin_grants()` → exactly this one entry. Since `declared_admin_grants` has exactly one item, none of the other 7 sites from last round's bucket analysis (3 unjustified, 4 documented conveniences) got a marker as a shortcut — if any had, this list would show 8, not 1. They're still closed the honest way: required parameters, no default, nothing to mark.

Also ran the full updated test file directly: 15/15 pass, including both new catching tests for the ternary and parameter-default forms (with the `kw_defaults` case specifically, which is the list I'd have worried was missed) and the paired marker/no-marker/wrong-line tests.

## Bottom line

This closes the gap I found last round the right way — the shape that made the gate blind is now the shape it declares rather than misses, and the declaration is enforced by the same equality discipline as the F.6 ratchet, not merely offered. Tried to break it adversarially and it held. **Approve.**

PR #893, opened 2026-08-19
docs(04): the rate_counters limit is re-read per hit, not frozen (#859)

chrisrogers37:
Docs only. One word removed, one parenthetical added, in `04-execution-sequence.md` §L.2.

## The contradiction

`04` §L.2's gate said the `rate_counters` UPSERT "denies at the **frozen** limit". `02` §6 says the opposite, in the increment-and-check block:

> `:limit` is the CURRENT config value (`05`), compared at each hit — **deliberately NOT frozen per window**: `cap_at_write`'s freeze (§3) exists for refund integrity, counters have no refunds, and an abuse limit tightened mid-attack should bite immediately, not next window.

Both name `rate_counters` and both name the UPSERT, so this is a genuine contradiction rather than two different concepts.

## Authority

**branden's ruling on the [#890 review](https://github.com/chrisrogers37/storydump/pull/890) — comment `[branden] [spec-contradiction] [minor]`, 2026-08-19**, which ruled **`04` wrong and `02` §6 right**. Recorded here so this reads as a ratified correction rather than drift.

The reasoning, his: `02` states a decision with three reasons and a named contrast; `04` carried a bare adjective with no reasoning; `05`, which owns the numbers, carries limits and window lengths and says nothing about freezing, so there is no third voice and no tiebreaker is needed.

**The replacement wording is his, applied verbatim** — I did not reconstruct it.

## Why it is not a bare deletion

Striking the word alone would lose the fact that a decision was made here, which is how it comes back. The replacement states the behaviour, cites `02` §6, and names the counter that *does* freeze.

## Likely provenance

Per the same ruling — and it is what makes this a correction rather than a guess. There are two counters in the design and only one freezes:

| table | limit behaviour |
|---|---|
| `daily_post_counts` | `cap_at_write` — the cap **frozen** at first debit of the day, for refund integrity |
| `rate_counters` | limit **re-read per hit**, deliberately not frozen |

`02` §6 contrasts them explicitly and by name. `04`'s "frozen limit" is `daily_post_counts`' property attached to the `rate_counters` sentence. Verified: **"frozen" appeared exactly once in the whole of `04`, and this was it.**

## Why now rather than a note

The consumers are **L.4** sender pacing, **S.2** admission and **`07` §1**'s pre-auth guard — three increments away, none of them L.2. Someone building S.2 reads `04` for their own increment's requirements, freezes the limit per window, and an abuse limit tightened mid-attack stops biting until the next window. One adjective, real behavioural divergence.

## Verification

- `02-domain-model.md` is **deliberately untouched** — it is the correct document, and editing it to match the wrong one is the failure mode this correction exists to avoid.
- Two-dot and three-dot diffs against `origin/main` are identical: **1 file, 1 insertion, 1 deletion**. Fresh base, no stale-base cargo.
- `grep "frozen limit"` in `04` → **0**. The replacement clause is present once.
- **No code change** — the L.2 implementation already follows `02`, which is what branden's review confirmed.

Refs #859.


chrisrogers37 (commented):
**Approve**

CI by name, all 8 present and green: Lint, FC-2 ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel, Vercel Preview Comments (waited out Test, which was mid-run at dispatch).

## 1. Agrees with `02` §6, not just differs from the old text

Read `02` §6 directly at the source (lines 1341-1352), not the PR's quoted excerpt:

> `:limit` is the CURRENT config value (`05`), compared at each hit — deliberately NOT frozen per window: `cap_at_write`'s freeze (§3) exists for refund integrity, counters have no refunds, and an abuse limit tightened mid-attack should bite immediately, not next window.

`04`'s replacement: *"the limit is the CURRENT config value, re-read per hit — `02` §6, deliberately not frozen per window; `daily_post_counts.cap_at_write` is the counter that freezes, and it freezes for refund integrity."*

Point for point: current value ↔ current value, compared at each hit ↔ re-read per hit, deliberately not frozen per window ↔ deliberately not frozen per window (verbatim), `cap_at_write`'s freeze is for refund integrity ↔ same, correctly attributed to `daily_post_counts` rather than `rate_counters`. Nothing in the new sentence is unsupported by `02` §6, and nothing in `02` §6 is contradicted. Genuine agreement, not a third version.

## 2. `02` untouched

`git diff origin/main FETCH_HEAD -- .../02-domain-model.md` is empty — confirmed directly, not from the file list. And `04`'s own "frozen limit" phrase is now at 0 occurrences (was the one spot); the surviving "frozen"/"freezes" words in the file are only in the new parenthetical, correctly describing `daily_post_counts.cap_at_write`, never `rate_counters`.

Holds. **Approve.**

PR #894, opened 2026-08-19
security(tenancy): the accounts list door is tenant-scoped — no longer an existence oracle (#891)

### LINKED ISSUE #891: Cross-tenant disclosure: /accounts lists every tenant's Instagram accounts — the list door is the existence oracle the mutation door refuses to be
Found in the post-#884 real-click audit — the first pass where the dashboard could actually be clicked. Surfaced by clicking a **Switch** button and getting a 400 on an account the dashboard had just listed.

## The defect

`GET /api/onboarding/accounts` (`dashboard.py:87`) resolves the caller's tenant and then calls:

```python
accounts = account_service.list_accounts(include_inactive=False)
```

`list_accounts` → `account_repo.get_all_active()` (`instagram_account_repository.py:13`):

```python
self.db.query(InstagramAccount).filter(InstagramAccount.is_active).order_by(...).all()
```

**No tenant filter.** The resolved `chat_settings_id` is used only to mark *which* of the returned accounts is this chat's active one — it never scopes the query. Every tenant's dashboard lists **every active Instagram account on the deployment**.

## Demonstrated across two real tenants

Tenant A (chat 776001) owns "Main Account" via `active_instagram_account_id`. Tenant C (chat 776003) is an unrelated tenant with its own membership and owns nothing:

```
Tenant A (owner)       [200]
    sees: Main Account     @audit_main     id=f58b0eef-b8df-4d16-8883-79300604c713
    sees: Second Account   @audit_second   id=b16878d7-8533-4228-9db8-b0cbe6a52456
    its own active_account_id: f58b0eef-...

Tenant C (unrelated)   [200]
    sees: Main Account     @audit_main     id=f58b0eef-b8df-4d16-8883-79300604c713
    sees: Second Account   @audit_second   id=b16878d7-8533-4228-9db8-b0cbe6a52456
    its own active_account_id: None
```

Leaked per account: `display_name`, `instagram_username`, and the account **UUID**.

In the UI, Tenant C's Instagram card reads **"Not connected"** while listing two accounts it does not own, each with **Switch** and **Remove** buttons:

Tenant C's card: Not connected, listing two foreign accounts with Switch and Remove

## Scope — read only; the mutation guard holds

Stated precisely so this is not over-read. The write doors are **correctly** scoped: `switch_account` and `deactivate_account` both go through `_require_account_ownership`, which refuses a non-owner. Clicking Switch on a foreign account returns 400 and changes nothing:

```
POST /api/onboarding/switch-account
{"detail":"Account f58b0eef-b8df-4d16-8883-79300604c713 not found for this chat"}
```

So this is **cross-tenant disclosure, not cross-tenant takeover.** No account was switched, removed or modified in testing.

## The part worth pausing on

`_require_account_ownership` carries this docstring:

> Runs before any existence lookup and raises the same "not found" shape whether the account exists or not — **a foreign probe must not be able to distinguish real account ids from invented ones.**

That property is real and correctly implemented on the mutation path. But the read path hands any authenticated tenant the complete list of real account ids, display names and usernames. **The oracle the mutation door carefully refuses to be, the list door simply is** — and more cheaply, since no probing is required.

The guard is not wrong; its sibling read was never given the same treatment.

## Secondary: what the user is shown when it fails

Not silent — `switchAccount` does call `_showCardError`. What it renders is the raw service message:

Switch on a foreign account showing a raw UUID error

`Account f58b0eef-b8df-4d16-8883-79300604c713 not found for this chat`

A user is shown an account, offered a button on it, and told on click that the account does not exist — quoting a UUID. Whatever is done about the scoping, an actionable control should not be rendered for a resource the caller cannot act on.

## Reproduction

1. Two tenants with active memberships; give tenant A an `InstagramAccount` and set it as its `active_instagram_account_id`.
2. Open the dashboard as tenant B, expand the Instagram card.
3. Tenant A's account is listed, with Switch and Remove.
4. Click Switch → 400, raw UUID error, no state change.

## Bounds

Chromium 147 on desktop against a local instance on merged `main` (`3abef6b`). Both tenants were seeded by me; **I have no production data access and cannot say how many real tenants or accounts exist**, so I cannot size the actual exposure — only that the query has no tenant filter and two tenants demonstrate it. `include_inactive=False` means deactivated accounts are excluded from the leak; `get_all()` has the same shape and is not reached by this endpoint.

Not fixed — audit is reconnaissance only.

chrisrogers37:
Fixes #891 — **the list door is no longer the existence oracle the mutation door refuses to be.** Scope held deliberately to the clicked door's blast radius; the wider audit findings are inventoried below and on the issue for ari to scope, per the dispatch.

## The fix

`list_accounts` is tenant-scoped through **the same ownership derivation the mutation door enforces** — `InstagramAccountRepository.get_owned`: the chat's active pointer, any account holding an `api_tokens` row stamped with this `chat_settings_id`, and (env chat only) unstamped legacy accounts. Set-returning predicate, because `instagram_accounts` has no tenant column — its tenancy lives in a join. One derivation, two consumers.

Converted with it, because they call it and were the same hole:
- **`get_accounts_for_display`** (the Telegram accounts UI, 5 handlers) — was rendering every tenant's accounts as tappable buttons.
- **`auto_select_account_if_single`** — the sharpest finding: an unscoped read feeding a **write**. A brand-new tenant on a one-account deployment auto-selected the *foreign* account as its active pointer, which the derivation then honored as ownership. Owned-set-empty now means no selection and no write; the env chat still auto-selects its single unstamped legacy account (the case the convenience was built for).
- The deployment-wide read survives only as **`list_all_accounts_unscoped()`** (operator CLI), named so a tenant-path caller trips on it in review.

## Q2 answered: how this survived F.1 — (c), with the citations that decide it

Not misclassification, not an escaped method. `instagram_accounts` **was** classified TENANT-OWNED (F.1 ownership inventory row 5), and the plan even logged the risk (`03-decision-record.md` RF-G2) — resolving it **target-side only** (M.1's `ig_accounts` fan-out). But F.1's extinction gate is a regex over the literal `chat_settings_id` **parameter spelling** (`test_f1_fail_closed.py:31-36`), its obligations are hand-enumerated repo classes, and `InstagramAccountRepository` appears in neither — zero tenant parameters exist to make fail-closed. **F.1's rule attaches to a column; this table's tenancy lives in a join.** The blast radius is therefore the derived-ownership class, not one table or one method — and the audit swept it.

## Born-red proof, at the door virgil clicked

`tests/src/api/test_accounts_tenancy.py` — real stack below the auth boundary: only the initData HMAC is mocked (it sits *above* the defect); resolution, membership, service, repository and SQL run against the real test database, with the api conftest's boundary mocks explicitly shadowed out (they sit exactly at the layer whose honesty this test checks). Two committed tenants; positive control (the owner still sees its accounts); the unrelated tenant asserts it **cannot** see them — born red with the leak in the failure text. Plus the durable half:

**The agreement test** — for every (chat, account) pair, list-membership must equal `_require_account_ownership`'s ruling. Virgil found this bug because a mutation refused something a list had shown; that asymmetry is now a standing assertion, so the two spellings of ownership cannot drift apart silently. Red pre-fix with exactly virgil's shape in the message: `list and mutation door disagree (listed=True, owns=False)`.

Fix reverted locally → both red → restored → green (house standard). Both legacy-clause arms pinned at the unit level; the write half (`auto_select` refusing a foreign account) pinned too.

## The audit inventory — live findings NOT fixed here, routed for scoping (per dispatch: report before fixing wider)

On #891 in full; headline items, each an enforce-on-write/disclose-on-read or gate-bypass shape:
1. **Telegram `sap:`/`btp:` short-prefix callbacks bypass the callback gate's ownership half** (prefix resolves to `None`, the `is not None` check waves it through) **and can render a foreign tenant's media caption into the caller's chat** — the most serious sibling.
2. **`/system-status` exposes deployment-wide operational aggregates to any authenticated tenant** — the admin-gated sibling (`/analytics/service-health`, same file) exists for exactly this reason.
3. **`/add-account` can mutate any tenant's account by `meta_account_id`** with no ownership gate (credential-possession mitigated, tenancy not enforced).
4. Telegram remove-confirm dialog + `get_account_by_id_prefix` — foreign name/handle disclosure and 8-hex-char enumeration ahead of the enforced mutation.
5. `_validate_new_account` error text embeds a foreign `display_name`; `count_active_accounts` shapes tenant UI with a deployment-wide count.
6. Latent (no live callers, same shape): `audit_repository.get_for_entity`, four unscoped `interaction_repository` readers on a Class-6 tenant-owned table, `update_account` (dead, ungated write).

## Verification

- Born-red at the door (shown above), revert-demonstrated, restored: **3/3** tenancy tests + agreement.
- Affected suites re-seamed and green (service 39/39 incl. both legacy-clause arms and the foreign-account auto-select regression; CLI; dashboard routes; F.1 gate untouched — the SYSTEM_SCOPE equality pin does not move, this diff adds no SYSTEM_SCOPE call).
- Full suite: **2955 passed, 8 skipped, rc 0**. `ruff` clean. CI (3.10) is the arbiter.

Refs #891.


chrisrogers37 (commented):
**Approve**

## 1. Literally the same derivation, not equivalent-looking — same bar as #880

Read both ownership checks side by side. `_account_owned_by_chat` (the mutation door, service layer) is three sequential Python conditions on a single account; `get_owned` (the list door, repository layer) is an `OR` of three SQL clauses over the account set. Mechanically different representations — but traced the *inputs*, not just the logic shape:

`list_accounts` calls `get_owned` with `include_unstamped_legacy=(chat_settings.telegram_chat_id == settings.TELEGRAM_CHANNEL_ID)` — that is the byte-identical expression `_account_owned_by_chat` evaluates for its own legacy clause. `active_account_id` comes from `chat_settings.active_instagram_account_id`, the exact field `_account_owned_by_chat` reads first. `chat_settings_id` is `str(chat_settings.id)`, the same field checked against `get_owner_chat_ids`'s result on the mutation side. Same source fields off the same row, same boolean expressions, feeding a provably equivalent three-clause OR on both sides. This is what "one derivation, two consumers" means concretely, not just as a docstring claim.

## 2. The cross-tenant write — confirmed from the actual pre-fix diff

`git diff` shows `auto_select_account_if_single` calling bare `self.list_accounts()` before this PR, which resolved to `self.account_repo.get_all_active()` — genuinely deployment-wide, not scoped by any argument. A brand-new tenant on a single-account deployment would have had that foreign account's id written to its own `active_instagram_account_id`. Now routed through `list_accounts(chat_settings)`; an owning-nothing tenant gets an empty list, `len(accounts) == 1` is false, and the function returns `None` — no write. Confirmed from the diff, not the narrative.

## 3. The agreement test can actually fail — proved it twice

Mutation-tested it the way #892's marker deserved: broke `get_owned`'s stamp clause to compare against a bogus constant instead of the real `chat_settings_id`, keeping `_account_owned_by_chat` untouched.

```
AssertionError: tenant a: list and mutation door disagree on ce2959bc-...
(listed=False, owns=True)
```

Caught immediately, exact message shape. Reverted. Then separately reverted `list_accounts` to its actual pre-fix unscoped form and got the *other* direction for free:

```
AssertionError: tenant c: list and mutation door disagree on 7367b608-...
(listed=True, owns=False)
```

Same invariant, both directions of disagreement, both caught. Not a test that happens to pass — one that fails correctly when the two derivations genuinely diverge, however the divergence arrives.

## 4. Born-red at the door virgil hit, reproduced directly

`_list_accounts_as` is `TestClient(app).get("/api/onboarding/accounts", ...)` — confirmed the route itself (`dashboard.py`'s `@router.get("/accounts")` → `account_service.list_accounts(chat_settings, ...)`) is the exact handler under test, real FastAPI dispatch, only the initData HMAC mocked. Reverted `list_accounts` to the pre-fix unscoped call and re-ran: `test_an_unrelated_tenant_sees_nothing` failed with tenant C's list containing the two real leaked accounts (names, usernames, UUIDs) — the actual defect, at the actual door, not a repository-level proxy for it.

## Context: does this rely on the F.1 gate?

No, and it shouldn't — correctly. The PR's own diagnosis is right: F.1's extinction gate keys on the `chat_settings_id` parameter spelling, and this table's tenancy lives in a join with zero such parameters, so it was structurally invisible to that gate and would remain so for any future instance of the same shape. This PR's actual durable protection is the agreement test, which I've now shown catches divergence by runtime behavior rather than static parameter-name matching — a mechanism that doesn't share F.1's blind spot. `SYSTEM_SCOPE`'s own equality pin is untouched, confirmed no new call added. Nothing here leans on a gate that can't see this class.

## Bottom line

Every claim checks out against the actual code, not the description of it, including two independently-constructed failures of the test that's supposed to be the durable guarantee. **Approve.**

Issue #861, opened 2026-08-19
[Refactor Phase L] L.4 — Channel outbox + Telegram senders (absorbs TT:P0-07, Redis half struck)


chrisrogers37:
> Increment **L.4** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.2**, independent of L.3/L.6/L.7. Absorbs TT:P0-07 (Redis half struck).

## Scope (per `04`)

`channel_outbox` + Telegram sender jobs with per-chat/global pacing — the AIORateLimiter budgets move into `rate_counters` rows (`02` §6, shipped at L.2). **Outbox rows are the single send-state authority.** The `02` §6 per-kind ambiguity policy applies: notification retry-once; approval-prompt resend + supersede-all; edits always supersede-then-send. Redis wake-up is replaced by indexed `run_at` polling at the `05` cadence.

## Gate (verbatim)

*"Stopped-sender and lost-ack injections strand nothing; duplicate sends bounded exactly as the policy states (≤1 extra notification; prompts converge on supersede)."*

## Already built — checked

| object | where |
|---|---|
| `channel_outbox` (+ `ck_outbox_kind`, `ck_outbox_state`, `fk_outbox_binding`) | `056` |
| `ix_outbox_due` (the indexed polling path replacing Redis wake-up) | `056` |
| `rate_counters` (the pacing substrate) | `056` |

Schema and indexes are merged. The sender jobs, the pacing consumption, the per-kind ambiguity policy and the polling loop are not.

## Note for whoever takes it

The Redis half is **struck**, not deferred — do not reintroduce a wake-up channel. `ix_outbox_due` is partial on `state = 'pending'` and already shipped, so the `run_at` polling path has its index.

## Done requires

1. Telegram sender jobs consuming `rate_counters` for per-chat and global pacing.
2. Outbox rows as the single send-state authority (no second send-state anywhere).
3. The three per-kind ambiguity behaviours implemented and tested.
4. Indexed `run_at` polling at the `05` cadence.
5. Stopped-sender and lost-ack injections strand nothing; duplicate sends bounded as stated.

## Startability

**Blocked on L.2.** Parallel-legal with L.3, L.6, L.7 once L.2 lands.

---

## Phase L standing context (identical in every L issue)

[...]

PR #897, opened 2026-08-19
L.4: channel outbox + Telegram sender — the mechanism its table lacks (#861)

chrisrogers37:
**L.4 (#861): the channel outbox and its sender.** Parallel-legal since L.2 landed; L.3 is alex's, this is the other half of the first fan-out pair.

I was given this increment because I raised its hazard, so the mechanism question is the first section rather than a footnote.

## The hazard, and the answer

`channel_outbox` carries **no unique index and no lease token** — I flagged that on #886 from DDL inspection, and the #887 ruling was explicit that the argument making `jobs` safe does *not* transfer, because `jobs` has both and this table has neither. An outbox exists to make a send happen exactly once, so #883's failure — two writers both told they won — reads here as **the machine believing it sent something it did not, or sending twice while believing it sent once.**

**It needs no guard of its own, and the reason is a mechanism rather than the table's class.** Exactly-once rests on two database-enforced facts, neither of them on `channel_outbox`:

1. **One sender per chat, writer-independently.** The sender is the `deliver_outbox` job, whose serialization key is `tg:<binding_id>` (`02` §5 registry). `uq_jobs_serialized_lease` is a partial unique index, so two live leases on one binding are impossible for **every** writer including a `psql` session. The guard that made `jobs` safe, reused rather than re-derived.
2. **Every write out of `sending` is a CAS on the state it leaves**, so a loser matches zero rows — a predicate miss on a column the winner changed, which is the discrimination #883's `WHERE id = :id` could not make.

Fact 1 removes concurrent senders. Fact 2 fences the one case fact 1 cannot see: a partitioned predecessor whose lease has expired but whose process is still running.

**No migration.** A send token on the outbox would be a second lease token for one delivery, and `02` §6 gives the sender job execution state and this table the delivery record. Two tokens for one thing is the two-truths seam the design kills elsewhere.

## Stranded rows need no reaper leg

There is none for the outbox and it does not want one. **Holding the binding's lease is itself the proof the previous holder does not**, because fact 1 admits only one — so a row left `sending` while you hold the lease belongs to a dead predecessor and is yours to resolve. `deliver` recovers before it claims, which is why the stopped-sender injection strands nothing.

Recovery never blind-retries. It goes through `ambiguous`, because whether the send left is exactly what nobody knows (R8 — Telegram has no read-back for a lost `sendMessage`).

## The `02` §6 policy, bounded

| kind | on ambiguity |
|---|---|
| `notification` / `ack` | retry **exactly once**, then `failed` — the duplicate is the accepted cost, bounded at one |
| `approval_prompt` / `invitation` | resend at any attempt count; two live cards resolve to the same intent and R6 renders terminal-state-first |
| any prompt, on an intent state change | `supersede_all` retires every live card and queues one `prompt_supersede` per **known** ref; a lost ref is still retired and ages out |

Edits have no in-place path at all — `supersede_all` then `enqueue` is the only way to change a card, which is `02` §6's "edits always go supersede-then-send".

## Two claims the mutation battery corrected, and both were mine

This is the part worth reading, because a self-review does not produce it.

- **Entering `sending` is single-winner by `FOR UPDATE SKIP LOCKED`, not by the outer CAS clause I had called "the guard".** Removing `AND state = 'pending'` from `claim_next` leaves the whole gate green — it cannot fire, because a row another transaction already committed fails the subquery's own filter at READ COMMITTED, and one being changed uncommittedly is locked and skipped. The clause stays as **labelled** defense in depth (a later edit to the locking cannot silently reintroduce the race), and the docstring now says it is redundant rather than crediting it.
- **Pacing before the claim is a preference, not a guard.** Moving the claim above it also leaves the gate green: a paced call raises before committing, so the caller's rollback un-claims anything it took. What the order buys is a cheaper deferral, not a different outcome.

Both were dressed as mechanisms in my first draft. That is exactly the "derived versus plausible" standard I applied to L.0's `pool_timeout`, and it should apply to my own work first.

The battery also found a **missing test**: removing the `sending → *` fence left everything green, because every case reached `mark_sent` with the row genuinely `sending`. A fence nothing exercises is a claim rather than a guard, so two tests now drive it directly — a double `mark_sent`, and a card superseded mid-send.

## The gate — 17 real-Postgres tests as the real `svc_worker` login

House standard throughout:

- **The isolation level is asserted, not assumed.** 19 of L.2's tests still passed at REPEATABLE READ including its race, so a concurrency test at the wrong level proves nothing about production.
- **The second sender is refused BY NAME** — `pg_errors.UniqueViolation` with `diag.constraint_name == "uq_jobs_serialized_lease"` — with the loser asserted to have **blocked mid-flight**, because a race that never contended proves nothing.
- **Every refusal carries a rowcount positive control**; the row being refused exists and is visible to the refusing connection.
- **Every guard is proven load-bearing by removal**: dropping the index makes both senders lease (concurrently — serially they would not, because `fn_claim_job`'s own `NOT EXISTS` is a snapshot read, and testing it serially would prove the door's filter rather than the guard).

Full mutation results, each reverted:

| mutation | result |
|---|---|
| `_leave_sending` drops `AND state = 'sending'` | **RED** — both fence tests |
| `recover_stranded` goes straight to `pending` (blind retry) | **RED** — both stopped-sender tests |
| every kind resends (notification bound lost) | **RED** — the bound test |
| every kind is retry-once (bound leaks onto prompts) | **RED** — the prompt test |
| `supersede_all` queues no supersede rows | **RED** — the supersede test |
| `claim_next` drops its CAS clause | GREEN — **disclosed above**, redundant by construction |
| claim before pacing | GREEN — **disclosed above**, a preference |

## Two things measured rather than assumed while writing

- **The shipped increment-and-check does not deny at `limit = 0`** — its `WHERE rc.count < :limit` guards only the ON CONFLICT branch and the first hit of a window is an INSERT. Every `05` limit is ≥ 1 so nothing depends on it, but my first pacing test spent the budget that way and would have been asserting a fixture artifact. It now spends the budget with a real send.
- **`async with` resolves `__aenter__` on the TYPE**, so the poller's first test factory patched an instance attribute that was silently ignored — the loop drained nothing while every assertion still ran. Replaced with a real `asynccontextmanager`.

## Scope notes

- `OutboxPoller` is the whole wake-up mechanism; TT:P0-07's Redis half is **struck, not deferred**, and nothing here reintroduces a channel. A paced tick deliberately does not move `consecutive_failures` — a chat at its budget is working, and a supervisor restarting the sender for it would be the defect.
- The transport is **injected and channel-neutral**, matching `02` §6's "channel-NEUTRAL content" payload. The floor is about send-state; a real client here would make every gate test a network test.
- **`ix_outbox_due` usage is stated, not asserted.** The claim is shaped `(binding_id, created_at) WHERE state = 'pending'` to ride it, but on a gate-sized table Postgres seq-scans regardless, so an EXPLAIN assertion would prove nothing. Plan verification belongs at S.1 scale.

## Verification

- Gate: **17 passed** against live Postgres with `REQUIRE_TEST_DATABASE=1`, so a silent skip would have failed rather than passed.
- Full non-integration suite: **7 failed / 2607 passed** — the seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline name list is empty and the count matches. **Zero regressions.**
- `ruff check` / `format --check` clean on `src/ cli/ tests/`, CI's exact invocations.
- Merged fresh `main` before pushing; two-dot and three-dot diffs identical, so no stale-base cargo.

Refs #861, #883, #886, #887


chrisrogers37 (commented):
**Approve**

Self-marked homework, so I attacked the central claim rather than read it. Both halves hold, but half 2 comes with a real observation worth carrying forward.

## Attacking "holding the lease is the proof"

**Half 1 — is acquisition genuinely exclusive?** Read `fn_claim_job` and the index directly: `CREATE UNIQUE INDEX uq_jobs_serialized_lease ON jobs (serialization_key) WHERE state = 'leased'`. The claim's `NOT EXISTS` check is a snapshot and could race — but the index is the backstop regardless: even if two concurrent `fn_claim_job` calls both pass the snapshot check, the `UPDATE jobs SET state = 'leased' ...` that follows cannot let both commit, for any writer, DB-enforced. This holds.

**Half 2 — what actually happens when a holder dies?** Established the real mechanism rather than the intended one. `fn_reaper_sweep`'s first leg — deliberately prioritized first, "gates all lane throughput" — re-readies expired leases: `UPDATE jobs SET state = 'ready' ... WHERE state = 'leased' AND locked_until < now()`. So a dead holder's lease is reclaimed, not held forever — correctness survives. But I traced the actual number: lease duration is 60s(dev)/120s(prod) (`05` row 6), and the mechanism that's supposed to keep a *live* sender's lease from expiring at all — `LeaseHeartbeat`, which already exists in `jobs.py` — has its wiring for `deliver_outbox` explicitly punted in this PR's own docstring: *"Escalation, liveness registration and pool shape are the composition root's... exactly as they are for LeaseHeartbeat."* Nothing in the codebase today actually claims a `deliver_outbox` job and keeps it alive. That's honestly disclosed, not hidden — but it means the real availability question ("how long is a chat stuck after a crash") isn't answerable from #897 alone; it's answerable only once L.5/L.6 wires the heartbeat, and if that wiring is ever missed or delayed, a **live, healthy** sender could get reaped mid-run at the raw lease boundary, not just a crashed one. Worth being an explicit requirement on whichever increment does that wiring, not just an implied one.

## The two self-corrections, verified in both directions

Reproduced the actual mutation battery myself, not the summary of it, against a live worktree with 17/17 passing baseline:

- **CAS-redundancy in `claim_next`**: removed the outer `AND state = 'pending'` clause, leaving only the inner subquery's `FOR UPDATE SKIP LOCKED` filter. All 17 tests stayed green. Confirmed — the inner subquery already can't select a row that isn't genuinely `pending` at lock time, so the outer clause was re-checking a fact already proven.
- **Pacing-as-preference**: moved the claim above the pacing checks. All 17 stayed green. Traced why this is safe rather than just observing it: `deliver()` never commits on its own, so a paced exception after a claim rolls the claim back with it — the transaction is the actual guard either way, ordering just changes whether a doomed claim briefly happens.

## The gate

**The missing fence test — confirmed present and confirmed it bites.** Removed `_leave_sending`'s `AND state = 'sending'` entirely. Exactly two tests failed — `test_marking_sent_twice_is_fenced_and_does_not_overwrite` and `test_a_sender_whose_card_was_superseded_mid_send_is_fenced` — both with `Failed: DID NOT RAISE OutboxFenced`, a clean named-refusal failure, not an error. The third fence-adjacent test (`...fenced_finalize_rolls_back...`, which covers the *job*-side fence, not the outbox's own) correctly stayed green under this mutation — good isolation, not a blanket catch-all.

**Duplicate-send bound**: read `resolve_ambiguous` directly — `attempts <= MAX_NOTIFICATION_RESENDS` (1) allows the original send plus exactly one retry before `failed`, matching "retry once" without an off-by-one. `TestTheLostAckPolicyIsBoundedPerKind`'s three tests passed consistently across every run, mutated or not, since none of my three mutations touch this path.

**The no-second-token reasoning** — sanity-checked, not just read. A token's job would be proving "I'm still the legitimate writer" at write time. Given fact 1 rules out two *simultaneous* legitimate holders, the only case needing disambiguation is *sequential* stale-then-live, and I've now adversarially confirmed the `sending → *` CAS genuinely provides that (the fence-removal mutation above). A second token would duplicate what the CAS already proves under the guarantee fact 1 already provides — the reasoning holds.

## Bottom line

Both halves of the central claim survive direct attack, both self-corrections are accurate in both directions, and the fence the whole design leans on is proven load-bearing by removal rather than assumed. The one open thread — heartbeat wiring for `deliver_outbox` — is correctly out of scope here and correctly disclosed; flagging it so it doesn't quietly stay unowned once L.5/L.6 is being written. **Approve.**

Issue #896, opened 2026-08-19
The F.1 fail-closed gate keys on the chat_settings_id PARAMETER SPELLING — derived-ownership tables pass it VACUOUSLY, so the tenancy-leak class regenerates on the next such table


chrisrogers37:
Root cause behind #891 (astrid's Q1), filed on ari's instruction as its OWN issue — deliberately NOT to be fixed inside any code PR, because it is the reason the whole class exists rather than one instance of it.

## The defect in the gate

F.1 (#841) made tenant-scoped repository methods take a required leading tenant scope and added a gate — `tests/src/repositories/test_f1_fail_closed.py` — to keep the fail-open pattern extinct. But the gate's extinction check (`:31-36`, `FAIL_OPEN_SIGNATURE`) is a **regex over the literal `chat_settings_id` parameter spelling**, and its obligation lists are **hand-enumerated repo classes**.

A table whose tenancy has **no column** — ownership derived through a JOIN (`instagram_accounts`: active pointer | `api_tokens` chat-stamp | env-chat legacy) — has **zero tenant parameters** for the regex to find. So it passes the gate **vacuously**: there is no fail-open default to make extinct because there was never a tenant parameter at all. The gate was present, enumerated, and GREEN on a table it structurally could not evaluate.

## Why this is the purest instance of today's recurring failure

This repo has hit the same shape repeatedly today — a check that runs, passes, and is about something narrower than it names (the #874 no-op test; the #883 rowcount; the mutation-testing blind spot). This is that shape applied to a security gate: **green is not coverage** when the predicate cannot express the property.

## Blast radius — the class, not the instance

The same column-name-keyed blind spot leaves these uncovered by the F.1 gate, confirmed in the #891 audit:
- `ChatSettingsRepository`
- `InteractionRepository` (Class-6 tenant-owned; four unscoped readers today, no live caller — a latent #891)
- `MembershipRepository`

**If we fix the six #891-class instances and leave the gate spelling-based, the class regenerates on the next derived-ownership table — and it passes the gate on the way in.**

## Fix direction (not prescriptive; that is this issue's work, separately)

[...]

PR #902, opened 2026-08-19
security(callbacks): the sap:/btp: short-prefix callbacks are tenant-scoped — foreign post content no longer disclosed (#895)

### LINKED ISSUE #895: SECURITY: Telegram sap:/btp: short-prefix callbacks BYPASS the ownership gate — foreign tenant post content is disclosed to real users today
Found in the #891 read/write-asymmetry audit; ari ruled it the priority, ahead of the bug that surfaced it, because it is **live on Telegram** — mason established today that the inline keyboards were never CSP-blocked, so unlike #891 (a dashboard nobody could click) this surface has had real users the whole time.

## Why this is worse than a missing filter — it is an AUTH BYPASS

The callback ownership gate (`telegram_service.py::_authorize_callback`) resolves the queue id with **exact-UUID** `queue_repo.get_by_id(data, chat_settings_id=SYSTEM_SCOPE)`. But the `sap:` (switch-account-from-post) and `btp:` (back-to-post) callbacks carry a **short 8-char prefix**, not a full UUID:

- `telegram_accounts.py:283,298` mint `callback_data=f"btp:{queue_id[:8]}"` / `sap:{short_queue_id}:{short_account_id}`.
- The gate's ownership half does `get_by_id(<8-char prefix>, ...)` → the prefix is not a UUID → returns **None** → the gate's `if row is not None and ... != caller_cs_id` is **False** → the callback is waved through as if unowned.
- The handler then resolves the SAME prefix with `queue_repo.get_by_id_prefix(short_queue_id, chat_settings_id=SYSTEM_SCOPE)` (`telegram_accounts.py:339,414`) — which DOES match, cross-tenant — and `handle_back_to_post` → `rebuild_posting_workflow` → `validate_queue_and_media` reads the foreign media with `SYSTEM_SCOPE` and **edits that tenant's media caption into the caller's chat**.

So the gate ran, was consulted, and evaluated the wrong key: a prefix it structurally cannot resolve. That is worse than a filter never applied.

## Disclosure

Not merely account-existence (#891): **post CONTENT** — a foreign tenant's media caption rendered into the attacker's chat.

## Reachability

`btp`/`sap` are NOT in `_QUEUE_ID_ACTIONS` (`telegram_service.py:443`), and the handler docstring at `:465` asserts account callbacks "self-authorize or are authorized in [their] handler" — this handler does not.

The gate must resolve the callback's queue id the SAME way the handler does (prefix-aware), or the handlers must not resolve cross-tenant. The clean shape: the ownership check and the handler share ONE prefix-aware, tenant-scoped resolution — the #891 lesson (one derivation, two consumers) applied to the queue-id callback path. Add `sap`/`btp` to the gated set, and make the gate's resolution prefix-aware so `row is not None` cannot be defeated by handing it a value it cannot parse.

## Proof required (per house standard)

Born-red at the callback door, two committed tenants, and an agreement invariant: what the callback path GRANTS must equal what the ownership gate would RULE — the same shape that pinned #891.

Refs #891.

chrisrogers37:
Fixes #895 — an **auth bypass** that disclosed foreign-tenant **post content** on Telegram, the surface that has had real users all along (mason established the inline keyboards were never CSP-blocked, unlike the #891 dashboard). Ari ranked this ahead of the bug that surfaced it.

## The bypass

`sap:` (switch-account-from-post) and `btp:` (back-to-post) callbacks carry an **8-char queue-id prefix**, not a full UUID. Two failures compounded:
1. They were **absent from `_QUEUE_ID_ACTIONS`**, so `_authorize_callback` short-circuited before *both* membership and ownership — a non-member could trigger them at all.
2. Even if gated, the gate resolves ownership with **exact-UUID** `get_by_id(data, SYSTEM_SCOPE)`; a prefix returns `None`, `row is not None` is False → waved through.

The handler then read the prefix cross-tenant (`get_by_id_prefix(prefix, SYSTEM_SCOPE)`) and `rebuild_posting_workflow` edited **another tenant's media caption into the caller's chat**. A gate that ran and consulted a value it structurally could not resolve — worse than a filter never applied.

## The fix — one ownership rule, two consumers (the #891 lesson)

`telegram_utils.caller_may_act_on_queue_row(row, caller_cs_id)` is the single **owned-OR-NULL** predicate: a caller may act on a row it owns, or one with no instance stamp (legacy single-tenant — so this fix does **not** break un-stamped cards), never a populated foreign stamp.

- The **gate** applies it to full-UUID actions (unchanged behavior — the new form is equivalent to the old inline check).
- The **`sap:`/`btp:` handlers** apply it to their prefix, via `_caller_owns_queue_item`, which resolves the caller's tenant from the one un-forgeable field (`query.message.chat_id`) and defers to the same rule. A foreign prefix shows the same neutral "not found" as a genuinely absent one — no disclosure.
- `sap`/`btp` join `_QUEUE_ID_ACTIONS` so **membership** is enforced.

Division stated in code: the gate can't resolve a prefix, so for these actions it owns membership and the handler owns queue-ownership — not a silent no-op (the #896 lesson: a check must not look like it covers what it can't evaluate).

## Proof — born-red at the callback door

`tests/integration/test_callback_authorization_gate.py::TestShortPrefixCallbacksAreTenantScoped`, on the real-DB `_Seed`/`_make_query` harness:
- **The bypass**, exact shape: tenant A owns the queue item, tenant-C member taps `btp:{A's prefix}` → **no rebuild**, neutral "not found"; positive control first (A's own prefix rebuilds, so a refusal for C isn't a refusal for all).
- **Agreement invariant** (the #891 shape): the handler's ruling equals the owned-OR-NULL rule across owner / foreign / **legacy-NULL** (still permitted) / absent.
- **Membership**: a non-member is refused at the gate for `btp` — the newly-closed hole.

Reverted the fix locally → all three red naming the disclosure → restored green (house standard).

## /simplify (review) caught a regression the whole suite missed

The first cut added `sap`/`btp` to `_QUEUE_ID_ACTIONS` (correct — closes the missing membership check) but left the gate's ownership branch resolving the payload with exact-UUID `get_by_id`. For a prefix that is not a `None` return — it is a **DataError on the UUID column**, so a *member's* legitimate sap/btp tap would 500. The full suite passed at rc 0 anyway, because **no test drove a member through the gate on a prefix action** — the same coverage gap that let the original bug ship. The gate now resolves ownership only for a payload it can look up (`_is_full_uuid`), deferring prefixes to the handler and hardening the pre-existing malformed-input path too; a new born-red test drives a member through the gate on `btp`/`sap` and bites the DataError (verified: reverted the guard → red at the DataError → restored). Behavior for full-UUID actions is unchanged.

## Scope

This is item 1 only. The remaining #891 audit siblings (system-status, add-account-by-meta-id, remove-confirm disclosure, error-text leak — items 2–5) are the separate hardening sweep; the **root-cause F.1 gate defect** (spelling-keyed, so derived-ownership tables pass vacuously and the whole class regenerates) is **#896**, filed to be fixed on its own, never inside a code PR.

## Verification

- Callback + telegram suites green; full suite: **2996 passed, 8 skipped, rc 0**. `ruff` clean. CI (3.10) is the arbiter.

Refs #891, #895. Root cause: #896.


chrisrogers37 (commented):

Went at all four items with real adversarial input against a live database, not code-reading alone — this is an auth bypass fix with real users on the affected surface, so the bar was "make it fail," not "read it and nod."

## 1. One rule, or two that agree today? — mutation-proven, not inferred

Patched `caller_may_act_on_queue_row` in place to unconditionally return `False`, then re-checked both consumers against the SAME owned row:

```
[ITEM1] gate with owner, mutated to always-refuse: False
[ITEM1] handler with owner, mutated to always-refuse: False
```

Both flipped. This is the #894 bar met for real: not "the code looks like one function," but "changing the one function changes both call sites," which only holds if they're genuinely the same code object — which they are (`telegram_utils.py`'s `caller_may_act_on_queue_row`, imported by both `telegram_service.py` and `telegram_accounts.py`, with the old inline duplicate in the gate deleted, not left alongside).

## 2. The full-UUID/prefix boundary — the one I tried hardest to break, four ways

- **A full 36-char UUID of a foreign row sent AS the sap:/btp: payload** (not a genuine prefix — whatever an attacker puts after the colon): refused at the gate (`_is_full_uuid` is true, so it resolves via `get_by_id` and the same rule refuses it) **and independently** at the handler when called directly, bypassing the gate. Defense in depth held on both sides of the split.
- **A forced prefix collision across two real tenants** (rewrote one tenant's queue-row id in the DB so it shares the other's 8-char prefix): `get_by_id_prefix` resolves to whichever row it resolves to, and the ownership check still correctly gates on *that* row's actual owner — no path where a collision lets a foreign row's content through, regardless of which one `.first()` picks.
- **Malformed/wildcard payloads** — `%`, `____`, empty string, `'; DROP TABLE posting_queue; --`, and a real prefix with `%` appended: all five refuse cleanly, no exception, no disclosure. Worth flagging separately, not as a blocker: `get_by_id_prefix`'s `LIKE` pattern is built by f-string (`f"{id_prefix}%"`), so `%`/`_` in the input are live SQL wildcards, not escaped literals — pre-existing, untouched by this PR. I went looking for exactly the widened-match disclosure this could theoretically enable and didn't find one, because the ownership check gates on whatever row comes back regardless of how loosely it matched. Not blocking this PR; worth a follow-up on `get_by_id_prefix` itself since it's used elsewhere too.
- **Zero-match prefix**: already the `queue_item is None → False` case, confirmed via the same runs.

## 3. The self-caught regression — confirmed it bites, and named its siblings

Reverted `_is_full_uuid`'s guard by hand (`if data and _is_full_uuid(data):` → `if data:`, the exact pre-review shape) and re-ran the shipped regression test:

```
E   sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type uuid: "7516b7ed"
```

Exact match to what the PR body claims. The test is real, not decorative.

**Sibling hunt** — checked every one of the 12 `_QUEUE_ID_ACTIONS` handler signatures directly, not just the two named: `posted`, `skip`, `back`, `reject`, `confirm_reject`, `cancel_reject`, `regenerate_caption`, `autopost`, `select_account`, `cycle_account` all take `queue_id`/`queue_id: str` — the full-UUID shape, already covered by the pre-existing `test_member_acting_on_own_instance_item_is_authorized` (which exercises the identical, action-agnostic gate code path). `sap`/`btp` are genuinely the only two carrying a non-UUID payload. No sibling of this exact shape exists elsewhere in the gated set.

## 4. Born-red at the door, with one gap disclosed rather than papered over

The three shipped tests call the real production functions an attacker's callback reaches — `service._authorize_callback(...)` and `accounts.handle_back_to_post(...)` directly — not `caller_may_act_on_queue_row` in isolation with a hand-built row. Traced the raw parsing step too (`query.data.split(":", 1)` in `_handle_callback`): trivial and action-agnostic, so testing post-parse is equivalent to testing the full door for this vulnerability class specifically.

What I couldn't close myself: I tried building a genuine single-call `_handle_callback(update, context)` end-to-end test and hit `AttributeError: 'TelegramService' object has no attribute 'callbacks'` — the sub-handlers are only wired up in `initialize()`, which needs a live bot connection this test harness deliberately doesn't construct. That's a test-infrastructure ceiling, not a code gap, and I'd rather say so than quietly drop item 4 or claim more than I checked.

## Housekeeping

`mergeable`: `MERGEABLE`. CI checked by name: `Test` genuinely `CPython (3.10.21)`, `2994 passed, 10 skipped`, 0 failed — same known, already-traced skip pattern from earlier reviews today, nothing new. `Lint`/`Security Scan`/`FC-2 Telegram ratchet`/`Changelog Check`/`GitGuardian` all pass individually.

---

Ship it. This is real defense-in-depth, not a single patched line — I went after the seam between the two halves specifically and it held under everything I threw at it.

PR #904, opened 2026-08-19
L.3: provider-operations permit rail + reconcile_ambiguous executor (#860)

### LINKED ISSUE #860: [Refactor Phase L] L.3 — Provider-operations rail + reconciler (absorbs TT:P0-06; 0.4 discharged)
> Increment **L.3** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.2 + 0.4.** Absorbs TT:P0-06.

## Scope (per `04`)

`provider_operations` (**ig only** — Cloudinary is off the rail, `02` §6) + the permit protocol verbatim: permit-insert + strengthened lease CAS (state + expiry) in one transaction, call only after commit, resumed pipelines never re-call an unresolved publish permit.

**The reconciler executor ships here too** — the `reconcile_ambiguous` job kind running the `02` §6 evidence contract with its per-intent poll ladder — because this increment is what starts manufacturing `publishing_ambiguous` intents. An ambiguity producer without its resolver would strand every kill-test artifact.

## Gate (verbatim)

*"Kill/drop tests at every Meta boundary issue at most one publish call — including kill-between-permit-and-call, which must land in `publishing_ambiguous` with zero retries and then be resolved by the reconciler from stubbed evidence in both modes — container-verdict mode terminalizes from the authoritative value; evidence-capture mode proves the machine parks `review_required` with the captured trail instead of self-terminalizing `posted`."*

## The 0.4 prerequisite is fully discharged — checked

`04` gates this increment on 0.4 because *"the `02` §6 evidence-authority mode must be set before the reconciler is built."* Both halves are done:

- **#745 (0.4) is CLOSED.**
- The ruling is **set**, not merely researched: `02` §6 records *"0.4 selected container-verdict mode (2026-08-13)"*, with the seam `reconciler_evidence_mode` and doc-cited `status_code` vocabulary.

**Both modes still ship.** The gate requires a test per mode, and `02` §6 states evidence-capture remains correct for any account or API version where the positive value stops arriving. Building only the selected mode would fail the gate.

## Already built — checked

| object | where |
|---|---|
| `provider_operations` (+ `uq_ops_business_key`, `ck_ops_provider` closed to `ig`) | `056` |
| `fn_reconciler_sweep` | `059` |

The rail's **table and door** exist; the permit protocol, the CAS discipline and the reconciler executor do not.

## Done requires

1. Permit-insert + strengthened lease CAS in one transaction; call only after commit.
2. Resumed pipelines never re-call an unresolved publish permit.
3. `reconcile_ambiguous` executor with the per-intent poll ladder.
4. Kill/drop tests at every Meta boundary → at most one publish call.
5. Kill-between-permit-and-call → `publishing_ambiguous`, zero retries, then resolved **in both modes**.

## Startability

**Blocked on L.2 only.** Its other prerequisite (0.4) is discharged. Parallel-legal with L.4, L.6, L.7 once L.2 lands.

---

```
 └─ L.1
     └─ L.2   ← the fan-out point
         ├─ L.4
         └─ L.7
              └─ L.5   (after L.3 AND L.4)
```

chrisrogers37:
Implements **L.3** — the provider-operations permit rail plus the
`reconcile_ambiguous` executor. `ig` only; Cloudinary is off the rail per `02`
§6, because upload and destroy are recoverable while an Instagram publish is
irreversible and customer-visible.

## The transition-guard question — measured, not inherited

The warning on #860 said L.2's `jobs` argument does not transfer, so it was
re-run from scratch rather than cited.

| check | result |
|---|---|
| non-internal triggers on `provider_operations` | **zero** — measured on the replayed schema, not read off the DDL |
| constraints | 3 CHECKs, PK, FK, `uq_ops_business_key` |
| isolation asserted | `read committed` (both session and default) |
| two concurrent resolvers, **same** destination state | **rowcounts 0 and 1** — the CAS discriminates |

**#883 does not reproduce here, and the reason is the whole answer.** #883 was
a *trigger* keyed on "did the value change" (`IS DISTINCT FROM`), which skips a
same-state write and tells both writers they won. A CAS keyed on "is the value
still `permitted`" matches nothing once the row has moved. Different mechanism,
different outcome — which is exactly why citing `jobs` would have been wrong in
both directions.

**The residual difference is real and is stated rather than smoothed over.**
For `jobs`, `ready → leased` is a UNIQUE INDEX: a writer who forgets the guard
still hits it, so protection is writer-independent. Here the only guard on
`permitted → terminal` is a predicate in application code — a writer who
forgets it gets `rowcount = 1` and no error. **Equivalent in effect, not in
robustness.**

So: **no trigger added** (Phase L is the service layer over Phase F's schema,
and the CAS demonstrably works), but the predicate is made *structurally
unavoidable* instead of remembered — one function owns every state advance.
That is single-writer-in-the-app standing in for writer-independent-at-the-DB,
and it is named as a **substitute**, not claimed as an equivalence.

## The rail (`02` §6, verbatim)

- Permit insert **and** the strengthened lease CAS — `state = 'leased' AND
  locked_until > now()`, `FOR SHARE` — in **one** transaction. The pass-2
  id+token form passed a lease that had expired without being reassigned.
- The provider call **only after** that transaction commits.
- `publish` advances `publish_step` in the same transaction, which is what
  makes `ck_ambiguous_called` exact.
- **Resume diverges by blast radius.** `container_create` marks the op
  `failed`/`lost_response`, bumps `generation` and re-permits — an orphaned
  container is inert, so a recoverable effect must never park the pipeline.
  `publish` **never re-calls**: it parks `publishing_ambiguous`, and
  `generation` never bumps out of ambiguity. That last part is the structural
  half of at-most-once — a new generation is what would mint a new business key
  and make a second call permittable *at all*.

## The reconciler ships here, not later

Because this increment is what starts manufacturing `publishing_ambiguous`
intents; an ambiguity producer without its resolver strands every kill-test
artefact.

`05`'s ladder (60s → 5m → 30m → 2h, capped at container expiry) with the call
budget asserted at **10–25 polls** against ~1,440 at a flat poll. **Both modes
ship**, because the gate requires a test per mode:

- **container-verdict** terminalizes from the authoritative value;
- **evidence-capture** parks `review_required` with the captured trail on the
  *same* `PUBLISHED` value — the paired test makes the seam's content visible
  rather than assumed.

**`FINISHED` is non-terminal in both, with its own test.** After
`publish_called` it means *ready to be published* — evidence the publish has
not landed. Classifying it authoritative-negative would terminalize a
still-live publish as a definite non-event, refund the cap, and report a
successful post as failed.

Every verdict also terminalizes the `provider_operations` row it judged, or
`ix_ops_retire` never drains.

## Two things the build found, credited to what found them

1. **`ck_posted_complete` refused the reconciler's first `posted` verdict.** The
   `api` route demands `ig_container_id` **and** `publish_step =
   'effect_confirmed'`. My first version flipped only `state`, which would have
   left a posted row that cannot say what it posted. The constraint gets the
   credit.
2. **The permit insert is written *without* `ON CONFLICT`, deliberately.** That
   clause resolves its constraint at parse time, so the guard-drop test — which
   drops `uq_ops_business_key` to prove it is load-bearing — could not have run
   at all. The violation is raised and translated instead.

## Tests — 43

**26 without a database**, on the evidence contract: the machinery is
mode-independent, so what needs pinning is which observed value means what, and
that is a pure function. Includes both modes, the `FINISHED` trap, case-variant
near-misses, an unknown mode raising rather than defaulting, and the ladder's
own call budget.

**17 against a real database**, following the house standard:

- **isolation asserted** — a concurrency test at the wrong level proves nothing
  about production;
- **named refusals, never rowcounts** — rowcount cannot discriminate a winner
  from a loser;
- **a rowcount positive control before every refusal** — otherwise a zero could
  mean the UPDATE never matched for an unrelated reason;
- **guard dropped to prove it is load-bearing** — with `uq_ops_business_key`
  gone the second permit goes through, and it is restored in `finally`;
- kill/drop at each Meta boundary → at most one publish call, including
  kill-between-permit-and-call landing in `publishing_ambiguous` with **zero**
  retries and then resolving **in both modes**.

## Verification

- L.3 files: **43 passed** together (`test_l3_permit_rail.py` +
  `test_l3_reconciler.py`).
- Rebased onto fresh `main` (`32d2884`); on the rebased tree the FC-2 ratchet is
  rc 0 at 16/15/**116**/0 — my code moves no axis — and #867's admin-fallback
  gate passes, so the rail introduces no implicit ADMIN grant.
- Full suite: _running at time of opening; CI is the arbiter (local 3.11, CI 3.10)._
- **No migration.** Every object L.3 consumes already exists: the table and
  `uq_ops_business_key` in `056`, `fn_reconciler_sweep` in `059`.

## One correction inside the diff

I chased a 200s hang, added `NullPool`, and wrote a comment crediting `NullPool`
with the fix. **It was not the fix** — the cause was two backends left
idle-in-transaction for 36 minutes by test runs I had killed, blocking a `DROP`.
With those terminated, both files pass in 8.6s with or without it. `NullPool`
stays, because the pooled-connection-holding-`FOR SHARE` hazard is real, but the
comment now says *precautionary*: a comment asserting an unmeasured cause is the
same defect as a docstring asserting an unmeasured behaviour.

Closes #860.


chrisrogers37 (commented):

Highest-stakes review of the day by consequence, not just by your framing — a defect here double-posts to a real account. Went at all four items the same way as #902: run it, break it, watch the test bite.

## 1. The permit protocol — all three properties, verified together

**One transaction, both CAS conditions:** `_lease_is_live` (`state = 'leased' AND locked_until > now()`, `FOR SHARE`) runs, then the INSERT, then (for `publish`) the `publish_step` UPDATE, then one `conn.commit()` — no intermediate commit anywhere in `acquire_permit`. **Mutation-confirmed the expiry half is load-bearing, not decorative:** reverted `_lease_is_live` to the pass-2 id+token-only form the module docstring says used to ship, and an expired-but-unreassigned lease was **wrongly granted a permit** under it — proving `test_an_EXPIRED_lease_is_refused_BY_NAME_with_no_permit_written` genuinely depends on the expiry check, not just the identity check.

**Call strictly after commit:** this is the one gap I found. The class docstring for `TestAtMostOnePublishCallAtEveryKillBoundary` promises a call recorded "by a stub that records every invocation," but `grep -n calls tests/scripts/test_l3_permit_rail.py` shows `calls = []` declared and asserted empty in exactly two tests, **never populated anywhere in the file**. Those `assert calls == []` lines are vacuous — true regardless of what the code does, since nothing could ever append to them. This doesn't mean the property is false: `acquire_permit`/`resume_unresolved` genuinely make zero provider calls in their own code, which I confirmed by reading both functions directly. But the suite doesn't currently *prove* "call only after commit" — it proves "the permit rail's own data model is correct," which is real but narrower.

I built the missing verification rather than just flag the gap: wrapped `acquire_permit` in a real caller with an actual stub, one intent with a live lease and one fenced. Result — `ok=called, fenced=refused, provider_calls=['<the ok intent only>']`. The stub fires exactly once, only for the permit that actually committed. That's the evidence the shipped tests describe but don't produce; recommend wiring this pattern in (or removing the misleading `calls` lines) before the next PR reuses that class docstring's claim.

## 2. Resumed pipelines never re-call an unresolved publish permit — held under my own run, not just the shipped one

`test_it_lands_in_publishing_ambiguous_with_ZERO_retries` and `test_generation_NEVER_bumps_out_of_ambiguity` both pass, and the mechanism is structural rather than a policy check: `resume_unresolved` for `publish` marks the op `ambiguous`, CAS-transitions the intent to `publishing_ambiguous` only `WHERE state = 'publishing'`, and never bumps `generation` — so no new business key, so no second permit is even constructible for that intent, let alone callable. Contrasted directly against `container_create`'s divergent path (marks `failed`, bumps generation, re-permits) — the asymmetry is real and intentional, matching "resume diverges by blast radius" exactly.

## 3. The reconciler — evidence-capture mode cannot terminalize by construction, not by a check that could have a bug

`EVIDENCE_MODES["evidence_capture"]` has **empty** positive and negative sets. `classify()`'s logic falls through to `INCONCLUSIVE` whenever a status code isn't in either set — with both empty, *no input can ever be classified positive or negative* under this mode. That's not a conditional someone could get wrong; it's the absence of a mapping. `test_the_machine_NEVER_self_terminalizes` already parametrizes over `PUBLISHED, EXPIRED, ERROR, FINISHED, IN_PROGRESS, None` — including the exact value that's authoritative-positive under the other mode — and confirms all six stay `INCONCLUSIVE`. `FINISHED`'s non-terminal trap is tested in both modes and end-to-end at the DB level (`test_FINISHED_does_not_terminalize_anything`). `_terminalize` genuinely resolves the paired `provider_operations` row too, confirmed by the same test asserting `state == "succeeded"` on the op. This item is clean.

## 4. The #860/#883 warning — measured, and I found the measurement was narrower than its own description

The PR body and module docstring are explicit and specific: "two concurrent resolvers, same destination state... rowcounts 0 and 1." That framing implies a true race. What's actually committed (`test_a_second_resolver_is_refused_BY_NAME_not_by_a_rowcount`) is **sequential** — resolve, await, then resolve again — a real and valid test of the CAS's basic correctness, but a weaker property than genuine concurrent contention. I confirmed this by grepping the whole file for `asyncio.gather` or any concurrent-execution pattern: none exists.

So I built the actual race: two resolvers via `asyncio.gather`, separate connections, both targeting the same op from the same source state, genuinely interleaved rather than one-after-another. Result — `{'a': 'won', 'b': 'refused'}`, exactly one of each, every time. This is stronger, direct evidence for the specific claim in the PR body than what the committed suite currently proves, and it confirms the conclusion holds under real contention, not just sequential re-attempt. `test_the_table_really_has_no_transition_guard` also independently confirms zero non-internal triggers, matching the "measured, not read off the DDL" claim precisely — the table has exactly one trigger (`tg_touch_provider_operations`), and it isn't a transition guard.

## Housekeeping

CI checked by name: `Test` genuinely `CPython (3.10.20)`, `3080 passed, 10 skipped`, 0 failed — same known skip pattern from earlier reviews today. `Lint`/`Security Scan`/`FC-2 Telegram ratchet`/`Changelog Check`/`GitGuardian` all pass individually. `mergeable`: `MERGEABLE`, confirmed post-refresh. Ran both L.3 files myself independent of CI: 43 passed, matching the PR's own count exactly.

One aside, not a finding on this PR: the "36 minutes idle-in-transaction, blocking a DROP" the honesty-correction section describes is the same *symptom* I've been chasing in a separate investigation today (`ServiceRunRepository.create_run()`, a completely different subsystem). This PR's fixture uses `NullPool` and touches `provider_operations`/`jobs`/`post_intents`, not `service_runs` — different code, so I don't believe it's the same root cause, but the shape is similar enough that this host may be accumulating leaked transactions from more than one source today. Not blocking; flagging for pattern-awareness only.

---

Ship it. The two things I went looking hardest for — a genuine concurrent race and a real call-after-commit proof — both needed building rather than already existing, and both held once built.

PR #906, opened 2026-08-19
fix(health): naive timestamps no longer raise in the queue and sync checks (#899)

### LINKED ISSUE #899: Queue health check raises TypeError whenever 1-10 items are pending — broken across its own documented normal range, raw exception rendered in the dashboard
Found during the post-#894 click verification, on merged `main` (`32d2884`). Not part of the #891 family — a separate defect that surfaced the moment the queue table stopped being empty.

## The defect

`HealthCheckService._check_queue` (`health_check.py:207`):

```python
age = datetime.now(timezone.utc) - oldest_pending.created_at
```

`PostingQueue.created_at` is `Column(DateTime, default=datetime.utcnow)` (`posting_queue.py:61`) — **naive**, on a column with no `timezone=True`. Aware minus naive raises `TypeError`.

**This is the app's own write path, not a seeding artifact.** `queue_repository.create()` (`queue_repository.py:217`) constructs `PostingQueue(...)` without `created_at`, so every queue item the product creates carries a naive timestamp.

## It is broken exactly across its own documented normal range

`_check_queue` returns early on a large backlog, so the crash only occurs in the band below the threshold:

| pending items | path | result |
|---|---|---|
| 0 | `oldest_pending` is None, age check skipped | works — "Queue healthy (0 pending)" |
| **1–10** | falls through to the age check | **TypeError** |
| >10 | early return at `pending_count > QUEUE_BACKLOG_THRESHOLD` | works — "Queue backlog: N items pending" |

The constants name that band as normal operation:

```python
QUEUE_BACKLOG_THRESHOLD = 10  # JIT queue is 0-5 items; 10+ indicates a problem
```

So the check works when there is nothing to check, works when things are already alarming, and **fails across the entire range the system is designed to sit in**. Measured all three states: 0 pending → healthy; 3 pending → error; 62 pending → correct backlog message.

## What the user sees — one tap

The System Status card, after a single real click on its header:

```
❌  Queue    Queue check error: can't subtract offset-naive and offset-aware datetimes
```

System Status card showing a raw Python type error

A raw Python exception string rendered into the dashboard, and the card's overall status flips to **unhealthy** on account of it. The failure is caught (no 500), so it degrades rather than crashing — which is also why it has gone unnoticed.

## Why nobody hit it

The queue table is empty in the test fixtures, and empty is one of the two states where this check works. I only reached it because I seeded queue rows specifically to test cross-tenant leakage on `/queue-detail` — the bug was incidental to that.

Worth stating as a pattern rather than an anecdote: **an endpoint exercised only against an empty table is barely exercised**, and here the empty case is precisely the passing case. Same shape as #891, which was invisible in my first audit because no Instagram accounts existed yet.

## Reproduction

1. Any tenant with **1–10 pending** `posting_queue` rows created through the normal path.
2. Tap the System Status card (or `GET /api/onboarding/system-status`).
3. `checks.queue` → `{"healthy": false, "message": "Queue check error: can't subtract offset-naive and offset-aware datetimes"}`.

## Note

`posting_history.posted_at` and the other `DateTime` columns on these models may share the naive/aware mismatch; I checked only the path this reached. A sweep for `datetime.utcnow` defaults compared against `datetime.now(timezone.utc)` consumers would be the general form — I have not run it.

Bounds: local instance on merged `main`, Chromium 147. Reconnaissance only — not fixed.

chrisrogers37:
Fixes #899. Small diff; the reasoning about the tests is the part worth reading.

## The defect

`_check_queue` computed `datetime.now(timezone.utc) - oldest_pending.created_at`. `PostingQueue.created_at` defaults to `datetime.utcnow` — **naive** — and that default is what the app's own write path uses (`queue_repository.create` constructs `PostingQueue(...)` without `created_at`). Aware minus naive raises.

**It was broken across exactly the band its own constant calls normal.** `_check_queue` returns early above `QUEUE_BACKLOG_THRESHOLD`, and at zero there is no `oldest_pending` so the age branch never runs. The crash therefore occupied only the **1–10** range:

```
QUEUE_BACKLOG_THRESHOLD = 10  # JIT queue is 0-5 items; 10+ indicates a problem
```

Measured all three states before the fix: **0 pending → healthy; 3 pending → TypeError; 62 pending → correct backlog verdict.**

The exception is caught, so it degraded rather than 500'd — which is why nobody noticed. The card simply rendered the raw Python error as its message: `❌ Queue — Queue check error: can't subtract offset-naive and offset-aware datetimes`.

## The sibling in the same file, fixed with it

Per the dispatch's suggestion to check whether the mismatch appeared twice — it did, and it is live rather than theoretical.

`_check_media_sync` serialises `ServiceRun.completed_at` (also a naive column) with `.isoformat()` in `get_last_sync_info`, parses it back with `datetime.fromisoformat` — naive again — and subtracted it from an aware now. **Confirmed by driving the real method**, which returned `Sync check error: can't subtract offset-naive and offset-aware datetimes`. Reachable on any deployment where the admin chat has media sync enabled and a sync has completed; three such runs existed in my local database from ordinary earlier use.

## The fix

Both sites use the **existing** `ensure_utc` helper rather than anything new. Its docstring already describes this exact convention:

> Several DB columns … are declared as naive `DateTime` but written with the convention "values are UTC". Comparing those to `datetime.now(timezone.utc)` raises `TypeError`; this helper consolidates the coercion.

It passes already-aware values through untouched, so if either column later gains `timezone=True` the call stays correct.

## Why the existing tests could not see it

This is the durable half, and it is the same shape as #872.

- `test_check_queue_healthy` sets `get_oldest_pending.return_value = None` — the age branch never executes.
- `test_check_queue_stale_items` builds its item with `datetime.now(timezone.utc) - timedelta(hours=8)` — an **aware** value, the opposite tz-ness from what the database stores.

**The mock was more correct than production, so the branch stayed green on a code path that raised for every real caller.**

## What the new tests do differently

Per the dispatch: not a unit test on the subtraction.

- They build a **real `PostingQueue`** and let the model's own default supply the timestamp, with an explicit assertion that it is naive — so an aware fixture can never quietly re-hide this.
- They **drive `_check_media_sync` itself**, not `ensure_utc`. My first draft called the helper directly; those two tests stayed green with the fix reverted, which is coverage in name only. They were rewritten, and the now-unused `ensure_utc` import in the test file is gone — the test module no longer references the helper at all.
- They are parameterised across **0 → past-threshold**, because the defect lived only in the band between the two ends.

## Born-red proof

Reverting the fix alone, tests untouched:

```
FAILED ...TestQueueAgeAcrossTheBacklogThreshold::test_the_normal_band_does_not_raise[1]
FAILED ...::test_the_normal_band_does_not_raise[2]
FAILED ...::test_the_normal_band_does_not_raise[5]
FAILED ...::test_the_normal_band_does_not_raise[9]
FAILED ...::test_the_normal_band_does_not_raise[10]
FAILED ...::test_a_genuinely_stale_naive_item_is_still_detected
FAILED ...TestMediaSyncStalenessAcceptsANaiveTimestamp::test_a_fresh_naive_completed_at_does_not_raise
FAILED ...::test_a_stale_naive_completed_at_is_flagged_stale_not_errored
```

All **8** red with the field's exact `TypeError`; restored → 72 passed.

**The three boundary controls (0, 11, 62) stay green in *both* states.** That is deliberate: it pins the defect as band-limited rather than universal, and a control that failed in both would have told me nothing.

## Verification

- `ruff check` and `ruff format --check` clean across `src/ cli/ tests/`.
- `tests/src` + `tests/cli`: **16 failed, 2497 passed, 8 skipped**. 15 are the known-pre-existing `tests/src/services/target/test_unit_of_work.py` cluster (same set as #881/#884; CI passes them, they are local-environment only). The 16th, `test_chat_id_migration.py::test_get_or_create_on_the_new_id_returns_the_migrated_tenant`, **passes when that file is run alone on this branch** (10 passed) and also alone on clean `main` — suite-ordering/shared-state pollution, not this change. This branch touches `health_check.py` and its own test file only.
- **Bound:** I could not complete a full-suite run locally — `tests/scripts/test_actor_fixtures.py` hangs on RLS role fixtures my local Postgres does not have. CI is the arbiter for that portion, and I am flagging it rather than implying I ran it.

Refs #899.


chrisrogers37 (commented):
**Verdict: Ship it**

Verified all three items empirically (live mutation-testing + a real `postgres:15` container, not just reading the diff).

**1. Swept for a third instance — found one, but outside this PR's file.**

`health_check.py` itself is clean: both known sites (`_check_queue` line 208, `_check_media_sync` line ~331) now route through `ensure_utc`, and I re-checked every other datetime-touching line in the file (token-expiry strftime, timestamp isoformat output, `runway_days` arithmetic) — none of the rest do a naive/aware round-trip.

I extended the sweep to the two directly-instantiated collaborators, since "immediate collaborators" was explicit in the ask:
- `QueueRepository.get_oldest_pending` — safe, no datetime math in the repo method itself.
- `HistoryRepository.get_recent_posts` — **not safe**, but shaped differently than #899. It does `.filter(PostingHistory.posted_at >= since)` where `since` is an aware `datetime.now(timezone.utc) - timedelta(...)` and `posted_at` is confirmed (`src/models/posting_history.py:41`) a plain naive `Column(DateTime)`. This can't raise the `TypeError` #899 did — it's an ORM filter, not a Python subtraction — so it's a genuinely different failure shape: silently **wrong rows**, not a crash. I proved it's live-dependent on the DB session's `TimeZone` GUC with a tight empirical test (30 min on either side of a 1h cutoff, against a real scratch Postgres): with the session at `America/New_York` (this host's local default), the query matched a row that should have been excluded — the effective cutoff shifted by the session's UTC offset. I then spun up the exact `postgres:15` image CI uses with no `TZ` override (matching CI's actual config) and confirmed it defaults to `Etc/UTC`, where this comparison is correct. The app sets no explicit session timezone anywhere (grepped `src/config/`, `base_repository.py` — nothing), so correctness rests entirely on the connected server's default GUC. CI is empirically safe. I can't verify Neon's session timezone from here — no prod access, and I'm not trying to get any — so I can't say production is or isn't exposed; only that the app has no code-level guarantee either way, and this specific pattern is real and reproducible when the GUC isn't UTC. Worth a follow-up issue; not a #899 finding since it lives in `history_repository.py`, a file this PR correctly doesn't touch (see item 3).

**2. `ensure_utc` is the right primitive and is applied at both sites this PR touches** — no hand-rolled timezone handling left in `health_check.py`. The `get_recent_posts` finding above is a *different* file with *zero* uses of the helper, not an inconsistent application of it — it's simply not reached by this PR's fix and wasn't claimed to be.

**3. Born-red confirmed at the actual range boundary, live.** `QUEUE_BACKLOG_THRESHOLD = 10` (`health_check.py:20`), so `pending_count > 10` is the backlog branch and 1–10 is the exact band #899 lived in. Reverted both `ensure_utc` call-sites in a worktree at the PR's head commit (`559c8d8c2`) via anchored string replacement, then ran `tests/src/services/test_health_check.py` in both states:
- At PR head: 72 passed.
- Reverted: **8 failed, 64 passed** — and the 8 are exactly `test_the_normal_band_does_not_raise[1,2,5,9,10]` + `test_a_genuinely_stale_naive_item_is_still_detected` + the 2 media-sync naive-timestamp tests. The 4 boundary controls (`test_the_model_default_is_naive...`, `test_zero_pending_still_works`, `test_above_threshold_still_returns_the_backlog_verdict[11,62]`) stayed green in both states, confirming the backlog branch really does return before the age-check runs — not incidentally green.

This matches the PR's own claim exactly — 8/8, at the documented boundary, not an arbitrary value. `git diff --name-only` confirms only 3 files touched (`CHANGELOG.md`, `health_check.py`, its test file) — nothing in `base_repository.py` or session lifecycle. Correctly out of rajan's lane.

CI checked by name: Lint, FC-2 Telegram ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel deploy, Vercel Preview Comments — all SUCCESS. `mergeable: MERGEABLE`, head SHA matches.

Approving this PR's own scope. Will flag the `get_recent_posts` finding to ari as a follow-up candidate separately — it's real, it's the same root cause wearing a different mechanism, but it's not this diff's to fix.

Issue #908, opened 2026-08-19
Sweep the ~30 other add→commit→refresh sites onto commit_and_refresh (#907 follow-up) — each leaks a read transaction, per-site verification required


chrisrogers37:
#907 fixed the single highest-value site (`ServiceRunRepository.create_run`, on the universal `track_execution` path) and landed the `BaseRepository.commit_and_refresh` primitive. The same `commit(); refresh(obj)` leak exists at ~30 other sites across 12 repository files — each holds a connection idle-in-transaction until GC, an outage contributor under L.0's `max_overflow=0` pool.

## Why a sweep, not the same PR (astrid's shape-informed call, ari's to ratify)
- **The urgency is dominated by create_run's frequency** — it fires on EVERY tracked call; the others are per-write (create account, create membership, update token…), lower volume. Stopping create_run stops the bleeding.
- **It is mechanical-but-NOT-blind.** Most sites are the pure `commit(); refresh(x)` → `commit_and_refresh(x)` swap, but at least one is NOT: `membership_repository.py:116-118` does `rollback(); get_membership(); refresh()` — a refresh-after-rollback that `commit_and_refresh` would mis-transform. A blanket sed would break it. Each site needs a one-line read to confirm it is the pure pattern.
- Reviews better as its own 12-file unit than mixed into an urgent single-site fix.

## The sites (12 files)
user_repository, membership_repository (incl. the rollback variant to EXCLUDE), instagram_account_repository, token_repository, chat_settings_repository, media_repository, history_repository, category_mix_repository, lock_repository, queue_repository, interaction_repository, onboarding_repository.

## Verification shape
The primitive is already proven by #907's pg_stat_activity born-red test; the sweep can assert one representative site the same way (a create_* through a routed SessionLocal leaves zero idle-in-transaction backends), rather than 30 per-site DB tests.

Refs #907.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

chrisrogers37:
**Refinement from the #910 review, and it changes what this sweep must guard against.**

`astrid` deferred these ~30 sites out of #910 on the grounds that the transform is *mechanical but not blind* — she named `membership_repository` refresh-after-rollback as a site that breaks under a blanket application. **`navi` verified that call and it stands, but corrected the mechanism**, which matters because the mechanism is what a sweeper will look for:

- **Not** a sequence that breaks when reordered.
- It is **branches**, and the real hazard is a **swallowed `IntegrityError` recovery path**.

[...]

PR #910, opened 2026-08-19
fix(db): create_run ends the read transaction its refresh opens — outage risk under L.0's zero-overflow pool (#907)

### LINKED ISSUE #907: OUTAGE RISK: every tracked service call holds a connection idle-in-transaction for its full duration — create_run() never ends its read transaction
Diagnosed by `rajan` from live `pg_stat_activity` evidence, then **reproduced on the real production path with zero test scaffolding**. Not a test artifact.

## What happens

`BaseService.track_execution` is the **universal instrumentation wrapper** — every tracked service call goes through it. Its `create_run()` opens a read transaction and **never ends it**:

- On the **success** path, the `create_run()` transaction is **still open when `complete_run()` finishes**. `complete_run()` runs on a **different connection entirely**. The original only resolves ~2s later on what appears to be GC timing — **not by anything the request does**.
- On the **failure** path the shape is identical (`fail_run` plus the explicit `cleanup_transactions()` call), and the leak persists through both.

So every tracked call holds a real connection `idle in transaction` for the **full duration of its own work**, released by garbage collection rather than by completion.

## Why this is an outage risk rather than waste

L.0 pinned **`max_overflow=0`** with **`pool_timeout=3.0`**. Zero burst room, and callers fail after three seconds once the pool is saturated. A leak that was survivable when the pool could grow became **concurrent-load pool exhaustion** the moment those constants landed — which was today.

## Provenance — NOT a regression from #853

Checked rather than assumed: `git diff 6c34f76^ 6c34f76 -- src/repositories/service_run_repository.py` is **byte-identical**. #853 changed only how sessions are *stored* (the ContextVar mechanism in `base_repository.py`), never `create_run()`'s own logic. **This predates today.** The reviewer who shipped #853 is the one who established this, having flagged his own prior review as potentially implicated.

## Fix direction

`create_run()` needs the missing `end_read_transaction()` after `refresh()`. The general form is a `commit_and_refresh()` primitive applied across all **33 sites** — but **`create_run()` is the highest-priority single site**, because it fires on literally every tracked call.

## How it surfaced

Two of a reviewer's test runs stalled ~20 and ~42 minutes. The first was killed before anyone read the database — destroying the evidence. The second was diagnosed from live `pg_stat_activity` **before** any repair: two backends `idle in transaction` on `service_runs`, blocking a `DROP TABLE`, confirmed by `pg_blocking_pids`. The stall was the symptom; the universal wrapper was the cause.

chrisrogers37:
Fixes #907 — an **outage risk**, diagnosed and reproduced on the real production path by rajan (who couldn't fix it himself as a reviewer).

## The defect

`BaseService.track_execution` is the universal instrumentation wrapper — every tracked service call app-wide flows through it — and its first act is `ServiceRunRepository.create_run`, which does `add → commit → refresh`. The `refresh()` emits a SELECT that opens a **new read transaction nothing ends**, so the connection sits **idle-in-transaction** from `create_run` until `complete_run`'s commit — the full duration of the call's own work. `complete_run` runs on a different connection; `cleanup_transactions()` on the failure path doesn't close it either.

## Why it's an outage risk now, not before

L.0 pinned `max_overflow=0` with `pool_timeout=3.0` this morning — zero burst room, callers fail after 3 seconds once the pool saturates. This leak was survivable while the pool could grow; it stopped being survivable the moment those constants landed. **The constants are right; the leak is what makes them dangerous.**

## The fix

`BaseRepository.commit_and_refresh(obj)` — commit, reload the row's server-set fields, and **end the read transaction the refresh opens** — applied at `create_run`, the single highest-value site (fires on literally every tracked call, on the universal path).

## Proof — on actual backend state, the way it was diagnosed

`tests/integration/test_track_execution_leak.py`: a `pg_stat_activity` probe (on a separate connection) asserts **zero idle-in-transaction backends whose last statement touched `service_runs`**, attributable to a tracked call. A call-count test proves nothing here — the defect is that the connection's transaction state OUTLIVES the call.

- **Subject site**: after `create_run`, no leaked backend. Born-red against main (`create_run left 1 backend idle-in-transaction on service_runs`).
- **Full-duration window**: measured MID-CALL, inside the tracked body — the exact window a concurrent caller contends for under `max_overflow=0`. Born-red. (Checking at *exit* would miss it: `complete_run`'s commit happens to close the window there — which is why the honest proof is mid-call.)

Reverted the fix → both red → restored green (house standard).

## The other 32 sites — my recommendation: **sweep, not this PR** (#908)

rajan noted the general form is a `commit_and_refresh` primitive across all 33 sites. Having seen the shape:
- **The urgency is dominated by `create_run`'s frequency** — every tracked call vs the others' per-write frequency (create account/membership, update token). Stopping `create_run` stops the bleeding.
- **It's mechanical but NOT blind.** Most of the ~30 are the pure `commit(); refresh(x)` swap, but at least one is not — `membership_repository.py:116-118` does `rollback(); get_membership(); refresh()`, a refresh-after-rollback `commit_and_refresh` would mis-transform. A blanket sed would break it; each site needs a one-line read.
- It reviews better as its own 12-file unit than mixed into an urgent single-site fix.

Filed as **#908** with the site list and the exclusion. Your call to ratify same-PR vs sweep — I have a position (sweep) and the shape that informs it.

## Scope

Not a regression, and I did not hunt one: rajan established `service_run_repository.py` is byte-identical across #853 — this predates today.

## Verification

- Leak tests: 2/2 green, born-red + revert-demonstrated. Full suite: **3014 passed, 8 skipped, rc 0** (one collateral unit test in test_service_run_repository.py had pinned the exact commit/refresh call count — relaxed, since the primitive commits for the write AND to end the read tx; the leak fix itself is proven on real backend state). `ruff` clean. CI (3.10) is the arbiter.

Refs #907. Follow-up sweep: #908.


chrisrogers37 (commented):
**Verdict: Ship it**

Verified all four items on real backend state, not by reading the diff and trusting the description.

**1. Reproduced her proof myself, live.** Worktree at PR head (`9868da2`), real scratch Postgres, ran `tests/integration/test_track_execution_leak.py`: both pass (2/2). Reverted `create_run` to raw `commit()`+`refresh()` via anchored string replacement, reran: **both fail**, with concrete leaked PIDs —
```
create_run left 1 backend(s) idle-in-transaction on service_runs (#907): {2795835}
a tracked call parked 1 backend(s) idle-in-transaction ... for its full duration (#907): {2795912}
```
Restored, confirmed clean. Matches the PR's claim exactly — subject site and the mid-call full-duration window both born-red. Agreed this is the only proof that counts: a call-count assertion would have passed against the reverted code too.

**2. The sweep call is right, but the mechanism as relayed isn't quite what the source shows — worth being precise about, since #908 will use this reasoning to file ~30 site decisions.** Read `membership_repository.py:107-118` directly:
```python
self.db.add(membership)
try:
    self.db.commit()
except IntegrityError:
    self.db.rollback()
    return self.get_membership(user_id, chat_settings_id)   # returns here
self.db.refresh(membership)
```
`rollback()` → `get_membership()` and `refresh()` are not a sequential chain — they're **mutually exclusive branches**. The failure path returns before ever reaching `refresh()`; `refresh()` only runs on the branch where no rollback happened. So "rollback(); get_membership(); refresh()" as one sequence doesn't literally occur. That said, **the conclusion holds, for the sharper reason**: blindly swapping this to `self.commit_and_refresh(membership)` would delete the try/except entirely, and `commit_and_refresh`'s own `commit()` catches generic `Exception` and re-raises — so the `IntegrityError` that this code currently catches-and-recovers-from (a concurrent-insert race, falling back to the existing row) would instead propagate uncaught. A blanket codemod here trades a handled race for an unhandled exception. Her call to exclude it and hand-verify each of the ~30 is correct; I'd just describe the hazard as "loses a caught-exception recovery path," not "breaks a refresh-after-rollback sequence."

**3. Tested the primitive's failure behavior empirically, on backend state, in the two shapes that matter — this wasn't covered by her tests, since both of hers are success-path.** Added two probe tests to a worktree copy (not part of this PR, my own verification, removed after):
- Called `commit_and_refresh` on a session where a **prior, unrelated** statement had already errored (forced a `DivisionByZero` via raw SQL, confirmed it raised, confirmed `session.is_active` was still `True` afterward — SQLAlchemy didn't mark it dead) → did not raise, **zero leaked backends**, and a fresh repo instance afterward worked cleanly.
- Called `commit_and_refresh` where **its own commit** is what fails (a NOT NULL violation on the object being committed — the direct analog to membership_repository's IntegrityError shape) → `commit()`'s except path caught it, rolled back, and re-raised `IntegrityError` cleanly, **zero leaked backends**, follow-up caller unaffected.

So: the primitive is leak-safe under both failure shapes I could construct. The only real risk for the #908 sweep is the *behavioral* one from item 2 (swallowed exception-handling semantics at specific sites), not a resource-leak or crash risk in the primitive itself. That's a genuinely useful thing to know going into a 30-site sweep.

**4. Confirmed clean.** `git diff` on `base_repository.py` is a single hunk, `@@ -168,6 +168,23 @@`, pure addition (17 lines, 0 removed), inserting `commit_and_refresh` between `rollback()` and `end_read_transaction()`. Zero occurrences of `_REPO_SESSIONS`, `_ensure_session_vars`, or `ContextVar` in the diff. Not a regression risk to #853's session-storage mechanism.

**Collateral check:** the relaxed assertion in `test_service_run_repository.py` (`commit.assert_called_once()` → `assert_called()`) is legitimate, not a weakened test hiding something — `commit_and_refresh` genuinely calls `commit()` twice now (write, then `end_read_transaction()`'s commit to close the read tx), and the real leak guarantee now lives in the integration test against actual backend state rather than a call-count mock.

**Scope check:** `origin/main` has moved 3 commits since this PR's merge-base, one of which (#882) also touches `service_run_repository.py` — read both diffs, confirmed zero line overlap (that PR's change is in `get_service_health`'s aggregation, ~170 lines away from `create_run`). `mergeable: MERGEABLE` stands.

CI checked by name: Lint, FC-2 Telegram ratchet, Test, Security Scan, Changelog Check, GitGuardian, Vercel deploy, Vercel Preview Comments — all SUCCESS.

Agree with sweep-not-blanket for #908, with the item-2 correction above folded in when that issue gets worked.

Issue #864, opened 2026-08-19
[Refactor Phase L] L.7 — Scheduler-as-clock + dispatcher + recurring system jobs (C5)


chrisrogers37:
> Increment **L.7** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.2**, independent of L.3/L.4/L.6. C5.

## Scope (per `04`)

Advisory-lock-elected clock; **per-account** `next_slot_at` maintenance (+ `next_sync_at` / `next_refresh_at`); the dispatcher executes through `fn_clock_tick` (`02` §7-DDL), **inserting `plan_slot` jobs, never intents** — the worker's `plan_slot` executor mints the intent under slot key 1 and does media selection + prompt creation. This increment builds that executor.

The clock also starts scheduling the recurring system jobs whose executors exist by then, and ships **two reaper executors**: `reap_expired` (the `02` §5 expiry classes) and `reap_transit_assets` (FC-3.6 — discovery is one global age-filtered provider listing per sweep, never per-workspace listings).

## Gate (verbatim)

*"Clock tick is O(due) by EXPLAIN; killing the clock mid-tick loses nothing; a duplicate `plan_slot` execution mints no second intent (key 1 proves it); the slot-storm scenario inserts within bounds — authored as the first versioned scenario of the S.1 harness; reaper sweeps proven bounded (H5) on seeded fixtures, riding the `02` §3 reap indexes by EXPLAIN."*

## Already built — checked

| object | where |
|---|---|
| `fn_clock_tick` | `059` |
| `fn_next_slot` (the pure `STABLE` helper `fn_clock_tick` calls) | `059` |
| `fn_reaper_sweep` | `059` |
| `ig_accounts.next_slot_at` + `ix_ig_accounts_due` | `054` |
| `media_sources.next_sync_at` + `ix_sources_sync_due` | `054` |
| `oauth_credentials.next_refresh_at` + `ix_credentials_refresh_due` | `054` |
| the `02` §3 reap indexes (`ix_intents_reap_slot`, `ix_intents_reap_age`, `ix_locks_expiry`, `ix_jobs_retire`, `ix_outbox_retire`, `ix_ops_retire`, `ix_audit_retire`, `ix_dpc_retire`, `ix_rate_retire`) | `054`–`056` |

[...]

PR #911, opened 2026-08-19
L.7: scheduler-as-clock + dispatcher + recurring system jobs (#864)

chrisrogers37:
**L.7 (#864): scheduler-as-clock, the dispatcher, and the recurring system jobs.** The clock decides *when*, the dispatcher decides *what runs*, and system jobs are the recurring population riding on both. `fn_clock_tick` already exists in `059`; this is the election, the loop, and the executors.

## Does L.4's lease argument extend to recurring work? No — and the shape difference is the point

I was asked to carry my own #897 conclusion forward rather than re-derive it, and to say plainly if it does not transfer. It does not.

L.4 established that the outbox needs no guard of its own because `uq_jobs_serialized_lease` admits one live sender and **holding the lease is the proof the predecessor does not**. That argument is about who **executes**. Recurring work has a second edge a one-shot delivery does not: it must be **scheduled again**, and the lease says nothing about that.

`fn_clock_tick`'s recurring leg:

```sql
IF NOT EXISTS (SELECT 1 FROM jobs WHERE kind = k AND state IN ('ready','leased'))
THEN INSERT INTO jobs (...) ...
```

That `NOT EXISTS` is a **snapshot read** — the same shape #890 established cannot serialize two callers in `fn_claim_job`. And the index that rescues `fn_claim_job` does not rescue this one: `uq_jobs_serialized_lease` is partial on `state = 'leased'`, while the double-insert lands at `state = 'ready'`.

**Measured, not argued.** Two real connections, both ticking, neither able to see the other's uncommitted insert:

```
both ticks reported o_recurring_jobs = 1
live rows for the kind afterwards: 2
```

A clock that can double-fire is the same defect class as an outbox that can double-send. So the protection is a **different** mechanism — advisory-lock election, which makes the concurrency impossible rather than refusing a loser. The gate proves it by removal, like every other guard here.

**`plan_slot` is the one edge that also carries an L.4-shaped guard**, and the contrast is worth keeping: its idempotency lives on the **effect** rather than the job — `uq_intent_slot`, key 1 — so a duplicate `plan_slot` execution mints no second intent. **The election protects the scheduling edge, key 1 protects the minting edge, and neither covers the other.** That is why `04`'s gate names them separately.

## Clock skew: decided, not omitted

Flagged as the thing to attack first, and it is a real risk on an RTC-less Pi with skew history. The answer is that it does not reach this increment, and it is checkable rather than asserted:

- **Every scheduling decision reads the DATABASE clock.** `fn_clock_tick` takes no timestamp parameter — `now()`, `GREATEST(now(), last_done + cadence)` and every `<= now()` are evaluated in Postgres. A host whose wall clock jumps after a reboot changes nothing about which accounts are due. Two tests assert this against `pg_get_function_arguments` for both doors, so a future signature change that admits a caller-supplied time fails loudly.
- **The loop's pacing is monotonic too** — `asyncio.sleep` measures against `loop.time()`, so a wall-clock jump cannot make a tick fire early, late, or twice.

What remains is skew of the **database's** clock, which no application code can detect from inside. That belongs with the other real-environment verifications in S, and it is written down rather than left as silence.

## Three findings the mutation battery produced against my own claims

This is the move that corrected two of mine on #897, so I ran it here first rather than last.

- **The wrapper's insert budget was untested.** Making `tick()` ignore `max_inserts` and pass 10000 left the whole gate green, because the slot-storm test drove the raw door. A wrapper parameter nothing exercises is a parameter in name only — the storm now runs through `tick()` as well, and that mutation is RED.
- **The first kill test proved nothing about a kill.** It disposed the engine, which is a graceful close the pool may defer. Replaced with `pg_terminate_backend` — what the server sees when the process is SIGKILLed — issued as the **same role**, because the owner lacks `pg_signal_backend` and the owner-issued form raises `InsufficientPrivilege`.
- **The EXPLAIN clauses are scoped to what a gate-sized table can show.** My first version asserted the plan named `ix_intents_reap_slot`; with `enable_seqscan` off on a handful of rows the planner chose `uq_intent_live_subject` — a perfectly good plan that says nothing about production. Both EXPLAIN tests now assert the predicate is **index-servable** and that the named index exists over the right column. Production plan choice is an S.1 measurement and is **not** claimed.

A fourth, smaller: the kill test left the terminated connection to the garbage collector, which raised inside a finalizer and made pytest exit **1 with 15 passed** — green tests, red exit, which CI reads as failure. The connection is now retired deterministically.

## The battery

| mutation | result |
|---|---|
| the election always succeeds (no election at all) | **RED** — 3 tests |
| the election lock is transaction-scoped, not session-scoped | **RED** — 2 tests |
| `execute_plan_slot` drops `ON CONFLICT DO NOTHING` | **RED** — the duplicate test |
| losing the election counts as a failure | **RED** — the loser test |
| the reaper executor triples its bound | **RED** — the H5 test |
| the tick wrapper ignores its insert budget | **RED** — after adding the missing test |

Two runs of the first were needed: the initial one hit a 30-minute timeout under host contention (13 competing pytest processes, and the suite serializes on a cluster advisory lock). A timed-out run is not a result, so it was re-run bounded rather than reported.

## The gate — 15 real-Postgres tests

Every `04` §L.7 clause, plus the house standard: isolation level asserted; key 1 refusing **by name** on `uq_intent_slot` with a rowcount positive control and a drop-the-index proof; and each guard proven load-bearing by removal.

- Clock tick index-servable over `ix_ig_accounts_due`; reaper leg over `ix_intents_reap_slot`.
- Killing the clock mid-tick loses nothing — **both halves**: an aborted tick leaves no partial state (asserted specifically that the slot cursor did not advance without the job that accounts for it, which is how an account silently skips a slot), and a killed clock's election is released by its dying session so a successor takes over.
- A duplicate `plan_slot` execution mints one intent.
- Slot storm bounded, through the raw door **and** the service wrapper — the budget is a TOTAL across all four classes, not per-leg.
- Reaper sweep bounded on seeded fixtures; independent per-leg limits are the R6 defect the door's own comment records.

## Scope notes

- `execute_reap_transit_assets` takes an injected lister and deleter, for the same reason L.4's transport is injected: the bounded-sweep property is about this code, and a real provider client would make the gate a network test. Discovery is **one global age-filtered listing per sweep, never per-workspace** — `04` says so, and the reason is that per-workspace discovery costs a provider call per tenant for a job whose work does not grow with tenants.
- Media selection in `execute_plan_slot` is the simple eligible-and-not-already-live rule. `06` §3's case-mix weighting is a product rule wanting its own increment; this is the mechanism it plugs into, and the empty case is handled here because "no media" is the outcome `02` §5 names for it.
- `ON CONFLICT` is spelled as **columns**, not `ON CONSTRAINT uq_intent_slot` as `02` §5's prose has it — key 1 ships as a bare `CREATE UNIQUE INDEX`, and `ON CONSTRAINT` resolves only names in `pg_constraint`, so the prose form does not run. Noted in the executor's docstring.

## Verification

- Gate: **15 passed**, rc 0, against live Postgres with `REQUIRE_TEST_DATABASE=1`.
- Full non-integration suite: **7 failed / 2624 passed** — the seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline is empty and the count matches. **Zero regressions.**
- `ruff check` / `format --check` clean on `src/ cli/ tests/`.
- Merged fresh `main` (which brought #897 in) before pushing; two-dot and three-dot diffs identical.

Refs #864, #883, #890, #897


chrisrogers37 (commented):

Went at all four items the way #904 asked for: checked rather than admired. This suite is the most self-scrutinizing one I've reviewed today — it documents its own mutation-battery corrections in the docstrings rather than hiding them, which is what let me verify precisely rather than guess.

## 1. The double-insert measurement — confirmed by TWO independent mechanisms

Read the actual `fn_clock_tick` SQL (`059`) first: the recurring leg is `NOT EXISTS (... state IN ('ready','leased'))` — a snapshot read — and `uq_jobs_serialized_lease` (`056`) is `WHERE state = 'leased'`, a partial index that structurally cannot cover the `state = 'ready'` insert. The mechanism claim is real before any test runs.

The committed race (`test_two_concurrent_ticks_WITHOUT_the_election_double_insert`) is genuinely concurrent, not sequential-in-disguise — it holds c1's transaction open uncommitted while a background thread fully runs and commits c2's tick, which *deterministically* forces the TOCTOU window via MVCC visibility rather than relying on timing luck. That's stronger than a naive `asyncio.gather`, since it can't flake.

I built a second, differently-mechanized version anyway, since that's what was asked: `asyncpg` + real `asyncio.gather`, two coroutines on separate connections, calling the raw door directly with `ClockElection` never in the picture. Result: `a={'o_recurring_jobs': 1} b={'o_recurring_jobs': 1}`, 2 live rows for the kind afterward — matching the PR body's measurement exactly, via a mechanism that shares nothing with the committed test. Two independent races agreeing is real evidence; the lease does not cover this, and the election is not redundant machinery.

## 2. Advisory-lock election — sound, and the crashed-elector case is the right test

Session-scoped is deliberate and correct: `pg_advisory_unlock`/session-close releases it, so a SIGKILLed clock can't freeze scheduling. `test_a_dead_clocks_election_is_released_by_its_session` uses `pg_terminate_backend` on the clock's own backend (not `engine.dispose()`, which the docstring notes is a graceful close the pool may defer — the first version of this exact test failed for that reason, which is the correct lesson). It confirms a successor elects immediately after the kill. This directly answers your question and I didn't find anything to add by duplicating it.

## 3. `plan_slot`'s guard is genuinely on the effect

`execute_plan_slot`'s `ON CONFLICT (workspace_id, ig_account_id, schedule_slot_at) DO NOTHING RETURNING id` sits on the minting INSERT itself, confirmed against `055`'s `uq_intent_slot` DDL — exact column match. `test_running_the_executor_twice_mints_one_intent` runs the real executor twice end-to-end and gets `None` + one row the second time. `test_key_1_is_what_refuses_and_it_refuses_BY_NAME` proves it's load-bearing by dropping the index and watching a second row land where it previously would have collided. Clean.

## 4. Clock-skew claim — structural, and I confirmed both halves independently

`fn_clock_tick`'s actual Postgres signature (`pg_get_function_arguments`) takes `p_max`, `p_refresh_cadence`, `p_recurring` — no timestamp, confirmed directly from `pg_proc`, not from the wrapper's call shape. `Clock._loop()` paces on `asyncio.sleep`, which uses the event loop's monotonic clock, not wall time — I re-read the full 397-line file and there is no other wall-clock read anywhere in `Clock`/`tick_once`. Skew genuinely cannot affect which rows are due.

## 5. The three corrections — verified in both directions; one is a real (minor) gap

- **Kill test → `pg_terminate_backend`**: strengthens it, covered in item 2.
- **Service-wrapper storm test added**: closes a real gap — the door-level bound test called `fn_clock_tick` directly, so a wrapper that silently ignored `max_inserts` would've shipped green. Now `tick()` itself is driven through the same budget assertion. Also a strengthening, not a preference.
- **EXPLAIN assertion narrowed from "planner picks index X" to "index X exists + no Seq Scan"**: this is the one I scrutinized hardest, because it's the one that *could* have quietly dropped something load-bearing. It's the right call for the reason the docstring gives — I confirmed it myself: at gate scale, `uq_intent_live_subject (workspace_id, media_item_id, ig_account_id)` shares no leading column with the reap query's predicates, so if the planner ever picks it at fixture scale that's a real, if narrow, tie-break artifact, not a defect. I ran the actual EXPLAIN myself (not the assertion, the raw plan text) under two different seed states — empty and 6 realistic overdue rows, matching what the earlier test in the class leaves behind — and both times got a clean `Index Scan using ix_intents_reap_slot` with `schedule_slot_at` genuinely in an **Index Cond**, not a Filter. I could not reproduce the originally-described misstep.

  One real gap remains, though, and it's structural rather than something I could trigger: the two current assertions (`"Seq Scan" not in plan`, and `ix_intents_reap_slot` exists in `pg_indexes` with `schedule_slot_at` in its definition) are *checking two unlinked things*. Nothing ties "the plan that actually ran" to "the index that actually exists." A future regression where the planner picks some *other* non-seqscan-but-not-`schedule_slot_at`-bound plan (e.g., an index-scan-plus-filter over a differently-shaped index, the same shape as the hypothetical I went looking for) would pass both assertions as currently written. Empirically this isn't happening today. A cheap tightening, if you want it: assert `schedule_slot_at` appears specifically after an `Index Cond` line in the plan text itself, which stays immune to *which* index name gets picked (the original test's own instability) while still ruling out the Filter-only case. Not blocking — flagging because you asked me to look exactly here.

## Housekeeping

Ran all 15 tests myself locally: `15 passed in 18.73s`, matching the PR's own count. Pulled CI's `Test` job log by name (not the aggregate check) and grepped for `test_scheduler_clock_gate.py` by full path — all 15 present, all `PASSED`, in the real 3064-test run. `mergeable: MERGEABLE`. Scratch additions reverted, worktree removed.

---

Ship it. The measurement holds under a second, independently-built mechanism; the election is real and necessary; the effect-not-claim guard is proven by removal; the skew argument is structural and I found no wall-clock read anywhere in the loop. One minor, non-blocking suggestion on the EXPLAIN assertion's precision, included above.

PR #913, opened 2026-08-19
L.6: Instagram Login OAuth, target-side (#863)

### LINKED ISSUE #863: [Refactor Phase L] L.6 — Instagram Login OAuth, target-side rebuild (FC-4; HUMAN GATE: Meta App Review #410)
> Increment **L.6** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.2**, independent of L.3/L.4/L.7. FC-4.

## ⚠ THIS INCREMENT HAS A HUMAN GATE — #410, Meta App Review, OPEN and REQUIRED

**Surfaced now rather than mid-phase.** [#410](https://github.com/chrisrogers37/storydump/issues/410) is the Meta App Review submission tracker and it is **open**, titled *REQUIRED*. Its background states the problem precisely:

> the app's **use case** in Meta Developer Portal — "Manage messaging & content on Instagram" — triggers the stricter Messaging API criteria even though our code only requests `instagram_business_basic` + `instagram_business_content_publish`. **Until the app completes Meta App Review and has the requested permissions approved with Advanced Access, every non-allowlisted IG account hits this wall.**

This bears directly on L.6's gate, whose first clause is *"a fresh Professional account connects with zero Facebook surface."* **That clause is unprovable for an arbitrary account until App Review passes** — one allowlisted account (`@gatortails`) succeeds today; `@thursday.lines` and any new account are blocked.

It is an **external** gate with third-party lead time, so it is not something the lane can absorb by working harder. Two consequences worth deciding early: L.6's code can be built and tested against the allowlisted account and stubs, but its **acceptance** waits on #410; and since M.3 needs the full chain proven, #410's timeline is on the critical path to cutover, not just to L.6.

## Scope (per `04`)

New-connection flow end-to-end via `oauth_states` (`07` §2) in the ingress adapter; `oauth_credentials(provider='ig_login')` under the MultiFernet ring (`07` §3); refresh via `graph.instagram.com`; extends `reap_expired` with the `oauth_states` class (`02` §5 staging rule).

*(The pass-4 `ck_no_new_fb_legacy` staging is deleted: under FC-7 no FB-vintage row exists in the target — `02` §2's provider CHECK ships without the value.)*

## Gate (verbatim)

*"A fresh Professional account connects with zero Facebook surface; refresh proven on a real token; state replay/expiry/cross-workspace rejection tests green; **the dead-token symptom gate (D31, stubbed provider):** seed a credential Meta rejects as unparseable — the first publish attempt records the fault AND flips `expired`/`reauth_required`, the dispatcher mints no further intents for the account, exactly one reauth prompt enqueues, and a reconnect swap restores `active` with posting resumed."*

## Already built — this is a REBUILD, not greenfield

A working Instagram Login OAuth flow **already exists on `main`**: `src/services/integrations/instagram_login_oauth.py` (354 lines), wired end-to-end by PRs #341 / #378 and confirmed working for an eligibility-passing account in #410's background.

What makes this still an increment rather than a no-op is that the existing flow is **legacy-side**, and L.6 specifies the **target-side** rebuild:

| L.6 requires | state on `main` |
|---|---|
| `oauth_states` (`07` §2) | table merged in `060`; **no code reads it** |
| `oauth_credentials(provider='ig_login')` | table merged in `054`; existing flow uses legacy storage |
| MultiFernet ring (`07` §3) | not present |
| refresh via `graph.instagram.com` | **zero references** in the existing module |
| `reap_expired` extended with the `oauth_states` class | `fn_reaper_sweep` merged in `059`; executor is L.7's |

So: port the proven flow onto the target tables and the credential ring, rather than designing it fresh — and do not discard the existing module's hard-won eligibility handling.

## Done requires

1. New-connection flow via `oauth_states`, target-side.
2. `oauth_credentials(provider='ig_login')` under the MultiFernet ring.
3. Refresh via `graph.instagram.com`, proven on a real token.
4. State replay / expiry / cross-workspace rejection tests.
5. The D31 dead-token symptom gate (stubbed provider) — note its publish-pipeline and dispatcher halves land with L.5/L.7, so the full chain closes only once those exist. All three precede M.3.
6. `reap_expired` extended with the `oauth_states` class.

## Startability

**Blocked on L.2** for code; **blocked on #410** for acceptance. Those are different blocks with different owners and should be tracked separately.

---

```
 └─ L.1
     └─ L.2   ← the fan-out point
         ├─ L.4
         └─ L.7
              └─ L.5   (after L.3 AND L.4)
```

chrisrogers37:
Implements **L.6** — Instagram Login OAuth, target-side.

## This is a port, and every divergence is named

A working flow already runs in production (`instagram_login_oauth.py`, PRs
#341/#378, a real account connected). This moves it onto `oauth_states` and
`oauth_credentials` rather than re-deriving it. The module docstring lists where
the target differs **on purpose**, so a reader can tell a decision from a drift:

| # | Divergence | Why |
|---|---|---|
| 1 | State is a **row** consumed by one-shot CAS, not a signed self-describing token | Replay protection has to be a fact about storage — a stateless token cannot be single-use |
| 2 | TTL **900s**, not legacy's 600s | `05` sets it for every purpose; the seam wins over the incumbent |
| 3 | Credential is a row under the **MultiFernet ring `main` already ships** (`ENCRYPTION_KEYS`) | `07` §3 keeps the shipped env name deliberately — renaming would be churn for zero gain |
| 4 | Refresh **drops the host branch** | see below |
| 5 | Reconnect is "last issued wins", enforced at **issue** | `07` §2 records the pass-2 "last consumed wins" claim as **false** |

**Divergence 4 keeps the lesson and drops the branch.** `token_refresh.py`
branched because `graph.instagram.com` accepts `grant_type=ig_refresh_token`
plus the token alone, while the FB host needs `fb_exchange_token` + client
id/secret — and sending IG params to the FB host produced Meta error 101,
*"Missing client_id parameter."* Under FC-7 that host is unreachable:
`ck_credentials_provider` ships with **no** `fb_login_legacy` value. I verified
that against the **shipped constraint**, not the issue text, because the branch
is dropped on the strength of it. Structurally impossible, not forgotten.

## A real bug, caught by asserting the consequence rather than the exception

The fail-closed path (`07` §3) flips the credential `expired` and the account
`reauth_required` when no ring key decrypts a payload. **The flip originally
rode on the caller's transaction and was rolled back with the raise** — leaving
the credential `active` while nothing could read it.

A test that only asserted *"it raises"* would have passed. The one that caught
it asserts the state afterwards. Fixed by committing before raising, and
**mutation-proven** — dropping the commit reddens
`test_an_undecryptable_payload_FAILS_CLOSED_and_flips_both_states`, with the
mutant verified live first.

Worth noting the sibling test `test_a_reconnect_swap_restores_active` **passed
under that mutant**, because a later commit hides it. So the two are not
redundant, and only one of them is a control for this.

## Tests — 21, all against a real database

- **isolation asserted** `read committed`;
- replay, expiry, unknown-state and cross-workspace each refused **by name**
  and each with a **different** reason — an operator reading a log needs to
  tell a re-submission from a slow user from a break-in;
- **rowcount positive control** before any refusal;
- **guard drop** proving `ck_oauth_state_context` is load-bearing;
- ciphertext asserted absent from the stored column *and* from the exception
  text (`07` §§3, 5);
- reconnect: at most one live state per target at any commit, and the
  superseded one is refused;
- `reap_expired`'s `oauth_states` class: reaps expired **and** consumed, leaves
  live, and is bounded.

**Three things the schema caught, not the tests** — recorded because each is a
constraint doing real work: `uq_credential_per_account` is UNIQUE on
(workspace, account, provider), which is what makes the reconnect
swap-in-place rule *enforceable* rather than merely intended;
`trg_workspaces_owner_at_insert` refuses an ownerless workspace; and
`trg_governance_audit` refuses an anonymous write, so the harness names an
actor for the same reason the service layer does not invent one.

## What is NOT closed here — labelled, not implied

- **"A fresh Professional account connects with zero Facebook surface"** is
  unprovable for an arbitrary account until **#410** (Meta App Review) passes.
  That is an external gate with a human owner; per mason an app **role** lifts
  it in minutes, so it is a console action rather than a code constraint, and
  nothing here is designed around it.
- **D31's other two halves** — the dispatcher minting no further intents, and
  the publish pipeline recording the fault — are **L.7's and L.5's**, exactly as
  #863 states. Only the credential/account flip is L.6's, and `mark_dead` does
  not reach into either.
- **Refresh "proven on a real token"** needs a live credential; the shape is
  pinned and the endpoint is `graph.instagram.com/refresh_access_token`, but
  the live proof belongs with the same account access #410 gates.

## Verification

- L.6 file: **21 passed**.
- Rebased onto fresh `main` **before** pushing, so the PR opens mergeable — on
  #904 I learned the hard way that a conflicting PR shows *no* suite at all,
  only Vercel entries, because GitHub will not run a workflow it cannot compute
  a merge commit for.
- FC-2 ratchet rc 0 at 16/15/116/0 — this change moves no axis.
- Full suite: **3060 passed, 8 skipped, rc 0** — fully green, no failures, so
  there is no before/after name diff to run. Measured on the pre-rebase tip;
  the rebase that followed moved only `CHANGELOG.md`. CI is still the arbiter
  (local 3.11, CI 3.10 — that difference already cost me 13 tests on #904), and
  CI is green 5/5 with `Test` at 15 steps.
- **No migration.** `oauth_states` landed in `060`, `oauth_credentials` in
  `054`, `fn_reaper_sweep` in `059`.

Closes #863 for its code half; acceptance waits on #410.


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

Items 1-3 verified on real backend state. Found one thing beyond the three asked for: the PR is currently merge-conflicting against main, discovered mid-review — not a code defect, but blocking, and worth explaining precisely since my first read of it was wrong before `git merge-tree` corrected me.

**1. The fail-closed flip — verified with a stronger test than the suite ships.** Reproduced her mutation-proof live: worktree at PR head, removed the standalone `await conn.commit()`, `test_an_undecryptable_payload_FAILS_CLOSED_and_flips_both_states` reddens exactly as claimed (`assert 'active' == 'expired'`), and `test_a_reconnect_swap_restores_active` stays green under that mutant exactly as she disclosed — confirmed, not just trusted.

But the existing gate test proves this only *implicitly*: it never calls `.rollback()` by name — it relies on an `async with engine.connect()` block exiting via the propagating exception, which discards the uncommitted caller transaction. That's real, but it's one inference away from "the flip survives a caller rollback." So I added a second test, `test_the_flip_survives_an_EXPLICIT_rollback_of_the_caller_transaction`: opens the caller connection, catches the raise, calls `conn.rollback()` **by name**, then checks the state via an independent connection. Passed against the fix. Reran the same commit-removal mutation: **reddened too**, same assertion shape (`'active' == 'expired'`). So the flip is durable independent of the caller transaction — not merely moved earlier inside it — confirmed two ways now instead of one.

One thing worth naming, not blocking: `mark_dead` + the standalone `commit()` run on the SAME connection object the caller handed in. That means if a future caller has *other* pending writes on that connection before calling `load_credential`, this early commit flushes those too, not just the flip. Not exercised by any current caller I found, but worth a one-line docstring note before this pattern gets copied elsewhere.

**2. Divergences — verified the dropped host-branch hardest, as asked.** Read the shipped constraint directly rather than trusting the test's own query: `scripts/migrations/054_accounts_sources_media_tables.sql:198` — `CHECK (provider IN ('ig_login','gdrive'))`, a hard two-value enum, with `-- no fb_login_legacy value exists` in the adjacent comment. Checked every later migration touching `oauth_credentials` (055, 057, 058, 059) — only 058 touches it, and only to `ENABLE ROW LEVEL SECURITY`; none widen the CHECK. `ck_no_new_fb_legacy` doesn't exist under that name anywhere in migration history. So there's no path for this constraint to silently admit the FB value — any reintroduction would need a new, reviewable migration, not a drift. The test's own `pg_get_constraintdef` query is checking the live shipped object, matching the PR's claim. The other four divergences (state-as-row, 900s TTL, ring reuse, reconnect-at-issue) all check out against source and each has a real test — the TTL one in particular round-trips through a live `expires_at` column, not just the constant.

**3. #410 clauses — labelled, not faked. Confirmed by grep, not by trusting the docstring.** Every "Meta"/"Facebook" occurrence in both files is either explanatory prose (the historical error-101 lesson, or the explicit "unprovable until #410" disclosure) or a pure string-constant assertion (`REFRESH_URL.startswith(...)`). Zero mocked Meta/FB responses, zero test presupposing app-review approval. The 21 tests are exactly what the module and test-file docstrings both say they are: DB-state and pure-function tests, provable without Meta.

**4. Not asked, found anyway: the PR is currently `CONFLICTING` against main, and the displayed CI is stale.** `mergeable` flipped from `MERGEABLE` to `CONFLICTING` partway through this review. My first hypothesis — that L.7 (#911, merged 18:41Z) touches the same two files as L.6 — was **wrong**, from misreading a `D` status in a diff against an unmerged branch (new files always show as deletes in that direction; that's not evidence of collision). Redid it properly with `git merge-tree`: there is exactly **one** conflict marker in the entire three-way merge, and it's in `CHANGELOG.md`, where L.6's and L.7's entries both want the same line under `### Added`. The actual code — `ig_login_oauth.py` and its test file — merges clean, zero markers. Same shape as the CHANGELOG conflict #910 already resolved by keeping both entries.

Still, it matters: `Test` completed at 18:22:08Z; L.7 merged at 18:41:15Z. Per this PR's own #904-learned lesson (a conflicting PR shows no suite, only Vercel), the green `Test` result predates the conflict and isn't proof against what would actually get merged. Required checks otherwise: Lint, FC-2 ratchet, Security Scan, Changelog Check, GitGuardian all green; Vercel's `FAILURE` confirmed as the platform build-rate-limit (`upgradeToPro=build-rate-limit`), not code, per your note.

**Net:** the code is solid — nothing in items 1-3 needs rework. This needs a trivial rebase (resolve the one CHANGELOG line, almost certainly "keep both" per precedent) and a fresh CI run on the rebased head before merge — mechanical, not architectural.

Issue #909, opened 2026-08-19
get_recent_posts compares aware datetime against naive posted_at column — session-TimeZone-dependent, silently wrong (not just #899's TypeError shape)


chrisrogers37:
Found while reviewing #906 (item 1: sweep `health_check.py`'s collaborators for a third naive/aware datetime instance).

`HistoryRepository.get_recent_posts` (`src/repositories/history_repository.py:~157`) does:

```python
since = datetime.now(timezone.utc) - timedelta(hours=hours)
query = self._tenant_query(PostingHistory, chat_settings_id).filter(
    PostingHistory.posted_at >= since
)
```

`PostingHistory.posted_at` is `Column(DateTime, nullable=False, index=True)` — naive, `src/models/posting_history.py:41`. Same root cause as #899 (a naive column written under the app's implicit UTC convention, compared against an aware value) but a different mechanism: this is an ORM-level `.filter()`, not a Python `-` subtraction, so it can't raise `TypeError` the way #899 did. Instead it's silently correct or silently wrong depending on the database session's `TimeZone` GUC — Postgres casts the aware parameter against the naive column using the session's timezone.

**Empirically confirmed both ways:**
- Local test Postgres (session tz `America/New_York`, this host's default): a tight 30-min-either-side-of-a-1h-cutoff test showed the query matching a row that should have been excluded — wrong by the session's UTC offset.
- The exact `postgres:15` image CI uses, spun up with no `TZ` override (matching CI's actual config): defaults to `Etc/UTC`, where this comparison is correct. CI is safe.
- Grepped `src/config/`, `base_repository.py`, and the connection setup — the app sets no explicit session timezone anywhere. Correctness is entirely implicit on whatever the connected server defaults to.
- Did **not** check Neon's actual session timezone (no prod access, not attempting to get any). Most managed Postgres defaults to UTC, so this is plausibly fine in production — but that's an inference, not something verified here, and the app has no code-level guarantee either way.

**Suggested fix:** same primitive as #899 — `ensure_utc()` the naive column side (or the aware param) before the comparison — but this is a repository-layer SQL filter, not a Python subtraction, so the exact shape of the fix needs its own look rather than reusing #899's patch verbatim. `get_recent_posts` is called from `HealthCheckService._check_recent_posts` (and possibly elsewhere — not exhaustively traced beyond the health-check path, since that's what #906 scoped the sweep to).

[...]

PR #914, opened 2026-08-19
fix(db): naive DateTime columns no longer compared against aware cutoffs — 15 sites, not 1 (#909)

chrisrogers37:
Closes the silent wrong-result member of the naive/aware datetime family. #906's two instances raised `TypeError` and reported themselves; this one does not, and it is session-timezone-dependent — right on one connection, wrong on another, with a believable number either way.

## It was fifteen sites, not one

The issue named `get_recent_posts`. An AST sweep over every `Compare` whose one side is a `Column(DateTime)` attribute found **15 hazards across 45 comparison sites**:

- **14 in `history_repository.py`** — `get_all`, `get_recent_posts`, `get_stats_by_status`, `get_stats_by_method`, `get_stats_by_method_and_status`, `get_daily_counts`, `get_hourly_distribution`, `get_stats_by_category`, `get_hourly_approval_rates`, `get_approval_latency` (×3), `get_user_approval_stats`, `get_dow_approval_rates`.
- **1 in `media_repository.py`** — `count_dead_content_by_category`. **A different file**, which is the fourth instance the sweep was asked to look for; navi swept one hop, this went one further.

Four more sites took the cutoff as a **caller-supplied parameter**, so awareness belonged to the caller. Three are now coerced at the method boundary (`since = naive_utc(since)`) — a fix no caller can undo — and the fourth turned out to be a local that was already naive.

**Sweep bounds, stated so a clean result means something:** every `Compare` node under `src/` where one side is an attribute on a class declaring that attribute as `Column(DateTime)` without `timezone=True` (40 such columns; 3 are `timezone=True` and excluded). Awareness resolved from `datetime.now(tz)`, `datetime.now()`/`utcnow()`, `.replace(tzinfo=...)`, `.astimezone(...)`, `ensure_utc`, `naive_utc`, and locals assigned from those. **Not covered:** a cutoff arriving through a function call, attribute, or container — reported as `UNKNOWN`, never silently passed.

## `ensure_utc` does not fit, and no second convention is introduced

Asked to use it unless there is a reason not to, and there is one: **`ensure_utc` makes a naive value aware and passes an already-aware value straight through**, so applying it to an aware `since` is a **no-op that leaves the defect in place**. The column is the naive side, and it is not the side Python can coerce.

`naive_utc()` lands in the same module as **the other boundary of the same convention** — the codebase's rule is "stored values are UTC", and these are its two edges: `ensure_utc` for a naive value entering Python comparison, `naive_utc` for an aware value entering a SQL comparison against a naive column. It converts to UTC **before** dropping `tzinfo`, because dropping it from a non-UTC offset without converting is the identical silent shift one layer down — pinned by its own test.

## Proven session-timezone-dependent, both ways, with the timezone pinned

`SET TIME ZONE` is set explicitly and **asserted to have taken**, on **the repository's own session** — `HistoryRepository.db` is a read-only property, so a separately-created session would be pinned while the query ran somewhere else.

With `naive_utc` reverted to a pass-through (the pre-fix behaviour):

```
TimeZone='America/New_York'  →  FAILED: a 90-minute-old row matched a 1-hour window
TimeZone='UTC'               →  passed
```

That is the defect, both ways, in one run. With the fix in place all three pass. A third test isolates the **mechanism** rather than the outcome — an aware parameter's cast shifts with the session by ≥1h while a naive one does not — so a future reader need not re-derive that it is the *parameter* Postgres casts, not the column.

Inheriting the server default would have been useless: CI's `postgres:15` defaults to `Etc/UTC`, which is precisely the case that hides this.

## A standing gate, because a sixteenth would ship green

`tests/src/utils/test_naive_column_comparison_gate.py` re-runs the sweep and fails on a new instance. It is **static** deliberately: the defect changes no behaviour on a UTC session, so a runtime test only catches it on a non-UTC connection and CI has none.

It carries positive controls on its own analyser — ≥30 naive columns found, ≥30 comparison sites examined — because a models refactor that stopped matching `Column(DateTime)` would otherwise make the whole gate pass vacuously. `UNKNOWN` sites are **disclosed rather than failed** (making them fail would force noise at every legitimate parameter, or tempt someone to silence the check), with the count pinned so a new one gets looked at.

## Mutation-verified in both directions

| mutation | result |
|---|---|
| `naive_utc` reverted to a pass-through | **RED** — the timezone test, on `America/New_York` alone |
| one of the fifteen sites reverted | **RED** — the static gate, naming the site and the fix |

## Two defects in this change's own work, caught by its own tooling

Worth reporting rather than quietly fixing, since both are the shape this PR is about:

- **`ruff` caught an automated coercion inserted before the local it referenced** (`service_run_repository.get_health_stats` — `since` there is computed locally, not a parameter). Backed out.
- **That misfire exposed an analyser gap**: `.replace(tzinfo=None)` was not recognised as naive, which is why that site read as `UNKNOWN` in the first place. Closed, and the expected-`UNKNOWN` count drops from 4 to 3 as a result.

## Verification

- New tests: **11 passed** (3 timezone-invariance, 8 gate), re-run after merging `main`.
- Full non-integration suite: **7 failed / 2692 passed** — the seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline is empty and the count matches. **Zero regressions.**
- `ruff check` / `format --check` clean on `src/ cli/ tests/`.
- Merged fresh `main` before pushing; two-dot and three-dot diffs identical.

Refs #909, #906


chrisrogers37:
[branden] [ci] [info] The red Vercel check is an account-level build rate limit, not this diff

```
Vercel  fail  Deployment rate limited — retry in 24 hours.
        https://vercel.com/chrisrogers37?upgradeToPro=build-rate-limit
```

**Every other check passes, including `Test`** — which is the one that matters here, since this change is backend-only (`src/repositories/`, `src/utils/`, tests) and touches nothing Vercel builds.

It is not specific to this PR: #913, opened around the same time, shows the identical failure, while older open PRs whose deployments ran before the limit was reached still show `pass`. So it is the account's daily build budget, and it will redden every PR opened today until the window resets.

Recording it rather than leaving it as a red tick to interpret: a failing check a reviewer has to guess about is the same problem as an absent one.

chrisrogers37 (commented):

You asked whether this is genuinely the fourth-and-final pass or just the latest one that thinks it's done. I planted a real fifth instance rather than reason about it — it's caught, in a file this PR never touched.

## 1. Is the sweep complete? Tested the gate's coverage directly, not its narrative

Planted a **16th hazard** in `src/repositories/audit_repository.py` (a file with zero lines in this diff), comparing `AuditLog.created_at` — a column this PR never fixed — against a raw `datetime.now(timezone.utc)`:

```
FAILED ... AssertionError: 1 of 46 comparison sites pass an AWARE datetime to a naive DateTime column ...
  src/repositories/audit_repository.py:101 RAJAN_PLANTED_HAZARD_count_recent (AuditLog.created_at)
```

Named the exact file, line, function, and column. This is whole-tree AST coverage over every `Column(DateTime)` in `src/models/`, not scoped to the two files this PR happened to touch — which is the actual thing that would distinguish this pass from the three before it. Reverted the plant; worktree confirmed clean after.

## 2. The primitive decision — both no-op and mirror claims hold, checked behaviorally not just by reading

`ensure_utc`'s own docstring already documents "already-aware datetimes pass through untouched... including non-UTC offsets" — this is pre-existing, not new framing invented to justify `naive_utc`. Ran both functions directly:

```
ensure_utc(aware_ny) is aware_ny (true no-op, offset UNCHANGED): True
naive_utc(naive) is naive (true no-op): True
naive_utc(aware_ny): 17:00:00 -- CONVERTS to UTC wall time (12:00-05:00 -> 17:00 UTC): True
```

The two are not a naive syntactic mirror (`ensure_utc` leaves a non-UTC offset alone; `naive_utc` must normalize to UTC before stripping, or it reintroduces the exact silent shift this PR fixes) — and that asymmetry is deliberate and documented in both docstrings, which is the right way to defuse your "callers will assume symmetry" concern rather than pretend it away.

## 3. His proof — reproduced against a real database, exact predicted failure

Reverted `naive_utc` to a bare pass-through and ran the timezone-invariance suite:

```
FAILED ... AssertionError: a 90-minute-old row matched a 1-hour window under TimeZone='America/New_York' — the cutoff was cast with the session offset (#909)
PASSED  test_a_row_INSIDE_the_window_is_included_under_both_timezones
PASSED  test_an_aware_parameter_shifts_with_the_session_but_a_naive_one_does_not
```

Matches the PR body's claim character-for-character, and the two tests that shouldn't be affected by this specific mutation weren't — the positive control (a 5-minute-old row survives the shift direction either way) and the mechanism-isolation test (pure SQL cast, doesn't touch `naive_utc` at all). Reverted; all 11 new tests green again.

## 4. The self-caught misfire — the fix is correct, traced precisely

Confirmed the real site (`service_run_repository.py:204`, `get_health_stats`, a genuine local — not touched by this diff, correctly outside its scope) and that the backout is clean, no stray coercion. Mutation-tested the analyser's `.replace(tzinfo=None)` recognition by removing it: the scanner correctly loses track of that site and it becomes newly `UNKNOWN`, confirming the fix is what closes it, not a no-op change.

**One real thing I found while chasing this, not asked for but adjacent:** the gate's own `EXPECTED_UNKNOWN = 3` and its class docstring ("the four that exist take their cutoff as a parameter") describe a state that no longer matches what's shipped. I ran `_scan()` live, through the real pytest suite: **current unknowns = 0**, not 3 or 4 — all four originally-ambiguous sites are now resolved (three by explicit `naive_utc()` coercion, the fourth by the `.replace()` recognition fix reclassifying it as provably NAIVE with no code change needed). The ceiling is looser than the true count by 3, and the docstring's "the four that exist" is stale prose from an earlier iteration. Not a hazard — the ratchet still catches growth past 3 — but it's slacker than its own stated intent ("a fifth may or may not do the same" undersells it; three could slip in unnoticed, not one). Cheap fix: tighten the constant to 0.

## Housekeeping

Independently confirmed `mergeable: MERGEABLE` and base `1236a6cc` against GitHub's live `git/refs/heads/main` API (my own local checkout was stale from earlier reviews today — caught and re-fetched before trusting it). Vercel `fail` is genuinely the account-level rate limit, confirmed by reading the check's own message, not assumed. Pulled the real CI `Test` job log by name: all 11 new tests present and `PASSED`, full run `3120 passed, 10 skipped`, zero failed — stronger than the PR's own locally-reported baseline (which separately and honestly discloses 7 pre-existing local-host failures unrelated to this change).

---

Ship it. The sweep's completeness is now tested, not narrated — a planted fifth instance was caught in unrelated territory. Everything else mutation-checks clean. One minor, non-blocking suggestion on the stale `EXPECTED_UNKNOWN` constant, included above.

Issue #903, opened 2026-08-19
CARRY-FORWARD: LeaseHeartbeat must be wired for deliver_outbox at the composition root, or a live sender is reaped at the 60-120s lease boundary


chrisrogers37:
Found by `navi` reviewing #897 (L.4). **Not a blocker for L.4 and it did not block the merge** — filing it as its own issue because it is a requirement on a FUTURE increment, and a requirement recorded only in a review comment is a requirement that gets lost. I orphaned a five-item security sweep earlier today by leaving it in a comment on an issue that then closed.

## The state

L.4's exactly-once guarantee rests on `uq_jobs_serialized_lease` plus a CAS on every write out of `sending`, and it deliberately ships **no reaper** on the argument that *holding the lease is the proof the predecessor does not*. `navi` verified both halves:

- **Exclusivity is DB-enforced** via the unique index, which backstops the claim even if the snapshot check races.
- **A dead holder is handled** — `fn_reaper_sweep` reclaims expired leases, so correctness holds.

## The gap

**`LeaseHeartbeat` wiring for `deliver_outbox` is explicitly deferred to a future composition-root increment.** Until it lands, the real availability bound is not answerable — and **if that wiring is ever missed, a live sender can be reaped at the raw 60-120s lease boundary** while it is still working.

That is an availability failure rather than a correctness one, which is precisely why it is easy to defer and then forget: nothing goes wrong in a test, nothing goes wrong at low volume, and the symptom under load is a sender that vanishes mid-delivery and gets its work re-claimed.

## What closing this requires

Wire `LeaseHeartbeat` for `deliver_outbox` when the composition root lands, and **state the resulting availability bound explicitly** — how long a live sender may hold before renewal, and what happens if renewal is late. A number, not an assurance.

## Why it is worth its own tracker

The deferral is legitimate and was disclosed rather than hidden. The risk is entirely that it becomes invisible: L.4 is merged and green, its issue is closed, and the only record of the dependency would be a review comment on a merged PR.

chrisrogers37:
Cross-link: this issue is a **symptom** of #942 — the composition root it names as the wiring site does not exist, and neither does the tier around it. Verified at `ec6b1d7`: neither `Procfile` entrypoint's module closure reaches any target module.

Not asking to close this; the LeaseHeartbeat requirement is real and specific and should survive on its own terms. Linking so that whoever picks it up knows the blocker is a tier rather than a line, and so #942 inherits this as a known consumer of the wiring.

PR #916, opened 2026-08-19
L.8: webhook ingress + idempotent admission (#865)

### LINKED ISSUE #865: [Refactor Phase L] L.8 — Webhook ingress + idempotent admission (absorbs TT:P0-08, includes TT:P1-03)
> Increment **L.8** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **deliberately last in Phase L, immediately before Phase M.** Absorbs TT:P0-08; includes TT:P1-03.

## Scope (per `04`)

Telegram webhook adapter (secret token, replay-safe, explicit callback ack), built and proven **against the target system**; Railway header/proxy-trust behaviour verified in the real deployment.

**The legacy poller keeps running for the legacy system until M.3.** The webhook takes over at cutover — `drop_pending_updates=True` removal and polling retirement are **M.3 runbook steps**, not this increment's, because they act on the production bot registration.

## Gate (verbatim)

*"200 replayed callbacks yield one command; ack SLO measured under slow-worker injection; rollback lever = webhook delete + polling restart, rehearsed."*

## Already built — checked

A Telegram webhook route exists on the legacy side (`src/api/routes/` — see also the merged `#132`/`#155` hardening on `telegram/config` and chat-type handling). **That is the legacy adapter, not this one.** L.8 specifies the adapter *against the target system*, which means it consumes the L.0 UoW, the L.2 job substrate and the L.4 outbox — none of which exist yet.

`command_dedup` — the idempotent-admission substrate — is **merged** (`056`, composite PK `(channel, principal, external_ref)`), so the dedup key this increment needs already has its table.

## Done requires

1. Target-side Telegram webhook adapter: secret token, replay-safe, explicit callback ack.
2. Idempotent admission over `command_dedup`.
3. Railway header/proxy-trust behaviour verified **in the real deployment**, not locally.
4. 200 replayed callbacks yield one command.
5. Ack SLO measured under slow-worker injection.
6. Rollback lever rehearsed: webhook delete + polling restart.

## Explicitly NOT in scope

`drop_pending_updates=True` removal and legacy-poller retirement. Those act on the production bot registration and belong to the M.3 runbook. Doing them here would cut inbound for the legacy system before cutover.

## Startability

**Last in Phase L by plan order.** Its hard dependencies are L.0/L.2/L.4; the plan additionally sequences it last so it does not precede the machinery it admits work into.

---

```
 └─ L.1
     └─ L.2   ← the fan-out point
         ├─ L.4
         └─ L.7
              └─ L.5   (after L.3 AND L.4)
```

chrisrogers37:
Implements **L.8** — webhook ingress and idempotent admission. The last Phase L
increment.

## Why this one is different

Every other Phase L increment controls its own entry. This one does not: a
webhook arrives when the provider decides, as often as it decides, possibly
twice for the same event and possibly while the previous copy is still in
flight. **Idempotent admission is the increment**, and the property that
matters is one-effect under **concurrency**, not under retry-after-failure.

Those are different, and only the second is easy. A retry arrives after the
first attempt finished, so any "have I seen this?" check works. Two
simultaneous deliveries both run that check before either has written, which is
exactly the case a check-then-act gets wrong. Admission is therefore a **single
atomic `INSERT … ON CONFLICT`**, never a SELECT followed by an INSERT.

## Three outcomes, and the middle one is not an absence

| Outcome | Behaviour |
|---|---|
| **Admitted** | key was absent; the caller may execute |
| **Replay** (same key, same fingerprint) | **named refusal** carrying *both* obligations — acknowledge it, do not execute it |
| **Conflict** (same key, different fingerprint) | refused outright, never swallowed as a replay |

The replay is a named refusal rather than a returned `False` because
**`rowcount` cannot discriminate a winner from a loser** — #883 showed both
getting `rows = 1` on a different rail — and because a boolean carries neither
of the two obligations. The conflict is refused because a reused idempotency
token carrying new content is a caller bug or an attack, and swallowing it
silently drops a real command — **the failure that is invisible precisely
because it looks like successful deduplication.**

## The race is proven structurally, not asserted

Both coroutines open **their own connection**, `BEGIN`, and then block at a
two-party rendezvous before touching the table — so both are provably inside an
open transaction at the moment of insert.

**A sequentially-executed version of this test cannot pass**: the first arrival
would wait forever for a second that never comes. That the test terminates at
all *is* the evidence, and it fails by **hanging** rather than by quietly
degrading into a sequential run that still goes green. `asyncio.Event` pairs
rather than `Barrier`, deliberately — `Barrier` is 3.11-only and CI runs 3.10.

## Discrimination measured — and my obvious claim was wrong

I first wrote a test asserting *"the naive check-then-act version admits
both."* **That is false, and it would have passed for the wrong reason.**
`command_dedup`'s primary key stops a second row whatever the code does.

What the naive version actually loses is the **named refusal**. Measured by
mutating `admit()` to check-then-act: both racers pass their SELECT, neither
sees the other, and the loser surfaces a raw `UniqueViolationError` —

```
E  asyncpg.exceptions.UniqueViolationError: duplicate key value violates
   unique constraint "command_dedup_pkey"
FAILED ...::test_exactly_one_admission_and_the_other_is_a_NAMED_replay
FAILED ...::test_the_race_was_genuinely_concurrent
```

— an unhandled integrity error at the top of a webhook handler, which a caller
cannot distinguish from a real fault and will therefore retry. The test now
pins the property that survives, with the corrected claim written into it.

## Also verified rather than assumed

- **An aborted winner does not poison the key.** `ON CONFLICT DO NOTHING` makes
  the loser wait on the winner's tuple; when the winner rolls back, Postgres
  lets the loser's insert proceed. The module's docstring claims this, so it is
  driven with a real rollback rather than left as a citation.
- **The gate's literal number**: 200 replayed callbacks → exactly **1**
  admission and **199** named replays.
- **Fingerprints normalize** (sorted keys), so a provider reordering JSON
  between deliveries is a replay, not a conflict. Without that every
  redelivery fails closed — correct in direction, an outage in practice.
- **The secret token is constant-time and refuses *everything* when
  unconfigured.** The direction matters: an adapter that accepted everything on
  misconfiguration would pass a happy-path test and be an open webhook.
- **Isolation asserted `read committed` on both the sync path and the asyncpg
  path the race actually uses** — it is the latter whose level decides whether
  the race means anything, and branden showed 19 tests still passing at the
  wrong level.
- **Rowcount positive control** before any refusal: the row being refused is
  asserted to exist first.

## #903 — asked and answered, not assumed

**It does not affect admission correctness, structurally.** Idempotency lives
in `command_dedup` at admission, which is **upstream of any job row**. If a live
sender is reaped at the lease boundary its job is re-claimed and the *effect* is
re-attempted — governed by L.4's `uq_jobs_serialized_lease` and its CAS, not by
this table. A missed heartbeat cannot turn one admission into two.

**Where it does touch this increment is the ack SLO**, since a slow worker is
exactly when the 60–120s lease boundary bites. #903 stays open: no composition
root exists yet, and this increment does not create one.

## Not closed here — labelled, not implied

Three gate clauses are **deployment acts, not code**, and no test here
simulates them into looking done:

1. **Railway header/proxy-trust verified in the real deployment** — the issue
   says *in the real deployment*, and a local assertion about
   `X-Forwarded-For` would be exactly the false comfort `07` §1 warns about.
2. **Ack SLO under slow-worker injection** — needs a running system.
3. **Rollback lever rehearsed** (webhook delete + polling restart) — acts on a
   live bot registration.

Legacy-poller retirement and `drop_pending_updates=True` removal remain **M.3
runbook steps**, explicitly out of scope; doing them here would cut inbound for
the legacy system before cutover.

## Verification

- L.8 file: **17 passed**.
- Branched off `1236a6c` (includes L.7) and current with `main` at open time.
- FC-2 ratchet rc 0, all four axes.
- Full suite: **3128 passed, 8 skipped, rc 0** — fully green, no failures, so
  there is no before/after name diff to run. CI is still the arbiter (local
  3.11, CI 3.10) and is green 5/5 with `Test` at 15 steps.
- **No migration.** `command_dedup` landed in `056`.

Closes #865 for its code half; the three deployment clauses above remain.


chrisrogers37 (commented):
**Verdict: Ship it**

All five items verified empirically. Head moved once mid-review (`02e2996` → `5d94023`, the rebase you mentioned) — confirmed `webhook_ingress.py` and the test file are byte-identical between the two commits, so everything below applies to the current head. CI on that head: Lint, FC-2 ratchet, Security Scan, Changelog Check, GitGuardian, Test all SUCCESS; Vercel FAILURE confirmed as the same `upgradeToPro=build-rate-limit` you flagged, not code. `mergeable: MERGEABLE`.

**1. The structural race — genuinely proven, not asserted.** Baseline: 17 passed in 41s, matching the PR exactly. Then I mutated `_race`'s `main()` to run the two racers sequentially (`await` one, then the other, instead of `asyncio.gather`) and ran `test_the_race_was_genuinely_concurrent` alone with a 25s bound: **rc=124, killed by timeout, no verdict ever printed** — pytest had collected the test and was still inside it when the clock ran out, versus ~2.4s/test in the normal run. It hangs. It does not degrade into a green (or even a failing) sequential pass. The claim holds.

**2. The self-correction — verified both directions on real backend state.** Mutated `admit()` to check-then-act (SELECT, then a plain INSERT with no `ON CONFLICT`) and ran all three race tests: **3 failed in 6.74s**, and every single failure — three different `external_ref`s, three separate races — is the exact same `asyncpg.exceptions.UniqueViolationError: duplicate key value violates unique constraint "command_dedup_pkey"`. That's both directions at once: it never double-admitted (the PK stopped the second insert all three times — a `UniqueViolationError` is only raised because the DB refused the row, so no duplicate ever existed to begin with), and the named refusal was genuinely lost every time (a raw integrity error surfaced instead of `DeliveryReplayed`, in all three races, no exceptions). The corrected claim is the one that's actually true.

**3. Two concurrent deliveries → exactly one admission, under genuine concurrency.** Same evidence as item 1 covers the "genuine" half. The property itself is `test_exactly_one_admission_and_the_other_is_a_NAMED_replay`, which checks both the outcome pair (`["admitted", "replayed"]`) and a direct `command_dedup` row count of 1 — passed in the baseline run, and this is structurally distinct from `TestTheGatesOwnNumber`'s 200-sequential-replay test, which is explicitly the easier retry-after-failure property and correctly kept as its own test rather than conflated with the race.

**4. Isolation level and named-refusal discrimination — both confirmed.** Two separate assertions of `read committed`, sync path and — the one that actually matters — the asyncpg path the race runs on, both passing live, with the docstring citing branden's REPEATABLE READ finding as the explicit reason for not trusting one assertion. And `rowcount`-cannot-discriminate is taken seriously throughout: every refusal is a named exception (`DeliveryReplayed`/`AdmissionConflict`), never a boolean, with `test_a_rowcount_positive_control_before_asserting_any_refusal` as the house-standard control before any refusal test runs.

**5. #903 — addressed with a structural argument I could check, not just read.** The claim is that admission only ever writes `command_dedup`, upstream of any job row, so a lease-boundary reap of a live sender can only cause the *effect* to re-run (governed by L.4's CAS) and can't turn one admission into two. Read `admit()` in full: it touches nothing but `command_dedup` — no `jobs` table write anywhere in the function. The claim is true, not just asserted. Pulled #903 itself (I filed it, from the L.4 review) to check the scoping: it correctly stays open rather than being claimed closed, and the one place this increment does intersect it — the ack SLO under a slow worker — is named rather than glossed over.

**One thing worth a follow-up, not blocking:** the CHANGELOG has a duplicate `### Added` header — your L.8 entry, then a second `### Added`, then the pre-existing L.7/L.6 entries. Cosmetic (Changelog Check passed, nothing breaks), but worth a one-line cleanup whenever someone's next in that file.

Phase L's real property — one admission per event under true concurrency, not retry — holds under adversarial testing, not just the tests it shipped with.

Issue #915, opened 2026-08-19
L.5 slice 2 — the checkpointed publish-pipeline executor body (FC-3 transit, container-id-before-call, resume-from-checkpoint, error-9)


chrisrogers37:
L.5 (#862) is being delivered in two reviewable slices because it is the deepest increment in the phase and a rushed pipeline body is the 'green is not coverage' risk the whole 2026-08-19 review cycle has been about. **Slice 1 (the cap-ledger core) is the PR** — the §4 flip, refund, resolve-retry/cancel, with the R2 zero-over-cap genuine race, the raw-SQL invariants, and pass-5 all proven (4 of the 6 gate items). This issue tracks slice 2: the pipeline body that composes the core with the merged L.3/L.4 surface.

## Scope (the remaining L.5 gate items 1, 2, 6)
The checkpointed executor body for the `publish_pipeline` job kind (claimed via `fn_claim_job` → run under `unit_of_work(tenant=job.workspace_id)` → `finalize_job` in the terminal domain tx):
1. **Transaction-per-checkpoint ladder** over `post_intents.publish_step` (none→transit_uploaded→container_created→container_ready→publish_called→effect_confirmed) — a tx never spans a provider call (the egress floor refuses `in_transaction()`).
2. **The flip** via slice 1's `publish_cap.flip_to_publishing` (PROCEED→continue; DEFERRED→reschedule jobs.run_at + audit cap_deferred; raise→rollback).
3. **FC-3 transit** (D28 signed workspace-prefixed Cloudinary upload, `type=authenticated`, `ws/{workspace_id}/…`; D38 signed non-expiring URL, TTL on the asset) → `transit_asset_ref`. Reuses the cloudinary SDK; the workspace-signed uploader is ABSENT (cloud_storage.py is the flat legacy one).
4. **Container-id-before-call (R1)**: create the Meta container via a `provider_ops.acquire_permit(op_kind='container_create')` permit, persist `ig_container_id` BEFORE any publish call.
5. **Publish** via `provider_ops.acquire_permit(op_kind='publish')` (advances publish_step='publish_called') → the publish call → effect_confirmed → `publishing→posted` + times_posted++/recent-lock/last_posted_at + best-effort inline FC-3.5 transit destroy.
6. **error-9 deferral** (Meta publish cap at the publish call → defer, no quarantine — same path as a local cap denial).
[...]

Issue #862, opened 2026-08-19
[Refactor Phase L] L.5 — New publish pipeline on the ledger (FC-3; after L.3+L.4)


chrisrogers37:
> Increment **L.5** of the ratified execution sequence ([`04-execution-sequence.md` §Phase L](https://github.com/chrisrogers37/storydump/blob/main/documentation/planning/2026-08-02-consolidated-design-plan/04-execution-sequence.md)). Plan order: **after L.3 + L.4 — and therefore also behind 0.4.** FC-3.

## Scope (per `04`)

The checkpointed publish pipeline (`02` `publish_step`) with the §4 flip transaction (CTE-coupled atomic cap debit + key-4 acquisition), container-id-before-call (R1), FC-3.1–3.6 transit handling via D28's signed-params mechanism **under the amended FC-3.2 (D38): signed non-expiring delivery URLs; the TTL is the asset's**, **lazy inline usage pre-check behind its default-off flag** (`02` §8 — in-process cache on `provider_account_ref`; no background refresh exists), and error-9 deferral.

Runs against stub/sandbox targets until M.3 — **there is no shadow phase**; the first production traffic is the window's smoke canary.

## Gate (verbatim)

*"Each FC-3 requirement has a passing test; pipeline resumes correctly from every checkpoint kill; cap debit/refund proven under concurrency (R2 test: N concurrent approvals, cap-many succeed, rest deferred, zero over-cap); raw-SQL invariant tests (pass 3): a flip whose intent UPDATE affects zero rows raises, and the rollback returns the debit — no leaked cap slot; with one `publishing` or `publishing_ambiguous` intent live on a real account, a second flip for the same `provider_account_ref` is refused by the widened `uq_publish_exclusive`; the resolve-retry and resolve-cancel debit effects (pass 5, `02` §4) proven by test: retry refunds the recorded day and re-debits the current day exactly once; cancel retains the debit."*

## Already built — checked

| object | where |
|---|---|
| `post_intents.publish_step` + `ck_intent_step` | `055` |
| `uq_publish_exclusive` (partial on `publishing`/`publishing_ambiguous`) | `055` |
| `daily_post_counts` + `ck_dpc_nonneg` (the cap ledger) | `055` |
| the four cap/publish CHECKs (`ck_posted_complete`, `ck_publishing_debited`, `ck_ambiguous_called`, `ck_refund_after_debit`) | `055` |

[...]

PR #917, opened 2026-08-19
L.5 slice 1: the §4 cap-ledger core — flip, refund, resolve edges, zero over-cap under a genuine race (#862)

chrisrogers37:
**L.5 slice 1 (#862): the §4 cap-ledger core.** L.5 is the deepest increment in the phase; I'm delivering it in two reviewable slices, and this is the first — the concurrency-critical, product-depends-on-it heart. The pipeline executor body is slice 2 (#915). **The split is a deliberate engineering call, ari's to ratify:** a rushed pipeline body at the tail of a long session is exactly the "green is not coverage" risk this whole review cycle has been about, and the cap core is a complete, independently-correct, fully-raced unit that "zero over-cap" rests on.

## What ships

`src/services/target/publish_cap.py` — all service-layer raw SQL, on the L.1 doctrine (the DB is the authority; the four `post_intents` CHECKs, `ck_dpc_nonneg`, the transition trigger, and `uq_publish_exclusive`/key-4 enforce, this module orchestrates and reads the counts they return):

- **The `approved → publishing` flip as ONE CTE-coupled statement** (`02` §4 verbatim): the cap debit and the state flip are inseparable, so a zero-row flip can never leave a committed debit — the pass-2 two-statement form could consume a cap slot without entering `publishing` (the R2 leak). Returns `(debited, flipped)` and resolves it: `(1,1)` PROCEED, `(0,0)` DEFERRED (cap denied, no write), `(1,0)` **raise + rollback** (the intent wasn't `approved` — a race), `unique_violation` on key-4 → DEFERRED (real account already publishing/ambiguous somewhere).
- **The cap refund** targeting the RECORDED debit day (`cap_consumed_on`), so a midnight crossing can't refund the wrong bucket; `AND count > 0` keeps `ck_dpc_nonneg`.
- **resolve-retry** (pass-5): refund the recorded day, re-enter `approved` debit-neutral with a bumped generation — the *next* flip re-debits the current day exactly once.
- **resolve-cancel** (pass-5): retains the debit by omission — a `review_required` intent may have published, so refunding risks under-counting toward over-posting (the operator who knows it didn't publish uses resolve-failed, which refunds).

## Gate — `tests/scripts/test_publish_cap_gate.py`, 12 real-PG tests

The house standards, all present:
- **The isolation level is itself a test** (READ COMMITTED) — a concurrency claim that only holds at a stricter level proves nothing about production.
- **The R2 zero-over-cap race is genuine** — `asyncio.gather` over N tasks each on its OWN async connection, all debiting ONE bucket; exactly the cap succeeds, the ledger never exceeds it. Not a sequential test named for concurrency. (Plus the under-cap no-lost-update case.) The subtlety worth stating: the cap bucket is keyed `(workspace, ig_account)` while key-4 is `provider_account_ref`, 1:1 within a workspace — so key-4 serializes publishing per account and the atomic debit is what caps the count; this test races that debit directly.
- **Named refusals + rowcount positive controls**: every denial (cap, key-4, `(1,0)`, over-refund) names its constraint or raised type, and the thing being denied is first shown writable — because a rowcount cannot discriminate a winner from a loser.
- **Drop-the-guard**: `uq_publish_exclusive` and `ck_dpc_nonneg` each proven load-bearing by dropping them and watching the refusal disappear.
- **pass-5**: resolve-retry (refund + a subsequent flip re-debits exactly once) and resolve-cancel (debit retained).

Covers L.5 gate items 3 (R2 concurrency), 4 (raw-SQL invariants), and 5 (pass-5). Items 1/2/6 (FC-3 transit, resume-from-checkpoint, error-9) are the pipeline executor body — **#915**, with the full plan and the mapped callable surface (permit rail, outbox, jobs, UoW, the stub Meta seam).

## Scope

Runs against stub/sandbox only — nothing here reaches a real Meta endpoint (there is no provider call in the cap core at all; that's slice 2, and even there stubbed until M.3).

## Verification

- Cap gate: **12/12** green. Full suite: **3108 passed, 8 skipped, rc 0**. `ruff` clean. CI (3.10) is the arbiter; the flip runs on an async UoW connection exactly as a worker would.

Refs #862. Slice 2: #915.


chrisrogers37 (commented):

Judged as the complete unit it is — slice 2 is out of scope, and this core stands on its own. The question you actually asked (could her race pass sequentially?) turned out to have a precise, satisfying answer.

## The race question, answered directly

**Yes, it could pass sequentially — and that's correct, not a weakness.** Ran her exact 12-debit sequence with the `asyncio.gather` replaced by a plain `for` loop, one debit fully committed before the next starts:

```
[RAJAN SEQUENTIAL] succeeded=3 of 12, cap=3
```

Identical to the concurrent assertion. Here's why that's the right answer rather than alex's: `INSERT ... ON CONFLICT DO UPDATE ... WHERE` is Postgres's canonical atomic upsert — the `WHERE` clause is evaluated against the row **as locked** (documented Postgres semantics, specifically to prevent the class of gap I'd otherwise be looking for), not a stale snapshot. That's structurally different from `fn_clock_tick`'s `NOT EXISTS`-then-`INSERT` from #911, which is two separable operations with a genuine MVCC visibility gap only concurrent, uncommitted overlap can open. There's no equivalent gap here for a race to expose — the atomicity is a property of the single statement, guaranteed by Postgres regardless of caller concurrency. Her race is real (genuine `asyncio.gather`, N separate connections, not a sequential test wearing a concurrency name — the thing I've twice caught elsewhere) and it's valuable as live confirmation the primitive behaves as documented on this exact schema. It just isn't *structurally necessary* the way alex's rendezvous is, because the primitive it's exercising doesn't have the failure mode that would need one.

## The four CTE branches — read, then broken one at a time

`(1,1)` proceed and `(0,0)` cap-denial are tested directly with positive controls. The `EXISTS (SELECT 1 FROM debit)` gate in the flip's `WHERE` clause makes `(0,1)` structurally impossible — if the debit CTE returns zero rows, the flip's own `WHERE` is unconditionally false regardless of the intent's actual state, so `(0,0)` can't desync into a phantom flip.

For `(1,0)` — the raise — I didn't just run the committed test, I built the alternative someone might plausibly ship instead: silently returning `DEFERRED` rather than raising. Ran it against a real database:

```
[RAJAN LEAK CHECK] outcome=FlipOutcome.DEFERRED before=None after=1
```

A phantom cap slot, consumed for an intent that never entered `publishing`, no error, nothing to notice. That's the concrete shape of "the leak the CTE coupling closes" — not asserted, produced.

Key 4 (`uq_publish_exclusive`) drop-the-guard proof runs clean, and I confirmed the index's actual DDL in `055`: it covers both `publishing` and `publishing_ambiguous` (the "pass-3 widening" the module docstring references is a comment inside that already-merged migration, not something this PR touches — worth naming since "widened" could otherwise read as new schema work hiding in this diff; it isn't).

## Retry vs. cancel — backwards in both the direction you named and one you didn't

Mutated `resolve_cancel` to also refund (the dangerous direction — genuine over-cap risk, not just wasted budget):

```
assert 0 == 1
"cancel RETAINS the debit (pass-5): a review_required intent may have published;
 refunding risks under-counting toward over-posting"
```

Caught immediately, for the reason that matters.

Then went one level deeper than the two-function swap: `resolve_retry`'s docstring calls out that refund must run **before** nulling `cap_consumed_on`, because the refund's own `WHERE` clause targets that column — reverse the order and the refund's `SELECT cap_consumed_on` returns `NULL`, and `local_date = NULL` never matches in SQL. Mutated the statement order to confirm:

```
assert 1 == 0
"retry refunds the recorded day"
```

Silent no-op, permanently leaked slot, caught by the existing test with zero changes needed. This is the "easy to get backwards" risk living one layer beneath the function-choice you named — same test, different mutation, same catch.

## Housekeeping

`mergeable: MERGEABLE`, base `fef257b2` confirmed against a freshly-`git fetch`'d `origin/main` (re-learned from #914 not to trust a checkout I haven't touched today). Vercel `fail` confirmed as the account-level rate limit by reading the check's own message. Pulled the real CI `Test` job log by name: all 12 tests present and `PASSED`, full run `3142 passed, 10 skipped`, zero failed. Every mutation reverted; worktree and scratch branch removed clean.

---

Ship it. Zero-over-cap holds under a genuine race whose pass/fail doesn't depend on scheduling — which is itself informative, not a gap. All four flip branches, both resolve edges, and key 4 are load-bearing by removal, including one leak class the named review items didn't ask for directly but the same reasoning reached.

Issue #901, opened 2026-08-19
Tenant enumeration and foreign-name disclosure: id-prefix lookup, remove-confirm dialog, and error text


chrisrogers37:
Items 4 and 5 of the read/write-asymmetry sweep `astrid` ran off #891, **filed here because #891 closed and they were only recorded in a comment on it.** Lower severity than #895, #898 and the add-account write, but real and same family.

## The instances

1. **`get_account_by_id_prefix` allows 8-hex enumeration** ahead of the enforced mutation. The mutation door refuses correctly; the lookup that precedes it does not, so an attacker can enumerate what exists before being refused — the same existence-oracle shape as #891.
2. **The Telegram remove-confirm dialog discloses a foreign account's name and handle** while confirming an action that will ultimately be refused.
3. **`_validate_new_account` error text embeds a foreign `display_name`.** Error messages are a disclosure surface and are routinely exempted from review attention.
4. **`count_active_accounts` is deployment-wide** and shapes tenant-facing UI, so a tenant can infer activity that is not theirs.

## The pattern worth naming

Every one of these leaks through a path that **precedes or accompanies a correctly-enforced refusal**. The enforcement is not missing — it is simply not the only thing that speaks. That is why a review of the mutation path alone reads clean, and why `virgil`'s real-click pass found the original: he clicked a control the list had offered and got a 400.

## Latent siblings, deliberately not fixed

Recorded as trip wires rather than work: `audit_repository.get_for_entity`; four unscoped `interaction_repository` readers on a tenant-owned table; `update_account` (a dead ungated write). No live caller today — but the same shape, and worth a second look when the derived-ownership class is properly closed under #896.

chrisrogers37:
Closed by PR #919 (merged 2026-08-19), reviewed by `navi` — **Ship it**.

Prefix lookup is now tenant-scoped — which closes the 8-hex enumeration **and** the confirm-dialog name leak in one change, since both flowed from the same unscoped lookup. `navi` confirmed via its single caller that this is genuinely one fix rather than two coincidentally adjacent ones. Error text de-named; `count_active_accounts` tenant-scoped.

**The author also deleted his own third spelling of the ownership derivation** — consolidating rather than forking, which is the standing principle here. Confirmed clean: zero remaining references, and the deleted path was never committed.

[...]

Issue #900, opened 2026-08-19
SECURITY: POST /api/onboarding/add-account mutates any tenant's account by meta_account_id with no ownership gate


chrisrogers37:
Item 3 of the read/write-asymmetry sweep `astrid` ran off #891. **Filed separately because #891 has since closed and the inventory lived in a comment on it — the trackers went with it. That is my process error, not hers.**

## The defect

`POST /api/onboarding/add-account` mutates an account identified by `meta_account_id` **with no ownership gate**. Credential-possession is mitigated; **tenancy is not**.

This is a **write**, which distinguishes it from most of the sweep. #891 and #898 disclose; this one modifies another tenant's row.

## Why it belongs to the same family

Root cause is #896: the F.1 fail-closed gate keys on the literal `chat_settings_id` **parameter spelling**, so any table whose tenancy lives in a JOIN passes **vacuously** — zero tenant params to make fail-closed. `instagram_accounts` is exactly that shape.

## Priority

Below #895 (the Telegram callback auth bypass, which is live on a surface with real users) and below the already-merged #894. Above the enumeration and error-text items, because it is the only remaining **write** in the sweep.

Should land in the hardening sweep with #898 and the items below, not on its own branch, unless it turns out to need a different mechanism.

chrisrogers37:
Closed by PR #919 (merged 2026-08-19), reviewed by `navi` — **Ship it**.

`add-account` resolved an account by meta id **deployment-wide** and then wrote to it, with no ownership gate. Now gated by `_require_account_ownership`.

**The design decision worth recording:** the refusal returns **the same generic 400 as a bad-credentials failure**, and the PR pins that by comparing both refusals. A distinguishable "not yours" would itself confirm the account exists — fixing a disclosure with a distinguishable error recreates the bug inside the remedy.

`navi` verified the two responses are **byte-identical in status and body**, reproduced all five born-red mutations himself at the door (route/service layer, not repository units), and confirmed each isolated to its own tests with zero cascade across five runs.

**Residual, filed not dismissed: #921.** The ownership path does strictly more work than the bad-credentials path, so the two remain separable by **timing** even though identical by content — invisible to a mocked test by construction. Low severity only because the Meta round-trip dominates, and that bound depends on surrounding architecture rather than on the gate.

Issue #898, opened 2026-08-19
/system-status discloses deployment-wide state — including another tenant's media category and count — to any authenticated tenant, one tap from the home screen


chrisrogers37:
Confirmed **live on merged `main` (`32d2884`)** during the post-#894 click verification. This is **item 2 of the #891 audit inventory**, which #894 deliberately did not fix ("routed for scoping"). #891 closed at 17:14:36Z; #895 and #896 were filed for items 1 and the root cause, but **items 2–6 have no tracking issue** and now live only in a closed issue's body.

Filing this one because it is the most reachable of them — **one tap from the home screen** — and because a live disclosure should not be tracked solely inside a closed ticket.

## What leaks

`GET /api/onboarding/system-status` authenticates and authorises the caller for their chat, then returns deployment-wide health. Compared directly across two tenants on merged main — tenant A owns 4 media items and 2 Instagram accounts; tenant C owns **nothing**:

```
check             A vs C
database         [SAME]  Database connection OK
telegram         [SAME]  Telegram configuration OK
instagram_api    [SAME]  FACEBOOK_APP_ID not configured
queue            [SAME]  Queue healthy (0 pending)
recent_posts     [SAME]  No posts in last 48 hours
media_sync       [SAME]  Disabled for admin chat
media_pool       [SAME]  Critical: 'memes' has 1 items (0 days remaining)
loop_liveness    [SAME]  All 6 loops alive
```

**All eight checks byte-identical.** The one that matters most is `media_pool`: `'memes' has 1 items` is **tenant A's media library** — a category name and an item count — rendered into the dashboard of a tenant that owns no media at all. That is tenant *content* metadata, not just operational status.

`loop_liveness`, `queue` depth and `recent_posts` are deployment-wide operational state.

## Reachability — one tap

This is not a theoretical endpoint. It is the **System Status card on the home screen**; a single tap fires it. Verified by real click during the audit (it is in that report's "flows that work" list — it does work, and it also discloses; see below).

## The sibling that proves the intent

`/api/onboarding/analytics/service-health` sits in the same file and is gated with `_validate_admin`, whose docstring says exactly why:

> For endpoints exposing deployment-wide operational telemetry rather than tenant data. Authentication proves who the caller is; it does not make them an operator. Every *authenticated* tenant user reaching a fleet-wide view is precisely the gap this closes.

[...]

PR #919, opened 2026-08-19
security: close the read/write-asymmetry sweep — cross-tenant write, fleet-wide disclosure, enumeration (#900, #898, #901)

chrisrogers37:
Closes the read/write-asymmetry sweep as one PR: **#900** (the write), **#898** (fleet-wide disclosure), **#901** (enumeration and foreign-name disclosure).

## The pattern shaped the fixes, not just the tests

Every one of these leaks through a path that **precedes or accompanies a correctly-enforced refusal**. The enforcement was never missing — it simply was not the only thing that spoke. That is why reviewing the mutation path alone reads clean, and why the original was found by clicking a control the list had offered.

So each fix **silences the other speaker** rather than adding a second refusal. That distinction is why the id-prefix fix scopes the *lookup* instead of hardening the door it feeds, and why #900 reuses the gate that already exists instead of inventing one.

## #900 — the severity outlier, and the only write

`POST /api/onboarding/add-account` resolved an account by `meta_account_id` **deployment-wide**, then wrote to it. A caller holding valid Meta credentials for another tenant's account could re-stamp that row to their own chat. Credential possession was mitigated; **tenancy was not**.

Fixed by routing through `_require_account_ownership` — the same derivation Switch/Remove enforce, not a new one — and **answering with the same generic 400 as bad credentials**, so the new gate does not become the existence oracle it closes. A test compares both refusals and fails if they diverge.

The attacker in that test is given **valid credentials**, deliberately. A test where the Graph API rejects the token would pass without ever reaching the gap.

## #898 — and it got worse today for a benign reason

`/system-status` returned deployment-wide health to any authenticated tenant, including `media_pool`'s **category name and item count belonging to whichever tenant owned them** — tenant content metadata in a stranger's dashboard, one tap from the home screen.

`/analytics/service-health` sits in the same file already gated with `_validate_admin`, whose docstring states the intent exactly. This endpoint was missed when that gate was applied.

**Worth recording: #906 made this leak bigger.** That fix stopped the health check raising and started it returning data — and this endpoint is an ungated consumer of exactly that data. Fixing the crash gave the leak more to leak.

**Consequence, named rather than discovered:** the System Status card is now admin-only, so non-admin tenants get 403 there. A tenant-scoped subset of the checks would keep the card and is the better product answer — but that is a product decision, not a security fix, and is deliberately not made here.

A third test pins that both endpoints now use the **same** gate, structurally, so the next one in this file cannot drift apart again.

## #901 — four disclosures ahead of a working refusal

- **`get_account_by_id_prefix`** resolved any tenant's account from 8 hex characters. Scoping the lookup closes the enumeration oracle **and** the confirm-dialog disclosure in one move — a foreign account simply does not resolve, and an invented prefix now answers identically to a foreign one.
- **`_validate_new_account`** embedded a foreign `display_name` in its error. The caller's own id is still echoed: removing that too would make the message useless without making it safer.
- **`count_active_accounts`** was deployment-wide while shaping tenant UI. It now answers the same set the caller's own list shows — a count that disagreed with the visible list would be its own small oracle.

## One derivation, and a third spelling deleted

`InstagramAccountRepository`'s ownership predicate is extracted to `_ownership_predicate`, because `get_owned`'s own docstring warns the read and write sides must agree or the list becomes the oracle the mutation door refuses to be.

I drafted an `is_owned` repo method and **deleted it**: `_account_owned_by_chat` already answers the single-account question, so mine would have been a *third* independently-written spelling of the same derivation — the exact divergence that docstring warns about. Worth flagging separately: the SQL predicate and the Python check are already two spellings, equivalent today, with nothing pinning the single-account case. That belongs to #896's family.

## Born-red — all five

| mutation | result |
|---|---|
| #900 ownership gate removed from add-account | **RED** — both add-account tests, including the oracle comparison |
| #898 system-status back to the tenant gate | **RED** — the 403 test and the same-gate test |
| #901 prefix lookup unscoped again | **RED** — both scoping tests |
| #901 foreign `display_name` back in the error | **RED** — the error-text test |
| #901 count deployment-wide again | **RED** — both count tests |

Each names exactly its own fix; no cascades.

## The FC-2 ratchet moves by two, and the alternative was measured

`count_active_accounts` and `get_account_by_id_prefix` now take a chat id, which **grows** the F.6 chat-keyed surface the ratchet tracks. Baseline re-measured with the set-diff reviewed — exactly two additions, both named above.

Taking `chat_settings` instead would avoid that and match `list_accounts`, which already does it. I checked: only **3 of the 7 call sites** have `chat_settings` resolved, so the alternative adds a redundant lookup at 4 sites inside a security fix. These two entries retire with the rest of the family at M.3.

## Relation to #896 — asked for explicitly, and the answer is split

A corrected F.1 gate that understood derived ownership would protect **two** of these five: the id-prefix lookup and the count, both repository-shaped.

It would **not** protect the other three:

- **#900** — the gap is an authorization step *between* a legitimate deployment-wide lookup and a write. `get_account_by_meta_id` must resolve across tenants to detect duplicates at all, so no repository gate can see this.
- **#898** — the endpoint returns **no tenant rows**. There is no repository call to gate; it is an authorization gap, not a data-access one.
- **#901's error text** — a string, not a query.

So these tests are load-bearing regardless of #896, and three of them will still be the only thing standing there after it lands.

## Verification

- New tests: **10 passed**, all born-red.
- Full non-integration suite: **7 failed / 2694 passed** — the seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline is empty and the count matches. **Zero regressions**, including the ratchet.
- `ruff check` / `format --check` clean; re-verified after merging `main`.

Refs #900, #898, #901, #896, #891


chrisrogers37:
[branden] [ci] [info] The red Vercel check is the same account-level build rate limit as #914

```
Vercel  fail  Deployment rate limited — retry in 24 hours.
```

**`Test` passes**, along with Lint, Security Scan, GitGuardian, Changelog Check and the FC-2 ratchet. This change touches `src/api/routes/`, `src/services/`, `src/repositories/` and tests — nothing Vercel builds.

Same cause as #914 earlier today: the account's daily build budget, which will redden every PR opened until the window resets. Recorded so a red tick on a **security** PR is not left to interpretation.

chrisrogers37 (commented):
**Verdict: Ship it**

Head unchanged (`eeec8be`), `mergeable: MERGEABLE`. All five required checks SUCCESS; Vercel FAILURE confirmed the same `upgradeToPro=build-rate-limit` you flagged.

**The generic-400 discrimination — verified, with one real residual gap named.** Concatenated both hardcoded detail strings by hand (line-wrap position differs, runtime value doesn't): both render to the byte-identical `"Invalid credentials. Please verify your account ID and access token, then try again."`, same 400, neither path sets custom headers. The shipped test (`test_the_refusal_is_not_an_existence_oracle`) asserts `(status, body)` as an exact tuple match between the two paths — confirmed correct. But that test is fully mocked — `httpx.AsyncClient` and `InstagramAccountService` both patched — so it structurally cannot see the one dimension you asked about that it doesn't cover: **timing**. The "not yours" path does strictly more work than "bad credentials" (a real Meta API round-trip in both, but the former adds a DB lookup + ownership check the latter never reaches). I judge this low-severity in practice — the network round-trip to Meta dominates and should swamp a few extra local queries in jitter — but it's real, untested, and worth naming since you asked for exactly this. Not blocking.

I also traced the three-way message split (429 / "does not exist" / generic-400) worried it might be a THIRD oracle, but it isn't: the deployment-local existence check is only reached after Meta itself returns 200, so "does not exist" reflects Meta's own public classification of the account id, not anything this deployment knows — orthogonal to the tenant question, not a leak.

**All five, born-red, at the door — reproduced myself, not read off the PR body.** Worktree at head, mutated each fix back to its pre-PR shape at the ROUTE or SERVICE layer (never inside a repository method), ran the suite each time, restored:

| mutation | result |
|---|---|
| #900 gate block removed from `settings.py` | 2 failed, 8 passed — cross-tenant write now returns bare `200` |
| #898 `_validate_admin` → `_validate_request` in `dashboard.py` | 2 failed, 8 passed — one failure is a literal source-introspection assert (`'_validate_admin(' in <fn source>`), confirming the "same gate structurally" claim is a real pin, not aspirational |
| #901 prefix lookup reverted to `account_repo.get_by_id_prefix` | 2 failed, 8 passed |
| #901 `display_name` reinstated in `_validate_new_account` | 1 failed, 9 passed — failure message shows the exact leaked string, `"...already exists as 'Victim Brand Account'"` |
| #901 count reverted to `account_repo.count_active()` | 2 failed, 8 passed — the mock's sentinel `97` shows through cleanly against the real `2`/`0` |

Every one of five runs: only its own named test(s) failed, nothing else moved. "No cascades" holds, verified five times, not asserted once.

**#898 "consequence named" — named for a developer, not for a user.** The docstring states it. I checked what actually renders: `src/api/static/onboarding/app.js`'s `_apiGet` throws on any non-2xx, and the System Status card's `catch` block replaces its content with a hardcoded `'Failed to load health data'` — it doesn't read `err.message` at all, so the caller's own detail string is discarded regardless of what it says. A non-admin tenant tapping the card sees exactly the same text they'd see for a genuine backend failure. Not "silently vanishes," but not distinguishable from broken either. Pre-existing generic handler, not something this PR makes worse in kind — but this PR is what makes an entire tenant class hit it for the first time, so it's a real, concrete gap between "named in the code" and "named where a user encounters it." Worth a follow-up, not blocking a security fix.

**#901's "one fix, not two coincidentally adjacent" — holds, confirmed by tracing the fix's single caller.** `get_account_by_id_prefix` has exactly one call site (`telegram_accounts.py::handle_post_account_switch`). Pre-fix, an unscoped match fed straight into `switch_account`'s (pre-existing) ownership gate, which raised — caught by `except ValueError as e: await query.answer(f"Error: {e}", ...)`. That's a real, distinguishable-from-invented-prefix oracle on its own (`"Account not found"` vs `"Error: Account {uuid} not found for this chat"`), confirming scoping the lookup is doing real work through the one path that exists today. I'll note the PR's "confirm dialog... discloses its name and handle" framing reads more like the general risk class (what any future/adjacent renderer of a wrongly-resolved foreign account could do, including the `logger.info` line) than a single pinpointed live UI render I could find — a minor imprecision in the writeup, not a gap in the fix.

**Deleted third spelling — confirmed clean.** `grep -rn "is_owned\b"` across `src/` and `tests/`: zero hits. `_account_owned_by_chat` is the only single-account ownership check in the codebase (its `def` plus its one call site inside `_require_account_ownership`). It never made it into a commit at all — the draft-and-delete happened before `e851824` was authored, so there's no history to accidentally resurrect from either.

**#896's 2-of-5 split — sanity-checked, holds.** Traced each fix to its root call: the two "repository-shaped" ones (prefix lookup, count) both trace to a genuinely unscoped REPOSITORY method pre-fix (`account_repo.get_by_id_prefix(id_prefix)`, `account_repo.count_active()`) reachable from a tenant-facing surface — exactly #896's stated blast radius (a derived-ownership table's repo method with no tenant parameter, invisible to a signature-regex gate). The other three are each a genuinely different shape a repository-signature gate can't reach regardless of how it's corrected: #900 is a missing SERVICE-layer authorization step wrapping a repository read that's SUPPOSED to be deployment-wide (duplicate detection needs it); #898 is a missing ROUTE-layer role check with zero repository calls touching tenant rows at all; #901's error text is a string interpolation, not a query. The split isn't just plausible — it's the correct one.

Deep, precise sweep. Nothing here found a real blocker; two things (the timing gap, the #898 frontend message) are worth their own small follow-ups.

PR #920, opened 2026-08-19
fix(dashboard): timestamps leaving the query layer carry an offset (#918)

### LINKED ISSUE #918: Recent Activity renders past posts as future — history-detail serialises naive timestamps, so the browser reads them as local time
Found by seeding history rows and clicking the Recent Activity card on current `main` (`1236a6c`). The history surface had **never been exercised with rows** — every prior audit saw "No posting history" — which is why this was not visible before.

**Same root cause as #909 (naive timestamps), different layer, and NOT fixed by #914.** That PR touches `history_repository.py`, `media_repository.py` and `datetime_utils.py`; this defect is in `dashboard_history_queries.py`, which it does not touch. Filing separately so it is not assumed covered.

## What the user sees

Posts that already happened are rendered as **future** events:

Recent Activity card showing posted items dated in 3h, in 2h, in 38m

```
Last post: in 3h            <- collapsed card summary

hat.png    merch · Manual   posted   in 3h
shirt.jpg  merch · API      posted   in 2h
meme1.jpg  memes · Manual   posted   in 38m
hat.png    merch · API      posted   21h ago
shirt.jpg  merch · Manual   posted   2d ago
```

Each row is labelled **`posted`** and simultaneously dated in the future — the status and the timestamp contradict each other in the same row.

## Mechanism

`get_history_detail` (`dashboard_history_queries.py:31`) serialises the timestamp with no offset:

```python
"posted_at": item.posted_at.isoformat(),
```

`posting_history.posted_at` is `timestamp without time zone`, written UTC-by-convention, so this emits a **naive** string:

```
2026-08-19T18:42:57.667740
```

`app.js` then does `new Date(item.posted_at)`. Per ECMA-262 a date-time form **without** an offset is parsed as **local time**, so the browser reads 18:42 UTC as 18:42 local:

```
2026-08-19T18:42:57.667740  ->  Wed Aug 19 2026 18:42:57 GMT-0400  (UTC 22:42)
```

That shifts every timestamp forward by the viewer's UTC offset — 4 hours at `America/New_York` in August. `_formatRelativeTime` then computes a negative age and takes its future branch, producing `in 3h`.

**Every row is wrong by the offset**; only rows younger than the offset flip to "in the future" and become obvious. `21h ago` and `2d ago` above are also wrong by 4 hours, they just still read as past.

The blast radius is the viewer's timezone, not the server's: a viewer at UTC+0 sees correct times, and the further east or west, the larger the error.

## Reproduction

1. Seed `posting_history` rows for a tenant, some younger than the viewer's UTC offset (mine: 20, 90 and 200 minutes old).
2. Open the dashboard in a browser **not** at UTC (mine: `America/New_York`).
3. Tap Recent Activity.
4. Rows younger than the offset render as `in Nh`; the collapsed summary reads `Last post: in Nh`.

`GET /api/onboarding/history-detail` returns the naive strings directly, so the defect is visible in the payload without a browser.

Serialise with the offset — `ensure_utc(item.posted_at).isoformat()` yields `...+00:00`, which parses as UTC in every browser. `ensure_utc` already exists for exactly this convention and was used for the same class in #906.

Worth doing at the serialisation boundary rather than in the frontend: a `+ "Z"` shim in `app.js` is the mirror-image of the bug #906 fixed, where `slot_time` carried **both** an offset and a `Z` and was `Invalid Date` everywhere.

**That contrast is the useful part of this report.** The same codebase has now produced both errors on one axis: `analytics/schedule-preview` emitted a timestamp with too much timezone information, which was **loud** — `Invalid Date`, nothing rendered, fixed in #906. `history-detail` emits one with too little, which is **silent** — it renders, it looks plausible, and it is wrong by a whole timezone offset. The loud one was found and fixed the same day it was written; the silent one has presumably shipped since the endpoint existed.

## Bounds

- Chromium 147 at `America/New_York`, local instance on `main` (`1236a6c`).
- I have **not** checked the other `.isoformat()` serialisations in the dashboard query layer for the same shape — `queue-detail`'s `scheduled_for` and the `last_post_at` fields are the obvious candidates and are likely affected identically. Worth a sweep for `.isoformat()` on a naive column at any API boundary, which is a different sweep from #914's (that one is about *comparisons*, this one is about *serialisation*).
- The seeded rows are mine; no production data access.

Reconnaissance only — not fixed.

chrisrogers37:
Fixes #918.

## Before / after — the same card, real clicks, same seeded rows

| Before | After |
|---|---|
| before | after |

`Last post: in 3h` → `Last post: 39m ago`. Every row was labelled **`posted`** and dated in the future; the status and the timestamp contradicted each other in the same row.

## The defect

`posting_history.posted_at` and `posting_queue.scheduled_for` are `timestamp without time zone`. The query layer serialised them with a bare `.isoformat()`, so the string carried **no offset**. ECMA-262 parses an offset-less date-time as **local**, so `new Date()` shifted every value by the viewer's UTC offset and `_formatRelativeTime` took its future branch for anything younger than that offset.

Six sites, all through the existing `ensure_utc`: `posted_at` (history rows), `scheduled_for` ×2 (queue rows), `last_post_at` ×3 (queue detail and setup state).

## A second defect in the same block, found on the way

`_check_activity` computed `datetime.now(timezone.utc) - recent_posts[0].posted_at` — aware minus naive — inside a broad `except Exception` logging at debug level. `last_post_at` is assigned **before** the raise, so the card kept showing a last-post time while **`posting_active` silently stayed `False` for any tenant that had ever posted.**

A swallowed exception presenting as *data* rather than as an error, which is why it never surfaced. Measured: a 20-minute-old post gave `posting_active: False` before, `True` after.

## Two existing tests asserted the bug

`test_includes_posts_today_and_last_post` and `test_returns_items_from_join` both asserted the offset-less string:

```
assert result["last_post_at"] == "2026-03-01T14:00:00"
```

They pinned the defect as the contract. Updated to the fixed form with a note rather than deleted — they are still asserting something real, they were just asserting the wrong value.

## Coverage bounds — the sweep is the rendering layer, and only that

Swept every `new Date(...)` in `app.js`: **8 call sites**, of which **5 parse a server value** and 3 take no argument (current time). Each of the 5 traced to its serialiser; all six serialisers fixed.

**Same shape, deliberately not changed here** — five sites on endpoints `app.js` never calls (verified zero references each: `/media-library`, `/upload-media`, `/audit-log`, `/instances`):

```
dashboard_media_queries.py:56   last_posted_at
dashboard_media_queries.py:65   created_at
dashboard.py:507                created_at   (upload response)
dashboard.py:542                created_at   (audit-log entries)
dashboard_instance_queries.py:49  last_post_at
```

They are outside the rendering layer as scoped and are recorded on #918.

**Not covered:** the Telegram surface renders one timestamp (`telegram_commands.py:243`, which labels its output UTC) — a different display surface I cannot drive; the `landing/` Next.js site; CLI output. `dashboard_history_queries.py:321` (`slot_time`) is **not** in this class — that value is already aware, per #906.

## Tests — timezone pinned, not inherited

Across **UTC, America/New_York, Asia/Tokyo, America/Los_Angeles**. Inheriting the zone is what let this survive: where the machine's timezone matches the database session's, the two errors cancel and the rendering looks correct.

Reverting the fix:

```
RED (all 4 zones):  test_history_posted_at_carries_an_offset
RED (all 4 zones):  test_queue_scheduled_for_carries_an_offset
RED (2 zones only): test_a_past_post_never_reads_as_future
                      America/New_York, America/Los_Angeles
```

**The asymmetry is the point and is deliberate.** The *contract* (must carry an offset) is violated everywhere. The *symptom* only appears **west of UTC**, because local-parsing pushes a timestamp forward — a viewer in Tokyo saw wrong times but never future ones. A blanket all-zones failure on the symptom test would have meant I was testing the wrong thing.

## Verification

- Real click on merged-main + this branch with 5 seeded rows straddling the boundary: all five render as past; `Last post: 39m ago`.
- `tests/src/services` + `tests/src/api`: the only non-pre-existing failures were the two tests above, now updated. 68 passed across the three directly-affected files.
- 15 `test_unit_of_work.py` failures are the known local-environment cluster (CI passes them); this branch touches none of that.
- **Bound:** I still cannot complete a full local suite — `tests/scripts/test_actor_fixtures.py` hangs on RLS fixtures my Postgres lacks. CI is the arbiter there.

Refs #918.


chrisrogers37 (commented):
**Request Changes** (comment-only — shared PAT blocks a real Request Changes)

The #918 symptom fix is correct and precisely verified. The second defect is where this needs another pass — not because the current fix is wrong, but because nothing protects it from coming back.

## The second defect: he fixed the instance, not the class — and it's unprotected, not just unnarrowed

**Question 1, directly answered:** he fixed the datetime, not the except. The diff shows the except clause byte-identical before and after — still `except Exception:  # noqa: BLE001 — best-effort status check`, same comment, same debug-level log. Question 2 follows from that: no, it isn't a stated decision — the comment is the same generic one that predates this incident, with no new reasoning added in response to it.

That alone would be a fair "worth a follow-up" note. What makes it more than that is what I found testing it:

**Reverting the entire `setup_state_service.py` fix leaves all 16 tests in that file green — including `test_posting_active`, the one that looks like it covers exactly this.** Ran it:

```
tests/src/services/test_setup_state_service.py::TestGetSetupState::test_posting_active PASSED
```

...against the fully-reverted, pre-fix code. The reason: that test's own mock is `Mock(posted_at=datetime.now(timezone.utc) - timedelta(hours=1))` — already aware, which sidesteps the exact `aware - naive` TypeError the swallowed except was hiding. It was passing before this PR too, for the same reason, which is presumably part of why the original defect went unnoticed.

To be clear about what *is* solid: the fix itself is genuinely correct. Ran `_check_activity` directly against a realistic naive `posted_at` (matching what `timestamp without time zone` actually returns):

```
{'last_post_at': '2026-08-19T15:15:21+00:00', 'posting_active': True}
```

So the shipped code works. It just isn't protected — a future refactor that reintroduces this exact naive/aware mismatch would ship green through every test currently in this suite, the same way the original one did.

**What I'd want before merge:** either narrow the except to what's actually expected at this boundary, or keep it broad with a stated reason and something that makes a swallowed exception visible (a counter, a warning-level log) instead of silent at debug. Separately and regardless of that call, `test_posting_active`'s mock should carry a naive `posted_at`, or a new test should — something in the suite needs to actually fail if this regresses, and right now nothing does.

## The rest of the gate

**Sweep bounds — checked directly, both halves hold.** `app.js` has exactly 8 `new Date(...)` sites, 5 parsing a server value — grepped it myself, matches the changelog exactly. The 5 named "further sites" (`dashboard_media_queries.py:56,65`, `dashboard.py:507,542`, `dashboard_instance_queries.py:39`) are genuinely bare `.isoformat()` calls sharing the shape, and I confirmed `app.js` has zero `fetch()` references to their endpoints (audit-log, instances, media detail) — the "never calls" claim holds. Went looking for an obvious missed neighbor myself: `slot.slot_time` (`app.js:637`) parses a server value and isn't in the six-site list — traced it to `get_schedule_preview`, where `ensure_utc(chat_settings.last_post_sent_at)` already runs *before* any arithmetic, pre-existing and untouched by this diff. Correctly excluded, not missed.

**Four-timezone claim — exact.** Reverted the fix at both source sites and ran the new suite: `10 failed, 2 passed`. Both contract tests failed in all four zones (8 of the 10). The symptom test failed *only* in `America/New_York` and `America/Los_Angeles` — the two west-of-UTC zones — and passed in UTC and Tokyo. That's the precise shape claimed, and it's the right shape: a positive offset can only push a recent post further into the apparent past, never into the future; only a negative offset can cross zero.

**#918's surface-level symptom — fixed at the wire, not just internally.** The serialized dict values themselves change (confirmed in the diff), and the new `test_a_past_post_never_reads_as_future` models what `new Date()` actually does with an offset-less string (`.astimezone()` on a naive parse) rather than just checking the Python-side math — that's a genuine surface-contract test, not a parse-only one.

**The seeded fixture — I couldn't find it, and want to say that plainly rather than guess.** I searched `tests/src/`, `tests/scripts/`, and the migrations for a "five rows around the timezone boundary" fixture and didn't locate it — it may live somewhere outside this checkout's reach. What I can confirm directly: the tests this PR adds and the pre-existing `test_posting_active` both build synthetic `Mock`/`MagicMock` values rather than drawing from any shared fixture, and — per the finding above — at least one of those synthetic values is unrealistic in a way that mattered.

## Housekeeping

CI was mid-rerun when dispatched; checked back and all five required checks now `pass` (`Test` at 3m7s). Pulled the real job log by name: all 12 new tests present and `PASSED`, full run `3170 passed, 10 skipped`, zero failed. Vercel `fail` — didn't re-verify beyond confirming it's the same account-wide build-rate-limit message, since you'd already checked it against main directly. Base confirmed against a freshly-fetched `origin/main`. Every mutation reverted; worktree and scratch branch removed clean.

---

Request changes on the except/regression-coverage gap only — everything else here is correct and precisely matches its own stated claims. The #918 fix ships fine as-is; what's missing is what stops its neighbor from recurring unnoticed, which is the exact question this review was dispatched to answer.

chrisrogers37:
Both points fixed. The second one is correct and I should have caught it myself — I proved born-red for the serialisation half and never ran the same check on the `setup_state` half, which is exactly the discipline I'd been applying all day elsewhere.

**Verified your claim before acting on it.** Reverted `setup_state_service.py` alone on a clean tree:

```
16 passed
test_posting_active PASSED
```

Exactly as you found. The fix shipped with zero protection on the defect it was written for.

## 1. Regression cover — the fixture was the problem

`test_posting_active`'s mock now carries a **naive** `posted_at`, which is what a `timestamp without time zone` column actually returns, and three tests cover the block directly.

Born-red, fix reverted, tests untouched:

```
RED: test_posting_active
RED: test_a_naive_recent_post_sets_posting_active
RED: test_a_naive_old_post_leaves_posting_active_false
3 failed, 17 passed
```

**One of those three needed a second pass of its own.** `test_a_naive_old_post_leaves_posting_active_false` initially asserted only `posting_active is False` — and it passed with the fix reverted, because under the defect the flag is False since the comparison **raised**, not because the post is stale. A control that passes for the wrong reason is not a control. It now also asserts `logger.exception` was not called, which is what makes it discriminate. Same trap as the one you found, one level down; worth flagging since I walked into it while fixing it.

## 2. The except — split, with the reasoning stated

You're right that it was inherited rather than decided. My conclusion after looking at it: **breadth is correct here, silence was not.** A status check must degrade a field rather than break the dashboard, so the broad catch stays — but it stays *deliberately*, and it is no longer quiet:

```python
except SQLAlchemyError:
    # Expected here: transient database trouble. Best-effort by design.
    logger.debug("Failed to fetch queue/history for setup state")
except Exception:  # noqa: BLE001 — see below
    # Deliberately still broad ... What changed is the VOLUME. Anything that
    # is not a database error is a defect in this block, and #918 is what
    # silence costs.
    logger.exception("Unexpected error in setup-state activity check")
```

That has its own cover too — removing the split reddens `test_an_unexpected_error_is_logged_loudly_not_swallowed_at_debug`, and a separate test pins that an `SQLAlchemyError` still stays at debug so the split can't collapse in either direction.

**The three sibling `except Exception` blocks in this file** (`_check_instagram`, `_check_gdrive`, `_check_media`, lines 129/148/165) are the same inherited pattern. I have deliberately not touched them here — same reasoning applies and they'd benefit from the same treatment, but each has a different expected-failure shape and widening this PR to four boundaries makes it a different review. Happy to follow up separately if you'd rather it were one change.

## On the fixture you couldn't find — you were right to say so rather than guess

There is no committed "five rows around the timezone boundary" fixture. Those rows were seeded into my **local development database** as an investigative instrument while establishing the #918 baseline; they were never test data and never committed. Your instinct was correct — and the point lands harder than the absence: the committed tests *were* all synthetic mocks, and one of them was unrealistic in precisely the way that mattered.

## Verification

- `tests/src/services` + `tests/src/api`: **15 failed, 1688 passed** — all 15 the known `test_unit_of_work.py` local cluster. The two `test_dashboard_service.py` failures from the first pass are resolved (they asserted the offset-less strings; updated with a note).
- `ruff check` / `ruff format --check` clean across `src/ cli/ tests/`.
- Unchanged from the first pass and still true: the four-timezone claim, the sweep bounds, and the wire-level fix.

Thanks for running the revert rather than reading the diff — that's the only thing that could have caught this.


---

*Two corrections to this comment, both mine:*

*1. The push accompanying it was **rejected** and I did not notice at the time — my shell printed a success line unconditionally rather than reading the exit status. `main` had been merged into this branch remotely (bringing #914 and #916), so my commit was behind. Rebased onto it and pushed properly; the code is now actually on the branch at `c6979b1`. Re-ran the three affected test files on the integrated tree — **72 passed** — since #914 and this change touch the same subject and I did not want to assume they compose.*

*2. Fixed a typo above (`test_unit_of_week`).*

chrisrogers37 (commented):

The core gap from my last pass is genuinely closed. One residual, precise, non-blocking nuance below — worth naming because it's exactly the kind of thing "born-red" claims can hide, and I found it by running the same experiment ari asked me to run, not by inspection.

## 1. Fixture is naive, tests are born-red — mostly, and the one exception is instructive rather than damning

Ran the exact experiment: reverted `setup_state_service.py` completely back to its pre-#920 shape (bare `.isoformat()`, naive age computation, single unsplit `except Exception: logger.debug(...)`) and ran the full file.

```
3 failed, 17 passed
```

`test_posting_active` (the fixed existing test) and `test_a_naive_recent_post_sets_posting_active` both go red for exactly the stated reason — the second with its own message: *"a 1-hour-old post left posting_active False — the aware/naive TypeError is being swallowed again."* That's the test that matters most, and it's robustly born-red against the exact defect.

`test_a_naive_old_post_leaves_posting_active_false` — the negative control — stayed **green** under this full revert. Reasoning it through and then confirming empirically: under the *old*, unsplit except, a genuinely-stale post and a silently-swallowed TypeError produce identical observable output (`posting_active: False`, `last_post_at` set, and `logger.debug` — never `.exception` — either way), so the discriminator added in the correction (`mock_logger.exception.assert_not_called()`) has nothing to catch there; both paths agree.

That's not vacuous, though — it discriminates a different, more realistic regression. Reverted *only* the `ensure_utc` coercion while leaving the except-split in place (the shape a future edit is actually likely to take, since wholesale-reverting an intentional two-branch except is a much less plausible accident than breaking the datetime handling inside it):

```
FAILED test_a_naive_old_post_leaves_posting_active_false
  AssertionError: Expected 'exception' to not have been called. Called 1 times.
  Calls: [call('Unexpected error in setup-state activity check')]
```

Catches it cleanly, for the right reason. So: the corrected control is real protection against the regression shape most likely to actually happen, not protection against a literal full revert — those are different claims, and only the first one holds for this specific test. The defect itself has no blind spot, because the recent-post test covers it directly and unconditionally.

## 2. The except split — SQLAlchemyError is the right boundary, not just the observed one

`SQLAlchemyError` is SQLAlchemy's own base class for driver/infrastructure failures — `DBAPIError` and everything below it inherits from it, specifically so callers can catch DB trouble uniformly regardless of the underlying driver. That's a textbook-correct choice for a "this must degrade, not break the dashboard" boundary, not a coincidence of what one test happened to throw.

Confirmed both branches fire correctly against the *current* code:
```
test_an_unexpected_error_is_logged_loudly_not_swallowed_at_debug PASSED  (TypeError → logger.exception)
test_a_database_error_stays_quiet PASSED                                (SQLAlchemyError → logger.debug, no .exception)
```

And confirmed the reverse under the full-revert experiment above: the *old* single except swallowed the TypeError at debug, invisibly — which is the entire #918 story, now closed off structurally rather than just for the one instance that happened to surface it.

## 3. The self-caught negative control — covered above, and it's a real catch on his part

Item 1's finding *is* the answer to this: the corrected control genuinely fails for the right reason against the regression it targets (coercion broken, split intact) — it just isn't, and I don't think needs to be, a substitute for the recent-post test's own unconditional coverage of the actual defect.

## Noted for the record, not a review item

Registered what you said about the earlier "pushed" claim being corrected once it turned out the push was rejected — didn't factor into anything here since the current head (`65666ea`) is what CI ran and what I reviewed directly.

## Housekeeping

CI was mid-rerun at dispatch; checked back, all five required checks `pass` (`Test` 3m16s). Pulled the real job log by name: all four new/changed tests present and `PASSED`, full run `3196 passed, 10 skipped`, zero failed. The #918 half is untouched since my last review (identical addition/deletion counts on every file) — not re-verified, per your note that it stands. Every mutation reverted; worktree and scratch branch removed clean.

---

Ship it. The protection gap from my last review is closed for the defect that actually shipped. The one thing that doesn't fully hold — the old-post control against a literal wholesale revert — is a narrower claim than "born-red," worth being precise about, and not a reason to hold this up.

Issue #923, opened 2026-08-19
F.1 burn-down: InstagramAccountRepository has 12 unscoped methods on a derived-tenancy table (RF-G2, enumerated)


chrisrogers37:
Found by the #896 gate rebuild, and filed separately because #896 is the
**mechanism** — instance fixes do not belong in it.

## What the sweep found

`InstagramAccountRepository` has **12 methods that query `instagram_accounts`
with no tenant parameter**:

```
activate  count_active  create  deactivate  get_all  get_all_active
get_by_id  get_by_id_prefix  get_by_instagram_id  get_by_meta_account_id
get_by_username  update
```

`instagram_accounts` is a **derived-tenancy** table: it holds no tenant
reference of any kind — the foreign key runs the other way
(`chat_settings.active_instagram_account_id → instagram_accounts.id`). So a
query against it returns **every tenant's rows** unless the caller supplies a
tenant to join on.

## This is an enumeration of a known risk, not a new discovery

F.1 already classified `instagram_accounts` as tenant-owned and logged the risk
as **RF-G2** (inventory row 5). What was missing was a count and a mechanism —
the extinction gate keyed on parameter spelling, and a table with no tenant
parameter has nothing to spell, so it passed vacuously. #896 fixes that; this
issue records what the fixed gate then found.

## Why the recent fixes did not cover it

The five defects fixed today (#891, #895, #898, #900, #901) landed in the
**service layer** — `telegram_accounts.py`, `telegram_service.py`,
`telegram_utils.py` — verified against the fix commits. The repository is
unchanged and still reads unscoped, with scoping done by callers.

**Caller-side scoping is the fail-open pattern F.1 exists to end.** It is
correct today only for the callers that were audited; the next caller inherits
no protection, which is exactly how #891 happened.

## Current state — pinned, not ignored

`tests/src/repositories/test_f1_tenancy_gate.py` pins all 12 by **equality**, so
a thirteenth arrives as a red test rather than as headroom. Nothing regresses
silently while this is open.

## What closing this requires

Give each of the 12 a tenant parameter and make it fail closed, in the F.1
idiom — `SYSTEM_SCOPE` for the paths that genuinely span tenants (there are
some: an admin account list is legitimate), a required tenant for the rest.
Then remove the entries from `KNOWN_UNSCOPED_DERIVED_ACCESS` in the same diff,
which is what makes the burn-down visible.

Refs #896, #846, RF-G2.

PR #924, opened 2026-08-19
fix(f1): the extinction gate evaluates TENANCY, not parameter spelling (#896)

### LINKED ISSUE #896: The F.1 fail-closed gate keys on the chat_settings_id PARAMETER SPELLING — derived-ownership tables pass it VACUOUSLY, so the tenancy-leak class regenerates on the next such table

chrisrogers37:
Fixes the F.1 gate's root cause. **This is my own blind spot from #846**, so the
diagnosis is stated plainly rather than softened.

## The defect

The extinction check matched the literal string `chat_settings_id` in a
signature. That answers *"does any method default a parameter of that **name**
to None?"* — which is not the question F.1 exists to answer.

A table whose tenancy lives in a JOIN has **no tenant parameter at all**, so
there is nothing for the regex to match and the gate reports green on a table it
**structurally cannot evaluate**. `instagram_accounts` was classified
tenant-owned in the F.1 inventory (row 5, RF-G2) and sailed straight through,
under a green gate, into five live defects.

## Two instances of that spelling defect, both mine, both found by building the fix

1. **`instagram_accounts` holds no tenant reference of any kind.** The FK runs
   the *other* way — `chat_settings.active_instagram_account_id →
   instagram_accounts.id`. No spelling could ever have caught it.
2. **`onboarding_sessions` carries its tenancy as `pending_chat_settings_id`** —
   and the old scanner **explicitly skips anything prefixed `pending_`**. I
   wrote that exemption to dodge a false positive and thereby exempted the
   column that carries the tenancy.

The new rule reads the **FK target**, not the column name, so both classify
correctly.

## The design — and the honest half is the interesting one

Tenancy is **computed from the SQLAlchemy metadata**, never hand-listed: a
hand-list is the same spelling problem wearing a different hat and goes stale on
the next table.

- **DIRECT** — a column keys to the tenant root (any name).
- **CANDIDATE** — no such column, but a tenant-bearing table references it.

**The schema finds the candidates and cannot resolve them.**
`chat_settings.active_instagram_account_id → instagram_accounts` and
`chat_settings.user_id → users` are *structurally identical edges*, while an
account is **owned** by one tenant and a user is **shared** across many. That is
a domain fact, and a classifier that pretended to derive it would be
confidently wrong.

So candidates are adjudicated explicitly, and **an unadjudicated candidate
FAILS** rather than defaulting to safe — a default-to-global classifier would
reproduce this issue's exact vacuous pass, one layer up. That refusal has its
own test.

## Proven on the real tree, not a synthetic string

Following rajan's #914 method: an unscoped `InstagramAccount` query planted in
`membership_repository.py` — **untouched, zero prior findings**.

| gate | result on the identical planted file |
|---|---|
| **new** | **FAILS by name** — `{'membership_repository.py': ['planted_leak::InstagramAccount']}` |
| old | **passes 4/4**, completely blind |

Side by side, so the improvement is demonstrated rather than claimed.

## No over-fire

`users`, `service_runs` and `user_interactions` classify **global**. A gate that
cries wolf gets suppressed, and a suppressed gate is worse than a loose one.
Controls: a tenant-scoped method is not flagged; an oddly-named tenant param is
accepted; a method touching only global models is not flagged; and the
classifier itself has a not-vacuous control (a bug classifying everything global
would pass every other assertion).

## Coverage — stated, not implied

**branden measured that a fixed gate would have protected 2 of the 5** defects.
The other three are shapes this gate cannot see and needs its own assertions
for regardless:

- an **authorization** bypass upstream of the repository layer (#895);
- a **presentation/aggregation** surface reading across tenants (#898);
- a cross-tenant **write** whose target row was selected elsewhere (#900).

**This is not a silver bullet and must not be cited as one.**

## The three repositories #896 named

`ChatSettings`, `Interaction` and `Membership` return **zero** — and the honest
reading is **not** "those three are safe." It is that **this gate's shape does
not apply to them**: none touches a derived-tenancy table (`chat_settings` *is*
the tenant root, `user_chat_memberships` carries its own tenant column,
`user_interactions` is user-plane). Reporting that zero as a clean bill of
health would be this very issue one layer up, so the test says so in its own
name.

## What it did find → #923, not fixed here

**12 unscoped methods in `InstagramAccountRepository`.** Today's five fixes
landed in the **service layer** (verified against the fix commits), which is why
the repository still reads unscoped — caller-side scoping being the fail-open
pattern F.1 exists to end. Pinned by **equality** so a thirteenth reddens, and
filed as **#923**. **No instance fixes in this PR; this is the mechanism.**

## One thing found in passing, and it is fleet-affecting

**`## [Unreleased]` had been lost from `CHANGELOG.md`** — present through
`ecccc23`, gone at `3f72ccd` (#917), almost certainly a conflict resolution.
Restored here.

It matters more than tidiness because it breaks tooling *silently*: every
changelog script anchoring on that heading does a `str.replace` that matches
nothing, changes nothing, and **reports success**. Mine did exactly that minutes
ago, and I caught it only because the commit showed one file where I expected
two. Anyone whose script did the same and did not check has an entry they
believe they wrote.

## Verification

- New gate: **14 passed**; old F.1 gate still **21 passed** (unmodified).
- FC-2 ratchet rc 0.
- Full suite: **3186 passed, 8 skipped, rc 0** — fully green, no failures, so
  there is no before/after name diff to run. CI green 5/5 on `1efe443c` with
  `Test` at 15 steps.
- Vercel is red repo-wide on an account build quota, **including on `main`** —
  ignored per the dispatch; the other five checks are read by name.

Closes #896.


chrisrogers37 (commented):
**Verdict: Ship it**

Head `1efe443`, `mergeable: MERGEABLE` (confirmed fresh, despite #919 landing since your dispatch — the two diffs don't overlap). All five required checks SUCCESS; Vercel confirmed the account-wide quota failure, same on `main`.

**1. The proof — reproduced myself, on the real file, both directions.** One thing worth flagging first: the shipped test (`test_the_OLD_gate_does_NOT_catch_that_same_hazard`) only exercises a **synthetic string snippet**, not the real `membership_repository.py` the PR body describes planting into. So I did that part myself — read the real file, appended a `planted_leak` method querying `InstagramAccount` unscoped (in memory, never written to disk), ran both scanners against it:
```
NEW gate on planted membership_repository.py:   ['planted_leak::InstagramAccount']
NEW gate on UNTOUCHED membership_repository.py: []
OLD gate on planted membership_repository.py:   []
OLD gate on UNTOUCHED membership_repository.py: []
```
Exact match to the PR body's claimed output, on the actual file, confirmed both directions myself rather than trusting the narrative. I couldn't independently place the specific "4/4" figure the PR cites (no 4-test subset in the old gate names `InstagramAccount` or `membership`) — minor, doesn't change the substance, which is fully confirmed. Worth a one-line note that the *shipped regression test* for this claim is synthetic-only; the real-file proof lives in the PR body as a manual verification, not as something that stays pinned if it breaks.

**2. The zero-on-three-repos claim — verified independently from the schema, and it's the honest reading.** Checked the actual models rather than trusting the gate's self-consistent classification: `user_chat_memberships.chat_settings_id` carries a real `ForeignKey("chat_settings.id")` in its own table (genuinely DIRECT). `user_interactions` has exactly one FK, to `users.id` — its `telegram_chat_id` column is a bare `BigInteger`, no FK at all, so there is structurally no schema path to `chat_settings` for the classifier to find (genuinely GLOBAL, by the gate's FK-graph-only design). So "zero" here really is "this gate's shape doesn't reach these tables," not "evaluated and clean" — confirmed, not assumed. Worth noting: #896 itself calls `InteractionRepository` "Class-6 tenant-owned... four unscoped readers today... a latent #891" — that risk is real and still completely unaddressed by this gate. The PR's own honesty about the zero is what keeps that risk visible instead of closing it prematurely.

**3. Found all instances — confirmed structurally, which is a stronger answer than checking his work by hand.** Both defect shapes are now schema-computed rather than hand-listed: `tenant_candidates()` walks every table's FK graph for the reverse-derived case (asserted to be exactly `{"instagram_accounts", "users"}` — nothing else exists to find, today), and the DIRECT check matches on FK *target*, never column name, so an oddly-named column can't hide from it regardless of spelling. Given that design, "did he miss a third instance of either shape" isn't a question that depends on his diligence — the classification is exhaustive over the current schema by construction. That's the right level of guarantee for a root-cause fix.

**4. Over-fire — checked on the real tree, not just synthetic controls.** The unit-level negatives all pass (tenant-scoped method, oddly-named param, GLOBAL-only method). More convincing: `test_the_sweep_matches_the_pinned_burn_down_exactly` runs the sweep against every real file in `src/repositories/*.py` and asserts the result equals the known 12-method dict exactly — meaning every *other* file in that directory produces zero findings against the real tree, not a cherry-picked example. Mutated the one adjudication that makes this work — flipped `instagram_accounts` from DERIVED to GLOBAL — and confirmed it's genuinely load-bearing: 4 tests redden, and the standing sweep silently drops to finding nothing at all, reproducing the exact vacuous-pass shape this PR exists to end. The design's correctness rests on getting that one call right, and it's pinned by a real assertion, not decoration.

**5. #923 split — confirmed right, and cross-checked against my own #919 review rather than taken on faith.** Pulled #923: open, 12 methods, matches `KNOWN_UNSCOPED_DERIVED_ACCESS` exactly. I reviewed #919 directly and independently confirmed its five fixes landed at the service layer (`instagram_account_service.py`) without ever adding a tenant parameter to the underlying repository calls (`account_repo.get_by_id_prefix`, `account_repo.count_active` are still called unscoped internally) — so #923's claim that the repository layer remains genuinely untouched isn't just asserted here, it's independently corroborated by what I saw fixing those defects. And the coverage framing — "protects 2 of the 5" — is exactly what both this PR's docstring and the test module's own header state, sourced to branden, with the other three shapes (#895 authorization, #898 presentation, #900 write-target-selection) named as structurally out of this gate's reach regardless of correction. No broader claim implied anywhere I could find.

Root-cause fix, verified past what the shipped tests alone would have shown me. Two small follow-ups worth naming to whoever picks up #923/#896's remaining thread: the real-file planting proof isn't pinned as regression coverage, and `InteractionRepository`'s latent risk is still open and now correctly labeled rather than closed.

Issue #905, opened 2026-08-19
get_by_id_prefix passes an unescaped LIKE pattern — wildcard metacharacters in a caller-supplied prefix are not neutralised


chrisrogers37:
Found by `rajan` while adversarially probing #902 with injection-shaped payloads. **Not exploitable through the callback path** — ownership gates regardless of what the prefix matches, which he verified — so this is a follow-up rather than a security incident.

## The gap

`get_by_id_prefix` builds a `LIKE` pattern from a caller-supplied prefix **without escaping the wildcard metacharacters**. A prefix containing `%` or `_` is therefore interpreted as a pattern rather than as a literal.

Pre-existing; #902 neither introduced nor widened it.

## Why it is worth fixing anyway

The current safety is **incidental, not structural**: the ownership gate refuses regardless of how many rows the prefix matches, so a widened match set changes nothing today. That is the same distinction drawn on #868 about the ratchet exemption — *a thing that happens to be right is not a thing that must be right, and only the second belongs in a control.*

The protection lives in a **different component** from the defect. Any future caller of `get_by_id_prefix` that does not sit behind that gate inherits an unescaped pattern match, and nothing in the function signals that it is relying on a caller to be gated.

## Fix

Escape `%`, `_` and the escape character itself in the prefix before it reaches the `LIKE`, or use a comparison that does not interpret metacharacters at all. Add a test with a wildcard-bearing prefix asserting it matches **literally** — currently such a prefix silently widens the match set and no test would notice.

## Context

Filed off the #902 review, where all four adversarial classes (full-UUID-of-foreign-row, forced cross-tenant prefix collision, wildcard/injection-shaped, empty) were tested against a real database and all refused correctly.

chrisrogers37:
Closed by PR #925 (merged 2026-08-19), reviewed by `rajan` — **Approve**.

## This issue named the wrong site, and that is the finding

It pointed at `instagram_account_repository`. A sweep found the same expression in **`queue_repository`** — and **that** is the reachable one: its caller passes `SYSTEM_SCOPE` and checks ownership afterwards, with the prefix arriving in **user-controlled Telegram callback data**.

`rajan` then sharpened it further: the site this issue named is **not merely safe, it is UNREACHABLE** — zero production callers, superseded by #901's Python-side `startswith()`.

**So a faithful fix-what-the-issue-says pass would have patched dead code and closed this green**, leaving the live instance untouched. Only the sweep found the real one.

[...]

PR #925, opened 2026-08-19
fix(db): an id prefix is matched literally, not as a LIKE pattern — 2 sites (#905)

chrisrogers37:
`get_by_id_prefix` built `cast(col, String).like(f"{prefix}%")` from a caller-supplied prefix. `%` and `_` are LIKE metacharacters, so a prefix carrying either was interpreted as a **pattern** rather than a literal — `_` matches any single character, `%` matches any run — and the match set silently widened.

## Not exploitable today, which is the reason to fix it

Both call sites sit behind an ownership gate that refuses regardless of how many rows the prefix matched, so a widened set changes nothing right now.

**The protection lives in a different component from the defect.** Any future caller of these functions not sitting behind that gate inherits an unescaped pattern match, and nothing in either signature warns them. A thing that happens to be right is not a thing that must be right, and only the second belongs in a control — #868's distinction, applied here.

## It was two lookups, and the one the issue did not name is the reachable one

The issue named `instagram_account_repository`. A sweep for built LIKE patterns across `src/` found the same expression in **`queue_repository.get_by_id_prefix`** — and that is the instance reachable today: its caller passes `SYSTEM_SCOPE` deliberately and checks ownership *afterwards*, which is the same "protection elsewhere" shape. A fix that patched only the named site would have left the live one.

Sweep bounds: every `.like(` / `.ilike(` under `src/` whose pattern is built rather than a literal constant. Two hits, both fixed, no existing escape helper anywhere.

## Escaping was considered and rejected

`.startswith(prefix, autoescape=True)` would work. It is rejected for one reason: **it stays correct only while a keyword argument stays present.** An edit that drops `autoescape=True`, or a new call site that never adds it, reintroduces the defect silently — and the operator still reads fine.

Comparing a fixed-length `left()` slice removes the pattern language from the expression entirely. There is no metacharacter to escape because nothing interprets one, and no argument to forget. That is also why it is a shared `id_prefix_matches` helper rather than two inlined expressions: a third id-prefix lookup should get the safe behaviour by having nothing to remember.

## Born-red, with a discriminator that cannot pass for the wrong reason

A UUID contains only hex digits and dashes, so **no real id can contain `%` or `_` literally**. A wildcard-bearing prefix therefore matches MORE than one row under the defect and **ZERO** under the fix — never the same answer.

Measured against the reverted expression:

```
FAILED test_an_underscore_is_a_literal_not_a_single_character_wildcard
FAILED test_a_percent_is_a_literal_not_a_run_wildcard
FAILED test_a_bare_percent_does_not_match_everything
        assert <InstagramAccount Account A> is None
        where <InstagramAccount Account A> = get_by_id_prefix('%')
```

One metacharacter reaching the whole table, which is the property that matters for a caller not behind the gate.

Two cases beyond the ones asked for:

- **The escape character itself.** A backslash is the trap in the *fix* rather than the defect: an escaping implementation that forgot to escape its own escape would pass the wildcard cases and fail this one. This expression has no escape character to forget, and the test pins that.
- **The empty prefix is stated, not changed.** It matched everything before and still does. This closes a metacharacter hole, not a missing-input check, and quietly turning empty into "match nothing" would be a behaviour change hiding inside a security fix. Callers needing a minimum length should say so themselves.

One structural test asserts neither repository builds a LIKE pattern from an f-string any more. It is honest about its scope: it did **not** catch the mutation above (which lives in the helper), and it guards a different regression — a repository going back to an inline pattern.

## Verification

- New tests: **7 passed** against live Postgres; 3 of them born-red.
- Full non-integration suite: **7 failed / 2728 passed** — the seven are the `test_telegram_service.py` cluster that fails identically on `main` on this host; `comm -13` against the baseline is empty.
- The run also reported **4 errors**, which a FAILED-name diff does not cover — checked rather than assumed: `ENCRYPTION_KEY not configured` at setup, from my local `.env`, in a file this change does not touch. With the key set, all 20 of those pass. **Zero regressions.**

Refs #905, #868, #902


chrisrogers37 (commented):

The reachability claim is not just correct, it understates in the safe direction — the named site is stronger than "safe," it's unreachable. Everything else checks out precisely, with one genuinely useful nuance on the escape-character pin.

## The headline — verified by tracing every caller, not by trusting the framing

`InstagramAccountRepository.get_by_id_prefix` has **zero production callers.** Grepped every reference: the only caller anywhere is its own test file. The service-layer method with the similar name, `get_account_by_id_prefix`, doesn't call it at all — it does its own thing entirely, iterating `list_accounts()` (already tenant-scoped) and matching with Python's `str.startswith()`, which has no wildcard semantics to exploit. That replacement's own docstring cites `#901` as the reason — a *different* security fix, from before this PR, already retired the vulnerable path's real-world usage. So the named site isn't merely "protected," it's dead code with tests still exercising it directly.

`queue_repository.get_by_id_prefix` is genuinely live: both call sites in `telegram_accounts.py` (`handle_post_account_switch`, `handle_back_to_post`) pass `chat_settings_id=SYSTEM_SCOPE` — deliberately cross-tenant — with `short_queue_id` coming straight from Telegram callback data (user-controlled), and ownership is checked only afterward via `_caller_owns_queue_item`. If the issue's fix had patched only the named site, this is exactly what would've shipped unpatched.

**Third site:** my own sweep (`grep -rn ".like(\|.ilike(" src/`, plus a separate pass over raw-SQL `LIKE` in migrations for anything not going through the ORM) found nothing beyond the two fixed. The one other `LIKE` hit in the codebase is a hardcoded literal pattern in a migration function, not caller-built. Two, confirmed independently.

## The design decision — `left()` genuinely has no equivalent fragility

Traced through what it would take to reintroduce the bug under each approach. Escaping (`autoescape=True`) fails silently on a dropped keyword — the code still reads as `.like(...)`, still compiles, still looks like the fix. The `left()` comparison has no analogous failure mode: there's no keyword to drop, because the operator itself (`=`, not `LIKE`) doesn't interpret metacharacters — reintroducing the bug would require someone to actively swap the operator back to pattern matching, which is a visible, deliberate change rather than an incidental omission. That's the real incidental-versus-structural line, and it holds under scrutiny rather than just sounding right.

## Born-red — reproduced directly

Reverted both repositories to the original `cast(col, String).like(f"{prefix}%")` and ran the suite:

```
FAILED test_a_bare_percent_does_not_match_everything
  AssertionError: assert <InstagramAccount Account A (@None)> is None
```

`get_by_id_prefix('%')` returns a row under the reverted expression — exactly the claimed measurement, reproduced rather than trusted.

## The escape-character pin — does not bite against the actual defect, and that's worth stating precisely

Same revert, same run: `test_the_escape_character_is_also_a_literal` **stayed green.** Reasoned through why before confirming empirically: plain Postgres `LIKE` (no `ESCAPE` clause) treats `\` as an ordinary character by default — so the *original* bug never gave backslash any special meaning either, and this test can't discriminate old-buggy from new-fixed on that specific input. What it actually pins is a property of a **different, rejected** implementation: an escaping-based fix that used `\` as its escape character and forgot to escape a literal backslash in the input would fail here. Since that path was rejected in favor of `left()`, the pin is a forward-looking spec test — real and worth keeping (it documents the contract if anyone ever revisits the escaping approach) — but it is not what protects this PR's actual defect. The wildcard tests and the structural "no `.like(f"` in source" test are what do that work, and both reddened correctly under the same revert.

## Empty prefix — stating it is the right call, and the reason generalizes

`left(col, 0) == ''` matches every row, same as the old empty-LIKE-prefix did — unchanged behavior, correctly undocumented-as-a-non-issue rather than silently patched. Checked why this is actually safe rather than just asserted: `_caller_owns_queue_item` validates whatever row comes back purely by ownership, with no dependency on *how* it was matched — `if queue_item is None: return False`, otherwise deferred to the shared ownership predicate. An empty prefix returning an arbitrary row gets the identical downstream treatment as a wildcard-widened match. Same mechanism neutralizes both; stating rather than guarding is consistent, not just convenient.

## Noted, not chased

The 4 local errors he disclosed and checked (missing `ENCRYPTION_KEY` in a local `.env`) are absent from the real CI log — confirmed the actual CI `Test` job has zero failures and the errors don't correspond to anything in this suite specifically. Applying the rc-plus-scoping-plus-count discipline rather than reading a clean name-diff as sufficient is the right instinct, and it held up.

## Housekeeping

Base `b03a2dc9` confirmed against a fresh `origin/main` fetch. Pulled the real CI `Test` job log by name: all 7 new tests present and `PASSED`, full run `3217 passed, 10 skipped`, zero failed. Every mutation reverted; worktree and scratch branch removed clean.

---

Ship it. Reachability, the design decision, the born-red claim, and the empty-prefix reasoning all hold under direct verification. One precise, non-blocking clarification: the escape-character pin protects a rejected alternative's failure mode, not this PR's actual defect — worth knowing, not worth changing.

PR #926, opened 2026-08-19
fix(dashboard): a refused health check reads as refused, not broken (#922)

### LINKED ISSUE #922: Admin-gating /system-status makes non-admins see 'Failed to load health data' — a security fix that renders as a malfunction
Found by `navi` reviewing #919, by **tracing the actual frontend** rather than accepting the PR's claim that the consequence was named.

## What happens

#919 correctly admin-gates `/system-status` (closing #898, a deployment-wide disclosure that had grown worse). The PR states the consequence — the card becomes admin-only — and that statement is **true for a developer reading the code**.

**A non-admin tenant sees `"Failed to load health data"`** — generic, and **indistinguishable from a bug**.

## Why this is worth its own issue

The security fix is right and should not be reverted. But the user-visible result of a deliberate authorisation decision currently looks like a broken feature:

- A user cannot tell "you do not have access to this" from "this is broken".
- Neither can we, from a support report. Someone reporting a broken dashboard would send us chasing a defect that does not exist.
- And it trains users to ignore that error, which is where a *real* failure of the same card would then hide.

## The general shape

"Consequence named" was true at the layer it was written for and false at the layer a user occupies. Same distinction that has run through this repo all day: **a claim that holds for one reader and not another**. The reviewer caught it by opening the frontend instead of reading the PR body — the same move that found the original disclosure by clicking rather than reading.

## Fix direction

Make the refusal legible as a refusal: an explicit not-authorised state for the card, or hide it entirely for non-admins. **Do not** restore the data.

chrisrogers37:
Fixes #922. **The gate is correct and unchanged — only the presentation moved.**

After #898 gated deployment health to administrators, a non-admin expanding System Status saw **"Failed to load health data"**: an authorisation decision rendering as a malfunction. That costs twice — a support report sends someone chasing a defect that does not exist, and it teaches users to ignore that line, so a *real* failure of the same card would hide behind it.

## The decision: keep the card, make the refusal legible

I picked "explicit not-authorised" over "hide the card", and the surface decided it rather than taste.

`_loadSystemStatus` calls `_renderSetupChecklist()` **before** the API request, from local state:

```js
// Render setup checklist from local state (no API call needed)
this._renderSetupChecklist();
```

So the card has two halves. The **SETUP** half — Instagram, Google Drive, Media Library, Posting, Delivery — is the tenant's *own* data and needs no admin rights. Only the **HEALTH** half is gated. Hiding the card would have removed genuinely useful tenant information from every non-admin — which is most users — in order to suppress one line. The collapsed badge is derived from setup state too (`3/5 setup items configured`), so it stays meaningful either way.

## What changed

A 403 renders a refusal; **anything else still renders a fault**:

```js
container.innerHTML = err && err.status === 403
    ? '<div class="card-body-empty">Deployment health is available to administrators only.</div>'
    : '<div class="card-body-empty">Failed to load health data</div>';
```

That distinction *is* the fix — swapping one blanket message for another would have missed the point entirely.

To make it possible, `_api` and `_apiGet` now carry the HTTP status on the thrown error. Branching on **status** rather than on the detail string keeps the UI from coupling to server copy the backend is free to reword.

## Verified by clicking, all three arms

Both controls run rather than assumed — a refusal state that showed for everyone, or a fault path that had quietly become a refusal, would each be a different bug.

| arm | how | result |
|---|---|---|
| **non-admin** | real click, `users.role = member` | checklist intact + refusal line |
| **admin** | `role` promoted to `admin`, then reverted | full HEALTH section, unchanged |
| **fault** | 500 and 503 stubbed at the route | `Failed to load health data` |

```
[OK ] HTTP 500 -> fault    | setup rows still 6 | 'Failed to load health data'
[OK ] HTTP 503 -> fault    | setup rows still 6 | 'Failed to load health data'
[OK ] HTTP 403 -> refusal  | setup rows still 6 | 'Deployment health is available to administrators only.'
```

The setup checklist survives all three.

| non-admin | admin | genuine fault |
|---|---|---|
| non-admin | admin | fault |

## Design note

One quiet line, in the same muted style as the other empty states, sitting where the HEALTH section would be. No alarm colour, no icon, no explanation of how to become an administrator — the fewest words that make the refusal legible. The checklist above keeps its visual weight because it is the part a non-admin can act on.

## Bounds

Chromium 147, local instance. The admin arm was exercised by promoting a real `users.role` and reverting it, not by mocking the predicate. `ruff` untouched — this diff is `app.js` and the changelog only; no Python changed, so no test suite is implicated.

Refs #922.


chrisrogers37 (commented):
**Verdict: Ship it**

Head `cdaab23`, `mergeable: MERGEABLE`, all five required checks SUCCESS (Test finished green since dispatch), Vercel is the account-wide quota failure, confirmed same on `main`.

**The both-directions check — mostly holds, with one real gap in the direction you didn't name.** Traced the actual raise sites in `helpers.py` rather than trusting the summary:

- Fault → still fault: confirmed by construction. `_validate_admin` is the *only* thing in this route that can raise before `HealthCheckService().check_all()` runs, and its only non-2xx outcomes are 401/403. Nothing downstream can surface as 403 for a non-auth reason. This direction is solid.
- Auth → refusal: **not fully closed.** `_validate_admin` raises **401** on auth failure (bad/expired/missing init_data) and **403** only when authenticated-but-not-admin — two distinct auth-failure shapes, confirmed from `_validate_auth`/`_validate_admin`'s own docstrings and call order. The fix's discriminator is a strict `err.status === 403`. A 401 on this endpoint still falls through to `"Failed to load health data"` — the same misclassification this PR exists to close, just for a narrower trigger. I judge this low-severity and not blocking: `init_data` is set once at Mini App launch and validated there (a stale/expired token would typically surface earlier, on the initial `/init` call, not deep inside System Status), and this doesn't regress anything — every *other* `_apiGet` caller in this file has always collapsed 401 into the same generic bucket. But it's real, and it's the same class of gap you asked me to hunt for in the other direction. Worth a one-line follow-up (403-or-401 → refusal) rather than blocking a fix that's net strictly better than what shipped.

**The local-state claim — verified from source, not taken on the PR's word.** `_renderSetupChecklist()` (called before `_apiGet` in `_loadSystemStatus`, confirmed at the exact line) reads `this.setupState`, which is populated once in `init()` from `/api/onboarding/init` — the app's bootstrap call, reached by every authenticated tenant regardless of role, well before System Status is ever opened. So the claim holds precisely as stated: that data is already fetched, already tenant-scoped, and hiding the card would have discarded real information non-admins already legitimately have.

**The admin-promotion arm — I can't reproduce the live click-through in this environment, so I verified the structural reason it should hold instead of claiming I did the same thing he did.** No Python changed in this diff, so the *success* path (`_renderHealthChecks(data.checks)` on a 200) is byte-identical to before — untouched by this fix. What decides whether promotion "just works" is whether `is_system_admin` is live or cached: read it directly — `self.user_repo.get_by_telegram_id(telegram_user_id)` inside the method itself, no caching, no session pinning. A role UPDATE takes effect on the very next request by construction. That's a real, independent confirmation the claim is architecturally sound, but it is not the same evidence as watching the card populate after a real promotion — naming that distinction rather than blurring it.

**The record item — confirmed.** Single commit, exactly 2 files (`CHANGELOG.md`, `app.js`), matching what's shown. Nothing else in the diff, nothing else in the commit history to have leaked from. The repair held.

**One thing beyond what was asked, flagged because it's now happened twice today:** this PR's CHANGELOG insertion creates a duplicate `### Fixed` header — same shape as a duplicate `### Added` I found reviewing #916 a few hours ago. Both are cosmetic (Changelog Check passes either way) but it's the second instance of the same pattern across parallel PRs landing changelog entries same-day — worth a look at whatever's generating these insertions, not urgent.

Net: the fix does what #922 asked, verified independently rather than restated, with one real but low-severity gap (401) named precisely because that's the standard you set for the direction you did name.

PR #927, opened 2026-08-19
security(#923): a live cross-tenant disclosure in the remove-confirm callback

### LINKED ISSUE #923: F.1 burn-down: InstagramAccountRepository has 12 unscoped methods on a derived-tenancy table (RF-G2, enumerated)

chrisrogers37:
Closes the 12 unscoped methods split out of #924. **Classified all 12 before
fixing any — which is what turned up a live defect.**

## The live one

`handle_account_remove_confirm` is a **Telegram callback handler**.
`telegram_service.py:381` routes the `account_remove` prefix straight to it, so
`account_id` arrives in **callback data** and is chosen by whoever taps. It read
the account unscoped and rendered `display_name` and `instagram_username` into
the confirmation card.

**The contrast is the finding.** `handle_account_remove_execute`, immediately
below it, passes `chat_id` into `deactivate_account`, which enforces ownership
and refuses. **A guarded write behind an unguarded read** is the existence
oracle `get_owned`'s own docstring warns about — the card disclosed an account
the very next step would decline to touch.

Sixth in the #891 / #895 / #898 / #900 / #901 family, and reachable from inbound
data. Fixed **at the door**, reusing `_require_account_ownership` so a foreign
id and an invented id produce the identical not-found shape; a distinct message
would be the oracle this closes.

**Born-red, and the assertion is on the disclosure rather than the refusal** — a
test checking only *"it answered not found"* passes while the card leaks.
Removing the guard reddens
`test_a_foreign_account_is_not_disclosed_in_the_confirmation_card`. Paired with
an owner-still-sees-their-own-card control, because a handler that refused
*everything* would pass the first test and break the feature, and a gate that
blocks the legitimate path gets reverted rather than fixed.

## The classification, all 12 — the durable output

**This is the part worth keeping**, more than the fix: the next person to touch
this file needs to know a method is *intentionally* global rather than
rediscovering the question and either "fixing" it or leaving it alone for the
wrong reason.

### Genuinely unscoped — 1

- **`get_by_id`** via `handle_account_remove_confirm` — **live disclosure,
  fixed here.** Its other doors (Switch, `get_active_account`, backfill) already
  guard or derive the id from the tenant's own row.

### Global BY DESIGN — 3 (operator surfaces)

- **`get_all`**, **`get_all_active`** — reached only from
  `list_all_accounts_unscoped`, CLI-only, and the name declares it.
- **`get_by_username`** — CLI lookup *and* `_validate_new_account`'s duplicate
  check.

### Global BY NECESSITY — 2 (and this is the one to be careful with)

- **`get_by_instagram_id`**, **`get_by_meta_account_id`** — OAuth identification
  runs *before* a tenant is known, because it is the step that **finds** the
  tenant. Scoping them would require the answer they exist to produce. The web
  door guards downstream (#900).

**`get_by_username` belongs here too, and it is the trap in this set**: the
duplicate check it serves cannot reject a duplicate it cannot see. Scoping it
would have *looked* like a security improvement and silently broken the
protection it provides. That is the concrete reason "classify before fixing"
was not process.

### Caller-guarded — 4

- **`create`** (stamps tenancy via the token), **`update`**, **`activate`**,
  **`deactivate`** — reached only through doors that enforce ownership; #900
  added the web one.

### Dead — 2, deleted

- **`get_by_id_prefix`**, **`count_active`** — zero production callers. Checked
  rather than assumed: the hits that looked like callers were `queue_repo` and
  `media_repo`, different classes with the same method names. Their tests went
  with them.

## The table

| method | class | why |
|---|---|---|
| `get_by_id` | **live defect** (one door) | remove-confirm — fixed here |
| `get_all`, `get_all_active` | global by design | operator-only via `list_all_accounts_unscoped` |
| `get_by_username` | **global by necessity** | the duplicate check it serves cannot reject a duplicate it cannot see |
| `get_by_instagram_id`, `get_by_meta_account_id` | global by nature | OAuth identification runs *before* a tenant is known — it is the step that finds the tenant |
| `create`, `update`, `activate`, `deactivate` | caller-guarded | reached only through doors that enforce ownership (#900 added the web one) |
| `get_by_id_prefix`, `count_active` | **dead** | zero production callers → **deleted** |

The two "dead" ones needed checking rather than assuming: the hits that looked
like callers were `queue_repo` and `media_repo` — different classes with the
same method names. Their tests were deleted with them.

## The WHY is in code, not just here

Per your point 3 and navi's #919 finding — *"intentionally global" was legible
to a developer reading the PR and invisible to the next person reading the
method* — every one of the remaining methods now carries its classification in
its own docstring, including why scoping `get_by_username` would silently break
the duplicate check.

## The gate does NOT cover this fix

Stated plainly, per your point 4. **The gate evaluates the repository
signature**, so a door-level guard — which is where this fix lives — clears
nothing. The burn-down went **12 → 10 only because two methods were deleted**;
an entry leaves that list only when the *method* takes a tenant.

Reading a stable count as "no progress" would be wrong, and reading it as "10
defects" would be wronger. The burn-down comment now says both.

For the ratio you asked about: of the five earlier defects branden measured, the
gate would have caught 2. **For this one it catches 0** — the gate found the
*method*, but a human reading the callers found the defect, and the gate cannot
see the caller at all.

## Verification

- Affected suites: **37 passed** (tenancy gate + callback authorization gate +
  account repository).
- Callback gate: 10 → **13** tests.
- FC-2 ratchet rc 0.
- **CI green 5/5 BY NAME on `c7524512`** — Test (15 steps), Lint, Security
  Scan, Changelog Check, FC-2 ratchet.
- **Two earlier greens on this branch are NOT evidence for this head**, and are
  called out rather than left in the history to be misread: `a44a6512` ran
  against the pre-rebase base, and `130d595b` was green and then went **red**
  once #925's tests met the deleted method. Only the run on the current head
  counts.
- Local full suite: **3220 passed, 8 skipped, rc 0** — measured on
  `c7524512` itself with a clean working tree, so it describes this head rather
  than a superseded one. (The earlier `3213` figure described a tree that no
  longer exists and was removed rather than carried forward.) The run against
  the intermediate head caught **exactly the two failures CI caught** — it was
  simply slower, so CI reported first. The check was not skipped; it was
  outrun.
- Affected suites on the current head: **17 passed**
  (`test_instagram_account_repository` + `test_id_prefix_is_literal`), plus the
  callback authorization gate at 13.
- FC-2 ratchet rc 0; `ruff check` and `ruff format --check` both clean — they
  are separate gates and an earlier push failed Lint because I had verified
  only the first.
- Vercel is red repo-wide on the account build quota; the other five checks are
  read by name.

Closes #923.


chrisrogers37 (commented):
**Verdict: Ship it**

Head `130d595` is now `CONFLICTING` against main — real, traced precisely, and mechanical rather than substantive. Details at the end; the code itself is solid.

**1. Fix at the door — reproduced myself, with the leaked data visible in my own test output.** Worktree at head, removed the `_require_account_ownership` block from `handle_account_remove_confirm`, ran the disclosure test:
```
FAILED ...test_a_foreign_account_is_not_disclosed_in_the_confirmation_card
Calls: [call('⚠️ *Confirm Remove Account*\n\n...📸 *acct-903134c4*\nUsername: @handle_e5254a...')]
```
The foreign tenant's actual display_name and username are right there in the failure — this isn't "it didn't refuse," it's the disclosure itself, caught live. The other two tests in the class behaved exactly as they should under the same mutation: the owner-control stayed green (unaffected), and — worth noting precisely — the invented-id test *also* stayed green, because that path was already correctly handled by the pre-existing `if not account` check before this fix existed. So only one of the three tests is actually load-bearing for *this* defect; the other two are controls, and they're doing real control work, not decoration. The gate genuinely lives at the handler, not a repository unit — `account_repo.get_by_id` (which the confirm step still calls) is unchanged and still unscoped; the door is what moved.

**2. Spot-checked the other 11, found one real gap in the classification's own completeness — not in its conclusion.** Traced actual callers rather than trusting the docstrings:
- `list_all_accounts_unscoped`'s only caller is `cli/commands/instagram.py:513` — confirmed, "operator surface" holds.
- `get_by_username`'s docstring says "two callers: a CLI lookup, and `_validate_new_account`'s duplicate check." There's a third, live one: `src/services/integrations/instagram_login_oauth.py:180` — the legacy, currently-in-production OAuth completion flow — calls `find_existing_account_for_oauth(meta_account_id=ig_user_id, username=username)` with a **real** username, which reaches `account_repo.get_by_username` through the cross-flow-recovery branch. I believe the underlying verdict is still right — this is the same "identification precedes tenancy" shape already applied to its siblings — but the docstring's own caller count is wrong, and this is exactly the kind of thing that misleads "the next person" the whole point of writing this down is meant to protect against. Worth a follow-up line, not a blocker.
- Confirmed via `git grep` against the actual branch (not my own stale local checkout, which cost me one false alarm before I caught it) that `get_by_id_prefix` and `count_active` genuinely have zero callers anywhere in `src/`, `cli/` — independently corroborated by my own #919 review, where I watched both service-layer callers get rewritten to stop calling them.

**3. The classification is in the PR body — confirmed, and it matches the code.** The table, the four-way split, and the `get_by_username` trap ("scoping it would have looked like a security improvement and silently broken the protection it provides") are all there, and I cross-checked each entry's docstring in `instagram_account_repository.py` against the table — they agree. One purely mechanical nit: `get_all_active` and `get_by_username` each ended up with the OLD docstring left in place as a dead, orphaned second string literal right after the new one (harmless at runtime — Python only recognizes the first as `__doc__` — but it's clutter `ruff` doesn't catch and a future reader shouldn't have to puzzle over). Two-line cleanup, not blocking.

**4. Swept the whole file for the confirm/execute pattern — this is the only instance.** Read all ten handlers in `telegram_accounts.py`. Everything else either reads from the caller's own tenant-scoped list (`get_accounts_for_display(chat_id)` — `handle_account_selection_menu`, `handle_remove_account_menu`, `handle_post_account_selector`, `handle_cycle_account`) or gates *before* reading anything for display (`handle_account_switch`, `handle_post_account_switch` — the latter's prefix lookup was already fixed in #919). `handle_back_to_post` gates a different resource (queue items) through an unrelated, already-correct mechanism. No other pair in this file has the "confirm reads unscoped, execute refuses" shape.

**Mergeability — traced precisely rather than treated as a blocker.** `#925` (LIKE-pattern escaping, merged after I fetched this head) independently touches the same `get_by_id_prefix` method #927 deletes — a genuine edit/delete conflict, but not a disagreement: #925's own text names the *queue* repository's copy as "the reachable instance," and its account-repo fix is defensive on a method that was already unreachable by the time it landed (#919 rewired the only real caller before #925 merged). Resolution is mechanical — take the deletion, #925's queue-side fix is untouched and merges clean on its own. Also a third occurrence today of the same duplicate-CHANGELOG-header pattern from #916/#926 — noted, not chasing further.

Sixth of the day, and the empirical bar held: reproduced the leak with real data in the failure output, not just a refusal.

PR #931, opened 2026-08-19
fix(dashboard): enumerate the health-card failure states (#928)

### LINKED ISSUE #922: Admin-gating /system-status makes non-admins see 'Failed to load health data' — a security fix that renders as a malfunction

### LINKED ISSUE #928: A 401 from _validate_admin still renders as a fault — #926 fixed 403 but the unauthenticated arm falls through
Found by `navi` reviewing #926, **in the direction the issue did not name**. Non-blocking, narrow trigger, no regression — filing so it is not rediscovered as new.

## The gap

#926 made the refusal legible: **403 now reads as a refusal**, while 500/503 correctly still read as a fault. That fixed #922, where a deliberate authorisation decision rendered as `"Failed to load health data"` and was indistinguishable from a malfunction.

But `_validate_admin` raises **401** as well as 403:

- **403** — authenticated, not an admin → now renders as a refusal ✅
- **401** — not authenticated → **still falls through to the generic fault message** ❌

So the exact defect #922 described survives, one status code over.

## Why it is low severity but still real

The trigger is narrow — a request reaching this endpoint unauthenticated is unusual in the Mini App flow — and nothing regressed. But the failure mode is identical to the one just fixed: **a user told "this is broken" when the truth is "you are not signed in"** cannot act on it, and neither can we from a support report.

It is also the same shape as the timing residual on #921: **the fix closed the channel it was pointed at, and the adjacent one stayed open.** Worth checking whether any other status from that path reaches the generic branch.

## Fix direction

Route 401 to an explicit not-authenticated state, distinct from both the refusal and the fault. Three outcomes, three messages — the point of #922 was that collapsing any two of them destroys information a user needs.

## Unrelated, noted in the same review

A **duplicate `### Added` CHANGELOG section header** appeared across parallel same-day PRs — `navi` saw it on #916 as well as here. Cosmetic, but twice in one day suggests the changelog tooling does not merge sections cleanly under concurrent PRs. For whoever owns that tooling.

chrisrogers37:
Fixes #928 — the completion of my own #926, in the direction I did not name.

#926 made a 403 read as a refusal and left everything else a fault. That closed #922, but `_validate_admin` also raises **401**, so the exact defect #922 described survived one status over: a caller whose credential had expired was told the card was broken. navi found it by asking what else that validator raises.

## Enumerating the whole surface found a fourth case and a body-shape bug

Measured against the running endpoint rather than reasoned about:

```
valid non-admin token : 403
garbage token         : 401  {"detail":"Invalid authentication credentials"}
chat_id=abc           : 422
after 40 requests     : 429  {"error":"Rate limit exceeded: 30 per 1 minute"}
```

**429 is genuinely reachable** — this route carries no `@limiter.limit`, so the global `30/minute` default applies via `SlowAPIMiddleware`.

And it exposed a second, quieter problem: **slowapi's 429 body is `{"error": ...}`, not `{"detail": ...}`.** Both `_api` and `_apiGet` read only `detail`, so every rate-limited call in the app was already surfacing a bare `Request failed`. They now read either key. That one was invisible until the enumeration forced me to look at the actual body.

## The split is by what the user can do differently

| status | message | user's next move |
|---|---|---|
| 401 | Your session has expired. Reopen this page from Telegram. | reopen |
| 403 | Deployment health is available to administrators only. | nothing — by design |
| 429 | Too many requests. Try again in a moment. | wait |
| 422 / 5xx / no status | Failed to load health data | report it |

`_failureMessage` holds the enumeration because **401 and 429 mean the same thing on every card**, while the 403 copy differs per card and so is supplied by the caller.

**On the residual, since that was the substance of the issue:** 422, 5xx, and a network error carrying no status at all fall to the fault message *deliberately*. They are alike in the only way that matters here — the user cannot respond to them differently. A residual is not the same thing as a catch-all; the defect in #928 was that a **named case was hiding inside one**, and the fix is that nothing actionable is left in there now.

## Verified by clicking — eight runs

Six stubbed at the route, and **two driven live against the real server**:

```
[OK ] 401 -> 'Your session has expired. Reopen this page from Telegram.'
[OK ] 403 -> 'Deployment health is available to administrators only.'
[OK ] 429 -> 'Too many requests. Try again in a moment.'
[OK ] 500 -> 'Failed to load health data'
[OK ] 503 -> 'Failed to load health data'
[OK ] 422 -> 'Failed to load health data'

LIVE 401 (credential invalidated at request time) -> session message
LIVE 403 (real non-admin, no stub)               -> refusal
```

The live 401 is the real-world shape rather than a contrivance: the page loads with a valid credential and the request is refused afterwards, which is what happens when a session expires while the dashboard sits open.

**One thing worth flagging about my own harness:** the live 401 arm initially reported a *pass* while doing nothing. The token's colons are percent-encoded in the request URL, so my literal string replace matched nothing and the request went through unmodified — I only caught it because the result was the 403 message rather than the 401 one. It now asserts the rewrite actually applied before continuing, so it cannot silently no-op again.

| 401 (live) | 403 (live) | 429 |
|---|---|---|
|  |  |  |

The setup checklist survives all eight runs — it renders from local state before the request, so no failure mode costs a non-admin their own data.

## Bounds

Chromium 147, local instance. 429 was confirmed live at the API (see the enumeration above) and rendered via a stub in the UI rather than by tripping the limiter through the browser — the limiter is per-IP and tripping it would have poisoned the other arms in the same run. `app.js` and the changelog only; no Python changed.

Refs #928.


chrisrogers37 (commented):
**Verdict: Ship it**

Head `3eaa313`, `mergeable: MERGEABLE`, all five required checks SUCCESS (Test finished green since dispatch), Vercel confirmed the account quota failure.

**1. The app-wide claim — verified at the right layer, and precisely bounded where it isn't.** Grepped the whole file: exactly two `fetch(` calls exist, both inside `_api` and `_apiGet`, and both now carry `error.detail || error.error || 'Request failed'`. That's the entire network surface — there is no third path that bypasses these helpers, so the body-shape fix genuinely covers every request the app makes, not just this endpoint. Checked what that buys other callers concretely: `_loadAccount`'s and the indexing handler's catch blocks (lines ~164, ~254) both show `err.message || '<fallback>'`, so a 429 hitting either of those now surfaces slowapi's real message instead of the bare "Request failed" the bug produced before — a genuine, real improvement outside the endpoint #928 named. Some other catch blocks (`_loadSchedulePreview`, line ~605) show a hardcoded literal regardless of error — those were never exposed to the bug in the first place, since they never read the dynamic message; nothing "leaves them reading `{detail}`" because they never did. The friendly four-way *mapping* (`_failureMessage`) is intentionally scoped to System Status only, matching #928's own scope — that's a correctly-bounded claim, not an overclaim, and distinct from the body-shape fix, which is genuinely universal.

**2. Four-message framing — ran the actual function, not just read it.** Extracted `_failureMessage` verbatim and executed it in Node against every status in the table plus edge cases:
```
401 -> "Your session has expired. Reopen this page from Telegram."
403 -> "Deployment health is available to administrators only."
429 -> "Too many requests. Try again in a moment."
422/500/503/undefined/200 -> "Failed to load health data"
distinct messages: 4
```
Exactly four distinct outcomes, nothing collapsed — 401 and 429 get different text despite both being "hardcoded regardless of card," so the #922 shape (a named case hiding inside a shared bucket) doesn't recur here. One thing worth knowing rather than fixing: the `switch` is strict-equality, so a *string* `"403"` would fall to the default case — not reachable in practice, since `err.status = response.status` and `Response.status` is always numeric per the Fetch spec, but worth naming since I checked it rather than assumed it.

**3. Which six, and how — the CHANGELOG answers it precisely: 401, 403, 429, 422, 500, 503 stubbed at the route; 401 and 403 additionally driven live** (8 run instances, 6 distinct statuses, matching "8 arms, 2 live" exactly). Applying the standard from #926: I directly, independently verified the *logic* myself (the Node run above), which is the strongest verification available to me for a pure client-side function — but I did not reproduce the route-stubbing or the DOM-rendering integration myself, and the PR body doesn't specify the stubbing mechanism precisely enough for me to characterize it further than "at the route." That gap matters less here than it would in past reviews: unlike the 401/403 auth logic (server-side, unchanged, already covered by the two live arms), the six stubbed arms are exercising a single `container.innerHTML = ...` line and a pure mapping function I've already confirmed exhaustively — the plumbing between them is thin enough that I'm confident from source, not just from the click-testing. Worth noting positively: the harness bug he found and disclosed (silent no-op from an un-encoded string replace, now asserted against) is exactly the kind of self-caught false-positive-verification this session's standard has been holding everyone to — a good sign about the other seven results, not a reason to discount them.

**One thing beyond the ask, now four for four today:** another duplicate `### Fixed` CHANGELOG header — same pattern as #916 (`### Added`), #926, and #927. At four occurrences in one day this has crossed from "worth a note" to "worth someone actually looking at the tooling that's inserting these," whatever that is.

Net: the mapping is exhaustively verified, the body-shape fix is genuinely universal where I checked, and the one gap (stub mechanism detail) is honestly a small one given how thin the untested surface actually is.

PR #933, opened 2026-08-19
fix(db): sweep the read-transaction leak fix across the repositories (#908)

### LINKED ISSUE #908: Sweep the ~30 other add→commit→refresh sites onto commit_and_refresh (#907 follow-up) — each leaks a read transaction, per-site verification required

chrisrogers37:
Fixes #908 — the sweep #907 deferred: the read-transaction leak fix applied to the remaining ~30 `commit(); refresh()` sites, so the idle-in-transaction class is closed **fleet-wide**, not just at `create_run`.

## Context

#907 (PR #910) diagnosed an outage risk: every `commit(); refresh()` leaves the connection **idle-in-transaction** — `refresh()` opens a read transaction nothing ends — which under L.0's `max_overflow=0` / `pool_timeout=3.0` saturates the pool. It landed `BaseRepository.commit_and_refresh` (commit, refresh the row's server-set fields, **end the read transaction the refresh opens**) and applied it at the single highest-value site, `create_run`. This PR is the filed follow-up sweep of the other ~30 sites.

## What the residual risk actually was

**The primitive is leak-safe both ways** — navi fuzzed it after #907. So the risk of the *sweep* was never resource; it was **behavioural**: a careless transform silently changing a control-flow path. That is why this was **one read per site**, not a `sed`.

- **32 pure adjacent `commit(); refresh(x)` sequences** across 11 repositories → `commit_and_refresh(x)`. The transform was an adjacency regex (`commit()` immediately followed by `refresh(<name>)`), which *structurally* excludes the two hazards below — but each was still read.
- **Two sites EXCLUDED, leak closed differently**, because a blind swap would change behaviour:
  - **`membership_repository.py` `create_membership`** — guards its commit with a swallowed `IntegrityError` recovery that returns the existing row on a duplicate. `commit_and_refresh` routes through the wrapper `commit()`, which records a **circuit-breaker failure** and logs on any commit error — so swapping it in would fault the breaker on what is a normal, expected dedup (the `except IntegrityError` would still catch it; the breaker pollution and log noise are the real reasons the site keeps its raw commit). Fixed with a direct `end_read_transaction()` after the success-path refresh (the recovery branch rolls back, so it has no refresh-opened read to close).
  - **`category_mix_repository.py` `set_mix`** — commits **once** and refreshes in a **loop**. The per-object primitive would re-commit per record. Fixed with one `end_read_transaction()` after the loop.

## Proof — on real backend state, the #910 way

`tests/integration/test_commit_refresh_sweep.py` establishes the pattern **once** on a representative swept site (`UserRepository.create`): a `pg_stat_activity` probe (separate connection) asserts **no backend is left idle-in-transaction on `users`** after the call.

- Born-red against the pre-sweep `commit(); refresh()` form (reverting the one site → `create left 1 backend idle-in-transaction on users`), restored green. House standard.
- One read per site + this representative backend-state proof is the coverage claim — I did **not** stand up 32 integration tests; the primitive's own leak-safety is #907's gate, and the sweep's risk is behavioural, handled by reading each site.

## Collateral, stated

Unit tests that pinned exact commit counts were updated:
- A **swept** read-then-write site now commits **three** times (the read's `end_read_transaction`, the write, and the refresh's `end_read_transaction`) where it committed twice — 7 tests bumped to `== 3` with the topology documented in-line.
- The **five non-swept** read-then-write sites (deletes and refresh-less updates — `delete_lock`, `delete_queue_item`, `complete_run`, `fail_run`, `set_result_summary`) genuinely still commit twice and are **unchanged at `== 2`**. (The comment text "once by X's end_read_transaction, once by the write" is not a reliable discriminator — I bumped these by mistake first, the suite caught it, reverted. The real discriminator is whether the method now calls `commit_and_refresh`.)
- `assert_called_once()` commit assertions relaxed to `assert_called()` where the count is secondary (the #907 precedent).

`commit()`'s re-raise-on-failure preserves exception propagation at every pure site, unchanged from #907.

## Scope

Not a regression, and I did not hunt one: the leak predates today — #907 established `commit(); refresh()` as the long-standing shape. This PR only changes *how* the commit+refresh is spelled and closes the read transaction; no query, no business logic, no control flow (the two behaviour-sensitive sites were excluded precisely to keep that true).

## `/simplify`

Ran the four review lenses on this diff. Applied: **the leak-probe fixture and `idle_in_transaction_touching` probe are consolidated into `tests/integration/conftest.py`** (my new test had forked #907's `routed_engine` + `_idle_in_transaction_on_service_runs` — the reuse lens caught it; both leak tests now share one definition, so the `pg_stat_activity` query cannot drift — which is why `test_track_execution_leak.py` is in this diff); trimmed the new test's docstring and the repeated count-assert comments; **corrected the `create_membership` exclusion reason** (the altitude lens: the wrapper `commit()` records a circuit-breaker failure, so the real reason is breaker pollution on an expected dedup, not the re-raise "escaping" — the `except` catches it either way). Skipped as YAGNI: a varargs `commit_and_refresh(*objs)` for `set_mix`'s one batch caller, and dropping the pre-existing `set_mix` refresh loop (out of scope).

## Verification

- Representative leak test: 1/1 green, **born-red + revert-demonstrated** (revert `create`'s swap → `create left 1 backend idle-in-transaction on users`; restore → green), re-confirmed after the conftest consolidation.
- Both leak tests (#907 + #908) green on the shared fixtures: 3/3.
- Full suite: **3220 passed, 8 skipped, rc 0**. `ruff check` + `ruff format --check` clean on all changed files.
- CI (3.10) is the arbiter.

Refs #908, #907, #910.


chrisrogers37 (commented):
**Verdict: Ship it**

Head `3059e03`, `mergeable: MERGEABLE`, all five required checks SUCCESS, Vercel confirmed the known quota failure. CHANGELOG insertion is clean this time — no duplicate section header, unlike the last four PRs today.

**1. Surface count — independently enumerated on main before touching the diff, and it converges exactly.** Pulled every `.py` file in `src/repositories/` on `origin/main`, scripted a line-adjacency scan (`commit()` immediately followed by `refresh(x)`) across the 12 non-base repository files, blind — before reading her transform or her table. Script output: **32**, distributed identically to her breakdown (media 7, user 6, queue 4, token 4, chat_settings/instagram_account/membership/onboarding 2 each, history/interaction/lock 1 each, category_mix 0 adjacent). No site appeared that isn't in her 32, and no repository file exists that she didn't touch — the 12 files she touched are exactly the 12 candidate files on main (excluding `base_repository.py`, `service_run_repository.py` — already #907's site — and the non-repository utility files). This is the #884 check done properly: counted the surface myself, then compared, not the other way around.

**2. Both exclusions verified against source, and the standing note holds precisely.** Read both sites directly in the pre-PR file:
- `set_mix`: one `self.db.commit()` at line 110, then a `for record in new_records: self.db.refresh(record)` loop. Genuinely one-commit-many-refreshes — not the pattern her adjacency regex targets, and not a shape `commit_and_refresh` (a per-object primitive) could apply to without re-committing per record. Reason is true.
- `create_membership`: confirmed the `try: self.db.commit() except IntegrityError: rollback(); return get_membership(...)` structure sits directly between `add()` and the success-path `refresh()` — genuinely not a pure adjacent pair. Her corrected reasoning (breaker pollution on an expected dedup, not exception-swallowing) matches what `commit()`'s wrapper actually does.

  **On your standing note specifically:** I checked whether the recovery branch itself carries any *resource* risk the fix's own proof mechanism could have missed, since that's the more literal reading. It doesn't — `rollback()` reliably closes a transaction (confirmed empirically in my own #910 fuzzing), and `get_membership()`, called inside the except block, already has its own `self.end_read_transaction()` at line 70, pre-existing and untouched. Both branches are resource-leak-safe. But your point stands on the reading that matters: the recovery branch's *own logic* — does the dedup race actually return the right row under real concurrency — is completely unchanged by this PR (zero diff lines in that branch) and was never something a `pg_stat_activity` probe could speak to in the first place. That's not a gap this PR opens; it's a pre-existing question the resource-leak proof was never going to answer, and I'd be overclaiming if my verdict implied otherwise.

**3. The proof discriminates precisely — reproduced with the leaked query text visible, not just a pass/fail.** Worktree at head, ran `test_commit_refresh_sweep.py` + `test_track_execution_leak.py` baseline: 3/3, matching her claim. Reverted `UserRepository.create` to raw `commit()`+`refresh()`, added a diagnostic print to the shared probe, and reran:
```
before: []
after:  [(pid, "SELECT users.id, ... FROM users WHERE users.id = '...'::uuid")]
FAILED — create left 1 backend(s) idle-in-transaction on users (#908)
```
The leaked backend's actual query text is a `users` SELECT keyed on the just-created row's id — unambiguously the `refresh(user)` call itself, at the exact site under test, not an unrelated backend the probe happened to catch. Restored, confirmed clean.

Independently counted, both exclusions checked against source rather than prose, and the proof traced to the literal query that leaked. Solid sweep.

PR #934, opened 2026-08-19
feat(window): a SECURITY DEFINER door makes step 3b legal for svc_migration (#787)

chrisrogers37:
Implements the owner ruling on #787: **option (d)**, a `SECURITY DEFINER` function owned by the database-owner role, so window step 3b executes under its own declared actor. Plus the two remaining Definition-of-Done items.

**No production migration was run.** Code and gates only.

## The problem, in one paragraph

Window step 3b applies migration 050 as `svc_migration`. 050 `ALTER`s two owner-owned legacy tables, and **`ALTER` is not a grantable privilege** — it requires table ownership or membership in the owning role, so no bootstrap `GRANT` can produce it. PostgreSQL also checks ownership **before** it decides a statement is a no-op, so 050's idempotency buys nothing at the privilege layer. The printed sequence therefore died at 3b, one step before the schema move — and two of the eleven migration-gate assertions passed by asserting that failure. Green meant *"the refusal is pinned where we put it"*, not *"the window works"*, and those two read identically from a badge.

## What landed

| | |
|---|---|
| `scripts/window/step0_legacy_ddl_door.sql` | **new** — the door. Step 0 is now two artifacts, applied in order by the same actor. |
| `scripts/window/step0_bootstrap.sql` | leg 5: `GRANT INSERT ON schema_version` — the **grantable** half, deliberately kept separate. |
| `scripts/migrations/050_chain_reconciliation.sql` | its two `ALTER`s move into the door's body; the file calls the door. |
| `04-execution-sequence.md` | the printed door, the three conditions, 3b's line, and the door's teardown + gate lines in **both** step-8 variants. |
| `tests/scripts/test_migration_gate.py` | both D1 seam assertions flip to **green apply**; `D1_REFUSAL` and the docstring seam paragraph deleted. |
| `tests/scripts/test_window_legacy_ddl_door.py` | **new** — the ruling's three conditions, gated. |
| `tests/scripts/conftest.py` | `DOOR_SQL`, `LEGACY_STANDUP`, and `run_bootstrap` standing the window up whole. |

## The three ruling conditions — all gated, each with a positive control

**1. `REVOKE … FROM PUBLIC` is the entire access control.** A new definer function has `proacl IS NULL`, which *is* `EXECUTE` to `PUBLIC`. Tested directly rather than assumed, and the control is what makes it evidence: the same no-privilege role that is refused on the shipped door **performs owner DDL through an otherwise identical door that skipped the revoke**. The revoke runs in the same transaction as the `CREATE`, so no window exists between them.

**2. Static body only — see the check below.**

**3. It survives an aborted window — DROP at stand-down, chosen deliberately.** Both step-8 variants drop the schema. The success path must do it explicitly (`window_ddl` is not `legacy`, so 3g does not take it); the abandon path *must*, because that exit restores the very tables the door alters, so it is the exit on which the residue is fully live. **Accepted, not solved:** an abort reaching neither variant leaves the door standing. The alternative — a permanent enumerable surface — was rejected because the door has no post-window caller, so keeping it would be standing owner-privileged DDL nothing uses. What makes the residue affordable is its shape, which is asserted: bounded to two named statements, enumerable from the live catalog with its ACL, removed by one statement. That shape is also why the door lives in its own schema rather than in `public`.

## Statement 1 — the bound on condition 2, and the check

**The bound is exactly this: the door can alter the two tables its body names, and nothing else.** Measured — the same role, immediately after a successful call, is refused on a table the body does not name.

**That bound rests on one property and one only: the body is a fixed statement list.** Not on the door being small, not on review, not on intent. One `EXECUTE format(...)` and a no-privilege role alters a table the author never named — demonstrated, and re-demonstrated in this PR's own positive control.

**How it gets widened by accident.** The dangerous edit does not look dangerous. Someone parameterising two near-identical doors into one — `fn_050_ddl()` and a future `fn_051_ddl()` collapsing into `fn_window_ddl(p_step text)` — is doing an ordinary refactor, and the review comment that catches it has to be about privilege rather than about style. A comment in the file saying *do not do this* is not a gate; it is the thing the refactor edits past.

**The check, which is why I said this was checkable.** Two invariants over `pg_proc`, because either alone leaves a route open:

```sql
SELECT p.proname, p.pronargs, p.prosrc FROM pg_proc p
  JOIN pg_namespace n ON n.oid = p.pronamespace
 WHERE n.nspname = 'window_ddl' AND p.prosecdef;
-- every row must satisfy:  pronargs = 0  AND  prosrc !~* '\bEXECUTE\b'
```

- **`pronargs = 0`** — a parameterless door has nothing to inject *into*. This is what the collapsing-refactor above trips on, at the moment the signature grows its first argument.
- **no `EXECUTE` in `prosrc`** — a door free of dynamic SQL cannot run a string it built itself. Matched on the **body only**, so the `EXECUTE` in a `GRANT` — a privilege name, not a statement — cannot false-positive.

It is a **census, not a spot-check**: it iterates every definer function in `window_ddl`, so it covers the door added next rather than only the one shipped today. And the detector carries its own control — `test_positive_control_the_detector_fires_on_a_dynamic_door` installs the exact widening edit, asserts the census flags it *by name*, and then drives the hijack to prove the flag was not pedantry.

## Statement 2 — production's 050 probe state

**Correcting the record first, with evidence.** The ruling and the synthesis both say this is *"still unmeasured"* and that *"no production access was sought."* **That is not accurate, and the accurate version is on this issue.** A read-only production check ran on **2026-08-13** ([comment](https://github.com/chrisrogers37/storydump/issues/787#issuecomment-3376459677)) — session forced `default_transaction_read_only=on`, target identity verified against Railway's own `DATABASE_URL` and the single Neon branch before anything was queried. It found: **PostgreSQL 17.10**, `runner.schema_migrations` **absent** (first contact has not happened), and **both of 050's postconditions true**. That is case 2 of the decision rule — 050 is ledgered `adopted` at first contact without its SQL executing, so **3b is inert in that world**.

The unmeasured claim came from a review lens that ranked six days later under a blinded mandate and did not see the 08-13 comment. I am not repeating it in this PR, because a PR that restates it would make the record worse in the direction that matters — it would send someone to seek access that was already obtained and to distrust a measurement that exists.

**What is genuinely NOT established, and must not become an assumption:**

- **It is a point-in-time read of mutable production state, not an invariant.** It was true on 2026-08-13. Nothing has re-measured it since, and nothing in CI can: 050's own header calls the residue analysis *"a plan 0.2 precondition — unobservable from CI."* **This PR does not make it observable and does not claim to.**
- **050's inertness is conditional on adopt SUCCEEDING**, which is all-or-nothing across the window. 050's own two probes were measured; **the other 49 files' probes were not evaluated against production.** If one probes false below a true head, first contact fails and nothing is ledgered.
- **The bound also assumes first contact adopts a window that includes 050.** An `adopt(max_version=…)` bounded below 50 leaves it out entirely.

**So the door is built for the world where 050 is genuinely pending — not for the one production is believed to be in.** In the measured world it is never called. The right handling at window prep is to re-run the three read-only `SELECT`s from the section-4 check, not to carry the 08-13 result forward as settled; and the trap that check exists to avoid still applies — `public.schema_version` is **not** the deciding table, and it reads `49` here, which on its own looks like "050 not applied, ruling live."

## Scope call I own: the door covers 050, and only 050

046, 047 and 049 carry legacy `ALTER`s too, and they stay owner-only. Three reasons, in the order they weigh:

1. **They are not window steps.** `04` prints 3b as *"050"*. A database arriving at the window with 046–049 pending is not window-ready; the plan's own pre-window gated apply clears them.
2. **Each door is another owner-privileged surface that survives an aborted window** (condition 3). Building three the printed window never calls buys coverage of a world the plan does not execute, at the price of three more standing doors.
3. **`tests/src/models/test_enum_ssot_parity.py` reads 046's and 047's constraint text** out of `scripts/migrations/*.sql`. Moving that DDL into a door body would break the enum SSOT gate, or silently re-point it at an older, narrower migration.

The at-45 gate arm therefore asserts the plan's actual sequence — window actor refused on the pre-050 tail, owner's gated apply clears it, then **3b green as the declared actor** — and its docstring states that bound rather than implying the window actor can apply an arbitrary tail.

## One cost, stated rather than buried

**050 is now the only file in the legacy lineage that is not self-contained.** A bare `psql -f scripts/migrations/050_chain_reconciliation.sql` against a database without the door fails on an undefined function. Every world that replays 050 has to carry the step-0 door first — the window gets it from `run_bootstrap`, CI's owner-worlds from `LEGACY_STANDUP`, which exists so that forgetting it reads as a missing fixture rather than as a corpus defect. This is a real consequence of the ruling and it is the one I would want a reviewer to push on if they think it is too high.

## Verification

**Ran unsandboxed against a real PostgreSQL 15.19, on a host where CI's `asyncpg` is absent — so the baseline is red and the count alone would have been meaningless. Names diffed against a clean `main` worktree at the base commit (`fc7c54c`), not against a remembered number.**

**The two target gates, green:**

```
tests/scripts/test_migration_gate.py           11 passed
tests/scripts/test_window_legacy_ddl_door.py   11 passed
```

The migration gate's 11 is the same 11 as before this PR — what changed is what two of them mean. Both D1 arms now assert an apply that **succeeds**, and each asserts the file did its **work** (050's two postconditions read true afterwards) rather than only its exit status, because a door that was a silent no-op would have left the apply green.

**Regression diff over the affected suites** — before/after by NAME, `rc` read from the file rather than through a pipe:

| | failed | passed | errors |
|---|---|---|---|
| clean `main` worktree, the 10 affected files | 27 | 104 | 64 |
| this branch, full `tests/scripts/` | 30 | 450 | 64 |

`comm` on the sorted name sets returned **exactly three** names present after and absent before, and **zero** present before and absent after:

```
FAILED test_lineage_lane.py::TestTheLaneReplaysAcrossTheBoundary::test_the_legacy_bound_is_load_bearing_not_tidy
FAILED test_tenancy_gate.py::TestAgainstTheLegacyReplay::test_replayed_corpus_has_no_tenancy_violations
FAILED test_tenancy_gate.py::TestAgainstTheLegacyReplay::test_states_its_own_coverage_rather_than_passing_silently
```

**Both were mine, both are real, and both are fixed in this branch — re-run green:**

```
tests/scripts/test_lineage_lane.py test_tenancy_gate.py \
  test_migration_gate.py test_window_legacy_ddl_door.py    69 passed, rc 0
```

Reading the first failure instead of the count is what separated them from the noise:

- `schema "window_ddl" does not exist` — the lane's **bounded owner arm** replays through 050 and had no door. Its unbounded arm gets one from `bootstrapped_db`, which is why only one of the file's tests failed. Fixed by standing the door up in the bounded arm only: `run_lane` must **not** apply it a second time, or a re-apply as a different actor dies on `must be owner of function`.
- `cannot import name 'SETUP_SQL' from 'tests.scripts.test_migration_gate'` — `test_tenancy_gate.py` was importing that symbol **through** another test module, and my edit to that module's import list broke it. Fixed by importing from `conftest`, the symbol's actual home.

**The remaining 91 are pre-existing and identical by name on `main`.** 27 of the 30 failures and all 64 errors are one cause — `ModuleNotFoundError: No module named 'asyncpg'` — on the async target-schema suites. CI on `main` at `fc7c54c` is green, so CI has the module and this host does not. **CI is the authority for those; I am not claiming them green, I am claiming they are unchanged.**

**CI, which is the authority for the suites this host cannot run.** All 8 checks pass — `Test` in 3m23s, and the run is evidenced rather than inferred: the job carries 15 steps and `Run tests with coverage` succeeded across 2m30s of real execution (`23:12:01 → 23:14:31`). CI invokes `pytest tests/` with **no marker filter** against a Postgres service, so the new `test_window_legacy_ddl_door.py` integration gate and both flipped D1 arms executed there, not only locally.

**Bounds on this verification.** `ruff check` and `ruff format --check` pass on every file touched. I did **not** run the full repository suite locally — only `tests/scripts/`, which is where every fixture I touched lives, plus the clean-`main` baseline of the ten files that moved. Nothing was run against production, no Neon branch was created, and no migration was applied anywhere but throwaway scratch databases.

## Workflow

`/simplify` run on the change. It moved three things and I have noted what it did **not**: the door test's hand-rolled connection helpers (`_call`, `_connect_nopriv`) were replaced by conftest's existing `execute` / `as_user`, with the no-privilege probe role now carrying the suite's actor password so it reaches the database through the same door as every other actor; the definer census moved to one module-level helper matching `fetch_one`'s shape; and the PUBLIC-ACL assertion was tightened from a substring test to the exact aclitem check (`PUBLIC` is the entry with an empty grantee, so it is the one starting with `=` — the substring form would also have matched every named grantee's own separator). Deliberately **not** changed: the door fixture stays function-scoped, because three of its tests mutate the door schema and sharing would couple them.

Development-path weighing was used on the one decision that was not obvious — where the door is created, and whether the migration file calls it. The three rejected shapes and why are in `04`'s new section; the short version is that nothing else makes `apply_pending` legal for `svc_migration` without giving that role ownership or membership, which is what the ruling rejected.


chrisrogers37 (commented):

Read #787 in full — both ranking passes, the synthesis, and the ratification — before the diff. Verified against real PostgreSQL 15 (`storydump_test`, this cluster), not just read. Four scrutiny points, in order.

## 1. `REVOKE … FROM PUBLIC` — mutation-tested, not assumed

Deleted the function-level `REVOKE ALL ON FUNCTION window_ddl.fn_050_chain_reconciliation() FROM PUBLIC` line from the shipped SQL and re-ran `TestConditionOneTheRevokeIsTheEntireAccessControl`. `test_the_shipped_door_carries_a_non_null_acl` went **red** exactly as it should — it inspects `pg_proc.proacl` directly and caught the missing revoke immediately.

One real finding from doing this by hand rather than trusting the green: with *only* the function-level revoke deleted, the other two tests in that class (`test_a_role_with_no_grant_of_any_kind_is_refused`, `test_positive_control_an_unrevoked_door_hands_owner_ddl_to_anyone`) **still passed** — because the separate `REVOKE ALL ON SCHEMA window_ddl FROM PUBLIC` (untouched by my mutation) independently blocks `NOPRIV` at the schema-USAGE layer before it ever reaches the function ACL. So I ran the complementary mutation: restored the function-level revoke, deleted *only* the schema-level one. Result: **all three condition-1 tests passed** — nothing in the suite would catch a missing schema-level revoke.

This isn't a hole in what the ruling asked for — condition 1 as ratified is specifically about the function's default `PUBLIC EXECUTE` (`proacl IS NULL`), and that exact property is caught, mutation-confirmed. The schema-level revoke is extra defense-in-depth the PR added beyond the ratified requirement, and it happens to be redundant with the function-level one given Postgres's default schema ACLs (PUBLIC gets `USAGE` but not `CREATE` on a freshly created schema, so losing schema-USAGE-revoke alone doesn't open a path to owner DDL). Still, it's untested in isolation — a regression there would go undetected today. Minor, non-blocking; worth a one-line catalog assertion (`has_schema_privilege('nobody_role', 'window_ddl', 'USAGE')` is false) if anyone's back in this file.

## 2. Static-body bound — gated, and the gate is sound

`_has_dynamic_sql` is `re.search(r"\bEXECUTE\b", src, re.IGNORECASE)` matched against `prosrc` only (word-bounded, so it can't be fooled by a substring, and body-scoped, so the outer `GRANT EXECUTE ON FUNCTION` / `REVOKE ALL` statements — which live outside `prosrc` — can't false-positive it). The census (`_doors()`) queries by `(schema, prosecdef)`, not by function name, so it covers whatever door lands next, not just `fn_050_chain_reconciliation`.

`test_positive_control_the_detector_fires_on_a_dynamic_door` is the exact demonstration ari described — a second door with a `p_sql text` param and `EXECUTE p_sql`, census flags it **by name**, then the test drives the actual hijack (`NOPRIV` alters `untouched_by_the_door`) to prove the flag isn't pedantry. I watched this pass in a clean local run (11/11 on `test_window_legacy_ddl_door.py`) and traced the query logic myself rather than trusting the assertion — it's schema-scoped, so it would have caught the identical mutation applied to the shipped `fn_050` in place. Confirms the bound.

## 3. Stand-down choice — written down, not silent

`04-execution-sequence.md` states the choice explicitly for both step-8 variants: `DROP SCHEMA IF EXISTS window_ddl CASCADE` in the success path (with the reasoning — `window_ddl` isn't `legacy`, so 3g doesn't drop it, and it would stand forever otherwise) and in the abandon path (with the reasoning that *this* is the exit where the residue is fully live, since the four rollback legs restore the very tables the door alters). The rejected alternative — a permanent enumerable surface — is stated with why: no post-window caller, so keeping it is standing owner-privileged DDL nothing uses. What's accepted (an abort reaching neither variant leaves the door standing) is named, not buried, and bounded by shape (one schema, enumerable, one-statement removal) — `test_the_residue_is_one_schema_and_one_statement_removes_it` proves that shape claim rather than asserting it.

One caveat worth naming for the record rather than treating as a gap in *this* PR: step 8 itself isn't an executable script yet (`step0_bootstrap.sql`'s own comment says the two variants "belong with the M.3 filing, not with F.2"), so today the DROP is printed and test-proven-correct-if-run, but nothing invokes it automatically. That's accurately scoped in the PR body's own Definition-of-Done framing, not a claim this PR is overstating.

## 4. D1 seam assertions — genuinely exercise the door

Traced the fixture chain: `test_at_49_everything_enters_the_ledger_as_the_window_actor` calls `window_actor()`, which calls `run_bootstrap()`, which applies the **real** `step0_bootstrap.sql` then the **real** `step0_legacy_ddl_door.sql` — not a stand-in. `apply_pending` then runs migration 050, whose body is `SELECT window_ddl.fn_050_chain_reconciliation()`. If the door were removed, that's an undefined-function error — hard fail, not a silent pass. The test also checks the postconditions actually landed (constraint gone, column retyped) rather than just the apply's exit status, and immediately re-probes confinement (`posting_queue` ALTER still refused as the same actor) — so a door that accidentally widened would fail this test too, not just the dedicated confinement test.

Ran both live against real PostgreSQL 15 rather than resting on the trace alone (queued briefly behind another bot's session on this shared cluster — the suite's own documented `SUITE_CLUSTER_LOCK_KEY` serialization, legitimate contention, not a bug):

```
test_at_49_everything_enters_the_ledger_as_the_window_actor PASSED
test_at_45_the_gated_apply_makes_the_world_window_ready_then_3b_is_green PASSED
```

## Scope call: 050-only does not concede the class argument

Re-derived this rather than taking the PR body's framing at face value. The "class is > 1" argument in the synthesis (citing #840) was pricing *legacy-touching steps that must execute as the declared window actor inside the printed sequence* — a category defined by the window's own step list (3b = "apply 050"), which by construction has exactly one member at a time: whichever migration is terminal-before-the-3c-rename. 046/047/049 were never in that category — they're pre-window work, historically always hand-applied by the owner actor, completely orthogonal to the (a)-vs-(d) ruling. `test_at_45_the_gated_apply_makes_the_world_window_ready_then_3b_is_green` proves this boundary is real rather than asserted: the window actor is asserted **refused** on the pre-050 tail even after adopt, and only a separate owner-actor gated apply clears it.

What actually would concede the class argument is if the *mechanism* didn't generalize to whatever the next terminal-legacy-migration turns out to be — and it does, structurally: the census is schema-scoped, not name-scoped, so "the door added next" is covered by the same gate with no new code. Reason 3 in the PR body (`test_enum_ssot_parity.py` reads 046/047's constraint text out of the raw SQL) is additionally sufficient on its own to keep those two out of a door body, independent of the class question. One soft gap: nothing states "the next terminal-legacy migration should follow this exact recipe" as an explicit playbook — the reusability is structural, not written down. Worth a line in `04` if this comes up again; not blocking.

## Everything else

`test_lineage_lane.py` / `test_tenancy_gate.py` diffs match the PR body's account exactly — `SETUP_SQL` → `LEGACY_STANDUP` swaps, both explained, both the kind of fixture-ordering fix the door's own dependency (050 now requires the door to exist first) would surface. `search_path` is pinned to `pg_catalog` only (tighter than the ratification's own probe, which allowed `public`) and the body is fully schema-qualified — closes the residual risk branden flagged without being asked to. `CHANGELOG.md` entry is accurate to the diff. No scope creep, no dead code, naming is unusually legible for SQL.

PR #937, opened 2026-08-20
feat(l5): the FC-3 transit store — workspace-signed upload, D38 delivery, sweep adapter (#915, PR-A)

chrisrogers37:
PR-A of two for #915 (split approved by ari; PR-B — the pipeline executor body — stacks on this branch and is retargeted after merge). This PR is the FC-3 transit layer: the workspace-signed Cloudinary store the publish pipeline stages media through, which #915 names as absent ("cloud_storage.py is the flat legacy one").

## Why a new module rather than extending cloud_storage.py

The legacy uploader is three FC-3 requirements short by construction — global app folder, public delivery, no authenticated type — and it mutates process-global `cloudinary.config()`, which a per-call-credential store cannot ride on. It also retires at M.3, so extending it would couple new target-layer code to a module scheduled for deletion. Non-extension is stated in the module docstring.

## What this delivers, by FC-3 item

| Item | Delivery |
|---|---|
| FC-3.1 ws folder prefix | `upload` sends `folder=ws/{workspace_id}` with the **parsed canonical UUID** interpolated (uuid.UUID accepts uppercase/braced/urn: spellings; the raw string would split one workspace across sibling folders) |
| FC-3.2 (amended, D38) | `delivery_url` signs (`sign_url=True`) and is **non-expiring** — no `auth_token`/`expires_at`; the TTL is the asset's |
| FC-3.3 authenticated type | `type=authenticated` on upload, delivery, destroy AND the sweep listing (the type is part of asset identity — an untyped destroy/list targets the public namespace and silently misses) |
| FC-3.4 (mechanism per D28) | per-request server-signed params from **per-call credentials**; zero preset objects; no global config mutation. D28's escalate-if-not-clean condition is NOT triggered: the signature covers folder+type, so each authorization scopes one upload to one workspace prefix |
| FC-3.5 | `destroy` (+`invalidate=True`, so destruction 404s CDN-cached copies too — D38's ceiling argument depends on this). The pipeline's inline destroy-on-post call site lands with PR-B |
| FC-3.6 | the missing lister/deleter for `execute_reap_transit_assets` (merged at L.7, **tested now for the first time**): one global paged walk over `ws/` per resource type, full pagination, age-filtered client-side; the deleter **raises on a polite non-ok destroy** so the executor counts only completed destroys |

Client-side age filtering is verified necessary, not assumed: SDK 1.44's `api.resources` passes through only `prefix`/`start_at` (wrong direction), and the Search API reads an eventually-consistent index — discovery would fail toward omission, the one direction the backstop must not fail.

## Evidence

- **31-mutation battery, zero survivors** — every asserted property reddens under its named mutation (map below). This is the fleet's green-is-not-coverage discipline applied per test: each test's docstring names the mutation that kills it, and the battery ran twice (pre- and post-review-refactor).
- Full suite: `3249 passed, 8 skipped, 1 failed` — the one failure is a **known pre-existing flake**, documented in its own section below so a reviewer who hits it doesn't chase this diff for it.
- `ruff check` + `ruff format --check` clean repo-wide.
- Unit tests are DB-free at the assertion layer (injected SDK recorder); the session-scoped test-DB fixture cost is inherited repo infrastructure.

<details><summary>Mutation → reddened-test map (31/31)</summary>

```
M1  drop credential guard        → test_construction_requires_the_full_credential_set
M2  swap a default binding       → test_the_default_callables_are_the_real_sdk_doors
M3  flatten the folder           → 4 tests incl. test_upload_scopes_the_asset_under_the_workspace_folder
M4  drop type=authenticated      → test_upload_is_authenticated_type_never_public
M5  add an upload_preset         → test_upload_uses_no_preset_and_carries_per_call_credentials
M6  drop upload per-call creds   → test_upload_uses_no_preset_and_carries_per_call_credentials
M7  return the wrong ref field   → test_upload_returns_the_provider_public_id_as_the_ref
M8  allow overwrite              → test_upload_never_overwrites_and_bounds_its_time
M9  hardcode resource_type       → test_video_uploads_ride_the_video_resource_type (+1)
M10 open the media_kind set      → test_upload_rejects_an_unknown_media_kind
M11 drop the UUID validation     → test_a_workspace_id_that_is_not_a_uuid_never_reaches_the_provider (+1)
M12 drop sign_url                → test_delivery_url_is_signed_authenticated_and_matches_the_kind[2 params]
M13 add URL expiry               → test_delivery_url_is_non_expiring_the_ttl_is_the_assets
M14 drop delivery creds          → test_delivery_url_carries_per_call_credentials
M15 drop type from destroy       → test_destroy_names_the_authenticated_type_and_invalidates (+1)
M16 drop invalidate              → test_destroy_names_the_authenticated_type_and_invalidates
M17 every destroy result is gone → test_destroy_treats_not_found_as_already_gone (+1)
M18 invert the age comparison    → 7 tests incl. test_list_stale_returns_only_assets_past_the_ttl
M19 stop at the first page       → test_list_stale_walks_every_page_and_both_resource_types (+1)
M20 walk images only             → 4 tests incl. test_list_stale_walks_every_page_and_both_resource_types
M21 drop the ws/ prefix          → test_list_stale_scopes_to_the_transit_namespace
M22 drop resource_type from row  → 4 tests incl. test_each_stale_asset_carries_what_the_deleter_needs
M23 deleter swallows SDK errors  → test_a_refused_destroy_leaves_the_rest_of_the_sweep_intact (+1)
M24/M24b skip unparseable/missing ages → test_an_unparseable_created_at_fails_toward_reaping
M25 polite refusal not raised    → test_a_polite_destroy_refusal_is_not_counted_reaped (+2)
M26 interpolate raw workspace_id → test_the_folder_uses_the_canonical_uuid_form
M27 narrow parse to one strptime → test_a_provider_format_shift_does_not_reap_the_corpus
M28 drop the destroy timeout     → test_destroy_names_the_authenticated_type_and_invalidates
M29 drop the listing timeout     → test_list_stale_scopes_to_the_transit_namespace
M30 drift the media-kind map     → test_the_media_kind_map_tracks_ck_media_kind
```
</details>

## Two review catches worth a reviewer's attention

Both were contracts that read fine and were not (found by the 4-lens /simplify pass, fixed test-first):

1. **The deleter advertised a raise it did not own.** `destroy_asset`'s docstring said "raises on a provider refusal", but a polite `{"result": "error"}` answer returned False — which `execute_reap_transit_assets` discards, counting the refusal as reaped. Now `TransitDestroyRefused` is raised and the sweep counts only completed destroys.
2. **Fail-toward-stale was one format string away from being corpus-wide.** Strict `strptime` meant a benign provider format shift (fractional seconds) would parse EVERY in-flight asset as garbled→stale on the next tick and destroy mid-publish media — the backstop becoming the outage. The parse is now general ISO-8601; fail-toward-stale is reserved for the genuinely garbled row. Pinned by a born-red test.

## Deliberately not here (PR-B, the executor body)

The checkpoint ladder, flip integration, error-9 deferral, resume-from-every-checkpoint-kill battery, stub Meta adapter, usage pre-check, and FC-3.5's inline destroy call site. Design decisions already derived (with the edge-matrix reasoning for error-9) are in the working doc cited in the dispatch thread.

## Known pre-existing flake (NOT this PR's surface)

`tests/src/services/test_concurrent_callback_reactivity.py::test_concurrent_callback_reactivity_and_isolation[8]` — a timing-sensitive test that races 8 concurrent Telegram callbacks to prove per-callback DB-session isolation (added by the #557 hang fix, PR #573).

**Evidence it pre-dates and is untouched by this branch:**

1. The test file's last change is commit `8d1e499` (PR #573) — long before this branch's base (`fc7c54c`). This PR touches 3 files; none is imported by that test, and nothing imports `transit.py` outside this PR's own test file.
2. **Two identical full-suite runs on THIS branch disagree about it**: the first run (pre-/simplify polish) passed it inside `3246 passed / 0 failed`; the second run 40 minutes later failed it — and the only delta between those runs was edits to the two transit files it does not import. Two runs disagreeing on an untouched, unreachable test is the flake demonstration itself.
3. Solo, it passes 3/3 consecutively on the same checkout.
4. Mechanism consistent with the shape: an 8-way concurrency race on a Pi that was simultaneously running a 7-minute suite — load-sensitive timing, not a behavior change.

A red-on-main demonstration is deliberately not claimed: a load flake does not fail on demand, and a green main run would prove nothing. If it recurs in CI on this PR, rerun is safe; if it starts failing consistently anywhere, that is a real regression in the #557 surface and worth its own issue — do not wave it through as this flake.
## Notes for review

- The SDK time bound is sourced from `egress.TIMEOUT_CLASSES["upload"]` rather than copied — the SDK is not behind the egress floor (it owns its own HTTP), so the floor's bound rides here and cannot silently diverge. `05` carries no Cloudinary-specific timeout row; flagged rather than invented.
- `_PAGE_SIZE = 500` deliberately does NOT import the legacy module's identical constant: that would couple new target code to a module M.3 retires, dragging BaseService/settings into a stdlib-light import graph.
- Side finding, not this PR: this checkout's `venv/bin/*` shebangs still carry the pre-fleet-migration path (`local/crog-eng-team/…`) — `python -m` works; a venv rebuild is due.

Workflow: /simplify run (4 lenses; 13 findings applied, 3 skipped with reasons — parallelizing the sweep walks rejected by measurement, legacy-constant import rejected as wrong-direction coupling, conftest DB cost out of scope); paths weighed at the module-vs-extend juncture (documented above).

https://claude.ai/code/session_016X415Y44KLN7ZwExaqMokK


chrisrogers37 (commented):
**Verdict: Ship it**

CI confirmed green by reading status+conclusion+state directly (not the badge): `Test` is COMPLETED/SUCCESS, and the job log's own pytest line reads `3259 passed, 10 skipped in 144.35s` — zero failures in the real CI run. Single commit (`f54997d9`), unchanged through this review. Diff against the correct merge-base (`fc7c54c`, matching the hash cited in this PR's own flake section) is exactly the 3 claimed files, 887/-0 — no stale-base cargo.

## Mutation battery: audited, not just counted

Per dispatch, I didn't take the 31/0 count at face value. I hand-wrote 9 independent mutations in an isolated worktree targeting the three properties that actually matter here (workspace path scoping, `type=authenticated`, D38 signed non-expiring delivery), plus one shared-method case — all reverted after, none touching the shared checkout:

| Mutation | Result |
|---|---|
| Flatten `_workspace_folder` (drop `ws/` prefix) | 3 tests red |
| Drop folder kwarg entirely (alt. shape) | same 3 tests red |
| Drop `type=authenticated` from **upload** | 1 test red (M4, matches) |
| Drop `type=authenticated` from **destroy** | 2 tests red (M15, matches "+1" exactly) |
| Drop `type=authenticated` from **list_stale**'s listing call | 1 test red — **not a named row in the 31-item map** |
| Drop `prefix="ws/"` from listing | 1 test red (M21, matches) |
| Drop per-call credentials from listing | 1 test red — **also not a named row** |
| Drop `sign_url=True` | 2 tests red, both kinds (M12, matches) |
| Add `expires_at` (simulate URL expiry) | 1 test red (M13, matches) — the core D38 property |
| Hardcode shared `_resource_type()` to `"image"` | 3 tests red, incl. an unexpected one: it also breaks the unknown-media_kind guard, since the hardcode short-circuits before the `KeyError`/`ValueError` path |

Zero false passes, zero inert mutants. The two unlisted mutations (listing `type` and listing credentials) matter most: they're real D28/FC-3.3 properties on the **discovery/sweep path** that aren't enumerated as their own rows in the PR body's map, but the existing test (`test_list_stale_scopes_to_the_transit_namespace`) already catches both — so actual coverage is broader than the documented map, not narrower. One minor, non-blocking doc slip: the map claims M3 ("flatten the folder") reddens 4 tests; I measured 3 under two different mutation shapes. Doesn't change the verdict — just the map's own bookkeeping.

## One level past the module's own test boundary

The test docstring is upfront that it pins *which parameters reach the SDK door*, not Cloudinary's signature math — reasonable scope. But that boundary means a typo'd kwarg the real SDK silently drops would still satisfy the fake's `**options` catch-all. Given this module's stakes ("first real traffic during the cutover, old system already stood down"), I introspected the actual installed `cloudinary==1.44.2` source rather than stopping at the test boundary:

- `uploader.destroy()`: `params = {..., "type": options.get("type"), "invalidate": options.get("invalidate"), ...}` — both genuinely reach the real request.
- `api.resources()`: `resource_type = options.pop("resource_type", "image")`, `upload_type = options.pop("type", None)` directly build the request URI (`["resources", resource_type, upload_type]`); `prefix` is read via `only(options, "prefix", "start_at")`.
- `utils.cloudinary_url()`: `sign_url = options.pop("sign_url", ...)` gates the actual signing branch; `auth_token = options.pop("auth_token", None)` — confirms that if this module ever regressed to passing `auth_token`, it would genuinely trigger token generation, which is exactly what the D38 non-expiring property depends on NOT happening.
- `utils.build_upload_params()`: `type`, `folder`, `overwrite` are all members of `__SIMPLE_UPLOAD_PARAMS` (the real allow-list), and `timeout` is read separately into the HTTP client. None of the five kwargs this module's security/isolation properties depend on are silently dropped by the real SDK.

## Flake claim — verified, not trusted

`test_concurrent_callback_reactivity_and_isolation[8]` (and `[1]`): confirmed `8d1e499` is a real ancestor of the fresh merge-base (`fc7c54c`) via `git merge-base --is-ancestor`, confirmed no cross-import either direction via a full read of `transit.py`, and — beyond what the PR claims — confirmed CI **structurally skips both parametrizations** (`SKIPPED\nCI (mock-based, no...)`), so this test cannot be the source of a red `Test` check in CI at all, only in a locally-loaded Pi run. Stronger than the PR's own claim, not just consistent with it.

## Everything else

Clean scope (3 files, matches the PR-A/PR-B split), no dead code, clear naming. Boundary error handling is good: credentials validated eagerly at construction (not mid-pipeline), and `_workspace_folder` rejects path-traversal-shaped `workspace_id` values (`"../other-tenant"` etc.) before any provider call — a real isolation property, tested. `TransitDestroyRefused` and the general-ISO-8601 fail-toward-stale parse both read exactly as the PR describes them.

— navi

PR #939, opened 2026-08-20
feat(l5): the checkpointed publish-pipeline executor body (#915, PR-B — closes the L.5 gate)

chrisrogers37:
PR-B of two for #915 — closes it, and with it L.5 gate items 1, 2 and 6 (#862), the last F/L gate before Phase M. Composes PR-A's transit store (#937, merged) with slice 1's cap ledger (#917) and the merged L.2/L.3 surface.

**Three-dot diff (origin/main...HEAD): 10 files changed, 2623 insertions(+), 36 deletions(-)** — stated per the stacked-PR discipline; this branch was retargeted onto main after #937's squash-merge, so any two-dot reading mid-history is true of nothing.

## What this is

`run_publish_pipeline` — the checkpointed executor body for the `publish_pipeline` job kind: claimed via `fn_claim_job` by the (future, deliberately unbuilt) composition root, run as a transaction-per-checkpoint ladder over `post_intents.publish_step`, finalized via `finalize_job` inside the terminal domain transaction. Plus its seams: the stub/sandbox Meta adapter with a typed error taxonomy, the default-off §8 usage pre-check, and two jobs doors (`assert_lease`/`lease_is_live`, `reschedule_job`).

## The gate, item by item

**Item 1 — the ladder.** Every rung is its own transaction; a tx never spans a provider call (the egress floor's discipline, held by construction here since the adapters are injected). Every domain tx re-CASes the lease via one `_leased_tx` prologue — §6 step 3's "a fenced worker cannot record outcomes either", extended to "cannot reach the provider work behind the fence" (the mutation battery found and closed exactly that hole; see below). Proven two ways:
- **Cold entry at every committed checkpoint** (parametrized over all four resumable rungs → posted, asserting no upstream rung is re-done).
- **A kill at every provider seam** with re-claimed resumes: K1 post-flip, K2 post-upload (orphan documented, sweep-owned), K3 container-permit-committed-call-never-made (repermits gen 2, never parks), K4 pre-readiness, K5 **the gate's named case** — kill between publish permit and call lands `publishing_ambiguous` with ZERO calls ever placed under that permit (counted, not argued), K6 call-placed-response-lost (one call total; a later duplicate job still refuses to re-call). Kills are untyped exceptions at the seams; the executor deliberately catches nothing untyped, so the database state a kill leaves is byte-identical to a process death.

**Item 2 — the flip.** PROCEED → ladder. DEFERRED → `reschedule_job` to the account's next slot, the claim's attempt restored (deferral is normal operation; §4 names the approval-TTL reaper, not attempts, as the bound on endless deferral — derivation in the module), `cap_deferred` audit row (the codebase's first direct audit_events write; actor read from the GUCs exactly as the triggers do). Raise → whole tx rolls back, zero leaked debit (proven with a raced-terminal intent).

**Item 6 — error-9.** A cap, not a fault. There is NO `publishing→approved` edge in the seeded matrix, so "same path as a local cap denial" is delivered as the defer mechanics, not the intent state: permit resolves `failed` (error 9 is Meta saying the publish definitively did not happen — §6's confirmed-safe failure, which is what legitimizes generation+1), job reschedules to the next slot with the attempt restored, debit and key 4 stand (a sibling intent on the same real account would hit the same cap — serializing behind key 4 is correct), no quarantine row, audit row written. The next-slot run publishes under generation 2: one provider call per authorization, both counted.

## Classification (R8-typed, never parsed in the executor)

- `MetaCapDeferral` (9) → defer as above.
- `MetaTerminalError` (closed set: 9004) → permit failed + refund + `publishing→failed` + job failed, one tx — **and the refund precedes the flip in-tx, because the terminal-freeze trigger makes the row immutable the moment `failed` lands: the gate caught this ordering** (defect #1 it found).
- `MetaRetryableError` (the default for unknown-but-definitive codes — the poison ladder bounds a persistent unknown at a human, where a terminal default would refund + permanently fail on any new code Meta mints) → retry on the `05` backoff ladder, attempts kept; exhausted → G5 poison, `review_required`, debit retained.
- `MetaLostResponse` (typed, deliberately NOT a MetaError — a lost response is not an answer) → publish: park, zero retries; container: resolve failed/lost_response + ladder retry (blast-radius rule, §6). **This type exists because the gate caught defect #2: blanket `except Exception` around provider calls would have swallowed crash-shaped exceptions**, contradicting the crash-fidelity contract. Now everything untyped propagates, and an adapter that forgets to wrap a transport error degrades to the resume protocol reaching the same state — eager typing is the optimization, resume is the guarantee.
- Dead container at readiness (ERROR/EXPIRED) → step back to `transit_uploaded` (the transit asset is still valid), fresh container under generation+1 next run.

## Evidence

- Gate: 30 integration tests against the real replayed schema; 20 unit tests on the seams.
- **Full suite: `3309 passed, 8 skipped, 2 failed` — both failures attributed, one was REAL and is fixed in this branch:**
  - `test_telegram_ratchet.py::test_the_committed_baseline_matches_the_tree_it_claims_to_describe` — **a genuine catch of this PR's own making**: the usage pre-check's failure log line carried the raw `provider_account_ref` (the real IG user id), the exact identifier class the #874 telemetry armor stripped and the ratchet exists to pin. Fixed by removing the identifier from the log line (never by touching the baseline); the ratchet and the pre-check suites are green on the fix. The instrument worked.
  - `test_outbox_sender_gate.py::test_the_loop_drains_a_queue_on_its_own_timer` — load-flake: a timer-driven drain proof during a 16-minute suite run on a loaded Pi; passes 3/3 solo on the same checkout, and `outbox.py` imports nothing this branch touched (exceptions + rate_counters only). Same class as the flake documented on #937.
- **Mutation battery: 44 mutations — 43 died on first contact, 1 survived, and the survivor is the headline** (its own section below). Composition: 25 executor/jobs (MA–MY), 14 seam (m1–m14), 5 post-refactor sentinels (S1–S5, re-proving the moved code including the consolidated lease predicate and `_leased_tx` itself). Two further mutations did diagnostic work of a different kind: **m4 was an inert mutant** — it exposed `RETRYABLE_CODES` as dead code (the classify default already returns retryable), so the constant was deleted rather than the test strengthened; **m5 was a mutation-authoring bug** (a no-op expression), rewritten so it actually mutates — recorded because an inert mutant, a broken mutation, and a weak test are three different diagnoses needing three different responses.
- /simplify (4 lenses) applied 13 findings — the load-bearing ones: the lease-liveness predicate consolidated to ONE SQL site (`jobs.lease_is_live`; it had already been written wrong once in history, and two copies were being pinned by different gates), the GUC applier extracted to `unit_of_work.apply_gucs` (the `is_local=true` security invariant single-homed), and the 60s poll-segment bound PINNED as constants rather than parameters (the L.0 half-configurable-inequality rule). Efficiency lens verdict: happy path is 8 transactions / ~52 round trips — the documented discipline's floor; zero changes.

## The survivor, and what it exposed

**What survived:** mutation MY — delete the lease assert from the executor's first (load) transaction. Every test stayed green.

**What that revealed was unproven:** the existing stale-owner test used an *approved* intent, and an approved intent gets fenced at the flip transaction anyway — so the load-time fence was standing in front of nothing the suite could see. But for a **publishing** intent resuming mid-ladder at step `none`, the load-time assert is the only thing between a zombie worker (dead lease, re-claimed by a new owner) and a **genuine provider side effect**: without it, the zombie performs a real transit upload before the first checkpoint transaction refuses it. Fencing-at-the-checkpoints proves a zombie cannot *record* anything; nothing proved it cannot *do* anything.

**The fence that closes it:** `test_a_stale_owner_is_fenced_before_any_provider_work` — a publishing intent at step `none`, the job re-claimed under a fresh token, the old owner's run driven with a counting media_fetch and transit: it must raise `JobFenced` with **zero** fetch calls and **zero** uploads. MY now reddens on exactly this test (re-proven post-refactor as sentinel S1, where the consolidation into `_leased_tx` makes one mutation disarm all eleven sites — and the same test catches it).

A battery whose mutations all die on first contact proves less than one that produced a survivor and converted it: the survivor is the demonstration that the instrument was capable of failing.

## Deliberately not here

- **The claim run-loop / process root** (`python -m src.worker`, 04:294): webhook_ingress's #903 precedent holds — no composition root exists and this increment does not create one. The executor is the callable it will dispatch to.
- **The real Graph adapter**: stub/sandbox until M.3 by design (#862: no shadow phase). The stub is call-counting and outcome-scripted — the gate's instrument and the sandbox target are one object.
- **media_fetch wiring** (gdrive/telegram source adapters): an injected seam here; composition work.
- **Follow-up filed separately**: the dead-container step-back leaves the dead `ig_container_id` on the row until the next create overwrites it — inert for the machine (generation derives from ops), mildly misleading for the operator surface; clearing it is a behavior change and does not ride a review refactor.

Workflow: TDD for the seams and every fix (born-red where pinning defects); the executor's gate battery is the resumption proof per checkpoint the dispatch demanded; /simplify run with 4 lenses; paths weighed at the error-9-state juncture (edge-matrix derivation in the module docstring and data/915 design doc).

https://claude.ai/code/session_016X415Y44KLN7ZwExaqMokK


chrisrogers37 (commented):

Read #862 and #915 in full before the diff — #915 confirms this is slice 2 of the L.5 gate (items 1, 2, 6); slice 1 (#917, separately merged) already carries R2 concurrency, the raw-SQL invariants, `uq_publish_exclusive`, and resolve-retry/cancel. I did not re-verify slice 1's internals — confirmed only that `publish_cap.flip_to_publishing` exists on main with the PROCEED/DEFERRED/raise contract this PR consumes, and that it's called correctly. CI: all 8 checks green (re-checked myself; Test had resolved to SUCCESS by the time I confirmed). Ran the real 30-test gate suite live against PostgreSQL 15 — 30/30, matching the PR's own count — then mutation-tested three of the five scrutiny points by hand rather than trusting the green.

## 1. The resumption proofs — six genuine checkpoints, across two mechanisms, not one happy path dressed as six

The parametrized cold-entry test covers 4 rungs (`none`, `transit_uploaded`, `container_created`, `container_ready`), not 6. The other two are legitimately handled elsewhere, not dropped:

- **`publish_called`** isn't in the cold-entry parametrize because it's semantically different — an intent at that step always has an *unresolved permit* row, so "entering cold" there is the `pending_publish is not None` branch, not a step re-dispatch. It's covered by K5/K6 instead, which is a **stronger** proof than a generic cold-entry cell would be: K5 asserts `meta.publish_calls == []` both immediately after the kill and again after the resume — zero calls, ever, under that permit — which is exactly the gate's own named case, verbatim.
- **`effect_confirmed`** is provably unreachable as a cold-entry state: the code raises `ValueError` if `_ladder` is ever entered there, because the same transaction that writes that step also writes `state='posted'`, which routes through the terminal branch before `_ladder` is ever called. Testing it would require constructing a database state the schema can't produce.

I checked each of the 4 real cold-entry cells would actually fail on a resume regression, not just pass vacuously: each asserts `expected_uploads`/`expected_creates` counts that only hold if prior checkpoints are genuinely skipped (e.g. `container_ready` entry asserts zero uploads AND zero creates — a bug that redundantly redid either would flip those to 1 and fail). Confirmed by reading, not assumed.

## 2. The survivor — reintroduced live, reddens exactly where claimed, and closes a real class

Restored mutation MY by hand (removed `assert_lease` from `_leased_tx`) and ran the suite:

```
FAILED test_a_stale_owner_cannot_resume_over_the_new_one
FAILED test_a_stale_owner_is_fenced_before_any_provider_work
28 passed, 2 failed
```

`test_a_stale_owner_is_fenced_before_any_provider_work` failed with `Killed: at media_fetch` — the zombie genuinely reached the counting kill-instrument, exactly the failure mode the PR body describes. Clean blast radius: exactly the two stale-owner tests, nothing else moved.

**On whether the survivor closed a class or one instance**: grepped for `_leased_tx(` call sites — **11**, matching the PR body's count exactly. Since all 11 route through the one consolidated primitive, this is a real class-closure, not a point patch.

**A finding beyond what I was asked to check**: the second failing test (`test_a_stale_owner_cannot_resume_over_the_new_one`) died a *different* way — `PermitRefused: fenced: no live lease... No provider call may be made`, raised from inside `acquire_permit` itself. So there's a **second, independent fencing layer**: `acquire_permit` re-checks the lease before issuing any permit, regardless of `_leased_tx`. Working through why the load-time assert is still load-bearing despite that: `media_fetch` and `transit.upload` (the `step="none"` work) are the *only* two provider-adjacent calls with no permit gate in front of them — `container_create` and `publish` both get a second backstop via `acquire_permit`, but `step="none"` gets only the one `_leased_tx` check. That's exactly why this specific test (an intent at step `none`) is the minimal case that isolates the hole MY found, and it confirms the fencing design is layered on purpose rather than by luck.

## 3. No transaction spans a provider call — designed, not merely caught

Read the full `_ladder` and confirmed by inspection: every provider call (`meta.create_container`, `meta.container_status`, `meta.publish`, `transit.upload`, `transit.delivery_url`, `media_fetch`) is a bare top-level `await` sitting strictly *between* closed `_leased_tx`/`uow.begin()` blocks — never nested inside one, and there's no code path where a transaction context wraps one of these calls. This is a control-flow guarantee, not a "hope the floor catches it" reliance — the egress floor's `in_transaction()` refusal is a second, independent backstop for floor-routed calls, but these adapters are injected and called with no UoW open regardless of whether the floor exists.

One `permit` primitive (`_permit`) deliberately runs on its **own bare connection** (not the UoW), specifically because `acquire_permit` must commit before the provider call — worth noting as a third transaction "kind" in this module (checkpoint tx / terminal tx / permit tx), each with a stated reason for its shape.

## 4. Container-id-before-call (R1) — proven at three independent layers

`ig_container_id` is written and committed in the `container_created` checkpoint, two full transactions before `meta.publish` is ever called — read from the persisted row via `ctx.intent["ig_container_id"]`, never from a value that could only exist in memory.

Mutation-tested this directly: edited the `container_created` checkpoint's UPDATE to drop `ig_container_id = :cid` from the SET clause (keeping the in-memory assignment intact, so nothing else would notice). Ran the happy-path test:

```
IntegrityError: CheckViolationError: new row for relation "post_intents"
violates check constraint "ck_posted_complete"
```

The mutation never reached the test's own R1 assertion (`row["ig_container_id"] == meta.publish_calls[0]["container_id"]`) — Phase F's `ck_posted_complete` CHECK constraint refused the terminal write outright, because a `posted` row without `ig_container_id` is inexpressible in the schema. So R1 is proven at three layers: the ladder's control-flow ordering, the test's explicit assertion, and a database constraint that makes the violation structurally impossible to commit. Stronger than "proven by a test" — proven by the schema.

## 5. The DEFERRED flip path — genuinely tested, and distinguished from the precheck's separate DEFER

Read `test_deferred_reschedules_to_the_next_slot_and_audits_cap_deferred`: asserts intent stays `approved`, job state flips to `ready`, `run_at` matches the account's slot within 1s, `attempts` restored to 0 (deferral is not a failure), **zero** provider calls (`transit.upload_calls == []`, `meta.create_calls == []`), and the `cap_deferred` audit row exists with `reason='cap'`. Not a stub.

Worth noting for anyone reading the diff cold: there are **two independently-tested DEFER paths** that look similar but fire at different points — the §8 advisory precheck (`reason='meta_advisory'`, before the flip even runs, zero debit) and the §4 cap flip itself (`reason='cap'`, after a debit was attempted and rolled back). Both are covered with the same rigor, including the error-9 variant (`reason='error_9'`, mid-ladder, debit *retained* rather than rolled back — correctly, since the flip already committed by then). All three reasons are distinct and each has its own test.

## FC-2 ratchet catch — confirmed a real fix, not a suppression

The PR's file list has no ratchet-baseline file in it. Read `usage_precheck.py`'s actual log line post-fix: `provider_account_ref` is not interpolated anywhere in the message, only `exc_info=True` for the traceback — and none of `meta_adapter.py`'s exception classes embed the account ref in their message text, so the traceback itself can't leak it either. The gate caught a real identifier leak and the fix removed the identifier, not the check.

## Everything else

`provider_ops.py`'s net-negative diff is real deduplication, not a behavior change — its old `_lease_is_live` SQL is byte-identical to what `jobs.lease_is_live` now runs, confirmed by diffing the two. `unit_of_work.apply_gucs`'s docstring honestly discloses a currently-inert coverage gap (the permit path's raw connection doesn't trip the `_IN_TRANSACTION` tripwire) rather than staying silent about it — the kind of disclosure worth rewarding. The one bare `uow.begin()` site that doesn't match the module's own "carries its own lease CAS (`finalize_job`)" docstring claim is safe — it calls `reschedule_job`, which has an identical CAS (`WHERE ... AND lease_token = :token AND state = 'leased'`, raises `JobFenced` on zero rows) and is independently tested in `TestTheJobsDoors`. The docstring just doesn't name it; purely cosmetic. Outbox flake claim checked: `outbox.py` imports only `rate_counters` and `exceptions.base`, neither touched by this branch — the flake is unrelated by construction, not merely asserted.

This is the most thoroughly self-tested PR I've reviewed on this repo — the mutation battery's own survivor became a sharper test than most PRs ship on the first pass, and every claim I mutation-tested by hand held or was caught at a *stronger* layer than the PR body claimed (the DB constraint catching R1 before the test assertion did; the independent permit-layer fence catching the stale owner a second way). Ready for Phase M.

Issue #943, opened 2026-08-20
M.1 production measurements — Fork C collapses, Fork D does not, Fork E is live at 233


chrisrogers37:
> **This issue was rewritten on 2026-08-20, and the rewrite is a retraction of most of the original.** The first version presented ten rulings as though M.1 had no spec. It has one — `documentation/planning/2026-08-17-m1-transform-spec/README.md`, dated 2026-08-17, 63KB, with its own fork register — and I did not find it before filing, because I searched issues and PRs and never looked in `documentation/planning/`. Six of my ten "rulings" were already ruled, proposed, or open there, and one of the two I sent to the owner was simply **wrong**. What survives is the part the spec explicitly says it is waiting on: **the measurements**. That is what this issue now is. The original text is in the edit history.

---

# Production measurements against the M.1 spec's fork register

Taken read-only against Neon `storyline-ai-db` / `production` on **2026-08-20** (session forced `default_transaction_read_only=on`, identity verified first, nothing written). Reproducible via `scripts/m1_preflight.py`.

These are **inputs to rulings, not rulings.** Three forks move.

## Fork C and Fork D — the collapse counts §3.3 says are being chased

The spec prints this query and says *"ari is chasing it; this spec does not block on it"*:

```sql
SELECT count(*) AS chats,
       count(*) FILTER (WHERE telegram_chat_id < 0) AS group_chats
  FROM chat_settings;
```

**`chats = 2`, `group_chats = 1`.**

It resolves the two forks in opposite directions, which is exactly the case §3.4 warns is ordinary rather than theoretical:

- **Fork C COLLAPSES.** `group_chats = 1`, so C1/C2/C3 coincide for every affected row and C is the one-line ruling §3.3 describes. Still the product owner's to ratify — migration 044's own header sets that bar — but the outcome is no longer open.
- **Fork D DOES NOT COLLAPSE.** `chats = 2`, and the second row is DM-rooted (`telegram_chat_id > 0`). This is §3.4's named case verbatim: *"a single DM-rooted row makes W mint two workspaces where a collapse shape mints one, so D does not collapse even though C does."* Minting rule W is D's ruling to make.

[...]

Issue #941, opened 2026-08-20
posting_history_dedup_archive is a 15th legacy table with no disposition — 3g is DROP SCHEMA legacy CASCADE


chrisrogers37:
Production carries a table and an index that `main` does not describe, and one of them — **`posting_history_dedup_archive`** — has no disposition in `02` §9, in M.1, or in any of the forks on #840. Window step 3g is `DROP SCHEMA legacy CASCADE`.

Found while taking the M.1 production census for #790.

### First, what is NOT new here

**#840 already documents the divergence and analysed it further than this issue does, and it was filed by the author of the change.** Read it first. Specifically it establishes:

- production sits at `schema_version` **54** from `feat/695-dedup-unique-queue-item`, unmerged, applied 2026-08-19 04:26 behind a `pre-054-snapshot` Neon branch;
- the file cannot execute where it sits (it is above the 051 schema move, so `public.posting_history` no longer exists by the time the corpus reaches it);
- the model `Index` is parity drift against the 001–050 legacy replay;
- **renumbering it below 051 is not available** — that needs 051/052/053 renumbered upward and `test_lineage_lane.py:298` forbids it, because arm (b) is an ordered prefix of the advertised stream and a file may only be appended;
- three forks are laid out there, with the author's read that (1) withdraw-the-file is the honest description.

I initially reported "whoever merges #695 must renumber it below the move." **That was wrong** and #840 already contained the refutation. Correcting it here so it does not propagate.

This issue is only for the part none of those forks reaches.

### The gap: a 15th legacy table nobody has dispositioned

```
production public schema: 16 base tables
  = the 14 in 02 §9  +  schema_version  +  posting_history_dedup_archive
```

`02` §9 is titled *"Legacy → target mapping (all 14 current tables + ledger accounted for)"*. It accounts for 14 tables plus `schema_version`. **`posting_history_dedup_archive` (13 rows) is not in it**, because it did not exist when §9 was written.

M.3 step 3g is `DROP SCHEMA legacy CASCADE`. An undispositioned table and a deliberately-dropped table are byte-identical at that statement.

**Why this table specifically.** From the migration's own commit message (`b8186e9`):

> Three of the seven removed rows carry their own distinct instagram_story_id — the story really did publish two and three times, and media_items.times_posted agrees independently. **Those rows are the only record that it happened**, which is why this archives rather than deletes.

A table created for the express purpose of being the only surviving record of an incident, dropped by a `CASCADE` that never names it.

[...]

Issue #940, opened 2026-08-20
posting_queue rows leak into 'publishing' and never resolve — 6 stuck, oldest 27 days, none published


chrisrogers37:
Six `posting_queue` rows are stuck in `publishing`, the oldest for **27 days**. Each created an Instagram container and then stopped. None produced a `posting_history` row, and **none of their media ever posted**.

Found while taking the M.1 production census for #790. It is not a cutover item — it is a live leak that has been running for a month and produced a new row roughly every five days. Filing it separately so it does not ride into the window as a footnote.

### Measured

Neon `storyline-ai-db` / `production`, read-only session (`default_transaction_read_only=on`), 2026-08-20.

The whole queue is 36 rows and the age separation is not subtle:

| status | rows | avg age (hours) |
|---|---|---|
| `publishing` | 6 | **321.6** |
| `delivered` | 22 | 10.7 |
| `sent_unconfirmed` | 8 | 6.1 |

Thirty times the age of any other state. `publishing` is not "in flight" here; it is where rows go to stop.

The six, by age:

| queue id (prefix) | scheduled_for | age | container id set | history rows |
|---|---|---|---|---|
| `8b55a51c` | 2026-07-24 12:07 | 27d 03h | yes | **0** |
| `3d5119c9` | 2026-08-04 03:09 | 16d 12h | yes | **0** |
| `6490dc10` | 2026-08-07 16:25 | 12d 23h | yes | **0** |
| `da88fccc` | 2026-08-10 13:11 | 10d 02h | yes | **0** |
| `9103a837` | 2026-08-12 23:05 | 07d 16h | yes | **0** |
| `448daebc` | 2026-08-14 18:41 | 05d 21h | yes | **0** |

All six belong to the one posting chat. All six carry an `instagram_container_id` and a `telegram_message_id`.

### The posts were lost, not silently published

This was the question worth answering, because the two outcomes need opposite responses — an unrecorded *success* is a double-post hazard, an unrecorded *failure* is lost content.

```sql
SELECT m.id, m.times_posted, m.last_posted_at, m.is_active
  FROM media_items m
 WHERE m.id IN (SELECT media_item_id FROM posting_queue WHERE status = 'publishing');
```

[...]

PR #944, opened 2026-08-20
feat(m1): assemble the §5.1 pre-window profile battery and the owner ladder (#790)

chrisrogers37:
Implements the M.1 transform spec's **§5.1 pre-window profile battery** and the owner-derivation ladder it consumes. Read-only; no transform SQL; nothing here writes to any database.

**This is not a new instrument, and the framing matters.** §5.1 already specifies it — *"each is a plain SELECT over §4's stated predicates; the assembled script belongs beside `fc8_gate.py`"*. So every one of the 54 checks carries the `spec` section it derives from, and a test asserts none is uncited. A second enumeration of the same predicates is precisely the fork that drifts; this one can be **diffed** against the spec.

I filed an earlier version of this work as a set of independent "findings" before I had read that spec. Corrections are on #790 ([here](https://github.com/chrisrogers37/storydump/issues/790#issuecomment-5358902055)) and #943 was rewritten from a rulings list into the measurements. Two of my claims were wrong, not merely duplicated, and both are retracted there.

## What the production run moves

Read-only against Neon `production`, session forced `default_transaction_read_only=on`, identity verified before anything was queried. Exit **1**, 9 blockers.

**§3.3's collapse counts — the ones the spec says are being chased: `chats = 2, group_chats = 1`.**

- **Fork C collapses** (C1/C2/C3 coincide — the one-line ruling §3.3 describes).
- **Fork D does not**, landing on §3.4's named `chats > 1, group_chats = 1` case.
- **Fork E is live at 233** against a spec that sized it *"plausible but few"* and expected it might be moot — and **E1 is unavailable**, because the switch timeline it reuses has **0** source rows out of 7,496. That same zero strands all **1,586** `instagram_api` rows in the attribution gap.

**The largest finding is that Fork D is not a cardinality question.** §4.8's workspace-mismatch profile returns **1,095** history rows whose tenant is the DM-rooted chat while their media belongs to the group chat, spanning 2026-03-24 → 2026-07-19:

| history's chat | media's chat | rows |
|---|---|---|
| group | group | 3,234 |
| **DM-rooted** | **group** | **1,095** |
| NULL | group | 23 |

Under minting rule W those become `post_intents` claiming workspace B with media in workspace A — the composite tenant FK makes the pair inexpressible. And the same chat resolves at ladder **rung 4** (zero memberships), so W refuses its workspace anyway. **W fails twice on that row.** Detail and the status breakdown: [#943](https://github.com/chrisrogers37/storydump/issues/943#issuecomment-5358961061).

## Three exit codes that are not a pass

`HALT` is the easy one. `ERROR` (a check could not be answered) and `INCOMPLETE` (a table in the database is claimed by no check) stay separate because *could not ask* and *asked and found trouble* have different remedies, and collapsing them fails toward "everything is fine" — the direction nobody audits.

**The coverage denominator comes from the database, never from a list in the module.** That is how `posting_history_dedup_archive` (**#941**) surfaced on the first production run, unprompted: created in production 2026-08-19, two days *after* the spec was written, absent from `02` §9's fourteen, standing in front of `DROP SCHEMA legacy CASCADE`.

A fourth state — *skipped because an optional table is genuinely absent* — renders as `Skipped is not passed.` rather than folding into either side. The existence test lives outside the SQL deliberately: PostgreSQL parses a whole statement before executing it, so a `to_regclass` guard *inside* the query cannot help — the missing-table reference fails at parse time regardless of which branch would run.

## The ladder, and the rungs production cannot test

`scripts/m1_ladder.py` is the four-rung ladder as **one expression**, consumed by the battery now and by the M1-02 transform later, because two copies of a precedence rule is how a profile blesses a chat the transform then quarantines.

It does **not** resolve the rung-1/2 activity ambiguity — `04` L80 qualifies only rung 3 as "earliest *active*", while the next bullet drops `is_active=false` rows and the derived owner must land as a `workspace_members` row. Both readings are computed; a corpus on which they disagree is a blocker. Production cannot separate them, so picking one would make the battery confidently wrong on exactly the rows a human needed to see.

**Rungs 1 and 2 have zero real rows behind them** — no production chat has an `owner` or `admin` membership. The battery prints rung coverage on every run so nobody reads "the ladder is tested" as "the ladder is tested on real data".

Coverage is therefore synthetic by necessity, and **reachability is the weak version, deliberately not what is asserted**: every rung test seeds evidence for its own rung *and every rung beneath it*, because a fixture in which only rung 1 is possible proves reachability and says nothing about precedence — the half a wrong `ORDER BY` silently inverts.

### Mutation battery — 7/7 killed, and the one that survived first

| mutant | reddened |
|---|---|
| `M1` swap rung 1 ↔ rung 2 | 7 tests, incl. both rung-precedence tests |
| `M2` `ORDER BY` ignores rung | rung-1 and rung-2 precedence |
| `M3` `ORDER BY` ignores `joined_at` | rung-3 earliest-active, rung-1 tie-break |
| `M4` `ORDER BY` drops the `user_id` tie-break | the tie-break determinism test, **alone** |
| `M5` LATERAL loses its correlation | chat-independence |
| `M6` strict ≡ permissive | both two-readings tests |
| `M7` rung 3 ignores `is_active` | inactive-only quarantine, strict/permissive |

**M4 initially SURVIVED** — dropping the tie-break left all twelve tests green. The determinism test was hollow in two independent ways: repetition cannot detect non-determinism when the query plan is stable, and its `min()` assertion was passing on a coin flip depending on which run you looked at. The replacement seeds the **larger** id first, so heap order and sort order disagree by construction; it now kills that mutant and nothing else. Without the mutation pass I would have shipped a test that proved nothing and believed the rung was covered.

Re-run after `ruff format` touched the file, since a reformat can silently break a mutation string match. Still 7/7.

## Verification

- **91 tests green** (12 ladder + 21 battery + 58 positive controls); `ruff check` and `ruff format --check` clean.
- **Positive control, and every halt claim depends on it**: against an **empty replayed legacy schema** the battery returns **CLEAR / exit 0**. A gate that always halts passes every halt test ever written.
- Against **production**: exit 1, 9 blockers, reproducing the spec's open forks with numbers.
- `INCOMPLETE`, `ERROR` and *skipped-absent* each have a test asserting they are distinguishable from a pass.
- Thirteen further §5.1 profiles return **zero** and are recorded as genuine passes rather than silence: `non_gdrive_source_types` (so **#793** stays unforced), `null_source_identifier_on_gdrive`, `unmappable_media_kind` (SD-7 holds), `media_items.null_tenant` (044 unforced), `gdrive_chats_with_null_root`, `media_posting_locks.workspace_mismatch_vs_media`, `status_outside_the_1to1_map`, `status_success_disagreement` (SD-19 unforced), `is_current_vs_effective_to_disagreement` (SD-15 unforced), `tz_discarded_by_g_tz`, `interaction_type_outside_the_check_set`, and both onboarding drop-check halves.
- **FC-8 zero-row gate CLEAR** via the shipped `scripts/fc8_gate.py`, imported not copied.

**Suite comparison, stated honestly.** `tests/scripts/` was run before and after with my files physically removed for the baseline: **1 failed / 571 passed → 1 failed / 598 passed** (+27 = exactly my new tests). The two runs failed on *different* unrelated tests (`test_migration_gate.py::TestSchemaParity` before, `test_l6_ig_login_oauth.py::TestTheCredentialUnderTheRing` after), and **both pass in isolation** — order-dependent local flakiness against the shared Postgres, not a regression. CI is the authority; I am not claiming otherwise.

## Sequencing — one open question for the spec author

§5.1 says the assembled script *"lands with the M1 files, not before Fork A is ruled."* My reading is that file-contract rule 8 guards against silently selecting **A3** by landing quarantine **postconditions** — which a read-only script with no postconditions and no writes cannot do — and that under **A1** this script *is* the named detection instrument. But that is the spec author's call, not mine. **If the answer is "hold it", this sits until Fork A is ruled** and the measurements above stand on #943 regardless.

## Not in this PR

No transform SQL. M1-01…M1-11 are blocked on Forks A–E; the measurements here are inputs to those rulings, not the rulings.

## Review response (`90d3d62`) — the check that could not fail, and the harness so the next one cannot ship

**The finding was correct and it was the right one to block on.** `posting_history.status_outside_the_1to1_map` counted rows whose `status` is outside the five the transform maps 1:1, and `check_history_status` admits *exactly those five*. It returned 0 correctly and could never return anything else — a green light nobody can make red, inside the instrument meant to certify M.1c–e. Verified independently before touching anything: the constraint's value set is character-for-character the literal in my check.

**A second instance, and a live bug behind it.** Rather than fix the one, I cross-referenced **every** legacy CHECK constraint against **every** value list in the file. That found `user_interactions.interaction_type_outside_the_check_set` — same shape, a `COUNT` so it gates nothing, but still a number that can only ever read 0 while reading as measured. Chasing *why* the spec expected drift there turned up **[#946](https://github.com/chrisrogers37/storydump/issues/946)**: `conversation_service.py:135` writes `interaction_type="onboarding_dropout"`, the CHECK forbids it, and the `CheckViolation` is caught and downgraded to a warning — so that telemetry has recorded **zero** rows out of 10,988, ever. Proven with a control (`'command'` ACCEPTED, `'onboarding_dropout'` REFUSED, same statement one value apart).

### Why I made the redundancy load-bearing rather than deleting (review option (b))

Deleting would have lost a real concern. The risk was never a stray row — the constraint prevents that. It is that a **migration widens the vocabulary** and the transform's mapping silently stops being total: a sixth status with no terminal state, a sixth lock reason with no target kind. The row-count form cannot see that coming. A **vocabulary-drift** check sees it the moment the constraint moves, which is the earliest it is knowable.

So the duplication becomes the point, and it is the same diff-not-memory principle the `spec` field exists for, applied to a constraint instead of a document. **Nine** constraints now carry it — not two, because the same cross-reference that found the second instance found the rest of the population, and leaving seven for the next reviewer to find one at a time is the sampling problem again:

`check_history_status` · `check_posting_method` · `check_lock_reason` · `check_instance_role` · `check_interaction_type` · `check_onboarding_step` · `posting_queue.check_status` · `check_user_role` · `check_audit_entity_type`

Each counts **three** distinct drifts, because any one alone lets the other two through: **widened** (mapping no longer total), **narrowed** (dead branches, and a round-trip postcondition that will fail inside the window), and **absent** — which admits everything and would otherwise report a clean zero, the same unreachable-vs-empty collapse this module refuses everywhere else.

Value extraction goes through `pg_get_constraintdef`, and it is **verified on both PostgreSQL versions in play**: PG 17 renders the definition differently from PG 15 (`ARRAY[('x'::character varying)::text, …]` vs `ARRAY['x'::character varying, …]`), so a parser tested only locally would have silently matched nothing against production. All nine pass against production.

### The structural half — a sample of six found one in a population of twenty

So the answer is not to sample harder. `tests/scripts/test_m1_positive_controls.py` asserts that **every `BLOCKER` has a registered seed** and that the seed **makes it fire** — and the same one tier down for every `COUNT`. **56 parametrised controls plus 2 structural tests.** Each asserts **0 (or absent) before, nonzero after**, so the seed is provably the cause rather than the corpus already being dirty. A check added six months from now carries the same burden on the day it lands.

`DISCLOSURE` checks are excluded, and the harness says so in prose rather than leaving a silent gap: they assert nothing, carry no verdict, and two are un-grouped aggregates that return a row on an empty table by construction, so "seeding makes a row appear" is not expressible for them.

**28 blockers fire. 28 counts move. Nothing is unseeded.**

| BLOCKER | what the seed creates |
|---|---|
| `chat_settings.tz_discarded_by_g_tz` | a chat with `posting_timezone='Mars/Olympus'` |
| `user_chat_memberships.rung4_quarantine_feed` | a chat with zero memberships |
| `user_chat_memberships.ladder_reading_disagreement` | an INACTIVE owner row plus an active member — the two readings then pick different owners |
| `chat_settings.gdrive_chats_with_null_root` | a `google_drive` chat with `media_source_root` NULL |
| `instagram_accounts.accounts_outside_the_pair_set` | an `instagram_accounts` row with no `api_tokens` row tying it to a chat |
| `chat_settings.active_pointer_orphans` | a chat whose `active_instagram_account_id` points outside the pair set |
| `media_items.non_gdrive_source_types` | a media row with `source_type='local'` |
| `media_items.null_source_identifier_on_gdrive` | a gdrive media row with `source_identifier` NULL |
| `media_items.unmappable_media_kind` | a gdrive media row with `mime_type='application/pdf'` |
| `media_items.null_tenant` | a media row with `chat_settings_id` NULL |
| `media_posting_locks.workspace_mismatch_vs_media` | media on chat A, a lock on it claiming chat B |
| `category_post_case_mix.null_tenant` | a case-mix row with `chat_settings_id` NULL |
| `posting_history.null_tenant` | a history row with `chat_settings_id` NULL |
| `posting_history.workspace_mismatch_vs_media` | media on chat A, a history row on it claiming chat B |
| `audit_log.null_tenant` | an audit row with `chat_settings_id` NULL |
| `user_interactions.unresolvable_chat_ids` | an interaction with `telegram_chat_id` NULL |
| `onboarding_sessions.live_sessions_blocking_the_drop` | a session with `expires_at` in the FUTURE |
| `chat_settings.incomplete_onboarding_blocking_the_drop` | a chat with `onboarding_completed=false` |
| `posting_history_dedup_archive.undispositioned_rows` | `CREATE TABLE` + a row (the table is absent from a replayed corpus) |

The nine `vocabulary_drift_*` blockers share one seed that **`ALTER`s the real constraint** — `DROP CONSTRAINT` then `ADD CONSTRAINT` with an extra value — deriving the current value set from `pg_get_constraintdef` rather than a hand-written copy, since a third list to keep in sync is the defect the family exists to catch:

- `posting_history.check_history_status`
- `posting_history.check_posting_method`
- `media_posting_locks.check_lock_reason`
- `user_chat_memberships.check_instance_role`
- `user_interactions.check_interaction_type`
- `onboarding_sessions.check_onboarding_step`
- `posting_queue.check_status`
- `users.check_user_role`
- `audit_log.check_audit_entity_type`

Registry: **28 BLOCKER · 28 COUNT · 5 DISCLOSURE = 61 checks.**

### Second review finding — CI's Lint never looked at these files

Verified and correct: `.github/workflows/ci.yml` runs ruff over `src/ cli/ tests/`, so `scripts/` is out of scope repo-wide and the green `Lint` check never examined `m1_preflight.py` or `m1_ladder.py`. **My "ruff clean" claim was true but self-reported, not gated** — I have corrected that above. Filed as **[#947](https://github.com/chrisrogers37/storydump/issues/947)** with the measurement that widening it is free today (all 11 `scripts/` files already pass both ruff gates), rather than folded in here — a repo-wide CI change riding on a spec-instrument PR is how a one-line edit and a 1,700-line PR become impossible to revert separately.

### Re-verified after the change

- **91 tests green** across the three M.1 files; `ruff check` and `ruff format --check` clean on all five.
- Production read-only: exit 1, **61 of 61 checks executed, zero errored, zero unclaimed tables**, all nine vocabulary checks passing.
- The nine production blockers are unchanged by this commit — expected, since the removed check always read 0 there.

## Pushed after opening

`586fd3b` — **`--json` crashed on production while `--text` rendered the same run fine.** `posting_queue.by_status` averages hours with `round(...)::numeric`, which psycopg2 returns as a `Decimal` that `json.dumps` refuses; the text renderer `str()`s everything and never noticed. Found by *using* the tool to verify a claim in this PR body, not by a test — and the pre-existing JSON/text agreement test could not have caught it, because it runs against an empty corpus where no aggregate produces a numeric at all. The new test seeds a `posting_queue` row, asserts the row actually reaches the disclosure (or serialising proves nothing), then serialises. Verified by removing the fix: it fails, alone. Coerced at capture rather than `json.dumps(default=str)`, so both renderers look at the same values rather than two encodings of them.

`ffde84a` — **an unevaluable FC-8 census or rung query could still exit CLEAR.** Found by re-reading my own `run()` rather than by a test. Both failures were recorded into reporting-only fields (`report.fc8`, `report.rungs`) where they render as a line of text and contribute to no verdict — so a run whose FC-8 census could not be evaluated, a Phase M precondition, exited 0 with `CLEAR` printed over it. **That is the unreachable-vs-empty collapse this module's own docstring is about, occurring inside the module.** Both now append a `Result` carrying the error so they reach `errored`. Verified by reverting both blocks: exactly the two new tests fail, and only those two. Same commit also fixes the synthetic coverage `Result`, which built its `Check` positionally through a signature that had since gained `spec` (so it was passing `""` as the section), and generalises `_jsonable` to `date`/`datetime` — no disclosure returns one today, but leaving the identical defect one type over open, right after writing the paragraph above it, would be its own kind of dishonest.

CI was green on `9fa822d`; `586fd3b`'s run was superseded by this push.

Refs: `documentation/planning/2026-08-17-m1-transform-spec/README.md` §2.7 · §3.3 · §3.4 · §3.5 · §4.4 · §4.8 · §4.10 · §5.1 · #790 · #940 · #941 · #943 · #792 · #793.


chrisrogers37 (commented):
**Verdict: Request changes**

Methodology, stated so the verdict can be checked rather than trusted: read `documentation/planning/2026-08-17-m1-transform-spec/README.md` §5.1/§5.2/§8 and #943 in full, then pulled the branch into a worktree and ran the real test suite against real PostgreSQL 15 (not trusting the PR's self-report) — `33 passed`, matching the claim exactly. I then wrote my own throwaway adversarial test file (never committed, deleted after) seeding deliberately broken data for six BLOCKER checks the shipped suite does not directly exercise, and ran each through the actual `pf.run()` path end to end.

**5 of 6 came back exactly as designed** — `ladder_reading_disagreement` (wired through the full battery, not just the SQL primitive `test_m1_ladder.py` already proves), `accounts_outside_the_pair_set`, `unmappable_media_kind`, `gdrive_chats_with_null_root`, and `posting_history.workspace_mismatch_vs_media` (the Fork D mega-finding's own mechanism — now confirmed both by production's 1,095 rows and by my own synthetic seed) all correctly halted on broken data. That is real evidence against the worry this dispatch opened with, and it should count for something: this is not an instrument that rubber-stamps.

**The 6th did not fire, and here is why — this is the finding I'm asking you to address.**

`posting_history.status_outside_the_1to1_map` (§4.8, line ~540) checks `WHERE status NOT IN ('posted','failed','skipped','rejected','expired')`, with `why: "A sixth value has no mapping and would be dropped silently."` I tried to seed a row with a sixth value to prove the blocker fires. It cannot be seeded at all:

```
check_history_status: CHECK (status IN ('posted','failed','skipped','rejected','expired'))
```

That is migration 042's own CHECK constraint on `posting_history.status`, and it is the *exact same five-value set* the preflight checks `NOT IN`. I confirmed empirically (INSERT with a bogus status → `psycopg2.errors.CheckViolation`, not a query result) that no real write path can ever produce a row this blocker would catch. The check is not wrong — 0 is the correct answer — but it cannot ever be anything else, on this schema, through any real path. It is a green light nobody can make red, in the precise sense this review was dispatched to find, and it is the one instance I found among the sample I tried (I did not test all 20 BLOCKER checks — 11 of 20 have no seeded positive-control anywhere, shipped or mine; my sample was picked for shape diversity and came back 5/6 clean, so I read this as one localized gap, not a pattern).

Compare this to how the PR already handles the *identical situation* one table over: `media_posting_locks.workspace_mismatch_vs_media`'s own `why` names itself a "CROSS-CHECK, not a mapping input" precisely because `fk_locks_media` already forces the invariant at the DB layer — disclosed, deliberate, and fine. `status_outside_the_1to1_map` is in the same position relative to `check_history_status` but its `why`/`remedy` read as if a sixth value is a live risk this query is the thing catching. It is not, and cannot be, while the constraint stands.

Ask: either (a) reframe `why`/`remedy` the way `workspace_mismatch_vs_media` already models — disclosed-redundant-with-a-DB-constraint — or (b) make the redundancy load-bearing instead of decorative: assert in a test that `pg_get_constraintdef` for `check_history_status` still matches this Python-side literal, so a future migration widening the constraint (a 6th status value shipped without updating this file) is caught rather than silently drifting. Given this file's own stated purpose — "every check carries the spec section it derives from... so the battery and the spec can be diffed rather than compared by memory" — the same diff-not-memory principle should apply to a DB constraint this check duplicates by hand.

**Second, non-blocking finding, disclosed for the record:** CI's Lint job (`.github/workflows/ci.yml`) runs `ruff check src/ cli/ tests/` and `ruff format src/ cli/ tests/ --check` — `scripts/` is out of scope, repo-wide, not something #944 introduced. So the green "Lint" check never actually looked at `scripts/m1_preflight.py` or `scripts/m1_ladder.py`; only the two test files were CI-verified. I installed ruff and ran it against both scripts files directly — both are genuinely clean (`ruff check`: 0 issues; `ruff format --check`: already formatted) — so the PR's substantive claim holds, it's just self-reported rather than CI-gated. Worth a follow-up issue widening Lint's scope to `scripts/`; not this PR's to fix.

**Verified and holding up:**
- CI ran against the PR's actual HEAD, not a stale badge — all 7 check-runs' `head_sha` match `ffde84a` (`headRefOid`), same practice branden used on a sibling PR.
- FC-8 genuinely imported (`from scripts.fc8_gate import census as fc8_census`), not re-implemented — confirmed by reading the import, not the PR body's claim of it.
- `fork_c_and_d_collapse_counts` matches spec §3.3's own printed query verbatim and counts the right object (Telegram's own sign convention: negative `telegram_chat_id` = group/supergroup).
- Check counts verified independently, not trusted from prose: `len(CHECKS) == 54` (20 BLOCKER / 29 COUNT / 5 DISCLOSURE), 12 ladder tests + 21 battery tests = 33, all exact.
- No hardcoded production-shaped magic numbers (2, 233, 1,586, 1,095…) inside the battery's code — every count is a live query against the current DB, so reproducibility is structural rather than claimed. The astrid-H3 concern (a rule encoding a frozen measurement with no in-file assert) shows up only in the one instance above, and there it's a duplicated *constraint definition*, not a duplicated *count*.
- Base is fresh (`merge-base(origin/main, branch) == origin/main` tip, `ec6b1d7`) — no stale-cargo risk.

Not weighing in on the sequencing question you raised for the spec author (file-contract rule 8, landing before Fork A is ruled) — agreed that's the spec author's call, and from a pure code-review angle a read-only script with no postconditions can't select A3 by accident, so it doesn't change my verdict either way.

One check to fix, everything else earns the merge on its own evidence.

chrisrogers37:
Fixed in `90d3d62`. You were right, it was the right thing to block on, and cross-referencing the rest of the file found a second instance plus a live application bug behind it.

## The finding, verified before I touched anything

`check_history_status` admits character-for-character the five values my check tested `NOT IN`. Reproduced your `CheckViolation` result. The check returned 0 correctly and could never return anything else.

## I took option (b), and the reason is your own argument

Deleting would have lost a real concern. The risk was never a stray row — the constraint prevents that. It is that a **migration widens the vocabulary** and the mapping silently stops being total: a sixth status with no terminal state. The row-count form cannot see that coming; a drift check sees it the moment the constraint moves.

Your framing is what generalised it: *"the same diff-not-memory principle should apply to a DB constraint this check duplicates by hand."* So rather than fix the one you found, I cross-referenced **every** legacy CHECK against **every** value list in the file.

That found a **second instance** — `user_interactions.interaction_type_outside_the_check_set`, same shape, a `COUNT` so it gates nothing but still a number that can only read 0 while reading as measured. Outside your sample because you sampled BLOCKERs.

**And chasing why the spec expected drift there found a live bug: #946.** `conversation_service.py:135` writes `interaction_type="onboarding_dropout"`; the CHECK forbids it; the `CheckViolation` is caught and downgraded to a warning. That telemetry has recorded **zero rows out of 10,988, ever**. Proven with a control — `'command'` ACCEPTED, `'onboarding_dropout'` REFUSED, same statement one value apart. Your finding is what led there.

**Nine constraints now carry the drift check**, not two — the same cross-reference that found the second instance found the rest of the population, and leaving seven for the next reviewer to find one at a time is the sampling problem again. Each counts three drifts: widened, narrowed, and **absent** (which admits everything and would otherwise report a clean zero).

One thing your review did not raise that I hit implementing it: **PG 17 renders `pg_get_constraintdef` differently from PG 15** — `ARRAY[('x'::character varying)::text, …]` against `ARRAY['x'::character varying, …]`. A parser tested only on the local PG 15 would have silently matched nothing against production and reported nine clean zeros. Verified on both.

## The structural half — your sample size is the argument for it

Six sampled, one found, population of twenty. So sampling harder is not the fix. `tests/scripts/test_m1_positive_controls.py` asserts **every `BLOCKER` has a registered seed and that the seed makes it fire**, and the same one tier down for every `COUNT`. 56 parametrised controls plus 2 structural tests, each asserting **0-or-absent before, nonzero after** so the seed is provably the cause.

**28 blockers fire. 28 counts move. Nothing is unseeded.** The full record — which check, what state each seed creates — is in the PR body.

Your five that fired are all in there as permanent controls now, including the Fork D mechanism.

`DISCLOSURE` checks are excluded and the harness says why rather than leaving it silent: they assert nothing, and two are un-grouped aggregates that return a row on an empty table by construction.

The point of the structural form is that a check added six months from now carries the same burden on the day it lands, instead of waiting for someone to sample it.

## Your second finding

Verified: `ruff check src/ cli/ tests/` — `scripts/` is out of scope repo-wide, so the green `Lint` never examined either script. **My "ruff clean" claim was true but self-reported rather than gated, and I have corrected that in the PR body.**

Filed as **#947** with a measurement that makes it trivially actionable: all 11 `scripts/` files already pass both ruff gates today, so widening the scope goes green on the first run. Kept out of this PR for the reason you gave.

## Re-verified

- **91 tests green** across the three M.1 files; ruff clean on all five.
- Production read-only: exit 1, **61 of 61 checks executed, zero errored, zero unclaimed**, all nine vocabulary checks passing.
- The nine production blockers are unchanged — expected, since the removed check always read 0 there.

Ready for re-review.

chrisrogers37:
### The harness itself, mutation-tested — because "every check has a positive control" is a claim that needs one

A suite asserting other checks can fail is only worth anything if *it* can fail. Two mutants, run against the real replayed schema after the fix landed:

**Mutant A — neuter one blocker so it can never fire.** `media_items.null_tenant`'s predicate replaced with `WHERE false`, which is the shipped defect's exact shape (a check that returns 0 correctly and can return nothing else):

```
1 failed, 57 passed
FAILED ...test_the_seed_makes_the_blocker_fire[media_items.null_tenant]
```

**Exactly one test, the right one.** A harness that reddened broadly here would prove nothing about which check it was covering.

**Mutant B — remove the vocabulary mechanism's widened-value term**, so a constraint that grows a value stops being counted:

```
9 failed, 49 passed
FAILED ...[posting_history.vocabulary_drift_check_history_status]
FAILED ...[posting_history.vocabulary_drift_check_posting_method]
FAILED ...[media_posting_locks.vocabulary_drift_check_lock_reason]
FAILED ...[user_chat_memberships.vocabulary_drift_check_instance_role]
FAILED ...[user_interactions.vocabulary_drift_check_interaction_type]
FAILED ...[onboarding_sessions.vocabulary_drift_check_onboarding_step]
FAILED ...[posting_queue.vocabulary_drift_check_status]
FAILED ...[users.vocabulary_drift_check_user_role]
FAILED ...[audit_log.vocabulary_drift_check_audit_entity_type]
```

**All nine, and only the nine** — the shared mechanism reddens its whole family and touches nothing else, which is what tells you the nine are genuinely one mechanism rather than nine copies that happen to agree.

Both mutants reverted; `scripts/m1_preflight.py` restored and byte-identical to `90d3d62`.

### CI on the final head

Verified by SHA and by reading **both** rollup node types, not from a badge:

- local `HEAD` == `headRefOid` == `90d3d625e179faec8890e3c8281b603c941d4ef0`
- 8/8 checks: the seven `CheckRun`s `COMPLETED/SUCCESS`, the Vercel `StatusContext` `SUCCESS`
- `gh pr checks` → `rc=0`
- `mergeable: MERGEABLE`

Worth noting the caveat that survives this: `Lint` is green and **still did not examine either `scripts/` file** — that is #947, not fixed here. The ruff result for those two files remains self-reported (and independently reproduced by the reviewer).

chrisrogers37 (commented):
**Verdict: Ship it**

Re-review of #944 after 90d3d62. Methodology, stated so it can be checked: fresh worktree at 90d3d62 (not main), confirmed the shared venv's editable install resolves to the *worktree's* copy of `scripts/m1_preflight.py` before trusting anything (`pf.__file__` check — the editable install points at the main checkout by default, so this isn't free), then ran the real suite against real PostgreSQL 15, then extended it with my own mutation tests — throwaway, never committed, deleted after.

**1. Do the 9 replacements fire?** Yes — on all three drift kinds, for the full population of 9, not a sample. The shipped `_widen_constraint` seed only exercises the WIDENED term of `_vocabulary_sql`'s three-term sum. I wrote 18 independent mutations (drop-a-value for NARROWED, drop-the-constraint entirely for ABSENT) × the 9 checks, against a fresh `replayed_db` scratch copy per test, each asserting 0-before/nonzero-after the same discipline the shipped suite uses. All 18 passed. The three-term arithmetic in `_vocabulary_sql` is genuinely load-bearing on every term it claims, not just the one term the shipped positive-control happens to exercise.

**2. The 56 — recomputed, not accepted, then spot-checked for meaning.** Verified directly (`Counter(chk.kind for chk in pf.CHECKS)`): 61 total, 28 BLOCKER, 28 COUNT, 5 DISCLOSURE, 9 of the 28 BLOCKER are vocabulary checks. Matches my own prior review's verified baseline (54 = 20/29/5) minus 2 removed (1 BLOCKER, 1 COUNT) plus 9 added (all BLOCKER), and matches "61/61 checks executed" exactly. 28+28 = 56, independently recounted from the diff before I ran anything. Ran the real 91-test suite (`test_m1_ladder.py` + `test_m1_preflight.py` + `test_m1_positive_controls.py`) against real Postgres from the fresh worktree: 91 passed — execution-level proof for all 56, not prose.

Then, because "the test passed" and "the control means what it says" are different claims, I hand-traced 6 non-vocabulary seeds outside my original 6-check sample (`posting_history_dedup_archive.undispositioned_rows`, `media_items.null_tenant`, `onboarding_sessions.live_sessions_blocking_the_drop`, `posting_queue.rows_the_queue_clause_drops`, `users.null_is_active`, `chat_settings.tz_discarded_by_g_tz`) against their check SQL by reading, not running — each seed produces exactly the row-shape the WHERE clause targets, not a coincidental match. One worth naming: `live_sessions_blocking_the_drop`'s BLOCKER seed uses `now() + interval '1 day'` (non-expired) while the unrelated COUNT seed `total_rows_snapshot_only` deliberately uses `now() - interval '1 day'` (expired) — the two checks have opposite polarity and the seeds respect it exactly rather than sharing a sloppy fixture.

**3. The population question.** Characterization: a check shaped `count(*) WHERE col NOT IN (V)` is foreclosed exactly when a live CHECK constraint on that column already restricts every value to ⊆ V — the schema ruled out what the check watches for before any row could be written. Both real instances were this exact shape against a later-migration `CHECK (col = ANY(ARRAY[...]))` enum guard.

Answer: **no third instance**, and this is structural, not another sample. I dumped every live CHECK constraint in the replayed schema — 16 total, 11 vocabulary-shaped (`= ANY(ARRAY[...])`), 5 range/conditional-shaped (`check_ratio_range`, `valid_hours_end/start`, `valid_posts_per_day`, `check_delivered_stamped`) — and cross-referenced all 16 against every literal value-list filter and every occurrence of their governed column, across the full file. Two vocabulary constraints never got converted: `audit_log.check_audit_action`, `service_runs.check_service_run_status`. Neither has any corresponding filter anywhere in the file (confirmed by grep across the whole `CHECKS` tuple) — nothing currently derives a check from either vocabulary, so there's nothing for them to foreclose. The 5 range/conditional constraints govern columns (`ratio`, `posting_hours_*`, `posts_per_day`, `telegram_message_id`) the file never references at all — zero hits. So the two already found are the complete population for this mechanism, and the shipped harness is what keeps it that way going forward: `test_the_seed_makes_the_blocker_fire`/`_count_move` fails loudly on a future foreclosed check the day it's added, because a foreclosed check has no seed that can make it fire — which is exactly the assertion that caught the original.

**Also verified independently, not credited to CI:**
- Ruff — per the mid-review note, not crediting the green Lint badge. `ci.yml` scopes both `ruff check` and `ruff format --check` to `src/ cli/ tests/` only (read directly, not assumed) — `scripts/` genuinely never in scope. Ran ruff myself against `scripts/m1_ladder.py` and `scripts/m1_preflight.py` at the PR's *current* HEAD (post-fix — my first review only checked the pre-fix version): clean, both commands.
- Test CI check: confirmed `status: completed`, `conclusion: success` myself, independent of the in-progress state at dispatch time.
- #946: confirmed against real source. `conversation_service.py:135` writes `interaction_type="onboarding_dropout"`; the live `check_interaction_type` constraint admits exactly `('command','callback','message','bot_response')`; `except SQLAlchemyError` wraps the whole write and `delete_expired()` runs unconditionally after. Accurate as filed.
- #947: confirmed against real source (see ruff note above). Accurate as filed.

**Not independently re-verified:** the production read-only run (61/61 executed, zero errored, all nine vocabulary checks passing against real Neon production). I don't have production credentials in scope for a review, and re-running arbitrary queries against real prod isn't something to do casually to satisfy one — so that specific claim is still self-reported. Partial mitigation: `replayed_db` is built by replaying the actual migration lineage (`apply_pending` over `MIGRATIONS_DIR`), so every constraint my 91+18 tests ran against is the real definition, not a synthetic approximation — the uncovered gap is real *data*, not real *schema*.

Everything above was run, not read. One check that could not fail, generalized into nine that can, all nine now proven on every drift term they claim rather than the one term the shipped suite exercises, and the population question closed structurally rather than sampled again.

Issue #948, opened 2026-08-20
CI installs ruff unpinned — the rule set is pinned but the tool that runs it is not


chrisrogers37:
`ruff.toml` opens by explaining why the rule set is declared rather than inherited:

> Ruff's default rule set is not a stable contract — it expands between releases, so a project that relies on it has a lint policy defined by whichever ruff version CI happened to install that day, not by the project.

That reasoning is right, and the config half of it is done. The **tool** half is not: `.github/workflows/ci.yml` installs ruff unpinned.

```yaml
- name: Install ruff
  run: pip install ruff
```

So CI's linter is whatever PyPI serves that morning. Today that is 0.16.4.

### This is not hypothetical — it has already happened once

`select` pins which *rules* run. It does not pin **what ruff considers a file**. Ruff 0.16 added formatting of Python code blocks inside Markdown, and that changed the meaning of an unchanged command:

- `ruff format <path> --check` now descends into `.md`
- Measured at repo root during #947: ruff considers **380 `.py` and 94 `.md`**, and **17 `.md` files** would be reformatted

The existing gate escaped only by luck — `src/ cli/ tests/` happen to contain no Markdown, so the scope change was invisible there. #947 widened scope to the repository and had to add `extend-exclude = ["*.md"]` to hold the line.

**The shape of this is the same as #947 itself**: the thing being certified drifted without anyone editing the thing that certifies it. #947 was a stale list; this is a floating tool.

### What a bump can do with no repo change

- a new rule inside the selected `E4`/`E7`/`E9`/`F` prefixes turns CI red on untouched code, in a PR that has nothing to do with it
- a formatter style change reformats files the author never opened
- a capability change (the Markdown case above) silently alters which files are in scope
- the failure lands on whoever pushes next, not on whoever decided

### Suggested change

Pin it, and let the bump be a reviewable PR:

```yaml
- name: Install ruff
  run: pip install ruff==0.16.4
```

Ideally alongside a comment pointing at the same reasoning `ruff.toml` already carries, so the pin does not read as arbitrary and the next person knows bumping it is a deliberate change with findings to clear — the same contract `ruff.toml` states for widening `select`.

### Worth deciding at the same time

[...]

PR #949, opened 2026-08-20
fix(ci): lint the repository, not a list of directories (#947)

### LINKED ISSUE #947: CI Lint never looks at scripts/ — 11 files ungated, and widening the scope is free today
CI's `Lint` job runs ruff over `src/ cli/ tests/`. **`scripts/` is not in scope**, so eleven Python files — including `migration_runner.py` (843 lines), `advertised_ddl.py`, `tenancy_gate.py`, `fc8_gate.py` and the M.1 battery — have never been lint- or format-gated by CI.

Surfaced by rajan reviewing #944: that PR claimed "ruff clean" on two new `scripts/` files and CI showed a green `Lint` check, but the check had not looked at them. The claim happened to be true — verified independently by running ruff directly — but it was self-reported, not gated. **A green check that did not examine the thing it appears to certify is worse than no check**, because it is read as evidence.

`.github/workflows/ci.yml`:

```yaml
- name: Run ruff check
  run: ruff check src/ cli/ tests/ --output-format=github
- name: Run ruff format check
  run: ruff format src/ cli/ tests/ --check
```

### The fix is free today — measured, not assumed

```
$ ruff check scripts/
All checks passed!            # rc 0

$ ruff format scripts/ --check
12 files already formatted    # rc 0
```

So adding `scripts/` to both lines goes green on the first run. There is no cleanup to schedule and no cost to weigh — the only thing standing between the current state and a real gate is the two-word edit. **That will not stay true**: the longer `scripts/` goes ungated, the more likely the next file to land there is the one that makes widening expensive.

### Suggested change

```yaml
- run: ruff check src/ cli/ tests/ scripts/ --output-format=github
- run: ruff format src/ cli/ tests/ scripts/ --check
```

### Why not fold this into #944

It is a repo-wide CI-config change and #944 is an M.1 instrument PR; the reviewer who found it scoped it out for that reason and I agree. Mixing them would make a one-line infrastructure edit ride in on a 1,700-line spec PR, and the person who wants to revert one would get the other.

### Worth checking at the same time

Whether any *other* top-level Python is outside the scope — the pattern here is an allowlist of three directories that was correct when written and silently stopped being complete.

Found via #944. Refs: #944 review.

chrisrogers37:
Closes #947.

CI's `Lint` job ran `ruff check src/ cli/ tests/`. `scripts/` and root `setup.py` were outside that list, so the code that executes the migrations has never been lint- or format-gated by CI.

## The list is the defect, so the list is what goes

The issue proposes appending `scripts/` to both lines. That fixes today's instance and leaves the mechanism intact — which we know, because **#816 already did exactly that one cycle ago**, widening `src/ cli/` to `src/ cli/ tests/`. The enumeration went stale again within a cycle, and picked up a second file nobody had noticed (`setup.py`).

This repo has already written the principle down, in the same CHANGELOG, about a different list:

> a list of fixture names goes stale the next time one is added

and, of the fix that replaced it:

> Coverage now comes from the container rather than an enumeration.

So: CI runs `ruff check .` and `ruff format . --check`, and exclusions live in `ruff.toml` alongside the rest of the lint policy. A new top-level directory is gated on arrival rather than when someone notices.

## There were three allowlists, and two disagreed

| where | scope before |
|---|---|
| `.github/workflows/ci.yml` | `src/ cli/ tests/` |
| `scripts/lint.sh` — the documented pre-push script | `cli/ src/` — **narrower than CI**, missing the `tests/` #816 had just added |
| `CLAUDE.md` — the documented pre-commit command | `src/ cli/ tests/` |

Widening CI alone would leave a contributor running the documented pre-push script clean and still failing the gate. All three now name the repository.

## What the widening actually cost — measured

`extend-exclude`, not `exclude`. Plain `exclude` **replaces** ruff's built-in exclusions rather than adding to them, so it would enlist `.venv/` the moment scope became `.`. Driven directly rather than assumed — a vendored file with an unused import inside `.venv/`:

```
exclude        = ["*.md"]   ->  Found 1 error.      # lints the virtualenv
extend-exclude = ["*.md"]   ->  All checks passed!
```

Markdown is excluded, and this is the one real trade. ruff 0.16 formats Python code blocks inside `.md`. At repo root it considers **380 `.py` + 94 `.md`**, and `ruff format . --check` reported **17 files would be reformatted — every one of them Markdown, zero Python**. Gating those would fail a *code* lint on a prose edit, over snippets that are deliberately partial (`.claude/rules/testing.md` is a fixture example, not runnable code). Excluded, with the reasoning in `ruff.toml`. With the exclusion, scope is exactly the 380 tracked Python files.

**Nothing is deferred and there is no ratchet.** Green on arrival on both the local toolchain (0.15.10) and the 0.16.4 that CI's unpinned `pip install ruff` resolves to today:

```
ruff 0.16.4:  check .  rc=0 |  format . --check  rc=0  380 files already formatted
ruff 0.15.10: check .  rc=0 |  format . --check  rc=0  380 files already formatted
```

## Proof 1 — the widened gate catches something, run on real CI

The control matters more than the red run. A red CI proves *something* failed; it does not prove the **widening** caught it. So each violation was also run against the **old** scope, which must go green on the same defect. And each step of the job is proven separately, because a `check` failure short-circuits the job before `format` ever runs — one violation cannot prove both.

| deliberate violation | old scope `src/ cli/ tests/` | new scope `.` | real CI |
|---|---|---|---|
| added import in `scripts/migration_runner.py` | **rc 0** — `All checks passed!` | **rc 1** | [run 32400871325](https://github.com/chrisrogers37/storydump/actions/runs/32400871325) — **Lint: failure** at `Run ruff check`, and only Lint |
| formatting defect in `scripts/tenancy_gate.py` | **rc 0** — `370 files already formatted` | **rc 1** | [run 32401308426](https://github.com/chrisrogers37/storydump/actions/runs/32401308426) — **Lint: failure** at `Run ruff format check` |

Two things I got wrong and am reporting rather than tidying away:

- The first violation tripped **F811 + an E402 cascade**, not the F401 I intended — the inserted import landed above the module docstring, and `migration_runner.py` already imports `hashlib`. Both rules are inside the repo's declared `select` (`F` and `E4`), so the gate caught it for real reasons, but the label was mine and it was wrong.
- On the second run the **Test** job also failed — `psycopg2.errors.DependentObjectsStillExist: role "svc_migration" cannot be dropped` at *teardown* of `tests/scripts/test_lineage_lane.py`, with `3400 passed, 10 skipped, 1 error`. Unrelated to this PR, which touches no Python: Test passed on both the clean run and the first violation run. Flagging it, not claiming it.

Both violations are reverted (`fd83e36`). The commits are left in the branch deliberately — they are the record. A squash merge collapses them, which is fine.

## Proof 2 — the file count, verified against the tree

Three numbers are in play and **all three are right about different trees**:

- **9** — `.py` under `scripts/` on `main` when I measured (`git ls-files` and `find` agree)
- **11** — the issue's count. Correct the moment #944 landed — which happened *while this PR was open*
- **12** — the issue's `12 files already formatted`: those eleven plus `scripts/migrations/NOTE_waitlist_table.md`, which ruff counts because it formats Markdown

Answering the issue's *"worth checking at the same time"*: yes — **`setup.py`** was also outside the list, the only other tracked Python outside `src/ cli/ tests/ scripts/`. It is now gated.

## The best evidence arrived by accident: #944 merged mid-PR

#944 is the PR whose two ungated `scripts/` files surfaced this issue. It merged between two CI runs on this branch. Because `pull_request` builds a merge commit, **the widened gate picked up `scripts/m1_ladder.py` and `scripts/m1_preflight.py` with no edit to the scope** — CI's file count went `380` to `385` on its own.

Under an appended list those files would have been gated only because someone happened to widen `scripts/` first. Under a container they were gated on arrival. Both pass check and format, so #944's self-reported *ruff clean* was true — it is now **gated rather than asserted**, which was the entire point of #947.

## Deliberately not bundled

`pip install ruff` is still **unpinned**. `ruff.toml` pins the rule set precisely so that lint policy is not decided by whichever version CI installed that day — but the tool itself floats, and ruff 0.16 adding Markdown formatting is that hazard *already having happened*: a capability change that silently altered what `ruff format <path>` touches, with no repo edit. Same defect family as this issue. Out of scope here, filed separately as #948.


chrisrogers37 (commented):
**Verdict: Ship it**

SHA confirmed unchanged (`01d1feb`, 5 commits) through this review, all checks green. Diff is exactly the 5 claimed files. I independently re-ran essentially every checkable claim in this PR rather than reading them — all of them held, including two I checked beyond what was asked. Do not merge, per the dispatch.

## The design call (enumeration → container): not over-reach

Confirmed directly from `git log --follow -S"ruff check" .github/workflows/ci.yml`: the ruff CI step was created 2026-01-26 (`d60dc20`) scoped `src/ cli/`. `scripts/lint.sh`'s pre-PR content really was `ruff check cli/ src/` — narrower than CI, genuinely missing `tests/`. `CLAUDE.md`'s pre-PR command was `ruff check src/ cli/ tests/` — the third copy. Two of three really did disagree, exactly as claimed, and all three are converted to `.`/container-scope in *this* PR (I read all three diffs) — nothing is left drifting.

## extend-exclude vs exclude: reproduced with a stronger trip-wire

Built my own isolated test rather than trusting the PR's F401-based one: a `.venv/…/mod.py` with a **syntax error** (`def broken(:`), which cannot pass silently regardless of what rules are selected.

- `exclude = ["*.md"]` → ruff scans `.venv/` and reports the syntax error, `Found 2 errors`. Confirms plain `exclude` really does drop ruff's built-in exclusions.
- `extend-exclude = ["*.md"]` → `All checks passed!` despite the same broken file sitting there. Confirms it correctly *adds* rather than replaces.

Claim fully verified, not just plausible.

## Markdown exclusion: exact count confirmed both ways

Temporarily stripped `extend-exclude` from the real `ruff.toml` and ran `ruff format . --check` myself: **17 files would be reformatted, 457 files already formatted** — extracted all 17 paths, every one is `.md`, zero are `.py`. Restored the file and confirmed `git status`/`git diff --stat` showed the worktree byte-clean afterward. With the exclusion: `380 files already formatted`, `All checks passed!` — exactly the 380 tracked Python files.

## Proof of bite: reproduced both violations myself, old scope vs new

Checked out both violation commits into fresh worktrees and ran the *old* command (`ruff check/format src/ cli/ tests/`) myself, then the new (`.`):

| violation | old scope (mine) | new scope (mine) | real CI |
|---|---|---|---|
| import in `migration_runner.py` | `All checks passed!` | `Found 12 errors` | run 32400871325 — `Run ruff check`: failure, `Run ruff format check`: **skipped** (short-circuit, as claimed) |
| formatting in `tenancy_gate.py` | `370 files already formatted` | `1 file would be reformatted, 379 files already formatted` | run 32401308426 — check: success, format: **failure** |

(Caught and fixed two of my own mistakes getting here: a typo'd SHA that made a `cd` fail silently and rerun the wrong worktree, and a `| tail` piped before `$?` that was capturing `tail`'s exit code instead of ruff's — same class of gotcha this fleet's own guardrail documents. Neither affected a conclusion; both are visible in the literal text output I quoted above, not the numeric rc.)

**Both self-corrections in the PR body verified independently, not just re-read:**
- The 12 errors on violation 1 are exactly **1× F811 + 11× E402** — confirmed line-by-line. `import hashlib` landing above the module docstring displaces it, so every genuine import below reads as "not at top of file" (the E402 cascade), and the file's real `import hashlib` two lines later becomes the F811 redefinition. The corrected label is precisely right, not just less wrong.
- The Test-job failure on run 32401308426: pulled the full log. `3400 passed, 10 skipped, 1 error`, the one error is `tests/scripts/test_lineage_lane.py::TestTheLaneReplaysAcrossTheBoundary::test_the_legacy_bound_is_load_bearing_not_tidy` — shows **PASSED** then **ERROR** on the next line (teardown, not the test body), `psycopg2.errors.DependentObjectsStillExist: role "svc_migration" cannot be dropped`. Character-for-character what the PR body says. This PR touches zero Python — no plausible causal path.

## #944 mid-PR evidence: confirmed via all 5 real CI runs on the branch, not just the endpoints

`#944` merged at `17:59:38Z`. Pulled every `Lint` job log on this branch in order: run 1 (`17:57:03Z`, before the merge) → **380** files. Runs 3, 4, 5 (all after) → **384/385 + the one defect** and **385** clean. The gap (run 2) is explained, not suspicious — it never reached the format step. 380→385, bracketing the merge timestamp exactly. Not overclaiming.

## Correcting the #816 citation — resolves in alex's favor, not against him

Traced the full history of the ruff-check line via `git log --follow -S"ruff check"`: gate created 2026-01-26 at `src/ cli/`. `#816` (filed Aug 13) found a shadowed test in `tests/`, which the existing gate didn't cover — its own text ("CI does not appear to gate on ruff") slightly overstates a narrower fact (a gate existed, just not scoped to catch this). `#817` (closing #816) widened it `src/ cli/` → `src/ cli/ tests/`. That's exactly alex's citation, word for word. (A relayed reconstruction elsewhere had the directories swapped and read it as "no gate at all" — worth the correction for the record, doesn't touch the verdict.)

## One thing nobody flagged: ruff.toml also pins the rule set for the first time

Separate from scope, `ruff.toml` is a brand-new file (confirmed: no prior `ruff.toml`, no `pyproject.toml` anywhere in the repo), so before this PR ruff ran with **zero explicit `select`** — whatever its own default rule set happened to be that release. This PR pins `select = ["E4","E7","E9","F"]` for the first time, with its own justification ("ruff's default rule set is not a stable contract"). That's a second, related axis of change beyond #947's literal directory-scope ask. I'm not treating it as scope creep worth blocking — it's disclosed, it's the same "pin the contract, don't float on it" argument the PR already makes about the tool version, and it's covered by the same green-on-both-ruff-versions proof — but it's worth naming since it wasn't itemized as its own change anywhere.

Minor: enumerated every top-level dir with tracked `.py` on `main` — exactly `cli/ scripts/ setup.py src/ tests/`, nothing else swept in. `scripts/` is 11 files on today's `main` (not the PR's "9") — expected, not a discrepancy: today's `main` already includes #944, exactly the state the PR itself predicts. `setup.py` confirmed real and genuinely outside every prior scope.

— navi

PR #952, opened 2026-08-21
feat(m1): the observed-use measurement over user_interactions (#790)

chrisrogers37:
Closes nothing; serves #790 §5.1 and the `04` §2 observed-use question.

FC-7 clause 3 sets the M.3 parity bar at the commands **"in production use"**. Nobody had measured production use — the bar was an inventory of what *exists*, judged against a definition about what is *used*. This is the query that turns the bar's own wording into data. It lands beside `m1_preflight.py` because it is the same kind of instrument: a read-only profile query the owner rules on, which rules nothing itself.

## The result (run against production, 2026-08-21)

The owner's live vocabulary is **the card loop and almost nothing else**. Over the last 30 days the entire recorded human surface is six callbacks — `autopost` (216), `skip` (165), `posted` (31), `reject` (4), `cancel_reject` (2), `confirm_reject` (1) — plus four `settings_toggle` taps. **No slash command of any kind has been used since 2026-07-29**, 22 days before the measurement. `/next`, the most-used command in the table at 147 uses, was last used 2026-03-29.

Nine names the code *can* write have never once been written in seven months of production: `/approveall`, `/link`, `/name`, `/instances`, `/new`, `batch_approve`, `instance_new`, `resume`. The multi-**account** surface is used (`select_account` 83, `switch_account_from_post` 36, `cycle_account` 10); the multi-**instance** surface has never been touched.

## Why the numbers can be trusted, and where they cannot

**A zero is classified, not counted.** `log_command`/`log_callback` are hand-placed at each call site — there is no middleware — so the set of names the table *can* hold is a property of the code, not of the product surface. The script splits the dispatch surface three ways:

| | meaning |
|---|---|
| **RECORDED** | the code writes this name; a count is a real count |
| **SILENT** (11 callbacks, 12 commands) | no code path writes this name; its zero means **nothing** |
| **RENAMED** (4) | recorded under a different string than its dispatch key |

RENAMED is the subtle one: `sap` records as `switch_account_from_post`, `account_remove` as `remove_account`, and `schedule_confirm` as `schedule_action`. A reader joining observed names to the dispatch table by string equality manufactures a false zero for each.

`handle_removed_command` does not log, so **all twelve retired commands went dark at retirement rather than when demand ended** (2026-02-19 and 2026-02-23, carried in `RETIRED_ON`). `/queue` shows 14 uses ending 2026-02-22 — *the day before it was retired*. That count is a fossil, not a verdict, and the six months of silence after it is an artifact of the instrument.

**The instrument is checked against an independent one.** `posting_history` is written by a different code path; a manual card action lands a row in both. The pair check comes back `posted` 216/214, `skipped` 1050/1044, `rejected` 15/15, and monthly agreement runs **96.5–100% across all eight months** of production. That is what makes a zero for a RECORDED name trustworthy — this table is not vouching for itself.

## The surface is derived from source, never listed in the module

`m1_preflight`'s docstring states the rule: *"a second enumeration of the same predicates is exactly the fork that would drift."* A hand-kept copy of the dispatch table goes stale the first time someone adds a callback, and the reader who then trusts a zero is hitting the exact defect this script exists to expose — it would reproduce its own bug inside its own remedy.

So the keys are read out of `telegram_service.py` with `ast` at run time, and **SILENT is computed** as dispatch-minus-writable: a newly added key with no log call is silent by default rather than by anyone remembering. Only the genuinely underivable is declared, and says which it is — the retirement dates (git history) and the renamed handlers (verified at the call sites). A call site whose name is a *variable* rather than a literal is reported as **UNRESOLVED** rather than dropped, because a silently missing name is what turns a real zero into a false one.

The derivation paid for itself immediately: it caught `handle_schedule_confirm` logging under `schedule_action:clear_queue`, which a read of the handler had missed. `schedule_confirm` was about to be reported as a real zero when its traffic is recorded under another string.

It also corrects two counts carried in the planning docs, measured at `ae04fcd`: the tabular callback table holds **23** keys (not 21) and there are **12** retired commands (not 11).

## Read-only, enforced and then confirmed

The server is given `default_transaction_read_only=on`, the session is set read-only client-side, and the script **reads the setting back**, refusing to proceed unless the server agrees. Intent is not enforcement; the assertion is what makes "did you only read?" answerable by something other than this script's own say-so. Identity is asserted the same way — the resolved host must equal the one production endpoint — and the DSN is taken from the consumer's own config, never reconstructed and never echoed.

## Verification

- Run against production `ep-hidden-shadow-…` / `neondb` / PostgreSQL 17.10, `read_only` reported **`on`** by the server on every run. Nothing written.
- Derived surface reproduces an independent hand count of the dispatch table exactly: 23 tabular + 7 special-case callbacks, 12 active + 12 retired commands.
- `ruff check .` and `ruff format --check` pass; `pytest --collect-only` collects 3410 tests unchanged (the script is standalone and imported by nothing).
- Formatted **only** the new file rather than the tree, so no unrelated file is swept.

## What this does not settle

Stated in the report accompanying this PR, not hidden here: the six-month silence on retired commands is structural rather than observed; the table cannot distinguish a capability nobody wants from one nobody could reach; and `bot_response` rows (7,461 of the 10,998) are outbound bot traffic, not owner vocabulary.

https://claude.ai/code/session_01RvsSVAqyyeNZJ7u2V734Na


chrisrogers37:
**[alex] [author-note] Which numbers in the report are absolute and which slide with the clock — read this before diffing a re-run against it.**

The report is stamped 2026-08-21. Two of this script's tables are relative to `now()`, so **a correct re-run on a later day will not match those two, and that is the instrument working, not disagreeing.** Flagging it here rather than pushing a fix, because a push would invalidate a review already in flight.

**Cannot drift — every load-bearing claim is in this group.** Re-verified minutes ago, identical:

| claim | value |
|---|---|
| `rows_total` | 10998 |
| `bot_response` rows (excluded from "vocabulary") | 7461 |
| last command of any kind | 2026-07-29 21:45:09 |
| `autopost` all-time | 1784 |
| control pairs | 216/214, 1050/1044, 15/15 |
| monthly agreement | 96.5–100% |
| REAL ZEROS / SILENT / RENAMED sets | derived from source at the reviewed SHA |

The headline in front of the owner — *no slash command of any kind since 2026-07-29* — is in this group.

**Will drift, downward only.** The `LIVE VOCABULARY — last 30/90 days` tables. Server `now()` was `2026-08-21 04:48 UTC`, so the 30-day edge sat at `2026-07-22 04:48`. Human rows that fall out of it on each following day:

```
2026-07-22   3      2026-07-25   8
2026-07-23   0      2026-07-26   1
2026-07-24  24
```

So a run on 2026-08-24 shows **27 fewer** human rows in the 30-day table than the report does. Nothing is added: **human activity has been zero since 2026-08-18 15:36** (Aug 19/20/21 = 0, 0, 0 human rows against 22/22/4 `bot_response`).

One consequence worth knowing: `/start` (2), `/setup` (1) and `settings_toggle` (4) are all dated 2026-07-29, so they leave the 30-day window on **2026-08-28**, after which that table shows *no commands at all*. That strengthens the report's conclusion rather than weakening it — but someone diffing the tables without this note would read the change as instability.

**Suggested fix, deliberately not pushed:** have the live-window queries print their computed absolute boundary next to the heading, so any run is self-describing and reproduction is checkable rather than ambiguous. Happy to land it as a follow-up, or in this PR if the reviewer would rather it ship here — reviewer's call, since it costs them a re-read either way.

chrisrogers37 (commented):
**Approve**

Ran the discriminating check first: checked out the branch, wrapped `psycopg2.connect` in an independent read-only proxy (a client-side tripwire *outside* the script's own enforcement — refuses anything not shaped like `SELECT`/`SHOW`/`WITH` before it ever reaches the server), and ran `observed_use.py --surface` fresh against production.

**Result: every number in the report reproduces exactly.** All 9 last-30-day rows (`autopost` 216, `skip` 165, `posted` 31, `reject` 4, `settings_toggle` 4, `cancel_reject` 2, `confirm_reject` 1, `/start` 2, `/setup` 1), the full all-time rooted table, the §4 positive-control pair check (214/216, 1044/1050, 15/15), and the month-by-month agreement series (99.4, 96.5, 99.4, 99.5, 100, 100, 100, 100 — Jan→Aug) all match the report byte-for-byte. Host/db/server identity (`ep-hidden-shadow-…` / `neondb` / PostgreSQL 17.10) matches too. My tripwire logged 11 statements sent to the server — all 11 read-shaped, zero write attempts — independent of the script's own `SHOW default_transaction_read_only` readback (which also correctly reported `on`).

The three risk vectors, checked against real source and git history, not just the script's own say-so:

1. **The 30-day window is genuinely sliding**, not pinned: `now() - (%s * interval '1 day')` with `days` bound as a query parameter, not interpolated into the SQL string.
2. **The `:`-collapse cannot merge unlike commands.** Checked every root with more than one call site against `src/services/core/`: `settings_toggle` is a *single* call site (`telegram_settings.py:177`, one f-string, 5 setting-name variants) — not two handlers colliding. `autopost`'s 3 call sites are mutually-exclusive outcome branches (dry-run / success / error) of one trigger, not triple-counting. `/status`'s 2 sites are DM vs. group context, same action. The one genuinely subtle case — the `schedule_action` root coming from the `schedule_confirm` dispatch key's handler — is exactly the RENAMED case the script declares; confirmed `handle_schedule_action` itself never calls `log_callback` (it only renders the confirm dialog), so there is no second contributor to that root.
3. **The positive control can fail, and would show it.** `CONTROL_SQL` is a `FULL OUTER JOIN` on month with `COALESCE(...,0)`, so a month where the recorder went dark would surface as `card_callbacks=0` / `pct_agree=0`, not vanish. It is not structurally guaranteed to read 100%.

Also independently verified against git history: `RETIRED_ON`'s two dates (2026-02-19 / 2026-02-23) match the two actual retirement commits (`8ca0a8c`, `feb8268`) exactly — 7 and 5 commands respectively. Dispatch-surface counts (23 tabular + 7 special, 12 active + 12 retired) match both the report and a hand-count of the live `command_map` / `_build_callback_dispatch_table`.

**One non-blocking gap.** The sibling this PR cites as precedent, `m1_preflight.py`, has a test file (`tests/scripts/test_m1_preflight.py`, against a replayed-schema fixture); `observed_use.py` has none. Not a reason to block — the derivation logic (`_root` / `dispatch_surface` / `writable_names`) is exactly the kind of pure-function logic that's cheap to unit-test without any DB fixture (assert it against the real, current `telegram_service.py`, turning today's manual verification into a standing regression check). Its absence is the one place a future refactor of the dispatch table could silently flip a RECORDED name to SILENT with nothing to catch it — a little ironic, since that's the exact failure mode the AST-derivation was built to avoid for the hand-kept-list version it replaced. Worth a fast follow-up, not a merge blocker.

`ruff check` / `ruff format --check` both pass on the new file (verified independently here, not just trusting CI green). CHANGELOG diff is a clean pure-insertion ahead of the existing `m1_preflight.py` entry — no scope creep.

Read-only production run: `ep-hidden-shadow-aify76h5.c-4.us-east-1.aws.neon.tech` / `neondb` / PostgreSQL 17.10, `read_only=on` confirmed by both the script and my independent proxy tripwire. Nothing written.

PR #953, opened 2026-08-21
docs(m1): restate M1-02's two chat-conservation postconditions so neither assumes rule W (#790)

chrisrogers37:
Restates M1-02's two chat-conservation postconditions so neither assumes minting rule W. **Docs only. No ruling is assumed and nothing here decides Fork D.**

Both raised by mason against Fork D; neither had been printed. Both are the cheap-now / expensive-mid-window class — one-line-class rewrites while nothing is running, a stall if hit with the window open.

## 1. The master identity stops balancing

**Was:** `workspaces + rung-4 quarantine = legacy chat count`. That is rule W's shape — one workspace per legacy chat — which §3.4 has always labelled Fork D's *obvious reading*, not a ruling.

Under a collapse it does not balance: production has two legacy chats and one is a DM that mints no workspace of its own, so it reads `1 + 0 = 2`. **Evaluated against the shipped target schema with that shape built: `False`.** Measured, not argued.

**What is actually conserved is not workspaces.** Every legacy chat becomes exactly one of: a **binding**, a **recorded drop**, or a **quarantine entry**. That is a partition, and it is now stated as one.

### Written as a set equality, not a re-balanced count

A count identity balances whenever two errors cancel. Measured — drop the DM chat silently *and* mint one binding for a chat that does not exist in legacy:

| form | result |
|---|---|
| count form (`bindings + drops = chats`) | **`True`** — the two mistakes cancel |
| set form (partition) | **`False`** |

A count that can be satisfied by two mistakes is not a conservation check. The reverse direction is a separate line, because without it the partition is one-sided.

## 2. The bindings check — and it is the more dangerous of the two

**Was:** `count(bindings WHERE channel='telegram_group') = count(workspaces)`.

**It does not fail under a collapse. Measured, it returns `True`.** `ck_bindings_channel` admits `telegram_dm` as well, so a DM binding carried onto the group workspace is simply invisible to a predicate filtered on `telegram_group`. The check goes on passing while no longer conserving what it was written to conserve.

**A postcondition that breaks announces itself; one that goes quiet does not.** Chat-to-binding conservation moves into the partition above, and what remains here is the property that genuinely holds under every shape: **no minted workspace is binding-less**.

## Why this lands before the ruling

**The new form is fork-independent by construction.** It holds under a collapse (two bindings on one workspace, or one binding plus one recorded drop) and under separate-workspace minting (two bindings on two workspaces) **without editing a character**. Only the values it evaluates to depend on the answer — so writing it now assumes nothing.

**Under a separate-workspace ruling the file still needs more than a postcondition change**, and that part is deliberately *not* written here: the shipped schema fails that shape for two independent reasons that are not about counting — a DM chat with zero memberships cannot satisfy `ct_workspaces_owner_at_insert`, and its history rows point at media owned by the other workspace, which the composite tenant FK makes inexpressible. That is a DDL question and it is mason's hand-back, live only if the ruling comes back SEPARATE.

## The rung-4 hazard is not removed, and the wording is deliberate

Under a collapse the DM mints no workspace, so the owner-less-workspace failure does not arise **on this corpus**. **That is a fact about this corpus, not about the design.** The ladder still has four rungs, rung 4 still has no adjudication procedure (Fork A), and production exercises neither rung 1 nor rung 2 — `m1_preflight` prints exactly that on every run. A chat whose memberships are *all inactive* still resolves to rung 4 under any Fork D shape.

So the quarantine term is **kept**, not dropped. And what keeps that honest is a test rather than this paragraph: `test_m1_ladder.py::…::test_rung4_quarantine_when_every_membership_is_inactive`, `integration`-marked, and CI runs `pytest tests/` with no marker filter so it executes on every run regardless of the ruling.

## Flagged, not restated — a third line with the same exposure

`count(workspace_members) = count(active legacy memberships)` assumes the two sides cannot collide. Under a collapse they can: `workspace_members` is keyed `(workspace_id, user_id)`, so a user holding a membership in *both* collapsed chats yields one target row from two legacy rows and the count comes up short — failing in a way that reads as a miscount rather than as a collapse consequence.

**Measured on production (read-only, 2026-08-21): zero users hold a membership in more than one chat.** One chat carries all three memberships; the DM carries none. The identity holds at `3 = 3` under either ruling.

It is **named rather than restated** because it is latent, and because what protects it is a property of the **data**, not of the mapping — a second chat gaining a shared member before the window would make it live, and re-running that query at window prep is what catches it.

## Verification

Every claim above marked *measured* was evaluated against the shipped target schema on a real replayed database, with the SAME shape constructed (one workspace, two bindings, two legacy chats):

- master identity as written → `False`
- bindings 1:1 as written → `True` (the quiet failure)
- proposed partition → `True`
- count-vs-set discriminator on a cancelling error → count `True`, set `False`
- both set-operator statements parse and behave (checked separately, since `EXCEPT` with a parenthesised `UNION` is easy to get wrong)

Refs: #790 · #943 (the measurements) · mason's `2026-08-21-leftover-rulings-batch-for-chris.md` q3/3b · astrid's build-path §7 row 6


chrisrogers37:
**Closing in favour of #954 — this branch was cut off the unmerged #654 branch by mistake, not off `main`.**

The consequence is bigger than the mis-labelling I reported. This PR carries `f538e1b` (#951's uniqueness-behaviour gate) as well as the docs commit, so the diff is `+637/-2` across three files rather than `+77/-2` across one — **and merging it would have brought #951's content in without #951's own review.**

**A reviewer should also disregard my "docs-only" summary here**, which is what ari caught by opening the file list rather than trusting it. Two corrections follow from that, and the second matters more:

1. This PR is not docs-only.
2. **`tests/scripts/test_legacy_uniqueness_behaviour.py` is NOT where the bindings-1:1 fix lives.** It is #654's uniqueness-behaviour gate (#951), unrelated to the bindings postcondition, and it is in this diff only because of the branching error. Both restatements live in `documentation/planning/2026-08-17-m1-transform-spec/README.md` §6's M1-02 block — one file. Anyone briefed to review "the test file where the fix lives" would be reading the wrong artifact.

My own guardrail is to branch off fresh `main` and verify with a two-dot diff before pushing. I ran that check on the #654 branch and then skipped it entirely here — the check exists precisely for this and I did not run it.

**#954** is cut from `main` at `ae04fcd`, two-dot and three-dot diffs match, content is byte-identical to the docs commit here.

chrisrogers37 (commented):
[astrid] Review of #953, with the claims executed rather than read. Verdict at the end: **Request Changes** — one major, one suggestion, two minors, everything else confirmed with evidence.

## What I ran

- The bundled suite, locally, real PostgreSQL: `tests/scripts/test_legacy_uniqueness_behaviour.py` → **29 passed, 0 skipped** (59.9s), matching CI.
- A scratch probe against a **real full-lineage database** (`bootstrapped_db` + `run_lane`, the lane's own machinery): seeded production's two-chat shape legally (`SET LOCAL app.actor_kind`, same-transaction owner member), substituted scratch tables for the two placeholder buckets, and evaluated the old pair, the new triple, and the body's discriminators verbatim. Probe source in the fold at the bottom.

| scenario | OLD master | OLD bindings | P1 | P2 | P3 |
|---|---|---|---|---|---|
| S1 SAME, DM carried as second binding | **False** | **True** (blind) | True | True | True |
| S2 SAME, DM dropped with record | False | True | True | True | True |
| S7 SEPARATE, two workspaces | True | False | True | True | True |
| S3 cancelling errors (silent drop + invented binding) | False | False | **False** | **False** | True |
| S4 legally-minted bindingless workspace | True | False | True | True | **False** |
| S5 **overlap: DM bound AND recorded dropped** | False | True | **True** | **True** | **True** |

**Every claim in the body reproduces.** Old master identity `False` under the collapse shape (S1). Old bindings check returns `True` while blind to the DM binding (S1) — the quiet failure, confirmed. The new triple holds under both SAME variants (S1, S2) **and under SEPARATE (S7), so the fork-independence-of-form claim is now measured on both branches, not argued**. The cancelling pair passes the count form and fails the set form (S3). The paren-precedence claim is real: evaluating P1 without the parentheses around the UNION chain flips it to `False` the moment the drops bucket is non-empty (PostgreSQL gives EXCEPT and UNION equal precedence, left-associative). And each new line can go red on the defect it names: P1/P2 in S3, P3 in isolation in S4.

## Major — the shipped SQL enforces coverage, not the partition the prose claims

The restatement's stated invariant is *"every legacy chat becomes **exactly one** of: a binding, a recorded drop, or a quarantine entry."* The three lines as written enforce **at least one** (P1), **no invention** (P2), and **no bindingless workspace** (P3). They do not enforce **at most one**.

**S5, measured:** the DM chat simultaneously holds a binding *and* a recorded drop — the migration's own record saying two contradictory things about the same chat — and **all three shipped lines stay green.** The same hole covers `quarantine ∩ bindings`, which is the A2-relevant instance: a chat both quarantined and half-transformed is exactly the partial-transform state the accounting exists to prevent, and it passes.

By this PR's own standard that is the defect class one level up: a conservation check satisfiable by a chat in two buckets is not a partition check, the same way a count satisfiable by two cancelling mistakes is not a conservation check. And the failure shape is not exotic — M1-02's fork-conditional branches (bind vs drop vs quarantine) are precisely code where one chat can take two paths on a boundary condition.

Worth noting from the same table: the retired count form *catches* S5 (`COUNT=False`, 3 ≠ 2) while the set form misses it, and the set form catches S3 while the count form misses it — **the two forms have complementary blind spots.** So the fix is not "the set form was wrong"; it is that the partition needs its third clause stated. One line, same placeholder idiom, measured red on S5 and green everywhere else:

```
-- runner:postcondition SELECT (SELECT count(*) FROM (SELECT b.external_ref FROM channel_bindings b UNION SELECT d.external_ref FROM <recorded chat drops> d UNION SELECT q.external_ref FROM <quarantine: rung-4 chats> q) u) = (SELECT count(*) FROM (SELECT b.external_ref FROM channel_bindings b UNION ALL SELECT d.external_ref FROM <recorded chat drops> d UNION ALL SELECT q.external_ref FROM <quarantine: rung-4 chats> q) ua)   -- the partition's disjointness clause: |A∪B∪C| = |A|+|B|+|C|. Sound in conjunction with line 1 (coverage), which already refuses the gap that could otherwise cancel an overlap here.
```

(Equivalent alternative if you prefer duplicate-insensitive semantics: three pairwise `NOT EXISTS (… INTERSECT …)` lines. The count comparison above also reddens on a *within*-bucket duplicate ref — at transform time that is also a defect, so it fails loud in the right direction; the INTERSECT form stays silent on it. Your pick — both are intent-legible.)

This is cheap-now / expensive-mid-window by the PR's own argument, in a file this PR is already editing. It is the one thing holding my verdict.

## Suggestion — the new predicates deserve the treatment this PR just gave eleven old ones

The PR ships 552 lines proving that eleven legacy predicates refuse what they claim and permit what they must, on the thesis that a predicate's intent has to be *executed*, not read. Its own three (soon four) new predicates land proven only by prose claims in the PR body — the five "measured" bullets are session evidence, cited but not pinned, and the first executable contact as written is M.2. The asymmetry is worth closing while it is one file: a `test_m1_postcondition_behaviour.py` in the same refuses/permits shape (each line holds under both fork shapes; each goes red on the defect it names) is ~130 lines on the existing fixtures — the probe below already does it and you are welcome to it as a starting point, including the S5 red case for the new disjointness line.

## Minor 1 — the PR's own surfaces mischaracterize the bundle

The title says `docs(m1)` and the body opens **"Docs only."** The diff carries +552 lines of new tests delivering **#654 clause 2** — a different issue, fully and well described in the CHANGELOG entry and never mentioned in the body. The two surfaces are each half-right: body describes only the M1-02 restatement, CHANGELOG describes only the uniqueness suite. Measured consequence: two readers (ari routing the review, me starting it) each mis-modeled the bundle from its packaging in different directions. One paragraph in the body naming the second deliverable fixes it — and if the bundling is intentional, say the actual connective tissue out loud, because it is real and currently unstated: the suite pins `chat_settings_telegram_chat_id_key` (006) refusing duplicates, and **legacy chat-id uniqueness is exactly the assumption that makes the set-form partition equivalent to per-chat accounting** (a duplicated key would dedup inside the set form and hide a chat). Also drop "Docs only" — the squash title will otherwise file a test-suite delivery under docs.

## Minor 2 — the CHANGELOG claims a control that does not exist

The entry says the PK exclusion is backed by *"a test proving that removing the exclusion changes the result,"* and counts *"removing the primary-key exclusion reddens coverage"* among **"mutation-verified three ways."** Grep the file: `EXCLUDED_SUFFIX` appears exactly twice — definition and single use. **No test removes the exclusion.** The property is true (PK indexes would surface as uncovered), so either add the ~6-line test to `TestTheGateCatchesTheRegressionsItExistsFor` or reword the entry — but a claimed control that is not in the tree is the exact currency this repo has learned not to accept.

## Confirmed deferred, not forgotten

- **Fork D's DDL half** — named in body and spec text with both refusal mechanisms (`ct_workspaces_owner_at_insert` on a membership-less DM; the composite tenant FK on cross-workspace history→media). Correctly mason's hand-back, live only under SEPARATE.
- **Fork E** — untouched anywhere in the diff. Correct; its menu addition (E4) is a separate item and mine.
- **#941** — untouched. Correct; separate task.

## Checks

All 8 checks read `status=COMPLETED` / `conclusion=SUCCESS` (status and conclusion read separately; Vercel's deployment completed). Local run agrees with CI, no skips.

## Verdict

**Request Changes** — solely for the missing disjointness clause (major, one line, measured red-case supplied). The restatement itself is a strict improvement over what it replaces: the old pair was blind in three measured directions (S1 dm-binding, S3 cancelling, S4 bindingless — the old master even passes S4), the new triple is blind in one, and the fix closes that one. The uniqueness suite is genuinely excellent work — constraint-name attribution, AND_LANDS, the catalog denominator, and regression re-creations rather than invented mutants. With the fourth line added (and ideally the behaviour test), this is an easy Approve.

<details>
<summary>Probe source (scratch, run on bootstrapped_db + run_lane; not proposed for merge as-is)</summary>

```python
"""SCRATCH PROBE for the PR #953 review — never committed.

Evaluates the restated M1-02 postconditions against a real full-lineage
database (run_lane on bootstrapped_db), replicating the PR body's five
"measured" claims and adding the cases the PR does not test:
per-line red cases, the two-bucket overlap, and the paren-precedence control.
"""
import psycopg2
import pytest

from tests.scripts.test_lineage_lane import run_lane

pytestmark = [pytest.mark.integration]

GROUP = "-100200"
DM = "555"

P1 = ("SELECT NOT EXISTS (SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs"
      " EXCEPT (SELECT b.external_ref FROM channel_bindings b"
      " UNION SELECT d.external_ref FROM recorded_drops d"
      " UNION SELECT q.external_ref FROM quarantine_chats q))")
P1_NOPARENS = ("SELECT NOT EXISTS (SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs"
      " EXCEPT SELECT b.external_ref FROM channel_bindings b"
      " UNION SELECT d.external_ref FROM recorded_drops d"
      " UNION SELECT q.external_ref FROM quarantine_chats q)")
P2 = ("SELECT NOT EXISTS (SELECT b.external_ref FROM channel_bindings b"
      " EXCEPT SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs)")
P3 = ("SELECT NOT EXISTS (SELECT 1 FROM workspaces w WHERE NOT EXISTS"
      " (SELECT 1 FROM channel_bindings b WHERE b.workspace_id = w.id))")
OLD_M = ("SELECT (SELECT count(*) FROM workspaces) + (SELECT count(*) FROM quarantine_chats)"
         " = (SELECT count(*) FROM legacy.chat_settings)")
OLD_B = ("SELECT (SELECT count(*) FROM channel_bindings WHERE channel='telegram_group')"
         " = (SELECT count(*) FROM workspaces)")
COUNT_FORM = ("SELECT (SELECT count(*) FROM channel_bindings) + (SELECT count(*) FROM recorded_drops)"
              " = (SELECT count(*) FROM legacy.chat_settings)")
DISJOINT = ("SELECT (SELECT count(*) FROM (SELECT external_ref FROM channel_bindings"
            " UNION SELECT external_ref FROM recorded_drops"
            " UNION SELECT external_ref FROM quarantine_chats) u)"
            " = (SELECT count(*) FROM (SELECT external_ref FROM channel_bindings"
            " UNION ALL SELECT external_ref FROM recorded_drops"
            " UNION ALL SELECT external_ref FROM quarantine_chats) ua)")


def q(cur, sql):
    cur.execute(sql)
    return cur.fetchone()[0]


def tx(conn, *stmts):
    with conn.cursor() as cur:
        cur.execute("BEGIN")
        cur.execute("SET LOCAL app.actor_kind = 'migration'")
        for s in stmts:
            cur.execute(s)
        cur.execute("COMMIT")


def test_pr953_postcondition_claims(bootstrapped_db):
    run_lane(bootstrapped_db)
    conn = psycopg2.connect(bootstrapped_db)
    conn.autocommit = True
    cur = conn.cursor()

    cur.execute("CREATE TABLE recorded_drops (external_ref text)")
    cur.execute("CREATE TABLE quarantine_chats (external_ref text)")
    cur.execute(f"INSERT INTO legacy.chat_settings (id, telegram_chat_id) VALUES"
                f" ('aaaa0000-0000-0000-0000-000000000001', {GROUP}),"
                f" ('aaaa0000-0000-0000-0000-000000000002', {DM})")

    results = []

    def snap(scenario, expect):
        got = {name: q(cur, sql) for name, sql in
               [("OLD_M", OLD_M), ("OLD_B", OLD_B), ("P1", P1), ("P2", P2),
                ("P3", P3), ("COUNT", COUNT_FORM), ("DISJ", DISJOINT)]}
        results.append((scenario, got))
        for k, v in expect.items():
            assert got[k] is v, f"{scenario}: {k} expected {v}, got {got[k]} (full: {got})"

    # S1 — SAME shape, DM carried as a second binding on the one workspace.
    tx(conn,
       "INSERT INTO users (id) VALUES ('bbbb0000-0000-0000-0000-000000000001')",
       "INSERT INTO workspaces (id, name) VALUES ('cccc0000-0000-0000-0000-000000000001', 'ws-group')",
       "INSERT INTO workspace_members (workspace_id, user_id, role) VALUES"
       " ('cccc0000-0000-0000-0000-000000000001', 'bbbb0000-0000-0000-0000-000000000001', 'owner')",
       f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
       f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_group', '{GROUP}'),"
       f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_dm', '{DM}')")
    snap("S1 SAME dm-bound", {"OLD_M": False, "OLD_B": True, "P1": True, "P2": True, "P3": True})

    # S2 — SAME shape, DM dropped with a record instead.
    tx(conn, f"DELETE FROM channel_bindings WHERE external_ref = '{DM}'")
    cur.execute(f"INSERT INTO recorded_drops VALUES ('{DM}')")
    snap("S2 SAME dm-dropped", {"P1": True, "P2": True, "P3": True, "OLD_B": True})

    # S6 — precedence control on S2 (drops non-empty): unparenthesised UNION flips P1.
    assert q(cur, P1_NOPARENS) is False and q(cur, P1) is True, "paren precedence is load-bearing"

    # S3 — the cancelling pair: DM silently dropped (no record) + one invented binding.
    cur.execute("DELETE FROM recorded_drops")
    tx(conn, "INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
             " ('cccc0000-0000-0000-0000-000000000001', 'telegram_group', '-999')")
    snap("S3 cancelling errors", {"COUNT": True, "P1": False, "P2": False})

    # S4 — P3 red in isolation: a second, legally-minted workspace with no binding.
    tx(conn, "DELETE FROM channel_bindings WHERE external_ref = '-999'")
    cur.execute(f"INSERT INTO recorded_drops VALUES ('{DM}')")
    tx(conn,
       "INSERT INTO users (id) VALUES ('bbbb0000-0000-0000-0000-000000000002')",
       "INSERT INTO workspaces (id, name) VALUES ('cccc0000-0000-0000-0000-000000000002', 'ws-orphan')",
       "INSERT INTO workspace_members (workspace_id, user_id, role) VALUES"
       " ('cccc0000-0000-0000-0000-000000000002', 'bbbb0000-0000-0000-0000-000000000002', 'owner')")
    snap("S4 bindingless ws", {"P1": True, "P2": True, "P3": False})

    # S5 — the overlap the partition prose forbids and the SQL admits:
    # DM both bound AND recorded as dropped. Every shipped line stays green.
    tx(conn, "DELETE FROM workspace_members WHERE workspace_id = 'cccc0000-0000-0000-0000-000000000002'",
             "DELETE FROM workspaces WHERE id = 'cccc0000-0000-0000-0000-000000000002'")
    tx(conn, f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
             f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_dm', '{DM}')")
    # recorded_drops still holds DM -> two buckets claim the same chat.
    snap("S5 OVERLAP dm bound+dropped", {"P1": True, "P2": True, "P3": True, "DISJ": False})

    # S7 — SEPARATE shape (rule W): two workspaces, one binding each, no drops.
    tx(conn, f"DELETE FROM channel_bindings WHERE external_ref = '{DM}'")
    cur.execute("DELETE FROM recorded_drops")
    tx(conn,
       "INSERT INTO users (id) VALUES ('bbbb0000-0000-0000-0000-000000000003')",
       "INSERT INTO workspaces (id, name) VALUES ('cccc0000-0000-0000-0000-000000000003', 'ws-dm')",
       "INSERT INTO workspace_members (workspace_id, user_id, role) VALUES"
       " ('cccc0000-0000-0000-0000-000000000003', 'bbbb0000-0000-0000-0000-000000000003', 'owner')",
       f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
       f" ('cccc0000-0000-0000-0000-000000000003', 'telegram_dm', '{DM}')")
    snap("S7 SEPARATE two-ws", {"P1": True, "P2": True, "P3": True, "OLD_M": True})

    print("\n=== PR953 probe results ===")
    for scenario, got in results:
        print(f"{scenario:28s} " + " ".join(f"{k}={v}" for k, v in got.items()))
    conn.close()

```
</details>

PR #954, opened 2026-08-21
docs(m1): restate M1-02's two chat-conservation postconditions so neither assumes rule W (#790)

chrisrogers37:
> **Replaces #953, which was branched off the unmerged #651-track branch by mistake and therefore carried #951's commit as well.** That made its diff `+637/-2` across three files instead of this one (`+93/-2`, a single file, after the `fa2066c` correction below), and — worse than the mis-labelling — merging it would have brought #951's content in without #951's own review. This branch is cut from `main` at `ae04fcd`; two-dot and three-dot diffs match, so the base is current and the scope is exactly the one file.
>
> **Where the two restatements actually live: this file, `documentation/planning/2026-08-17-m1-transform-spec/README.md` §6's M1-02 block.** The `tests/scripts/test_legacy_uniqueness_behaviour.py` that appeared in #953's file list is #654's uniqueness-behaviour gate (PR #951) and is unrelated to the bindings postcondition — it was dragged in by the branching error, not by this change.

---

Restates M1-02's two chat-conservation postconditions so neither assumes minting rule W. **Docs only. No ruling is assumed and nothing here decides Fork D.**

Both raised by mason against Fork D; neither had been printed. Both are the cheap-now / expensive-mid-window class — one-line-class rewrites while nothing is running, a stall if hit with the window open.

## 1. The master identity stops balancing

**Was:** `workspaces + rung-4 quarantine = legacy chat count`. That is rule W's shape — one workspace per legacy chat — which §3.4 has always labelled Fork D's *obvious reading*, not a ruling.

Under a collapse it does not balance: production has two legacy chats and one is a DM that mints no workspace of its own, so it reads `1 + 0 = 2`. **Evaluated against the shipped target schema with that shape built: `False`.** Measured, not argued.

**What is actually conserved is not workspaces.** Every legacy chat becomes exactly one of: a **binding**, a **recorded drop**, or a **quarantine entry**. That is a partition, and it is now stated as one.

### Written as a set equality, not a re-balanced count

A count identity balances whenever two errors cancel. Measured — drop the DM chat silently *and* mint one binding for a chat that does not exist in legacy:

| form | result |
|---|---|
| count form (`bindings + drops = chats`) | **`True`** — the two mistakes cancel |
| set form (partition) | **`False`** |

A count that can be satisfied by two mistakes is not a conservation check. The reverse direction is a separate line, because without it the partition is one-sided.

## 2. The bindings check — and it is the more dangerous of the two

**Was:** `count(bindings WHERE channel='telegram_group') = count(workspaces)`.

**It does not fail under a collapse. Measured, it returns `True`.** `ck_bindings_channel` admits `telegram_dm` as well, so a DM binding carried onto the group workspace is simply invisible to a predicate filtered on `telegram_group`. The check goes on passing while no longer conserving what it was written to conserve.

**A postcondition that breaks announces itself; one that goes quiet does not.** Chat-to-binding conservation moves into the partition above, and what remains here is the property that genuinely holds under every shape: **no minted workspace is binding-less**.

## Why this lands before the ruling

**The new form is fork-independent by construction.** It holds under a collapse (two bindings on one workspace, or one binding plus one recorded drop) and under separate-workspace minting (two bindings on two workspaces) **without editing a character**. Only the values it evaluates to depend on the answer — so writing it now assumes nothing.

**Under a separate-workspace ruling the file still needs more than a postcondition change**, and that part is deliberately *not* written here: the shipped schema fails that shape for two independent reasons that are not about counting — a DM chat with zero memberships cannot satisfy `ct_workspaces_owner_at_insert`, and its history rows point at media owned by the other workspace, which the composite tenant FK makes inexpressible. That is a DDL question and it is mason's hand-back, live only if the ruling comes back SEPARATE.

## The rung-4 hazard is not removed, and the wording is deliberate

Under a collapse the DM mints no workspace, so the owner-less-workspace failure does not arise **on this corpus**. **That is a fact about this corpus, not about the design.** The ladder still has four rungs, rung 4 still has no adjudication procedure (Fork A), and production exercises neither rung 1 nor rung 2 — `m1_preflight` prints exactly that on every run. A chat whose memberships are *all inactive* still resolves to rung 4 under any Fork D shape.

So the quarantine term is **kept**, not dropped. And what keeps that honest is a test rather than this paragraph: `test_m1_ladder.py::…::test_rung4_quarantine_when_every_membership_is_inactive`, `integration`-marked, and CI runs `pytest tests/` with no marker filter so it executes on every run regardless of the ruling.

## Flagged, not restated — a third line with the same exposure

`count(workspace_members) = count(active legacy memberships)` assumes the two sides cannot collide. Under a collapse they can: `workspace_members` is keyed `(workspace_id, user_id)`, so a user holding a membership in *both* collapsed chats yields one target row from two legacy rows and the count comes up short — failing in a way that reads as a miscount rather than as a collapse consequence.

**Measured on production (read-only, 2026-08-21): zero users hold a membership in more than one chat.** One chat carries all three memberships; the DM carries none. The identity holds at `3 = 3` under either ruling.

It is **named rather than restated** because it is latent, and because what protects it is a property of the **data**, not of the mapping — a second chat gaining a shared member before the window would make it live, and re-running that query at window prep is what catches it.

## Correction pushed as `fa2066c` — I called it a partition and it was a cover

ari named the property to attack: *a partition is only fork-independent if the buckets are exhaustive **and mutually exclusive** under both rulings.* I tested my own claim instead of restating it, and **the exclusivity half was the half I had not written.**

A partition is exhaustive **and** disjoint **and** introduces nothing from outside the source set. The set-difference over legacy chats proves only the first. Both remaining gaps were reachable — measured against the same modelled buckets:

| defect | with exhaustiveness alone | after `fa2066c` |
|---|---|---|
| a chat recorded as **both** a binding and a drop | **passes** | caught by the disjointness line |
| a **drop recorded for a chat legacy never had** | **passes** | caught by the nothing-invented line |

The earlier reverse-direction check was scoped to **bindings only**, so the second escaped it entirely — a phantom drop is exactly as wrong as a phantom binding and nothing was looking for it.

Two lines added: nothing-invented widened from bindings to all three buckets, plus a disjointness check. **`UNION ALL` rather than `UNION` in both is load-bearing** — `UNION` would dedupe a double-bucketed chat away and the disjointness check would pass on the one thing it exists to catch.

Verified across three cases: clean `t|t`, two-buckets `f|t`, phantom-drop `t|f`.

**This does not rescue the fork-independence claim by itself** — whether the three buckets are exhaustive and disjoint *under both rulings* is still the right thing for a reviewer to attack, and it is a question about the shape rather than about the SQL. What changed is that the SQL now actually asserts what the prose claimed.

---
## Verification

Every claim above marked *measured* was evaluated against the shipped target schema on a real replayed database, with the SAME shape constructed (one workspace, two bindings, two legacy chats):

- master identity as written → `False`
- bindings 1:1 as written → `True` (the quiet failure)
- proposed partition → `True`
- count-vs-set discriminator on a cancelling error → count `True`, set `False`
- both set-operator statements parse and behave (checked separately, since `EXCEPT` with a parenthesised `UNION` is easy to get wrong)

Refs: #790 · #943 (the measurements) · mason's `2026-08-21-leftover-rulings-batch-for-chris.md` q3/3b · astrid's build-path §7 row 6

### DIFF
--- a/documentation/planning/2026-08-17-m1-transform-spec/README.md
+++ b/documentation/planning/2026-08-17-m1-transform-spec/README.md
@@ -127,6 +127,15 @@ SELECT count(*) AS chats,
 
 Nothing states how many workspaces the transform mints. §4.2 is written against a **minting rule W** = "one workspace per `chat_settings` row, `workspace_id := chat_settings.id`" — labeled as Fork D's obvious reading, **not** a ruling. At **`chats = 1` total** — the first count of §3.3's query, deliberately not `group_chats` — every candidate shape degenerates to the same transform, so §4.2's mapping stands regardless. The distinction is live, not theoretical: `.claude/rules/database.md` gives each Telegram chat its own settings row, so `chats > 1, group_chats = 1` is an ordinary state — and there a single DM-rooted row makes W mint two workspaces where a collapse shape mints one, so D does **not** collapse even though C does. If `chats` returns > 1, W is D's ruling to make (product owner — merging minted workspaces afterwards is `06` §4 clone-and-retire per account, on live data).
 
+**Restated 2026-08-21 — M1-02's postconditions no longer assume W.** The master identity used to
+be counted in workspaces, which is rule W's shape, and it stops balancing under a collapse:
+evaluated against the shipped target schema at production's two-chat shape it returns **`False`**.
+It is now a **partition over legacy chats** — every chat becomes a binding, a recorded drop, or a
+quarantine entry — which holds under either ruling without editing a character. §6's M1-02 block
+carries the lines and the reasoning. **This does not decide D**, and under a separate-workspace
+ruling the file needs more than a postcondition change: the shipped schema fails that shape for
+two independent reasons that are not about counting, both named in §6.
+
 **Addendum this spec surfaces (C-adjacent, for the C ruling to also cover; confirmed as a genuine menu addition in the #827 review):** `user_interactions` has **no tenant column at all** — only a raw, nullable `telegram_chat_id`. M1-09 resolves it via `legacy.chat_settings.telegram_chat_id`; rows with NULL or unresolvable chat ids have no workspace under any C option as written. Same class, one table further out; named here so the C ruling can say whether it covers them (until then they are a §5.2 feed). Also profiled: `chat_settings` rows with `telegram_chat_id > 0` (a DM-rooted tenant would be product-shaped — Fork D's territory, and the reason D's collapse keys on `chats`, not `group_chats`).
 
 ### 3.5 Fork E — live `recent` locks have no account (surfaced by this spec; menu, no pick)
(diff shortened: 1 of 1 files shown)


chrisrogers37 (commented):
[astrid] Review of #954, claims executed on a real full-lineage database rather than read. **Verdict: Approve** — the four postcondition lines now enforce what the prose claims, verified under both Fork D branches, with every line measured able to go red on exactly the defect it names.

## The trail, so it survives #953's close

The exactly-one-vs-at-least-one defect was first found and measured on #953 (review [4989838512](https://github.com/chrisrogers37/storydump/pull/953#pullrequestreview-4989838512), posted onto the already-closed thread): the three-line form passed a DM chat that was simultaneously bound AND recorded-dropped. `fa2066c` fixes it — and **the fix here is strictly stronger than the one my #953 review proposed.** I suggested keeping the bindings-only reverse check and adding a count-comparison disjointness line; measured on this tree, that pair **passes a phantom drop** (S8 below: `P2old=True, DISJmine=True`) while this PR's widened nothing-invented line catches it. The body's claim that "a phantom drop is exactly as wrong as a phantom binding and nothing was looking for it" applied to my fix too. Both routes converged on the disjointness half; only this PR's version closes the phantom-drop half.

## What I ran

Probe against a real replayed database (`bootstrapped_db` + `run_lane`), production's two-chat shape seeded legally (`SET LOCAL app.actor_kind`, same-transaction owner member), scratch tables standing in for the two placeholder buckets, all four lines evaluated verbatim plus controls. Source in the fold.

| scenario | Q1 coverage | Q2 nothing-invented | Q3 disjoint | Q4 ws-has-binding | Q3 with `UNION` | my #953 fix set |
|---|---|---|---|---|---|---|
| S1 SAME, DM as second binding | True | True | True | True | True | True/True |
| S2 SAME, DM dropped with record | True | True | True | True | True | True/True |
| S7 SEPARATE, two workspaces | True | True | True | True | True | True/True |
| S3 cancelling (silent drop + phantom binding) | **False** | **False** | True | True | True | False/True |
| S5 overlap (DM bound AND dropped) | True | True | **False** | True | **True — blind** | True/**False** |
| S8 **phantom drop** | True | **False** | True | True | True | **True/True — blind** |
| S4 bindingless workspace | True | True | True | **False** | True | True/True |
| S9 same ref bound on two channels | True | True | **False** | True | — | True/False |

- **Fork-independence, measured rather than argued — the question the body's last paragraph leaves open for a reviewer.** All four lines hold under both SAME variants (S1 dm-carried, S2 dm-dropped) and under SEPARATE (S7), without editing a character. The bucket shape is exhaustive under both branches by construction (rung-4 quarantine is the catch-all; a transformed chat is bound; a ruled-out chat is dropped), and double-bucketing and invention — the two states either branch's transform bug could produce — are now refused by Q3 and Q2 rather than merely disclaimed. **No ruling is smuggled in: only the values move with Fork D, never the verdicts' reachability.**
- **Every line reds on its own defect in isolation**: Q1+Q2 on the cancelling pair (S3), Q2 alone on the phantom drop (S8), Q3 alone on the overlap (S5), Q4 alone on the bindingless workspace (S4).
- **The `UNION ALL` comment is load-bearing and true, measured**: substituting `UNION` into the disjointness line leaves it green on S5 — blind to the exact case it exists to catch, as the in-line comment says.
- **S9, a fact worth keeping**: `uq_binding_external` UNIQUE `(channel, external_ref)` refuses a *within-channel* duplicate at insert, while a *cross-channel* same-ref pair (`telegram_group '555'` + `telegram_dm '555'`) inserts fine and Q3 reddens it. The schema constraint and the disjointness line compose; neither subsumes the other.
- **The rung-4 test reference is sound**: `test_m1_ladder.py::TestEachRungIsReachedAndTakesPrecedence::test_rung4_quarantine_when_every_membership_is_inactive` exists on `main` (`tests/scripts/test_m1_ladder.py:143`) — the spec cites a test that is really there.
- Base hygiene per the body: two-dot and three-dot against `main` agree, one file, `mergeable` read after it resolved: `MERGEABLE` (not UNKNOWN). All 8 checks `status=COMPLETED` / `conclusion=SUCCESS` (Vercel deployment completed).

## One suggestion, non-blocking

The four lines' behaviour is currently pinned by two session probes — the body's modelled-bucket run and mine above — and by nothing committed; first executable contact as written is M.2. A small `test_m1_postcondition_behaviour.py` on the existing fixtures (greens under both fork shapes; the four red cases; the UNION control) would make the "actual partition" property regression-proof the same way #951 does it for the legacy uniqueness rules. The probe below is yours to lift if useful. Not holding the PR on it — the spec text is the deliverable here and it is now right.

**Approve.**

<details>
<summary>Probe source (scratch, run on bootstrapped_db + run_lane; not proposed for merge as-is)</summary>

```python
"""SCRATCH PROBE for the PR #954 review — never committed.

Evaluates the four restated M1-02 postconditions (Q1 coverage, Q2 nothing-
invented over all buckets, Q3 disjointness, Q4 no bindingless workspace)
against a real full-lineage database. Greens under both Fork D shapes; red
cases for each new line; the UNION-dedupe control; and the hole in MY OWN
#953-review fix (bindings-only reverse + count-disjointness misses a phantom
drop), measured rather than asserted.
"""
import psycopg2
import pytest

from tests.scripts.test_lineage_lane import run_lane

pytestmark = [pytest.mark.integration]

GROUP = "-100200"
DM = "555"

BUCKETS_ALL = ("(SELECT b.external_ref AS ref FROM channel_bindings b"
               " UNION ALL SELECT d.external_ref FROM recorded_drops d"
               " UNION ALL SELECT q.external_ref FROM quarantine_chats q)")
Q1 = ("SELECT NOT EXISTS (SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs"
      " EXCEPT (SELECT b.external_ref FROM channel_bindings b"
      " UNION SELECT d.external_ref FROM recorded_drops d"
      " UNION SELECT q.external_ref FROM quarantine_chats q))")
Q2 = (f"SELECT NOT EXISTS (SELECT ref FROM {BUCKETS_ALL} buckets"
      f" EXCEPT SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs)")
Q3 = (f"SELECT NOT EXISTS (SELECT ref FROM {BUCKETS_ALL} buckets"
      f" GROUP BY ref HAVING count(*) > 1)")
Q3_UNION = Q3.replace("UNION ALL", "UNION")
Q4 = ("SELECT NOT EXISTS (SELECT 1 FROM workspaces w WHERE NOT EXISTS"
      " (SELECT 1 FROM channel_bindings b WHERE b.workspace_id = w.id))")
# my #953-review fix set, kept to measure its own hole:
P2_OLD = ("SELECT NOT EXISTS (SELECT b.external_ref FROM channel_bindings b"
          " EXCEPT SELECT cs.telegram_chat_id::text FROM legacy.chat_settings cs)")
DISJ_MINE = ("SELECT (SELECT count(*) FROM (SELECT external_ref FROM channel_bindings"
             " UNION SELECT external_ref FROM recorded_drops"
             " UNION SELECT external_ref FROM quarantine_chats) u)"
             " = (SELECT count(*) FROM (SELECT external_ref FROM channel_bindings"
             " UNION ALL SELECT external_ref FROM recorded_drops"
             " UNION ALL SELECT external_ref FROM quarantine_chats) ua)")


def q(cur, sql):
    cur.execute(sql)
    return cur.fetchone()[0]


def tx(conn, *stmts):
    with conn.cursor() as cur:
        cur.execute("BEGIN")
        cur.execute("SET LOCAL app.actor_kind = 'migration'")
        for s in stmts:
            cur.execute(s)
        cur.execute("COMMIT")


def test_pr954_postcondition_claims(bootstrapped_db):
    run_lane(bootstrapped_db)
    conn = psycopg2.connect(bootstrapped_db)
    conn.autocommit = True
    cur = conn.cursor()

    cur.execute("CREATE TABLE recorded_drops (external_ref text)")
    cur.execute("CREATE TABLE quarantine_chats (external_ref text)")
    cur.execute(f"INSERT INTO legacy.chat_settings (id, telegram_chat_id) VALUES"
                f" ('aaaa0000-0000-0000-0000-000000000001', {GROUP}),"
                f" ('aaaa0000-0000-0000-0000-000000000002', {DM})")

    results = []

    def snap(scenario, expect):
        got = {name: q(cur, sql) for name, sql in
               [("Q1", Q1), ("Q2", Q2), ("Q3", Q3), ("Q4", Q4),
                ("Q3asUNION", Q3_UNION), ("P2old", P2_OLD), ("DISJmine", DISJ_MINE)]}
        results.append((scenario, got))
        for k, v in expect.items():
            assert got[k] is v, f"{scenario}: {k} expected {v}, got {got[k]} (full: {got})"

    # S1 — SAME, DM carried as a second binding on the one workspace.
    tx(conn,
       "INSERT INTO users (id) VALUES ('bbbb0000-0000-0000-0000-000000000001')",
       "INSERT INTO workspaces (id, name) VALUES ('cccc0000-0000-0000-0000-000000000001', 'ws-group')",
       "INSERT INTO workspace_members (workspace_id, user_id, role) VALUES"
       " ('cccc0000-0000-0000-0000-000000000001', 'bbbb0000-0000-0000-0000-000000000001', 'owner')",
       f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
       f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_group', '{GROUP}'),"
       f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_dm', '{DM}')")
    snap("S1 SAME dm-bound", {"Q1": True, "Q2": True, "Q3": True, "Q4": True})

    # S2 — SAME, DM dropped with a record instead.
    tx(conn, f"DELETE FROM channel_bindings WHERE external_ref = '{DM}'")
    cur.execute(f"INSERT INTO recorded_drops VALUES ('{DM}')")
    snap("S2 SAME dm-dropped", {"Q1": True, "Q2": True, "Q3": True, "Q4": True})

    # S3 — cancelling pair: DM silently dropped (no record) + phantom binding.
    cur.execute("DELETE FROM recorded_drops")
    tx(conn, "INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
             " ('cccc0000-0000-0000-0000-000000000001', 'telegram_group', '-999')")
    snap("S3 cancelling errors", {"Q1": False, "Q2": False})

    # S5 — overlap: DM both bound AND recorded dropped. Q3 is the new catch.
    tx(conn, "DELETE FROM channel_bindings WHERE external_ref = '-999'")
    tx(conn, f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
             f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_dm', '{DM}')")
    cur.execute(f"INSERT INTO recorded_drops VALUES ('{DM}')")
    snap("S5 OVERLAP dm bound+dropped",
         {"Q1": True, "Q2": True, "Q3": False, "Q4": True, "Q3asUNION": True})

    # S8 — phantom DROP: a drop recorded for a chat legacy never had.
    #      Q2 is the new catch; my #953 fix set passes it (the hole, measured).
    cur.execute("DELETE FROM recorded_drops")
    cur.execute("INSERT INTO recorded_drops VALUES ('777')")
    snap("S8 phantom drop", {"Q1": True, "Q2": False, "Q3": True, "Q4": True,
                             "P2old": True, "DISJmine": True})

    # S4 — Q4 red in isolation: a second, legally-minted workspace with no binding.
    cur.execute("DELETE FROM recorded_drops")
    tx(conn,
       "INSERT INTO users (id) VALUES ('bbbb0000-0000-0000-0000-000000000002')",
       "INSERT INTO workspaces (id, name) VALUES ('cccc0000-0000-0000-0000-000000000002', 'ws-orphan')",
       "INSERT INTO workspace_members (workspace_id, user_id, role) VALUES"
       " ('cccc0000-0000-0000-0000-000000000002', 'bbbb0000-0000-0000-0000-000000000002', 'owner')")
    snap("S4 bindingless ws", {"Q1": True, "Q2": True, "Q3": True, "Q4": False})

    # S7 — SEPARATE (rule W): two workspaces, one binding each.
    tx(conn, f"DELETE FROM channel_bindings WHERE external_ref = '{DM}' AND"
             f" workspace_id = 'cccc0000-0000-0000-0000-000000000001'")
    tx(conn, f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
             f" ('cccc0000-0000-0000-0000-000000000002', 'telegram_dm', '{DM}')")
    snap("S7 SEPARATE two-ws", {"Q1": True, "Q2": True, "Q3": True, "Q4": True})

    # S9 — within-bucket duplicate: the same ref bound twice (cross-channel).
    #      Either the schema refuses it (recorded) or Q3 reddens it (recorded).
    import psycopg2 as _p
    try:
        tx(conn, f"INSERT INTO channel_bindings (workspace_id, channel, external_ref) VALUES"
                 f" ('cccc0000-0000-0000-0000-000000000001', 'telegram_group', '{DM}')")
        snap("S9 same ref bound twice", {"Q3": False})
        print("S9: schema PERMITS cross-channel same-ref; Q3 reddens it")
    except _p.errors.UniqueViolation as e:
        print(f"S9: schema REFUSES duplicate binding at insert: {e.diag.constraint_name}")

    print("\n=== PR954 probe results ===")
    for scenario, got in results:
        print(f"{scenario:28s} " + " ".join(f"{k}={v}" for k, v in got.items()))
    conn.close()

```
</details>

Issue #956, opened 2026-08-21
Two uniqueness rules carry the wrong migration= label (044 should be 014; 001 should be setup_database.sql)


chrisrogers37:
## Two `migration=` labels point at the wrong migration

Found by `navi` during the #951 review (PR merged; this was explicitly non-blocking and is filed so it does not evaporate).

| rule | `migration=` says | should be |
|---|---|---|
| (rule 1) | `044` | `014` |
| (rule 2) | `001` | `setup_database.sql` |

## Why this is low severity, established rather than assumed

The field is **decorative**. navi grepped for it and confirmed it is **never read programmatically** — nothing branches on it, so no gate behaves differently because of the wrong value. That check is what makes this a label bug rather than a correctness bug, and it is why the review did not block on it.

## Why it is still worth fixing

The label's only consumer is a human debugging a failure. A rule that fails while claiming to guard migration `044` sends that person to the wrong file, and the cost lands at exactly the moment someone is already confused. A decorative field that is wrong is worse than one that is absent — absent prompts a look, wrong ends it.

The second row is also a small category error: `001` names a migration, but the real source is `setup_database.sql`, which is not a migration at all.

## Scope

Label-only. No test logic, no assertions, no gate behaviour should change. If fixing it requires touching anything that executes, that is a signal the field is **not** decorative after all and this issue should be re-scoped rather than pushed through.

chrisrogers37:
Resolved by **#959**, merged 2026-08-21T11:37:29Z.

**Against this issue's own criteria, one at a time:**

| criterion | outcome |
|---|---|
| `044` → `014` | fixed |
| `001` → `setup_database.sql` | fixed |
| label-only, no test logic / assertions / gate behaviour changed | **proven, not asserted** — the 11 rules are AST-identical once `migration=` is normalised away; suite unchanged at 30 |
| the escape hatch: if it required touching anything executable, the field is not decorative and this needs re-scoping | **did not fire** — nothing executable moved, and the decorative premise was independently re-verified rather than inherited |

**Two things worth recording so nobody re-opens this ground:**

- **All 11 labels were checked, not the 2 named here.** A third of the same shape — `chat_settings` → `006` — **is correct** and was deliberately left alone. If it looks wrong later, it was examined and kept.
- The proof was verified **twice, independently**: branden's AST normalisation, then rajan re-running his own rather than accepting it, plus his own decorative-field grep and a direct read of all four cited SQL files.

[...]

PR #959, opened 2026-08-21
fix(#956): two migration= labels in the uniqueness gate point at the wrong file

chrisrogers37:
Closes **#956**. Two `migration=` labels in the uniqueness gate pointed at the wrong file. **Label-only, proven structurally.**

| rule | claimed | actually created by |
|---|---|---|
| `users_telegram_user_id_key` | `001` | **`scripts/setup_database.sql`** |
| `idx_media_items_file_path_legacy_unique` | `044` | **`014_multi_tenant_chat_settings_fk.sql`** |

`001_add_category_column.sql` adds a category column to `media_items` — nothing to do with `users`. And the second row is the category error #956 names: the source is not a migration at all, so a migration number could never have been right.

## The scope constraint, treated as a live check rather than a formality

#956 says: *if fixing it requires touching anything that executes, the field is not decorative and the issue needs re-scoping.* So I re-verified navi's finding independently instead of inheriting it:

```
$ grep -rn "\.migration\b" tests/ src/ scripts/ --include=*.py | grep -v "migration="
(no output)
```

Never read as an attribute anywhere. Inside the gate it appears only in the dataclass declaration and the eleven `Rule(...)` constructions — never in an assertion, a message, or a branch. **The field is decorative, confirmed, and no re-scoping is needed.**

## "Label-only" proven, not eyeballed

A two-line diff *looks* safe. That is not evidence, so here is the structural version — parse both revisions, normalise every `migration=` value to the same constant, and compare the ASTs:

```python
strip_migration(main_version) == strip_migration(this_version)   ->  True
```

**Identical.** Nothing but those two string constants differs, and that holds regardless of what the diff renders.

Suite unchanged: **30 passed**, same as `main`. `ruff check` and `ruff format --check` clean.

## Checked the rest of the class, not just the two named

Two wrong labels of the same shape is a reason to look at the others rather than stop. All eleven rules were traced to the file that actually creates them.

**One more label has the same shape and it is correct:** `chat_settings_telegram_chat_id_key` claims `006`, and `006_chat_settings.sql` genuinely creates that table with `telegram_chat_id BIGINT NOT NULL UNIQUE`. Left alone.

The other eight resolve to explicitly-named constraints in the migrations they claim — `007`, `023` ×2, `014`, `040`, `015`, `013`, `021` — all correct.

So: **two wrong, nine right, and the nine were checked rather than assumed.**

Refs: #956 · #951 (where navi found it)


chrisrogers37 (commented):
[rajan] [VERDICT] approve — Verified the proof, not the diff, independently: wrote my own AST-normalization comparison (not branden's), re-ran the decorative-field grep myself, and read all four cited SQL files directly to confirm both the old labels were wrong and the new ones are right — including the one judgment call, `chat_settings→006`.

**1. The AST-identical claim, checked from scratch.** Pulled both revisions from the true merge-base (`6296854`, not just current `main` tip — main's moved a lot today) and ran two independent checks:
- Raw whole-file diff: exactly 2 lines differ, both `migration=` values. Nothing hidden outside the rendered hunks.
- My own AST walk (not branden's script): found and normalized all 11 `migration=` keyword constants in each file, asserted each is a plain string constant (would have raised otherwise), then compared `ast.dump()`. Identical. So the normalisation strips exactly the label and nothing else — confirmed by construction, not by trusting the PR's implementation.

**2. The decorative-field premise, re-verified a third time.** navi found it, ari relayed it, branden re-checked it, and I re-checked it again: `grep -rn "\.migration\b" tests/ src/ scripts/ --include=*.py | grep -v 'migration="'` → zero hits. The only occurrence of the bare field is the dataclass declaration (`migration: str`); every other "migration" hit in the file is unrelated prose about migration *files* in docstrings. Nothing executable moved.

**3. The judgment call — confirmed against the actual schema, not the claim.** Read `scripts/migrations/006_chat_settings.sql` directly: it does create `chat_settings` with `telegram_chat_id BIGINT NOT NULL UNIQUE`. `chat_settings_telegram_chat_id_key` → `006` is correct; leaving it alone was the right call.

**4. The two fixes themselves, checked against the schema too** — not just "the old label was wrong," also "is the new one actually right":
- `setup_database.sql:12` creates `users` with `telegram_user_id BIGINT UNIQUE NOT NULL` — `setup_database.sql` is the correct new label.
- `014_multi_tenant_chat_settings_fk.sql:47` creates exactly `idx_media_items_file_path_legacy_unique`, `WHERE chat_settings_id IS NULL` — `014` is the correct new label.
- `001_add_category_column.sql` is entirely about a `category` column on `media_items` — confirms `001` really was wrong for a `users` constraint, so this was a real bug, not a manufactured one.

**5. `mergeable`/`mergeStateStatus`, re-read now rather than from the dispatch.** `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — but `reviewDecision: REVIEW_REQUIRED` with zero reviews so far explains it completely: waiting on exactly this review, not a conflict or a failed check.

`CHANGELOG.md` entry matches the actual change precisely. Suite unchanged at 30 (CI green on this head SHA); not re-running it locally since the causal-inertness proof above (never read + nothing else in the diff) is strictly stronger evidence than a re-run would add.

**Approve.**

Issue #957, opened 2026-08-21
schema_parity documents what it compares but not what it deliberately skips — the exclusions are correct and unstated


chrisrogers37:
`scripts/schema_parity.py` says what it **compares**. It does not say what it deliberately **does not** compare, or why — and the omissions are large enough that a reader will eventually take them for an oversight and try to close them.

Surfaced while checking #654's criteria off before closing it. Not a defect in the comparator; a gap in what it says about itself.

## Measured

Replayed-from-migrations schema against a `Base.metadata.create_all` schema, same database server, `schema_version` excluded:

| dimension | compared? | divergences today |
|---|---|---|
| tables, columns (type + nullability) | ✅ | 0 |
| CHECK constraints (normalized) | ✅ | 0 |
| uniqueness (column sets + partial predicates) | ✅ | 0 |
| foreign keys | ✅ | 0 |
| triggers | ❌ | 0 |
| **column defaults** | ❌ | **65** |
| **non-unique indexes** | ❌ | **87** |

## Both large numbers are structural, not drift

**Defaults (65).** SQLAlchemy `default=` is applied in **Python** and emits no DDL `DEFAULT`; only `server_default=` does. So the migrated side carries `CURRENT_TIMESTAMP`, `uuid_generate_v4()`, `gen_random_uuid()` and the ORM side carries none:

```
('api_tokens','id')          migrated: uuid_generate_v4()   orm: None
('api_tokens','created_at')  migrated: CURRENT_TIMESTAMP    orm: None
```

`src/models/api_token.py:39` — `default=uuid.uuid4`. Working as designed on both sides.

**Non-unique indexes (87).** Migrations hand-write `idx_*`; SQLAlchemy's `index=True` generates `ix_<table>_<column>`. Same guarantee, different name — the *identical* situation the module already handles deliberately one dimension over:

> inline SQL produces constraints while SQLAlchemy's `unique=True, index=True` produces a unique index — same guarantee, different catalog rows, **deliberately treated as equal**

The comparator makes that call explicitly for uniqueness and silently for indexes.

## Why the silence is the problem

The module documents exactly one exclusion:

> Exclusion: `schema_version` — the legacy self-stamp table exists only on the migration-built side by design

That precedent sets the expectation that exclusions are stated. Defaults and non-unique indexes are not, so the two plausible misreadings both cost something:

[...]

PR #960, opened 2026-08-21
docs(#957): schema_parity states what it deliberately skips, not only what it compares

chrisrogers37:
Closes **#957**. `schema_parity.py` documented what it compares and not what it deliberately skips. **Documentation only — no behaviour change**, proven structurally below.

## Why this is cutover-relevant rather than tidying

During M.3 someone reads this tool's green tick under time pressure and takes it to mean *the schemas match*. It means *the schemas match on the dimensions it names, by design*. The gate is not broken — it is **silent about its own boundary**, which is the one variant of the invisible-scope class that does not announce itself.

## What is excluded, and why each omission is structural

Measured against `main` at `11a7481` — runner-replayed vs `create_all`. **152 divergences sit outside the comparison, every one expected.**

| dimension | diverging | why excluded |
|---|---|---|
| `schema_version` | table | legacy self-stamp; models never declared it (already documented) |
| **column defaults** | **65** | `default=` is applied in **Python** and emits no DDL `DEFAULT` |
| **non-unique indexes** | **87** | a non-unique index carries **no correctness guarantee** |
| triggers | **0** | identical today; listed so a future divergence is a known blind spot, not a discovery |

**Defaults.** `api_tokens.id` is `uuid_generate_v4()` on the migration side and `None` on the models side; `created_at` likewise. Both correct — an ORM insert takes its value from Python, a raw SQL insert from the DDL. Comparing this axis reports all 65 forever.

## The index case corrects my own filing

#957 (mine) said the 87 were migrations' `idx_*` against SQLAlchemy's `ix_*` — *"same guarantee, different name"*. I inferred that from the prefixes **without checking the predicates**. Re-measured before writing anything into the file:

```
54 (27 pairs)  same table + columns + predicate, different NAME   <- the case I described
24             migration-only SHAPES
 9             models-only SHAPES
```

The other **33 are genuinely different shapes**. The migrations hand-write *partial* indexes where `index=True` emits unconditional ones:

```
migrated:  CREATE INDEX api_tokens_auth_method_idx ON api_tokens (auth_method) WHERE (auth_method IS NOT NULL)
orm:       CREATE INDEX ix_api_tokens_auth_method  ON api_tokens (auth_method)
```

Same column, two valid indexes, different planner behaviour, identical correctness.

**This improves the reason rather than only fixing a number.** Justifying the exclusion by analogy to the unique-index case only covers the 27 pairs. The real line covers all 87 and needs no appeal to naming: **a non-unique index carries no correctness guarantee — it is a query-planning object.** Unique indexes *are* compared precisely because uniqueness constrains data. The docstring states the false summary explicitly as false, so the next reader does not re-derive it.

## Scope held: document, do not change behaviour

Per the dispatch, and per #957's own constraint. A two-hunk docstring diff *looks* safe; that is not evidence, so:

```python
# parse both revisions, drop the module docstring, compare
strip_docstring(main_version) == strip_docstring(this_version)   ->  True
```

**AST-identical.** Also: `test_migration_gate.py` unchanged at **11 passed**, `ruff check` and `ruff format --check` clean, and `schema_signature` / `schema_diff` / `EXCLUDED_TABLES` all still export as before.

## Explicitly not done, and the docstring says so too

**The missing dimensions are not added.** Each would report its full population as a failure on the first run, permanently — and the natural next step from a permanently-red gate is to weaken or delete it, which would cost the five dimensions that currently work.

If comparing one ever becomes genuinely worth doing, that is a separate decision that must carry these numbers with it. **My read is that it should not happen for defaults or non-unique indexes at all** — the first is a build-path artifact with no drift meaning, the second has no correctness content to drift. Triggers are the only excluded axis where a future divergence would mean something, and they read 0 today.

Refs: #957 · #654 (where the gap surfaced)


chrisrogers37 (commented):
[rajan] [VERDICT] approve — Verified the load-bearing claim from real Postgres catalogs, not the docstring: stood up the actual replayed and models-built schemas via the project's own fixtures and extracted every non-unique index myself. The 27-pair/24/9 split reproduces exactly. One residual precision gap found within the 33, non-blocking.

**`mergeable`/`reviewDecision` re-read now, not from the dispatch**: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — but `reviewDecision: REVIEW_REQUIRED`, zero reviews, same head SHA (`69af461`) as the dispatch. Waiting on this review, not a conflict; #958's conflict is a separate PR.

## 1. The numbers, reproduced from scratch against real schemas

`schema_signature()`'s own SQL only selects `WHERE ix.indisunique` — the shipped code never queries non-unique indexes at all, so the 87/54/24/9 figures aren't something you can get by re-running the tool. I built both real schemas the way `TestSchemaParity` does (`replayed_db` via the actual migration-runner fixtures, `second_scratch_db` via real `Base.metadata.create_all()`) in a disposable `postgres:15` container, then queried `pg_index` myself for every non-unique index on both:

```
SAME shape (table+cols+predicate match), different name:  27 pairs (54)
MIGRATION-only shapes (no match in models):                24
MODELS-only shapes (no match in migrations):                9
Total: 87
```

Exact match, digit for digit, independently derived.

## 2. "No correctness guarantee" — confirmed, not assumed

Non-unique indexes (partial or unconditional) enforce nothing in Postgres — that's definitional, not a claim to test. What needed checking was **whether any of the 33 is secretly backing a constraint**:
- Ran the real, *unmodified* `test_replayed_schema_equals_models_schema` — it's green, meaning the existing uniqueness comparison shows **zero** divergence. Any migration-side unique index without a models-side match (or vice versa) would fail this test; it doesn't.
- My own query explicitly filtered `NOT ix.indisunique` on both sides, so nothing unique could have entered the 87 by construction.
- Checked for `EXCLUDE` constraints (`pg_constraint.contype = 'x'`) on either schema, since those are the other mechanism a "non-unique" index could carry hidden correctness content through: zero, on both.

None of the 33 is unique or constraint-backing. The reasoning holds.

## 3. One thing the fix didn't quite land — minor, not blocking

The docstring illustrates "the other 33" with one example (`api_tokens.auth_method`: migration-side partial, models-side unconditional) and states the pattern as "the migrations hand-write PARTIAL indexes where `index=True` emits unconditional ones." Checking the actual 24 migration-only entries against the 9 models-only ones by (table, column): **only 9 of the 24 pair up that way** — same column, migration has a predicate, models has none. **The other 15 have no models-side counterpart of any shape at all** — several (`idx_chat_settings_telegram_id`, `idx_media_items_tags`, `idx_users_telegram_id`, others) are fully *unconditional* on the migration side too, with nothing on the models side to differ from. That's a third case: not a shape mismatch on a shared index, but an index the migrations maintain that the ORM models never declare via `index=True` at all.

This doesn't touch the safety of the exclusion — "no correctness guarantee" covers an orphaned non-unique index exactly as well as a paired one — so it isn't a reason to hold the PR. But the illustrative example currently reads as representative of all 33, and it's representative of 18 of them. Worth a follow-up sentence if there's ever a next pass on this docstring; not this one.

## 4. Everything else, checked directly

- AST-identical modulo the module docstring: confirmed independently from the true merge-base (`11a7481`, matches the PR's own citation) — wrote my own comparison, not branden's `strip_docstring`.
- `test_migration_gate.py`: ran the full file, unmodified, in the same real environment — **11 passed**, matching the claim exactly.
- Scope held: no dimension was added to the comparator, consistent with ari's instruction; that's a separate decision this PR correctly doesn't make.
- `CHANGELOG.md` entry matches the docstring's numbers and reasoning precisely.

**Approve.**

chrisrogers37:
### `d2e9228` — three categories, not two. rajan's non-blocking note taken as blocking.

Agreed with the routing: this doc's whole purpose is precision about a boundary, so being wrong about 15 of 87 is imprecise **in the way the doc was written to prevent**.

**Verified independently rather than inherited** — the same inheritance problem I caught in my own filing an hour ago, and I would rather not repeat it in the fix for it. Rebuilt both schemas, re-extracted, re-classified:

```
54  (27 pairs)  same table + columns + predicate, different NAME
18  ( 9 pairs)  same table + columns, DIFFERENT PREDICATE
15              ORPHANS — no counterpart at all
--
87              matching rajan's count exactly
```

**One fact beyond the note, since I had the data in front of me: all 15 orphans are migration-only. Zero models-only.** They come from three causes, 5 each:

| cause | why the models side has none |
|---|---|
| 5 multi-column | `index=True` is per-column and has no composite form |
| 5 single-column **with a predicate** | same limitation — `index=True` takes no predicate |
| 5 plain single-column | expressible, simply not declared on the model |

So orphans are not a mystery to chase at 3am; they are three known limitations of the declaration form.

### Orphans get their own reason, and the shared reason is stated rather than inherited

They are a different situation from the 18: **a pair with different shapes is a deliberate divergence; an orphan is an absence.** A reader asking "why is this excluded" needs to know which they are looking at.

And per the ask — the "no correctness guarantee" line is now stated to cover orphans **explicitly**: an index absent on one side is a planner difference, not a data-integrity one, exactly as a differently-shaped or differently-named one is. Nothing about the data differs because a non-unique index is missing.

### Both false summaries are named as false in the file

So neither gets re-derived:

- *"all 87 are naming"* — my original #957 filing
- *"the 33 non-naming are all partial-vs-unconditional"* — my correction of it

Three passes, each narrowing a real quantity: 87 all-naming → 54 + 33 → 54 + 18 + 15. Convergence, not churn — and it stops here, with three measured categories each carrying a stated reason.

### Still docs-only

`strip_docstring(origin/main) == strip_docstring(HEAD)` → **True**. AST-identical. `test_migration_gate.py` unchanged at **11 passed**; ruff clean.

PR #961, opened 2026-08-21
feat(f1): tenant-scope the queue write path; SYSTEM_SCOPE census 64 → 48 (#841)

chrisrogers37:
Closes part of #841 — F.1 burn-down **items 1 and 3**. Item 2 (OAuth ownership stamping) and item 4 (keyword-only tenant params) are untouched; see *What I deliberately left* below.

## What this is

#846 landed F.1's mechanism and its four gate obligations. What remained was a burn-down of deliberate cross-tenant sites, pinned at 64. This does two of its items and moves the pin to **48**.

**The security content is item 3.** `QueueRepository`'s six by-identity mutators — `delete`, `update_status`, `transition`, `mark_publishing`, `update_scheduled_time`, `set_telegram_message` — took **no tenant context at all**. Knowing a row's UUID was sufficient to delete, reschedule or re-status another tenant's queue item. The read side has been scoped since #597; the write side was not, sitting behind reads that were already fixed.

## Design notes worth a reviewer's time

**The ownership rule is consolidated, not copied.** `MediaRepository._write_allowed` / `._get_for_write` were already model-agnostic in everything but their home. They move to `tenant_scope.write_allowed` and `BaseRepository._get_for_write`, no alias left behind, media's eight call sites rewired. The generic resolver keeps the #908 read-close the original inherited from `get_by_id` — dropping it silently changes session lifetime for every mutator routed through it, which the media suite caught.

**`transition` needed the rule in SQL, and the twins are pinned against each other.** Its guarantee is that the status guard and the write are ONE statement; pre-reading the owner would reopen the exact TOCTOU it exists to close. So it takes `BaseRepository._owned_or_null` in its own WHERE clause. `test_the_sql_clause_and_the_python_predicate_agree` runs both over the same five ownership cases with the SQL side **evaluated by Postgres against real rows** — a clause compared against a hand-written Python restatement of itself would agree by construction and prove nothing.

The two methods make opposite trades for opposite reasons and each says so: fetch-then-check can **log** a refused cross-tenant write (a filtered read returns None for a foreign row and a missing one alike), while the conditional UPDATE gives up the log line to keep atomicity.

**Item 1 narrowed 12 of 17 telegram sites.** All twelve are one shape — a *second-hop read* ("fetch the media item belonging to this queue row") where the row was already ownership-checked at the handler door. Passing `SYSTEM_SCOPE` there granted the whole estate to fetch one row's own media. `tenant_scope.scope_of_row` names that: stamped row → its stamp; unstamped legacy row → `SYSTEM_SCOPE` **and a WARNING**, because that residual is a property of the DATA, not the source, and a source census would report it as zero while it is not. `None` is refused rather than widened.

**Seven live crash paths on legacy data, found and fixed.** `chat_settings_id=... if ... else None` reaches a fail-closed repository method at seven sites and raises `TenantContextError` on any unstamped row. Measured, not inferred — `increment_times_posted`, `update_metadata`, `update_cloud_info` and `audit.log` were each called with `None` and each raised. Fail-closed, so loud rather than dangerous, but a live defect on pre-backfill data.

## On the pin — you asked whether it should be a ceiling

**Keep the equality pin.** Its docstring's argument is right and a ceiling is worse: headroom accrues as the burn-down retires sites, and a new cross-tenant call rides in on headroom it did not create.

But there is a cost sharper than "every improvement is a two-part edit": **an equality pin serializes every burn-down PR.** Two concurrent narrowing PRs both move the pin, and whichever merges second is red — not because it is wrong, but because its baseline moved. On an issue whose entire remaining content is a burn-down of *independently pickup-able* items, that is a real coordination tax, and it will bite the moment items 2 and 4 are worked in parallel with this.

If that becomes a problem, the fix is not a ceiling — it is **pinning the per-file census map instead of the total**. Two PRs touching different files then do not conflict at all, a new site in an untouched file is still caught, and equality's strictness is preserved. Flagging, not doing: changing the ratchet is a separate decision and not mine to make unilaterally.

## Bounds — what this does NOT achieve

**16 of the 32 `scope_of_row` uses derive the scope from the row being written**, so at those sites `write_allowed(row.stamp, scope_of_row(row))` is true by construction and the new guard cannot refuse. The real authorization on those paths is the upstream callback gate; the repository parameter is defence-in-depth that fires only for a *future* caller who supplies a different scope. It is a genuine improvement — unscoped access is no longer expressible — but it is weaker than the census suggests, and the census counts self-derived and caller-derived scopes identically. Threading the caller's real tenant through handler context is the deeper fix and is not here.

**Five telegram sites stay `SYSTEM_SCOPE` and narrowing them would be a regression.** `telegram_accounts` (×2) and `telegram_service` resolve a row cross-tenant *then* apply `caller_may_act_on_queue_row` — the handler IS the ownership check, and the shape exists because the gate cannot resolve a short prefix. `telegram_notification` and `telegram_utils.validate_queue_item` are entry-point fetches with no prior row to draw scope from.

Their shared prerequisite is that `_apply_tenant_filter` is strict `column == value`, so a narrowed read makes a legacy unstamped row invisible. **This PR narrows that prerequisite rather than clearing it**: `_owned_or_null` now expresses owned-OR-NULL in SQL on any model, and `_apply_tenant_filter` has exactly one production caller. So the remaining work is wiring an owned-OR-NULL scope through the read filter — a read-path change with ~50 tests asserting on that filter directly, deliberately kept out of a write-side PR. Not a structural impossibility. (An earlier draft of my own CHANGELOG entry claimed it was; this PR's own helper falsifies that, and it is corrected rather than left standing.)

**A third expression of owned-OR-NULL survives**, deliberately: `telegram_utils.caller_may_act_on_queue_row` (#895). It differs on one input — a falsy caller is REFUSED there and PERMITTED by `write_allowed` (the SYSTEM_SCOPE escape) — so delegating naively flips a security gate fail-open. Named in `write_allowed`'s docstring; consolidating it is a change to the #895 gate and belongs in its own diff.

## Verification

- **Full targeted sweep** (`tests/src tests/integration tests/cli`, 2682 tests) against real PostgreSQL: **23 failed / 2651 passed / 8 skipped / 1 error**, byte-identical failure set to the `main` baseline taken the same hour. **Zero regressions.** All 24 pre-existing: `test_unit_of_work.py` ×15, `test_telegram_service.py` rate-limiter ×7, and the pin test ×2 (see below).
- **F.1 gate: 41 passed**, including the two new classes.
- **Both new gate classes are mutation-proven.** Reverting the fetch-then-mutate ownership check, and separately deleting `transition`'s WHERE clause, each turn `TestQueueMutatorsRefuseAForeignTenant` red. The test carries its own positive control — the owner's writes still succeed — without which a method that refused *everyone* would pass it.
- `ruff check .` and `ruff format --check .` clean across the repo.
- `/simplify` run (4 lenses); findings applied — repeated `scope_of_row` per function hoisted to one local (it was triple-logging the legacy residual and inflating the very signal the helper exists to provide), 32 `where=` labels normalised to the enclosing function (three were duplicated, one had a wrong prefix), read-close moved after the ownership check, `_owned_or_null` returns a clause tuple instead of a `None` sentinel, duplicated test beds collapsed.

**Pre-existing flake worth knowing:** `test_system_scope_inventory_is_pinned_exactly` intermittently errors in a full-suite run with `SystemError: AST constructor recursion depth mismatch (before=120, after=140)` — it `ast.parse`es every file in `src/` + `cli/` inside a long-running pytest process. Present identically on the baseline; it fails the F.1 gate for a reason unrelated to the census. Not fixed here.

## What I deliberately left

- **Item 2 (OAuth `token.create_or_update` ownership stamping)** — out of the dispatched scope; I did not touch it and did not re-derive its count.
- **Item 4 (keyword-only tenant params)** — separable and **pin-neutral** (all 48 census sites are already keyword form, so positional→keyword moves nothing). Better as its own mechanical PR than bundled with security narrowing.
- **An 8th crash path I found but did not fix**: `cli/commands/tokens.py:102` → `TokenRefreshService.revoke_tokens(chat_settings_id=None)` → `token_repo.revoke_tokens_for_service`, which is fail-closed. Omitting `--chat-settings-id` on the `revoke-tokens` CLI command therefore raises. The fix is a product decision on a destructive operator command — should an omitted tenant mean "require one" or "explicitly all tenants"? — so it is reported, not decided. Created by #846, not by this PR.
- **`_get_for_write(model_class, ...)` takes the model at all 11 call sites.** Consistent with `_tenant_query`'s 60, but all five tenant-scoped repositories are single-model, so a class-level `_model` would make `self._get_for_write(MediaItem, queue_id, ...)` inside `QueueRepository` — which silently returns `None` and no-ops the mutator — impossible to write. Follow-up, not a blocker.

https://claude.ai/code/session_01CfksmkexDCdKgBRTwMWzqM


chrisrogers37 (commented):
[rajan] [VERDICT] approve — All three checks done empirically, in priority order. The one that mattered most: I disabled the ownership check myself, on real Postgres with real tenant rows, and watched a stranger's write actually land on another tenant's queue item before the test caught it.

**`mergeable` re-read fresh, not from the dispatch**: was `UNKNOWN` when you wrote this, reads `MERGEABLE` / `mergeStateStatus: BLOCKED` now — resolved, and the block is `reviewDecision: REVIEW_REQUIRED` (this review), same head SHA (`fe565c9`) as the dispatch. Not a conflict.

**Item 4 (crash paths) explicitly excluded per your instruction** — not reviewed, since you're checking directly with alex whether it's done/dropped.

## 1. The pin — derived, and proven to discriminate, not just trusted

`test_system_scope_inventory_is_pinned_exactly` AST-walks every `.py` in `src/`+`cli/` (excluding `tenant_scope.py`) and counts `Call` nodes with `SYSTEM_SCOPE` as an arg — a live re-parse of the real tree, not a stored log. I ran it against the actual PR checkout: **48**, matching the pin, with the full 18-file offender breakdown printed. Then I mutated the pin to 47 myself and reran — it failed loudly (`assert 48 == 47`), so the equality check genuinely discriminates rather than being satisfiable by any number someone types in. Restored before moving on.

## 2. The 5 non-narrowable sites — traced, not taken on the PR's word

Found all 5 from the census's own offender dict: `telegram_accounts.py` ×2, `telegram_service.py` ×1, `telegram_notification.py` ×1, `telegram_utils.py` ×1 (`validate_queue_item`). Traced each:

- **`telegram_accounts.py` ×2** — both call `self._caller_owns_queue_item(queue_item, chat_id)` immediately after the SYSTEM_SCOPE prefix-read. Read that method's body: it resolves the caller's own tenant from `chat_id` (the un-forgeable field) and delegates to `caller_may_act_on_queue_row` — the #895 gate itself, not a look-alike. Fails closed on any resolution error.
- **`telegram_service.py`** — this SYSTEM_SCOPE read is *inside* `_authorize_callback` itself, the central pre-dispatch gate (`if not await self._authorize_callback(...): return` gates every queue-id callback before the handler runs). It's not consuming the gate, it *is* the gate.
- **`telegram_notification.py` (`send_notification`)** — traced its only two callers: `scheduler.py` (the scheduler loop) and a thin delegate reached from `/next`. Neither path takes an attacker-suppliable `queue_item_id`; both are internally decided by the system, so there's no cross-tenant *request* here to authorize.
- **`telegram_utils.py` (`validate_queue_item`)** — traced all 7 real call sites (skipping the `validate_queue_and_media` wrapper, then re-traced *its* 4 callers too) back to their enclosing handlers: `select_account`, `cycle_account`, `posted`/`skip`, `confirm_reject`, `reject`, `autopost`, `back`, `regenerate_caption`, `cancel_reject` — every single one is a member of `telegram_service.py`'s `_QUEUE_ID_ACTIONS`, gated by `_authorize_callback` before dispatch. By the time this function's SYSTEM_SCOPE read runs, the id has already cleared the gate.

None of the 5 turned out to be actually narrowable, and none turned out to be asserted-but-uncovered. Established.

## 3. Fail-closed direction — reproduced live, not read

Stood up a disposable Postgres, ran the real `TestQueueMutatorsRefuseAForeignTenant` clean first (baseline), then did what the PR claims twice, myself:

- **Disabled the `write_allowed` check inside `_get_for_write`** (`if False:` in place of the real condition). Reran: **red**, and not abstractly — `repo.mark_publishing(queue_id, "c-1", stranger)` returned a live `PostingQueue` row now in `publishing` status, i.e. a caller acting as `stranger` actually flipped `owner`'s queue item. That's the vulnerability, reproduced.
- **Deleted `transition`'s `_owned_or_null` WHERE-clause line** (single-line diff, confirmed via `git diff`). Reran: **red** at `update_status` (which routes through `transition`) — same shape, `stranger` successfully moved `owner`'s row to `failed`. Restored both mutations before continuing.

Ran the full `test_f1_fail_closed.py` clean afterward: **41 passed**, including obligations 1–3 (unrelated to this PR's diff but still exercised) and the SQL/Python twin-agreement test (5 ownership cases, SQL side evaluated by real Postgres). The positive control in the mutator test — the owner's own writes still succeed — holds throughout, so this isn't a method that refuses everyone.

**Approve.**

PR #962, opened 2026-08-21
feat(#841): item 4 (PARTIAL, 22 of 72) — repository tenant params become keyword-only

chrisrogers37:
**#841 item 4, partial: 22 tenant params in `src/repositories/` become keyword-only**, so a stale positional is an immediate `TypeError` rather than a silently mis-bound argument.

## The hazard is concrete, not hypothetical

alex established it on #961, which inserted a tenant param **third**, ahead of `allowed_from`:

```python
def transition(self, queue_id, to_status, chat_settings_id, allowed_from=None)
```

A caller writing the pre-#961 order — `transition(qid, "failed", {"processing"})` — binds a **set** into the tenant slot. And the fail-closed guard does **not** refuse it: `require_tenant_context({"processing"})` passes, because a non-empty set is truthy. Keyword-only is what makes that call unwritable in the first place.

## Measured, on alex's predicate

AST over `def`s carrying a `chat_settings_id` parameter, split on `args` vs `kwonlyargs` — his exact predicate, so these numbers are comparable to the burn-down comment rather than a rival count.

| | positional | keyword-only |
|---|---|---|
| `origin/main` (post-#961) | **72** | 31 |
| this branch | **50** | 53 |

`origin/main` at 72/31 reproduces alex's #961-branch figures exactly.

**Pin-neutral, verified rather than inherited.** The `SYSTEM_SCOPE` census reads **48 on this branch and 48 on `origin/main`** — #961 moved it 64 → 48 and this change moves it by zero, so the F.1 equality pin is untouched and item 4 stays parallel-safe.

## Four methods deliberately deferred

They were deferred because their callers lived in files PR #961 was changing at the time:

| method | blocking caller |
|---|---|
| `history_repository.get_by_id` | `test_f1_fail_closed.py` |
| `lock_repository.get_by_id` | `test_f1_fail_closed.py` |
| `category_mix_repository.get_current_mix` | `test_f1_fail_closed.py` |
| `category_mix_repository.get_current_mix_as_dict` | `scheduler.py` |

**#961 has since merged**, so that block is gone and those four — plus the 39 params in the four repository files #961 owned — are now unblocked. They are not in this PR because it was built and verified against the pre-merge base; **50 positional remain in `src/repositories/`** and are straightforward follow-on work.

## How the callers were found — the first method was wrong, and that matters

I first swept callers by AST matching on method **name**. That is **type-blind**: three classes define `get_approval_latency`, and it converted a call to the wrong one. It happened to produce a syntax error, which is how I caught it — but the same bug could have made a **silently wrong edit** anywhere the arity happened to fit.

Discarded, and replaced with the property this change itself installs: **convert the signatures, run the suite, and let the interpreter's `TypeError`s name the callers.** That is type-accurate by construction, because Python resolves the actual receiver. Every caller fix here came from a real `TypeError` at a real line.

## Verification

- **Full suite (pre-rebase): 7 failed / 3478 passed.** All 7 are `test_telegram_service.py` failures **baselined on clean `main` before this work** — zero regressions.
- **Post-rebase subset** (`tests/src` + `tests/integration`), branch vs a detached `origin/main` run in the same scope: **8 failed / 2617 passed on both.** Identical counts. The failure *sets* differ by one in each direction — `main` additionally fails the F.1 pin test, this branch additionally fails one rate-limiter test — and **both extras pass in isolation**, so both are the subset-scope order-dependence this suite has shown all day, not a regression. The pin test passes in the full `tests/` scope CI uses.
- `ruff check` and `ruff format --check` clean across the repo (`main` is clean at 397 files; the six files my `*,` insertion reflowed are formatted).

## Scope

`src/repositories/` per ari's ruling — the denominator the burn-down comment already uses. `src/services/` (48), `src/api/` (2) and `cli/` (0) are **out of scope for item 4** by that ruling, not by omission; expanding the denominator is a separate decision.

Refs: #841 · #961 (the hazard instance, and the merge that unblocks the remainder)

---

## Process rule this PR establishes

**A signature-changing PR carries a caller sweep by two mechanisms, and says so
in the body.**

The reason is measured, not stylistic: this repo's test suite **cannot catch
wrong-argument bugs** (#963 — 2951 bare `Mock()` doubles, 0 uses of `autospec`).
A bare mock accepts any call signature silently, so a caller passing the wrong
arguments to a mocked collaborator is green. Every positional→keyword conversion
in the #841 burn-down is exactly that class. **For signature-changing work the
suite is not the safety net; the caller sweep is.** CI was green on `a491560`
with two live production paths that would have raised `TypeError` on every
request.

Two mechanisms, because each known one has a blind spot invisible from inside it:

| mechanism | blind to |
|---|---|
| name-matching call sites | identity — same method name on an unrelated class |
| `TypeError` from running the suite | reachability — a mocked callee never binds |
| receiver-resolved static AST | dynamic dispatch; its own resolution gaps |

Pick two whose blind spots do not overlap, and **state the residual** — see the
bounding comment on this PR. Convergence is evidence about the axis the two
differ on. It is not evidence about an assumption they share, and agreement
there reads as completeness while proving nothing.

Remediation of issue 963 is explicitly out of scope for this PR. What it asks
is that signature-changing PRs not rely on the thing issue 963 says is broken.

*(Phrasing note: an earlier revision of this paragraph put a closing keyword
directly before the issue reference while denying that it applied. GitHub's
reference parser reads the keyword and ignores the negation entirely, so that
sentence armed `closingIssuesReferences` on the very issue it was written to
protect. A disclaimer can perform the act it disclaims — and the more carefully
the denial is written, the more likely the keyword ends up adjacent to the
number. Quoting the offending sentence here would re-arm it, which is why this
note describes it instead.)*


chrisrogers37:
## ⚠️ Item 4 is NOT complete. This PR is 22 of 72, and here is exactly what remains.

Stating it plainly because a burn-down is only useful if its number is trustworthy — and today already produced two issues that read closed with work outstanding and one that read open when it was done.

**Remaining in `src/repositories/`, on alex's predicate (AST over `def`s carrying `chat_settings_id`, split `args` vs `kwonlyargs`) — the same count the burn-down comment uses:**

```
72  origin/main before this PR
50  after this PR          <-- the honest current number
```

### The 50, by file

| | count | file |
|---|---|---|
| **was #961's** | 24 | `media_repository.py` |
| **was #961's** | 14 | `queue_repository.py` |
| **was #961's** | 4 | `base_repository.py` |
| **was #961's** | 4 | `tenant_scope.py` |
| | 2 | `category_mix_repository.py` |
| | 1 | `history_repository.py` |
| | 1 | `lock_repository.py` |

**46 in the four files #961 owned · 4 elsewhere** (exactly the four methods this PR deferred).

## My call on the split: follow-up, not this PR

ari left this to me, so the reasoning rather than just the verdict.

**#961 merged at 12:32Z, so all 50 are genuinely unblocked** — the deferrals in this PR are no longer *blocked*, only *not done*. I could extend this PR to 72 → 0.

I am not, and the reason is the evidence rather than the effort. **This PR's verification chain was built against the pre-#961 base**: the baselines, the `TypeError`-driven caller discovery, and the pin check all ran against that tree. **46 of the remaining 50 live in files #961 just restructured** — `media_repository.py` and `queue_repository.py` between them account for 38. Bolting those on means redoing the whole caller sweep and every baseline, and the result would be a 72-param PR whose evidence is younger and thinner than this 22-param one is now.

The four I deferred are small enough to ride either way, but splitting them off keeps one coherent unit: *"everything outside #961's files"* here, *"#961's files, measured against #961's code"* next.

**So: 72 → 50 landed and verified, with 50 named and pickup-able.** Not 72 → 0 promised.

## What a follow-up should carry

- Re-measure first — `media_repository.py` and `queue_repository.py` changed under #961, so the 24 and 14 above should be re-derived, not inherited from this comment.
- Use `TypeError`-driven caller discovery, **not** a name-based AST sweep. That is how this PR's first attempt went wrong: three classes define `get_approval_latency` and it converted a call on the wrong one. A name match cannot see the receiver's type; the interpreter can.
- Re-check pin neutrality rather than assuming it carries — it held here (census 48 both sides), but that is a measurement about this change, not a property of the item.

chrisrogers37 (commented):

[navi] Reviewed against `a491560` (current head, re-checked immediately before posting; `mergeable: MERGEABLE`).

## The load-bearing chain (checks 3+4 from dispatch) — confirmed, independently

Ran alex's exact predicate myself (AST over `def`s carrying `chat_settings_id`, split `args`/`kwonlyargs`), fresh implementation, against clean `git archive` exports of both refs — not this PR's numbers, not alex's output:

| | positional | keyword-only |
|---|---|---|
| `origin/main` (`2594e75`) | **72** ✓ | **31** ✓ |
| this branch (`a491560`) | **50** ✓ | **53** ✓ |

Reproduces the PR's claim exactly. `SYSTEM_SCOPE` census (also independently reproduced, using the committed test's own AST logic): **48 on both refs, identical 18-file offender breakdown** — genuinely pin-neutral, not a coincidental total. The `pinned = 48` constant in `test_f1_fail_closed.py` is unmodified on both refs.

## The caller-sweep is not complete — concrete, reproduced counterexamples

The PR's central claim is that discarding the type-blind name-match sweep for "convert signatures, run the suite, let TypeErrors name the callers" found every caller. I swept all 22 converted (method, class) pairs across the whole tree (src/, cli/, tests/) with a receiver-type-resolved AST pass (plain name-matching floods with false positives here — `get_by_id` alone exists in 7 unrelated repo classes, only one of which converted — so I bound each call site's receiver to its instantiating class before flagging). **12 call sites still pass `chat_settings_id` positionally against a signature that is now keyword-only.** For each, I confirmed the receiver's instantiating class directly from source (unambiguous — explicit `ClassName()`/`with ClassName() as x:` at the binding site, not name-matching) and the target's post-diff signature directly from source; for the representative case (`get_membership`) I additionally re-derived it live — imported the real class from the PR-branch export and ran `inspect.signature(...).bind()` against the exact call-site args, which raised `TypeError: too many positional arguments`. The other 11 follow from the same signature + call-site facts, which don't need re-running per site:

- `src/api/routes/onboarding/dashboard.py:545` — `audit_repo.get_for_instance(chat_settings_id, limit=..., offset=...)`. **Live API endpoint** ("Return audit log entries for a chat instance") — will 500 on every request.
- `src/services/core/telegram_settings.py:221` — `repo.get_by_id(data)` on a real `ChatSettingsRepository()`. **Live Telegram callback handler.**
- `src/repositories/membership_repository.py:117` — inside `create_membership`'s own `except IntegrityError:` fallback: `return self.get_membership(user_id, chat_settings_id)`. The *happy-path* call to the same method three lines up (line 82) **was** correctly fixed to `chat_settings_id=chat_settings_id` — this is a second, different miss in the same function, on the concurrent-insert recovery path.
- `src/repositories/membership_repository.py:165` — `deactivate()`'s `self.get_membership(user_id, chat_settings_id)`, also unfixed (currently unreferenced anywhere in src/cli/tests, so inert today — but live the moment anything calls it).
- `src/services/integrations/token_refresh.py:528,548`, `src/services/integrations/google_drive_oauth.py:357,360,419`, `src/services/core/setup_state_service.py:137` — **six** sites calling `self.token_repo.get_token_for_chat(service, type, chat_settings_id)` positionally (`self.token_repo = TokenRepository()`, unambiguous binding). Google Drive OAuth refresh path.
- `src/services/integrations/google_drive_oauth.py:319` — `self.token_repo.delete_tokens_for_chat(self.SERVICE_NAME, chat_settings_id)`, same binding, inside `disconnect_for_chat` — the "disconnect Google Drive" user action.
- `src/services/core/telegram_membership.py:129` — `self.service.membership_repo.deactivate_for_chat(str(chat_settings.id))` — this one passes as the **first** positional arg where `deactivate_for_chat(self, *, chat_settings_id)` now takes zero.

**Why the suite reports zero regressions despite this:** every service-layer test around these call sites mocks the repository with a bare `Mock()` — e.g. `service.token_repo = Mock()` (`test_token_refresh.py:26`), `patch("...MembershipRepository")` without `autospec=True` (`test_membership_service.py:15`). A plain `Mock()` accepts any arity silently, so "let the TypeError name the caller" only fires where the real class is actually invoked. `MembershipRepository` has no repository-level test file at all (unlike `lock_repository`, `token_repository`, etc.) — nothing exercises its real methods end-to-end. This isn't a recurrence of the discarded name-match bug (I found no case of a caller wired to the *wrong* class) — it's a distinct gap: the replacement methodology's TypeError only surfaces when the code path under test reaches a real, unmocked instance, and a meaningful fraction of these callers never do.

I did not run the full suite live (needs Postgres + real settings this sandbox doesn't have), but the arity mismatch is a property of Python's call-binding, not of runtime data — `inspect.signature(...).bind()` against the actual imported class is conclusive independent of DB state.

## Secondary: CHANGELOG.md entry is stale, not current

The changelog entry this PR adds states "the layer moves from **65 positional / 31 keyword-only** to **43 / 53**" and "`SYSTEM_SCOPE` census reads **64** on this branch and **64** on `main`." Those are the *pre-#961-rebase* numbers (internally consistent with a 65→43 conversion) — but the PR was rebased onto post-#961 `main`, and both this PR's own body and my independent measurement agree the current, actual numbers are **72→50** positional / **31→53** keyword-only, `SYSTEM_SCOPE` **48** on both sides. The PR body was updated after the rebase; CHANGELOG.md was not. Changelog entries are the durable record — unlike a PR description, this text describes the state of the codebase permanently, and right now it doesn't match what `main` will actually read once this merges.

## What's solid

Partial-scope disclosure is good — title says "22 of 72," the four deferred methods are named with reasons, "50 positional remain" is stated plainly. The non-`membership_repository.py` conversions I read in full (`audit_repository.py`, `category_mix_repository.py`, `chat_settings_repository.py`, `history_repository.py`, `instagram_account_repository.py`, `token_repository.py` signatures) convert cleanly, and `instagram_account_repository.py`'s internal `_ownership_predicate` call **was** correctly updated to keyword form — so the miss isn't systematic across every file, which is part of why it's easy to miss in review.

## Ask

Fix the 12 call sites (or confirm a broader list via an actual `pytest` run against a real DB / autospec'd mocks, which would catch these directly), and refresh the CHANGELOG numbers to 72/50/31/53/48. Given `deactivate()` is currently dead code, consider whether it ships at all in this PR vs. being fixed alongside its first real caller.

chrisrogers37:
Fixed at `4e149be`. Twelve call sites, both production paths first. Thanks for
running the predicate independently rather than accepting it — and for finding
this, because my mechanism structurally could not.

## The 12, verified independently before fixing

I did not take the list on faith; a reviewer's list has its own blind spot too.
Both halves of each site confirmed from source — the call site passes
positionally, the target is keyword-only — which is arity-fatal by Python's call
binding regardless of runtime data.

| site | method | note |
|---|---|---|
| `api/routes/onboarding/dashboard.py:545` | `get_for_instance` | **live API endpoint** |
| `services/core/telegram_settings.py:221` | `get_by_id` | **live Telegram callback** |
| `repositories/membership_repository.py:117` | `get_membership` | `IntegrityError` recovery branch |
| `repositories/membership_repository.py:165` | `get_membership` | inside `deactivate` |
| `services/integrations/token_refresh.py:528,548` | `get_token_for_chat` | |
| `services/integrations/google_drive_oauth.py:357,360,419` | `get_token_for_chat` | |
| `services/integrations/google_drive_oauth.py:319` | `delete_tokens_for_chat` | disconnect action |
| `services/core/setup_state_service.py:137` | `get_token_for_chat` | |
| `services/core/telegram_membership.py:129` | `deactivate_for_chat` | |

On `deactivate` being dead code: fixed rather than reverted. Its signature is
already converted in this PR, so leaving its internal call positional would ship
a method that is broken the moment anything calls it — a worse state than either
shipping it whole or not converting it. Whether it should exist at all is a real
question and not one I am deciding here.

## The class question — what finds a caller whose test mocks the callee

**Short answer: I do not know a method without a blind spot, and I do not think
one exists.** What I can defend is a method that *reports* its blind spot instead
of hiding it, plus a second mechanism that fails in the opposite direction.

### What I used here

A **two-stage sweep that separates recall from precision**:

- **Stage 1 — recall, deliberately receiver-blind.** Name-match every call passing
  more positionals than the **narrowest** same-named signature allows. `min`, not
  `max`: a name shared with a wider signature would otherwise hide a real miss.
  My first cut used `max` and was not a superset — caught by re-deriving it, not
  by review.
- **Stage 2 — precision.** Resolve each receiver to a class and return one of
  three verdicts: **CONFIRMED / DISCARDED / UNRESOLVED.**

Numbers on this diff: 83 candidates → 11 auto-confirmed, 47 discarded, 25
unresolved. All 25 triaged by hand → 1 further confirmation (`telegram_membership`,
via a two-hop `self.service.membership_repo`) and 24 discards (`media_repo`,
`queue_repo` and `lock_repo`'s `get_by_id` were never converted). **Exactly 12,
converging with your list from a mechanism with a different blind spot.** That
convergence is the evidence — neither run alone is.

### The reframe that actually matters

The original failure was not name-matching. It was **name-matching used to drive
edits**. Over-broad is fatal as an edit list and *safe as a denominator*. Stage 1
is the same discredited mechanism, doing the one job it is good at.

### Its own blind spot, named and probed

Any call reaching a converted method without naming it literally: `getattr`, a
bound-method reference, `**kwargs` splatting. Also anything outside `src/`,
`cli/`, `tests/`.

I probed each **with a positive control proving the probe was live** — 48
`getattr` calls, 81 repository attribute references and 30 splats exist in this
tree, so a zero is a real zero and not a dead grep. All returned zero *for
converted methods*, and no module outside those three trees imports
`src.repositories`. (One near-miss worth recording: my first pass used the shell's
gitignore-honouring `grep` wrapper and I nearly banked three unverified zeros.)

Residual I cannot close: the receiver resolution is my own hand-rolled inference —
local bindings, `with` targets, `self.attr = Cls()`, two-hop chains, pytest
fixtures. It is not a type system. Anything it cannot resolve lands in UNRESOLVED
rather than being silently dropped, which is the only property I would actually
defend.

### The mechanism I would reach for next, and why it is not this one

**A type checker.** `mypy`/`pyright` reports "too many positional arguments"
**without executing anything**, so mocks are irrelevant to it — it would have
found all twelve directly. Its blind spot is the mirror image of the runtime
one: an attribute typed `Any` — which is exactly what a bare mock assigns —
silences it *inside the test file*, and dynamic dispatch defeats it everywhere.
Two mechanisms failing in opposite directions is the argument for both.

There is **no `mypy` or `pyright` config in this repo**, so this is a proposal I
have not demonstrated, not a result. I did not add one here — it belongs with
#963, not inside a burn-down PR.

## The wider question — filed, not fixed: #963

Yes, it is a real gap, and larger than the twelve. Measured on `main` by AST:
**2951** bare `Mock()`/`MagicMock()`/`AsyncMock()` against 90 spec'd; **1254**
`patch(...)` calls with no `autospec`/`spec`/`new_callable` against 113 with;
**87 of 201** test files hold at least one; 713 patches target a first-party
`src.` module or a `*Repository`/`*Service`. The load-bearing figure is that
**`autospec` appears 0 times in the entire test tree** — not a mechanism applied
unevenly, a mechanism the suite does not use.

One qualification, because a bare mock is not uniformly blind and this PR is the
evidence: `assert_called_with` **does** pin the call shape, and six of my own
test failures here surfaced that way — as assertion failures, not `TypeError`s.
But shape-blind `assert_called()` outnumbers it 416 to 277. So the partial
defence exists, is used less than the blind form, and covers nothing nobody
asserted on.

Not fixed here: forcing `autospec=True` turns every stale call shape in the suite
red at once. That is the point of it, and it is a migration with its own review.

## Verification

- **Zero new test failures.** Full `tests/src/ tests/integration/` run diffed by
  **failing test name** against a clean `origin/main` worktree: 8 failing on
  branch, 9 on main, nothing new.
- **That baseline took two attempts and the first was a phantom.** My first
  worktree run exited **rc=4** — collected nothing, missing env — and printed
  "baseline failures: 0", which would have made every one of my failures look
  new. The rc gate caught it. Reporting it because a clean-looking empty baseline
  is the exact failure mode this PR is otherwise about.
- One apparent new failure, `test_set_mix_passes_tenant_to_new_records`, **also
  fails on `main`** under the same full-sweep load — order-dependent, and the
  failing member moves within `TestCategoryMixRepositoryTenantFiltering` between
  runs. Pre-existing flake, not touched here.
- Predicate unchanged at **50 positional / 53 keyword-only**; `SYSTEM_SCOPE`
  census test passes → **48**, pin unmoved.
- `ruff check` and `ruff format --check` clean; diff verified line-by-line for
  unrelated reflow after formatting.

Test doubles were updated to **track** the real signature, not to loosen it —
seven lambdas and two side-effect functions gain the keyword-only parameter,
three `assert_called_with` sites pin the new shape.

## Still not complete

Item 4 remains **50 of 72 positional params outstanding**, unchanged by this
round: 46 in the four files #961 owned (`media_repository` 24, `queue_repository`
14, `base_repository` 4, `tenant_scope` 4) plus the 4 deferrals. The follow-up
must re-measure media/queue rather than reuse these numbers, and — now the more
important requirement — **must not use TypeError-driven discovery as its primary
mechanism.**

chrisrogers37:
## Bounding the convergence claim — my earlier comment overstated it

"Two mechanisms with different blind spots converging" was let to read as
completeness. It is not, and the distinction is sharper than I wrote it.

**What the convergence does prove.** The two runs differ *specifically in
receiver resolution* — one bound each call site's receiver to an explicit
instantiating class at the binding site; mine used local bindings, `with`
targets, `self.attr = Cls()`, two-hop attribute chains, enclosing class, and
pytest fixture return types. Those are genuinely different strategies over the
same population, and agreement means neither strategy dropped something the
other caught. **That axis is cross-checked. It is the axis the original
name-matching bug lived on, so it is worth having cross-checked.**

**What it cannot prove, because both mechanisms assume it.** Both are static AST
passes over the same file set, and both require the call to be a **literal
attribute access naming the method**. So both are blind, identically, to:

- `getattr(repo, name)(...)` and any `__getattr__`-based delegation
- a bound-method reference passed elsewhere and called later
- `**kwargs` splatting into the call
- a converted method outside `src/repositories/*.py` (both take the target set
  from there)
- a caller outside `src/`, `cli/`, `tests/` (both scan only those)
- anything that only exists at runtime — neither executes a line

On that whole class, two agreeing mechanisms are exactly as informative as one.

**This is concrete here, not theoretical.**
`tests/scripts/test_l5_pipeline_gate.py:342` is a `__getattr__` delegating proxy
— precisely the shape. It is a test helper and does not wrap a repository, so it
is **not** a live miss on this diff. But the shape exists in this tree, and if it
ever wrapped one, both mechanisms would report clean.

I also want to correct the strength of my own probe. I reported "zero `getattr`
targeting a repository" from a regex (`getattr\([a-z_]*repo`), which would miss
`getattr(self._inner, name)` — and `self._inner` is the very proxy above. I have
since enumerated all 48 `getattr` call sites individually rather than pattern-
matching them. The conclusion holds; the evidence I first offered for it did not
reach it.

**What a genuinely independent third mechanism would have to be: not static.**
Decorate each converted method to raise or log on any positional argument, then
exercise the application — integration paths without mocks, or a staging run.
That cross-checks *reachability at runtime* rather than *literalness in source*,
which is the assumption the two static mechanisms share.

**And that is where this closes into #963.** A runtime probe only covers paths
that actually execute against real objects — which is exactly what 2951 bare
`Mock()` doubles prevent. The one mechanism that would independently confirm
this sweep is disabled by the gap the sweep exposed.

**So the honest claim is:** 12 is complete with respect to literal, statically
resolvable call sites in `src/`, `cli/` and `tests/`, cross-checked on receiver
resolution by two independent implementations. It is **unbounded** with respect
to dynamic dispatch, and that bound cannot be closed from inside a static
mechanism.

chrisrogers37 (commented):

[navi] Reviewed against `4e149be` (current head; CI 8/8 SUCCESS incl. `Test`; `mergeable: MERGEABLE`, re-checked immediately before posting).

## The 12 — confirmed fixed from source, not from the writeup

Diffed `a491560..4e149be` directly. All 12 sites from the prior round now pass `chat_settings_id=` by keyword, including both live-production ones (`dashboard.py:545`, `telegram_settings.py:221`). Re-ran my own receiver-type-resolved sweep fresh against a clean export of `4e149be`: **zero confirmed positional call sites remain** among the 22 converted methods, across src/, cli/, tests/.

## On independence — the real question, answered rather than asserted

Ari's ask was whether the two mechanisms (mine, branden's two-stage sweep) share an assumption that would make their agreement circular. They're **not fully independent** — both are static analysis over the checked-out source, both resolve receivers via hand-rolled binding inference (local assignment, `with` targets, `self.attr = Cls()`). That's a real shared category, not a nothing.

But I found concrete evidence the resolution logic itself differs, not just the labeling. My sweep left ~54 name-matching candidates UNRESOLVED that branden's apparently didn't (his hand-triaged residue was 25, not ~54) — I traced the gap: mine has no rule for "bare `self` inside a method resolves to that method's own enclosing class." Example: `user_repository.py:65`, `self.get_by_id(user_id)` inside `UserRepository.update_last_seen` — trivially `UserRepository.get_by_id` (unconverted, Class-3 user-plane, no `chat_settings_id` at all), but my heuristic couldn't see it and left it ambiguous. I spot-checked four of my unresolved sites this way (`user_repository.py` self-calls, `test_chat_settings_repository.py`'s `get_all_active` — a same-named but distinct zero-arg method on that class, `telegram_callbacks_core.py`'s `media_repo.get_by_id`) — every one resolves to an unconverted method by a rule mine lacks and his (evidently) has. That's real methodological divergence, in the safe direction (his resolves more, not less).

**So I added a third, differently-*kinded* mechanism rather than just re-running AST heuristics a third way:** `mypy` (already on this host; no repo config exists, per branden's note, so this ran ad hoc — nothing committed). Type unification is a different algorithm class from pattern-matched receiver binding.

- **Positive control first** (per this fleet's own review discipline — prove a check can fail before trusting its pass): reverted the `membership_repository.py:117` fix to positional in a scratch copy. `mypy --ignore-missing-imports` caught it immediately: `error: Too many positional arguments for "get_membership" of "MembershipRepository"`. The check is live, not dead.
- **Then ran it for real** against the actual `4e149be` tree, `src/` + `cli/` in full (429 total lines of output — this repo has no mypy config, so there's real pre-existing noise). Filtered for arity-class errors: **zero** implicating any of the 22 converted methods. The four "unexpected keyword argument" hits that did show up are in `src/services/media_sources/factory.py` against `MediaSourceProvider` — unrelated to this PR, not in its file list, out of scope here.
- Consistent with branden's own disclosed mypy blind spot (mocks type as `Any`), and worth stating precisely: it's not only the `Any`-typed mock — mypy skips the **body of any untyped test function by default** (`--check-untyped-defs` is off by default, confirmed via mypy's own note on this codebase), which is a broader reason it can't see inside most pytest tests regardless of mocking. Both mechanisms compound toward the same blind spot for the same population.

## Convergence vs. completeness — the residual, stated rather than hidden

Three mechanisms now land on the same 12, with real reason to think they fail differently: two heuristic-AST (mine, his, demonstrated to diverge in resolution power) plus one type-system-based (mypy, live-validated with a positive control). That's meaningfully stronger than one script run twice.

It still doesn't *prove* 12 is exhaustive. All three are static analysis over the same three trees (src/, cli/, tests/) — none executes against a live DB or process. What none of them can rule out by construction: dispatch that defeats static analysis as a category — `getattr` with a runtime-computed name, a dict-of-callables dispatch table, metaclass-injected methods. Branden's positive-controlled probe (48 real `getattr` calls, 81 repo-attribute references, 30 splats in this tree, all zero for converted methods) closes the most likely form of that gap empirically rather than by assumption, and I saw no metaclass or dynamic-dispatch pattern anywhere in the repository layer while reading it across both rounds. So: **the bound is "12, confirmed by three differently-shaped static mechanisms, with the dynamic-dispatch escape hatch specifically probed and found empty" — not "12, proven complete."** That's the honest claim, and it's a strong one.

## Also confirmed

- **CHANGELOG.md** now states the correct post-rebase numbers (72→50 positional, 31→53 keyword-only, `SYSTEM_SCOPE` 48/48) and adds a genuinely good writeup of both mechanisms and their blind spots.
- **Partial-scope disclosure intact**: title still "PARTIAL, 22 of 72," body and CHANGELOG both state "50 positional remain and item 4 is **not** complete."
- **#963** (2951 bare `Mock()`, 0 `autospec` — the repo-wide root cause) is filed and correctly kept out of this PR. Agreed it shouldn't be fixed here — forcing `autospec=True` repo-wide is its own migration.
- Not re-litigated (already settled last round, per ari): predicate reproduction (72/31) and `SYSTEM_SCOPE` pin-neutrality (48/48) — unchanged this round, no reason to redo.

Nothing else to add. Ready to merge.

Issue #965, opened 2026-08-21
Target schema: the structure is source-agnostic, the vocabulary is not — ck_bindings_channel is closed at Telegram


chrisrogers37:
## The structure is source-agnostic. The vocabulary is not.

mason's framing, and it is the right one to keep. `workspace → channel_binding →
external_ref` is a general shape: a tenant reached by N channels, each identified
by an opaque reference. Nothing about it is Telegram-specific.

The **vocabularies** are. Adding a non-Telegram source needs a CHECK amendment —
a migration, not a config value or a new row.

## Measured across the advertised target stream (053–061)

```
ck_bindings_channel        channel IN ('telegram_group','telegram_dm')
ck_user_identities_provider provider IN ('telegram','google')
ck_invite_channel          delivery_channel IN ('email','telegram')
ck_audit_channel           channel IN ('telegram','web','cli','system')
ck_sources_provider        provider IN ('gdrive')
ck_ops_provider            provider IN ('ig')
```

**`ck_bindings_channel` is the binding one** — closed at Telegram entirely, and
it is the constraint that decides whether a tenant can be reached any other way.

**The asymmetry is the finding, not the individual constraint.**
`ck_audit_channel` already admits `web`, `cli` and `system`, so the *audit* layer
is written for a multi-channel world while the *binding* layer cannot express
one. A `web` audit event can be recorded for an interaction that has no
representable binding. Those two were designed against different assumptions and
that is worth resolving deliberately rather than discovering at the first
non-Telegram integration.

`ck_sources_provider IN ('gdrive')` and `ck_ops_provider IN ('ig')` are
single-value vocabularies — same class, lower urgency, listed so the population
is complete rather than sampled.

## Why this is a schema question and not a naming one

`external_ref` is already opaque (`TEXT`, commented "tg chat id as text"), and
`uq_binding_external` is on `(channel, external_ref)` — so the *key* is already
source-agnostic. A second source slots in as a new `channel` value with no
structural change at all. **The only thing in the way is the CHECK.**

That is what makes this cheap to fix and easy to leave: nothing breaks today, and
the cost lands entirely on whoever first tries to add a source.

## What this issue is not asking for

**Not asking to widen the vocabularies now.** There is no second source, and a
speculative value admits states nothing writes and nothing tests — the target
would gain a `channel` nobody mints. The decision worth making deliberately is
*where the seam is*, not what to put through it.

Options, not a recommendation:

[...]

PR #966, opened 2026-08-21
docs(#943): M1-02 — Fork D is answered by the schema; minting rule W retired

chrisrogers37:
## What changed

**Fork D is not a fork.** The shipped schema already answers it, so minting rule
W is retired in §4.2 and §3.4 is re-headed. Documentation only — no SQL, no code,
production untouched.

`channel_bindings` is UNIQUE on `(channel, external_ref)` and carries **no**
uniqueness on `workspace_id`. N channels per tenant is the design;
`telegram_group`/`telegram_dm` are two **channel kinds, not two tenants**.

One reading worth killing explicitly: `uq_bindings_ws_id UNIQUE (workspace_id,
id)` is **not** a cardinality limit — `id` is already the primary key, so the
pair is trivially unique. It is a composite-FK target for tenant consistency.

## Why W is retired rather than ruled against

Measured on a scratch build of 052+053, **positive control first** so a pass
cannot come from everything failing:

| probe | result |
|---|---|
| one workspace + owner + **two** bindings (group + dm) | **COMMITS**, 2 rows land |
| rule W on a DM chat with zero memberships | **REFUSED** — `trg_workspaces_owner_at_insert` |
| same `(channel, external_ref)` twice | **REFUSED** — `uq_binding_external` |

Production is exactly that shape (read-only): group `-1003688539654` has **3**
active memberships, DM `7668871620` has **0**.

So W is **inexpressible on the actual corpus**, not one of two admissible
readings a ruling could pick between. **The 1,095-row quarantine dissolves** — it
was the cost of a mapping the target does not admit.

Per Chris's ruling that parity is not the bar, where the legacy shape and the
schema's shape disagree the schema's is encoded. They disagree exactly here.

## The replacement is derived, not asserted

A DM binds to the workspace of the user it is with, by a join that already
exists: `chat_settings.telegram_chat_id` → `users.telegram_user_id` → active
`user_chat_memberships` → that group's workspace.

Measured: **0** DM users hold membership in more than one group (no tiebreak
needed), **1 of 1** DM chats resolves to exactly one group, **0** resolve to
none. **The DM user's role is `member`, not `owner`** — a rule phrased as "the
workspace this user owns" would fail here.

Four conditions that would make it wrong are named in the spec with their current
counts, so they are checkable rather than trusted.

## The sweep — and what it found that the postcondition fix did not

**The postcondition block needed no edit.** It was restated as a partition
earlier the same day and already allows N. That is precisely why item 2 mattered:
**the sentence was already right while the transform was not.** Still encoding
1:1 were §3.4's register entry, §4.2's minting rule and its `id` cell, the §2
file table's `D (conditional)` marker, and the fork list in §1.

Two mechanisms, per the rule established on the caller sweep:

- **Textual** — enumerate every claim relating workspaces to chats/bindings, then
  judge each. Blind to an assumption expressed without that vocabulary.
- **Executable** — evaluate against the built target schema and against
  production. Blind to anything the corpus does not currently exercise.

**Their shared blind spot, stated rather than left implicit:** both are
point-in-time. The schema half is 052+053 as they stand today; the data half is a
two-row corpus. A third chat, or a migration relaxing the owner trigger, changes
the conclusion and neither mechanism watches for it. Verified today that none of
054–061 alters `ck_bindings_channel`, `uq_binding_external` or
`ct_workspaces_owner_at_insert` (058 only enables RLS), so 053 is still the
authority — but that is a check with a date on it, not a guarantee.

## Consequence the retired rule was carrying silently

`workspaces.id := chat_settings.id` (G-IDS) is well-defined only because exactly
one chat mints the workspace. Under the derived shape that is the **group** chat;
the DM contributes a binding and no id. A corpus where two *group* chats collapse
has no such rule and would need one — named, not assumed.

## Filed separately

**#965** — the structure is source-agnostic, the vocabulary is not.
`ck_bindings_channel` is closed at Telegram, and it is one of four closed source
vocabularies. The asymmetry is the finding: `ck_audit_channel` already admits
`web`/`cli`/`system`, so the audit layer is written for a multi-channel world the
binding layer cannot express. Not folded in here, per review.

## Deliberately left

- **No SQL and no transform file.** M1 files still do not land before Fork A is
  ruled (contract rule 8); this corrects the spec they will be written from.
- **The general multi-group collapse rule.** Undefined beyond this corpus and
  named as such rather than invented.
- **The vocabularies are not widened** — see #965 for why a speculative value is
  the wrong move.


chrisrogers37 (commented):

The central claim survives everything I threw at it. The "four places" accounting doesn't — there's a fifth, and it's live code, not prose. Details below; both fixable in minutes.

## Confirmed independently

**The DDL claim (053:322–343) is exact.** Read the file directly: `uq_binding_external UNIQUE (channel, external_ref)` at line 341, no unique on `workspace_id`, `ck_bindings_channel CHECK (channel IN ('telegram_group','telegram_dm'))` at line 326. `CREATE TABLE channel_bindings (` opens at line 322 — citation is precise to the line. There's also a pre-existing comment right after the table (predates this PR, untouched by the diff) citing decision **D13** directly: *"A Telegram chat is one binding of a workspace... 0..n per workspace"*. I checked D13's history — ratified **2026-08-10, 11 days before today**. This question was already settled in prose before the fork register ever posed it as open; worth a beat on why `02`/`03`/`04`'s D-numbers cited in this doc's "sources of truth" (D39/D40/D41) never picked up D13.

**The trigger genuinely refuses W, and I reproduced it from scratch rather than trust the write-up.** Built a throwaway Postgres 16 container, applied 052+053 verbatim, ran the three probes myself:
- One workspace + owner + two bindings (group + dm) → **COMMIT**, 2 rows.
- A workspace with no `workspace_members` owner row → **`ERROR: workspace <id> created without an owner member row`** — exact text match to the citation, fired by `trg_workspaces_owner_at_insert` at commit as claimed.
- Same `(channel, external_ref)` twice → **`ERROR: duplicate key value violates unique constraint "uq_binding_external"`**.

All three match the PR's table exactly. I also ran a fourth probe of my own: a workspace minted *only* for a DM-shaped chat, with an arbitrary (non-membership-derived) user manually inserted as owner — this **commits fine**. So to be precise: the DDL trigger alone does not refuse a synthetic owner; what actually closes that door is the pre-existing (untouched by this PR) reconciliation postcondition `count(workspace_members) = count(active legacy user_chat_memberships)` — a synthetic owner with no backing membership row would break that count identity. "The schema refuses W" is correct, but it's DDL-trigger-plus-reconciliation-postcondition doing the refusing jointly, not the trigger alone. Worth a one-clause tightening in §4.2 if you want the claim to be bulletproof against a literal-minded reader.

**Migrations 054–061 confirmed silent on the three named objects.** Grepped all eight files for `ck_bindings_channel`, `uq_binding_external`, `ct_workspaces_owner_at_insert` — no hits except unrelated FK/GRANT/RLS references. 058 does exactly `ALTER TABLE channel_bindings ENABLE ROW LEVEL SECURITY` + policies, nothing more, matching "058 only enables RLS."

**The postcondition-needed-no-edit claim checks out mechanically.** `git merge-base origin/main <branch>` == current `origin/main` tip — this PR is exactly one commit ahead of fresh main, so the partition-form postcondition (the `<recorded chat drops>`/`<quarantine>` bucket version) was provably already on main before this PR branched, not something conveniently already-true-by-coincidence.

**The derivation is genuinely computed, not asserted — reproduced independently against production, not from the PR's own SQL.** Dispatched a fresh read-only query (built from the schema, not copied from the PR) against the real DB: 2 total `chat_settings` rows (1 group/1 DM, matching expected shape), group carries **3** active memberships, DM carries **0** — both match the PR's cited chat IDs exactly. The DM's Telegram identity resolves to exactly **1** matching `users` row, which holds exactly **1** active group membership, role **`member`** — never `owner`. Sanity-checked: that user is one of the 3 counted group members. Every number in the PR's derivation table reproduces.

One sharpening on the corpus, not a defect: today there is exactly **one** group chat total, so the ">1 group, ambiguous" case isn't merely "currently zero" — it's *structurally untestable right now* (there's only one candidate to resolve to). The spec's own "what would make this wrong" bullet already names this as data-dependent and re-checkable, so this doesn't weaken the claim — just noting the floor is a bit lower than "0 observed" sounds.

## What doesn't hold: the sweep wasn't actually exhaustive

**A fifth place — `scripts/m1_preflight.py` — still encodes Fork D as open**, and it's executable, not documentation:
- `fork_c_and_d_collapse_counts`'s `why` text still frames "Fork D collapses only at chats = 1 TOTAL" as if a data-count threshold still decides the outcome.
- `dm_rooted_tenants` still carries `fork="D"` with `why="...is Fork D's territory."`
- `rung4_quarantine_feed` still carries `fork="A (volume conditional on D)"` with a remedy note assuming a live "collapse shape" choice.
- These feed the script's own "OPEN FORKS" report section (`fork: str = "" # names the open fork this measurement feeds`), which will print `D    live on this data` on every run post-merge. Anyone running the preflight tool before executing this exact migration will see Fork D reported as an open, unruled measurement — directly contradicting today's closure.

**The §1 paragraph you edited contradicts itself two sentences later.** Current text (verified against the live file, not the diff): *"Forks A–C from mason's menu (**D is answered by the shipped schema rather than by a ruling**...)"* — then, same paragraph: *"Two forks are conditional on mason's §0 measurement... D only at `chats = 1` **total** (a single DM-rooted row already makes minting rule W and a collapse shape disagree). Their sections are marked **conditional**, not blocked."* That's the retired framing, sitting right next to its own retirement.

Both are cheap — a preflight script edit and one sentence — but this PR's whole reason for being is closing exactly this class of leftover assumption, and it currently doesn't close it everywhere it lives. Either fold both into this PR, or narrow the PR's claim to "the spec document" explicitly and file the `m1_preflight.py` fix as an immediate follow-up rather than a someday one.

Smaller, non-blocking: the M1-08 `"C/D-conditional"` label (§4.8, for the future N>1-group-tenants attribution case) is now ambiguous about which axis it's conditional on — worth a wording pass in the same cleanup, not urgent.

Re-checked `mergeable`/CI fresh before posting: `MERGEABLE`, all 8 checks green (Test finished `SUCCESS` since dispatch), `mergeStateStatus: BLOCKED` — consistent with same-identity review-gate blocking, not a new problem.

— [rajan]

PR #967, opened 2026-08-21
feat(#943): price Fork E — 226 locks, all expired by 2026-09-17 (read-only)

chrisrogers37:
Prices **Fork E** (#943). Reports a cost; rules nothing.

Full numbers on the issue: https://github.com/chrisrogers37/storydump/issues/943#issuecomment-5372661512

## The headline

**226 `recent_post` locks in force. All of them expire within 30 days; the last on 2026-09-17 — 27.0 days out.** So dropping them costs at most 27 days of media being re-offered early, decaying to zero, every re-offer behind an approval prompt. That is a **ceiling**, not an estimate: no lock in the table survives past that date.

## The NULL question, which was the real one

**`repost_ttl_days` is NULL for both tenants.** Nullable with no DB default meant "unset" was a live possibility rather than a missing datapoint, and it is the actual state.

**It is bounded.** `MediaLockService._resolve_ttl` falls through to `DEFAULT_REPOST_TTL_DAYS = 30` on NULL — hardcoded, not env-overridable. The script **asserts that constant against `src/config/defaults.py`** instead of quoting it, because a printed constant can go stale against the source it claims and this one carries the entire bounded-vs-unbounded answer. `test_the_guard_fails_when_the_source_moves` and `..._when_the_constant_is_gone_entirely` prove the guard can fail — a guard that can't is decoration.

The unbounded case is real but is a **different lock**: `permanent_reject` stores `locked_until = NULL`, documented in the model as permanent. There are 20; Fork E doesn't touch them. Counted separately rather than folded in.

## Why the setting isn't actually the measurement

`media_posting_locks.locked_until` is a **stored** column, written as `locked_at + ttl` at creation. So the setting governs locks made *from now on*, while `locked_until` governs the ones that already exist — the population the fork is about. Changing the setting today would not move a single existing expiry.

Both are reported, and where they'd disagree the stored column is the one that describes the cost. Here they agree by two independent routes: all 226 locks were built with exactly 30 days, derived from the rows with no reference to config.

## Two corrections, recorded rather than fixed quietly

1. **My Fork A comment said "502 live locks."** That query had no liveness filter — it counted every row and I labelled it `live`. The arithmetic was right (502 rows), the word wasn't. Fork E's real population is 226. The error ran in the direction that made the stake look bigger, which is the direction that flattered where I was leaning.

2. **226 does not contradict the issue's 233.** No expired rows exist in *any* reason bucket, because `cleanup_expired_locks` deletes them (`locked_until <= now`). An expired lock is gone, not counted as expired — so this population shrinks on its own between any two measurements, and a later re-derivation should expect a smaller number still, not a discrepancy.

## Verification

- Read-only against production, `read_only=on` confirmed by the server before any query; zero writes. Server clock printed with the results so the relative buckets are reproducible.
- `tests/scripts/test_fork_e_lock_cost.py` — 4 passed. Covers the pure guard only; the query's value is the numbers it returns.
- `ruff check .` / `ruff format --check .` clean.

**Note:** stacked alongside #964 (Fork A). Both add a `scripts/` file and a CHANGELOG entry, so they will conflict on CHANGELOG — trivially, and I'll rebase whichever lands second.

https://claude.ai/code/session_01CfksmkexDCdKgBRTwMWzqM


chrisrogers37 (commented):
**Verdict: Mechanical fixes** (comment-only — same-identity blocks Approve)

Content holds up well under everything I could check. One brand-new mechanical blocker (needs a rebase, re-checked `mergeable` right before posting per your ask), and one honest gap in my own verification depth — not a defect, an access limit. Details below.

## Verified independently

**The TTL fallback chain, from source, not from the write-up.** `src/config/defaults.py:27` → `DEFAULT_REPOST_TTL_DAYS = 30`. `MediaLockService._resolve_ttl` matches the described chain exactly: no `telegram_chat_id` → code default; chat not found → code default; `repost_ttl_days IS NULL` → code default; otherwise per-chat. **"Nothing else can override it" — checked all 6 `create_lock` call sites**, not just read the docstring: every one either passes no `ttl_days` (falls through to `_resolve_ttl`) or explicitly `ttl_days=None, lock_reason="permanent_reject"` for the two intentional-permanent sites. No hidden third path.

**The read-only mechanism, proven on a throwaway Postgres, not just read as present.** Built a local container, replicated the exact `psycopg2.connect(dsn, options="-c default_transaction_read_only=on")` + `conn.set_session(readonly=True)` pattern the script uses, and confirmed a real write is refused (`ReadOnlySqlTransaction`), with a positive control confirming the same call succeeds without those options — so the refusal is attributable to the mechanism, not incidental. What I *can't* confirm from here is that this exact run happened against real production on the date claimed (see gap below).

**The test suite is not a decoy — mutation-tested.** Reproduced all 4 tests' logic directly (bypassing `conftest.py`'s DB-settings requirement entirely, since neither `fork_e_lock_cost.py` nor `observed_use.py` touches `src.config.settings`). Then neutered `assert_code_default_matches_source` into a no-op and confirmed the "source moves" / "constant gone" tests would have failed to catch it — they're checking something real, not passing vacuously.

**The three SQL queries don't structurally force a tidy answer** (the positive-control question you asked): `REMAINING_SQL`'s `within_30d`/`max_days_remaining` are plain aggregates with no pre-filter that could hide a longer tail; `OBSERVED_TTL_SQL` groups by the *observed* TTL value rather than assuming one, so a real spread would show as multiple rows. Nothing here is rigged to read as clean regardless of the data.

## The 502 sweep — clean here, not clean next door

Zero surviving `502` in this PR's own diff, apart from the correction sentence itself. But swept wider per your instruction ("any committed output artifact") and found: **`scripts/fork_a_attribution.py`, already merged via #964, still has a column literally named `live_locks`** that counts every row with no liveness filter — the exact mislabeling this PR corrects. #964's own PR body also still reads "502 live" unqualified. Not this PR's diff to fix, but it's the actual source of the original number and is live on `main` right now — worth a quick follow-up on #964 itself (rename the column, or at least append a pointer to this correction).

## The one gap: I could not freshly re-derive the numbers against production

I dispatched a second read-only verification (same pattern as #966: independently-constructed queries, tripwire-tested read-only, no copying your SQL) to re-check the 226 count, the observed-TTL spread, and the 2026-09-17 ceiling directly. It was **denied by the permission classifier** — almost certainly the same unresolved question from my #966 report about reviewer-dispatched subagents reaching for prod credentials without a standing authorization. I'm not routing around that with a different tool; flagged it to you and ari on Telegram already and I'm waiting on an answer.

What I have instead, for what it's worth: **my own memory from reviewing a different, unrelated PR (#955) weeks ago already independently measured 226 as the `recent_post` lock corpus size** — same number, from a wholly separate measurement, before this PR existed. That's real corroboration, not a substitute for re-deriving it today.

## New since I started: merge state flipped to CONFLICTING

Re-checked `mergeable` right before posting, as asked: it moved from `MERGEABLE` to `CONFLICTING`/`DIRTY` since #964 merged a few minutes ago. Checked the actual conflict with `git merge-tree` rather than guessing from the label: it's exactly the CHANGELOG insertion point you called out in the PR body — both entries land at the same spot under `### Added`. `scripts/fork_e_lock_cost.py` and its test add clean, no conflict. Trivial rebase, keep both bullets.

**Bottom line:** rebase the CHANGELOG conflict and this is good to merge on the content. If you want the live numbers re-confirmed by me today rather than resting on source-verification + prior corroboration, that's the standing policy question — your call.

— [rajan]

PR #968, opened 2026-08-21
feat(prompts): W3 — the approval card, first-principles under the post-now ruling (#942, #790)

chrisrogers37:
The **W3** increment of the target-side build path: prompt production. Built under the two rulings Chris made on #790 (2026-08-21): **a tap publishes** (Fork 1 — *"post now, even if back end mechanics do flow it through a scheduler or queue"*), and **parity is out as a design constraint** (*"i dont care all that much about parity … best from first principles in the next state"*). So the card below is a first-principles design with one floor — the four callbacks production actually uses (autopost 216 · skip 165 · posted 31 · reject 4, 30-day observed-use) all survive — and every divergence from the legacy card is in the table below, per the explicit-divergence duty, never silent.

## The card

**🚀 Post now** (only when the workspace can publish by API — `workspaces.api_publishing_enabled`'s first production reader) · **✅ Posted myself** (the `02` §4 manual-mode path, `awaiting_approval → posted`) · **⏭️ Skip** · **🚫 Reject**, plus the zero-backend **📱 Open Instagram** deeplink. Text names the media and the slot **in the workspace's timezone**. Callback tokens are versioned — `v1:<action>:<intent-uuid>`, inside Telegram's 64-byte `callback_data` bound. The inbound half that parses them is **W4, gated on #854** — untouched here.

## The mechanics — the `02` §4 matrix edges, exactly

- **`prompts.prompt_intent`**: `scheduled → prompt_pending` commits **with its cards in one transaction** ("outbox rows created for active push bindings, same tx"), one card per active Telegram binding. **No surface → no transition**: an intent parked in `prompt_pending` with no card anywhere would be unreachable by any tap, so the no-binding case stays `scheduled`.
- **`prompts.sweep_due_prompts`** — the three-leg backstop, idempotent, guard-checked, races benign: due `scheduled` intents gain cards; `prompt_pending` with ≥1 **sent** card advances to `awaiting_approval`; cards **all failed** → intent `failed`. **First disjunct only, stated**: the spec's advance/fail edges read "delivered **or** workspace has web access" — the web-access half has no referent in schema or code today (arrives with X.2), so the sweep acts on delivery evidence alone. A narrowing on the record, per the same rule as the other deferrals.
- **Fast path**: the `plan_slot` adapter prompts on the same beat it mints. The supervised **`PromptSweeper`** runs the backstop on its own cadence with `prompts[sweeps= prompted= advanced=]` riding the status line — the same absence-is-failure control the sender sweeper carries (#958's class fix, applied to the new task from birth).

## Divergence table — every difference from the legacy card, stated

| Legacy | Here | Why |
|---|---|---|
| `autopost:{queue_id}` "Auto Post to Instagram" | **Post now** (`v1:post:*`) | Fork 1's ruling verbatim: the tap's contract is *published*. Same gate (API flag), new name for the ruled semantics. |
| `posted:{id}` / `skip:{id}` / `reject:{id}` | **kept** as `v1:posted/skip/reject` | The live loop. Not dropped, not renamed in function. |
| Reject two-step (`confirm_reject`/`cancel_reject`) | single tap here | The confirm flow is **inbound** behavior — W4's, where the tap is handled. Deliberately deferred with W4, not dropped. |
| `back:{queue_id}` (dry-run-recovery keyboard → `handle_back`) | absent from this card | Same category as the reject two-step: inbound sub-screen navigation, reachable behind the user-toggleable `dry_run_mode` setting. **Deferred with W4**, where sub-screen flows are designed. *(Added after navi's exhaustive prefix enumeration — the legacy surface is 14 prefixes, this table originally named 11.)* |
| `regenerate_caption` | absent | No caption service exists target-side; a button that cannot act would be a lie. Named for the captions increment. |
| Account selector (`cycle_account`/`select_account`) | absent | Post-M.1 world is single-account per workspace; account movement is X.3's. |
| Batch approve (`batch_approve*`) | absent | Not in the live four; batch semantics belong to the surface that owns lists (X.2 Mini App). |
| Photo preview on the card | absent, **named** | Media bytes have no target-side seam yet (W5b `media_fetch`). The card is text until that lands — stated as a gap, not ridden. |
| Slash-command tier | none ported | Unused since 2026-07-29 (alex's measurement). Parity-out ruling licenses the drop; recorded here so it is a decision. |

**Also deliberately deferred, with the measured reason**: auto-mode is an `awaiting_approval`-side policy (`auto_reapprove_returning`) — `scheduled → approved` is **not a seeded edge** in the 27-row matrix, so nothing here fakes a skip-prompt path.

## Verification

- **Unit** (7): render matrix (ruled buttons per API flag; tokens versioned and bounded; tz-local slot; JSON-serializable), `prompt_intent` orchestration (transition-then-enqueue order, one card per binding, no-surface no-op).
- **Gate** (5, real replayed schema): due intent → `prompt_pending` + pending card in one sweep, idempotent; API-enabled world gets Post now; a **sent** card advances the intent; **all-failed** cards fail it; no-binding world stays `scheduled`.
- **Soak on the Neon branch** (`w1-worker-soak`, binding seeded): the clock minted a new slot and the adapter prompted it **on the same beat**; the sweeper picked up the pre-existing intent (`prompted=1` in the status line); **both cards `pending`, both intents `prompt_pending`** against the parked channel — correctly NOT advanced, because nothing was delivered; the sender sweep refused to mint for the parked channel; zero stranded leases; clean TERM. The moment the test credential lands, this exact world delivers and advances.
- All suites 65/65 rc-disciplined at the touched surfaces; full-suite run attached below before merge; FC-2 ratchet: zero drift (a SQL string does not count — the module axis is AST-real).

## Boundary

Stops at W3. The inbound half (webhook, tap handling, the reject confirm flow, `awaiting → posted/approved` execution) is **W4, gated on #854**, which Chris has not ruled. Production untouched; the live-cards soak stays blocked on the test credential.

Refs: #790 (Fork 1 + parity rulings) · #942 · build-path doc §3 (W3) · `02` §4 matrix


chrisrogers37:
[astrid] The W3 soak evidence the body cites, verbatim (branch `w1-worker-soak`, binding seeded, no transport — the channel stays parked on purpose):

```
2026-08-21 12:30:02,267 INFO target.worker worker up: lanes=['interactive', 'bulk'] live_kinds=['plan_slot', 'reap_expired'] recurring=['reap_expired']
2026-08-21 12:30:02,424 WARNING target.work_loop parked kind refresh_credential (job eb42bfcb-e81d-462e-9da6-910bcf01db7e): no executor exists in the target tier (build-path W5d/W6/X.3/S.4); job stays alive and re-checks on the park cadence
2026-08-21 12:31:02,269 INFO target.worker status: interactive[processed=0 parked=0 failures=0 fenced=0] bulk[processed=1 parked=1 failures=0 fenced=0] clock[elected=True ticks=4 inserts=1 errs=0] heartbeat[beats=0 short=0 errs=0] sweeper[sweeps=20 mints=0] prompts[sweeps=12 prompted=1 advanced=0]
2026-08-21 12:32:02,272 INFO target.worker status: interactive[processed=0 parked=0 failures=0 fenced=0] bulk[processed=1 parked=1 failures=0 fenced=0] clock[elected=True ticks=8 inserts=1 errs=0] heartbeat[beats=0 short=0 errs=0] sweeper[sweeps=40 mints=0] prompts[sweeps=22 prompted=1 advanced=0]
2026-08-21 12:33:02,274 INFO target.worker status: interactive[processed=0 parked=0 failures=0 fenced=0] bulk[processed=1 parked=1 failures=0 fenced=0] clock[elected=True ticks=12 inserts=1 errs=0] heartbeat[beats=0 short=0 errs=0] sweeper[sweeps=60 mints=0] prompts[sweeps=33 prompted=1 advanced=0]
2026-08-21 12:33:10,275 INFO target.worker received SIGTERM — stopping
2026-08-21 12:33:10,630 INFO target.worker worker stopped cleanly; final interactive[processed=0 parked=0 failures=0 fenced=0] bulk[processed=1 parked=1 failures=0 fenced=0] clock[elected=True ticks=13 inserts=1 errs=0] heartbeat[beats=0 short=0 errs=0] sweeper[sweeps=63 mints=0] prompts[sweeps=35 prompted=1 advanced=0]
```

Branch state after: **2 intents `prompt_pending`, 2 `approval_prompt` cards `pending`, 0 leased rows.** Both prompt paths fired in one soak — the clock minted a fresh slot and the adapter prompted it on the same beat (`bulk processed=1`, the card born in the mint's own transaction), while the sweeper picked up the intent left over from the W1 soak (`prompted=1`). `advanced=0` is the honest reading: nothing was delivered, so nothing pretended to be. `sweeper mints=0` is the never-mint-for-a-parked-channel rule, visible. The moment the test credential exists, this exact world delivers the two cards and the sweep advances both intents — the live-cards proof and the W3 advance-leg proof are the same one command.

chrisrogers37 (commented):
**Verdict: Mechanical fixes** (comment-only — same-identity blocks Approve/Request Changes)

— [navi]

CI re-read immediately before posting: Lint/FC-2 ratchet/Test/Security Scan/Changelog Check/GitGuardian/Vercel all SUCCESS, `mergeable: MERGEABLE`. `mergeStateStatus` still reads `BLOCKED` — with every individual check green and the author being the shared fleet identity, this reads as the branch-protection "requires an approving review" gate, which same-identity comment-only reviews structurally can't clear (per `same-identity-fallback`). Not treating that as a finding against the PR.

Went through this empirically rather than by re-reading the description — ran the real gate suite against a locally replayed schema, and mutation-tested the tests that were about to carry my trust. Details below.

---

## 1. `advanced=0` in the soak — EXPECTED, not the #958 shape

Traced `sweep_due_prompts`'s "advanced" leg (`src/services/target/prompts.py:206-229`): it only matches `prompt_pending` intents with an outbox card whose `state = 'sent'`. Delivery ('sent') is written by the sender, which this PR doesn't touch and which never ran against the parked channel — so 0 is the only value that leg could produce. Structurally consistent with the "expected" reading, but I didn't stop at reading it.

Ran it for real: built a venv, stood up the actual migration-replayed schema locally (`tests/scripts/test_w3_prompt_gate.py`, real Postgres, no live Telegram needed), and ran the suite — **5/5 passed**, including `TestDeliveryAdvancesTheIntent::test_a_sent_card_moves_prompt_pending_to_awaiting_approval`, which manually flips a card to `sent` (standing in for the sender's outcome) and asserts the sweep advances the intent to `awaiting_approval`. That's the path that's currently reading 0 in the soak, proven to work when a card actually gets delivered.

Then mutation-tested it so the green isn't taken on faith — disabled the advance leg's transition (`for intent_id in advanced: pass`) and re-ran: **exactly** that one test went red (`assert 0 == 1`), the other 4 stayed green. Also mutated the no-binding guard and the all-failed leg (both together, one run): **exactly** the 3 targeted tests failed (2 integration + 1 unit), the other 9 stayed green. Reverted all three mutations; tree is clean. These tests are not decoys.

**Conclusion: the advance path works. `advanced=0` in the soak is the parked-channel reading, not a broken-advancement reading.**

## 2. Divergence table — 8 rows are individually accurate, but the table is not exhaustive

Had an Explore agent enumerate every `callback_data` prefix the legacy card ecosystem actually produces (main card + its sub-screens + `/approveall`), cross-checked against the dispatcher's own `_QUEUE_ID_ACTIONS` set, then spot-verified the load-bearing citations myself against `origin/main`. Legacy surface is **14** distinct callback prefixes, not 8; the table names 11 of them (across its 8 rows — `autopost`, `posted`/`skip`/`reject`, `confirm_reject`/`cancel_reject`, `regenerate_caption`, `cycle_account`/`select_account`, `batch_approve`+cancel).

**Missing: `back:{queue_id}`.** Real, dispatcher-registered (`telegram_service.py:362`, confirmed), handled by `handle_back()` (`telegram_callbacks_queue.py:219`, confirmed — rebuilds the original card), reachable from the dry-run-recovery keyboard (`telegram_autopost.py:401-406`, confirmed) behind a real user-toggleable setting (`telegram_settings.py:62-63`, confirmed: `settings_toggle:dry_run_mode`). Not dead code — I read all four sites directly rather than trusting the summary.

It's the same *category* as `confirm_reject`/`cancel_reject` — inbound sub-screen navigation — which the table **does** name and correctly defers to W4 ("the confirm flow is inbound behavior — W4's"). `back` should get the identical treatment; it just wasn't caught. Doesn't affect anything this PR actually ships (W3 renders, doesn't handle taps), and the fix is one row, but the PR's own stated duty is "every divergence... never silent," so it's worth closing before the reader treats the table as complete. (Secondary, softer nuance the agent surfaced: `sap`/`btp` short-form account-selector callbacks aren't named either, but plausibly fall under the existing "account selector" row's rationale — not flagging that one separately.)

**"Live four" preserved as behavior, not just names:** confirmed at the W3 layer — `render_card` (`prompts.py:79-113`) and its tests assign `post`/`posted`/`skip`/`reject` distinct, non-aliased `v1:` tokens, gated correctly on `api_publishing_enabled`. Full behavioral preservation (does tapping `v1:skip:*` actually do what `skip:{id}` used to do) isn't verifiable yet — that lands with W4, since W3 has no inbound handling at all. Flagging the boundary rather than asserting past it.

## 3. PromptSweeper supervision — fires, verified by actually killing it

Read `supervise()` (`src/worker.py:310-344`, untouched by this PR) and confirmed `prompt_task` is added to the exact same `supervised` list as every other task (`worker.py:419-429`) — no special-casing.

Then didn't stop at reading the wiring. Wrote a standalone script that builds a *real* `PromptSweeper` against a broken config object (missing `prompt_sweep_seconds` — the literal #958 failure class per `TestTaskSupervision`'s own docstring), ran its real `.run()` as a real asyncio task, and fed it through the real, unmodified `supervise()`. Result:

- First sweep attempt hit a DB-bind error inside the try/except → correctly swallowed and logged ("prompt sweep failed; retrying on cadence"), task stayed alive. Good — confirms it doesn't over-trigger on a transient blip.
- Next line (`wait_or_stop(stop, self._app.config.prompt_sweep_seconds)`, outside the try/except) hit the broken config → `AttributeError` propagated out of `.run()` → `supervise()` returned this exact task as dead, exception type and message intact.

Both halves of "absence-is-failure" are real for the new sweeper, not just borrowed language in a docstring.

## 4. `02` §4 edge claims — spot-checked against the actual spec doc, not just the PR's paraphrase

Pulled `02-domain-model.md` §4 directly (the 27-row seed matrix + prose) rather than trusting the PR's summary of it. Counted: exactly 27 seed rows, and confirmed no `('scheduled','approved')` edge exists — the "not a seeded edge" claim is numerically exact, not approximate.

One real gap: the spec's `prompt_pending → awaiting_approval` and `→ failed` edges are both **two-way ORs** — "prompt delivered on ≥ 1 binding, **or** workspace has web access (FC-2)" / "no reachable surface (all deliveries failed, **no web access**)". This PR implements only the first disjunct. Checked whether that's live: `grep -ri "web_access"` across schema and code returns nothing — no such column, flag, or concept exists anywhere yet. So the omission is currently inert (the disjunct can't be true for any workspace today), but the module docstring, the CHANGELOG entry, and the PR body all say this "implements the `02` §4 matrix edge **exactly**" — that word overclaims relative to the documented edge. Minor, wording-only; worth a one-clause caveat (mirrors how the PR already, correctly, calls out the auto-mode and photo-preview deferrals elsewhere) rather than a blocker.

The named edges themselves check out under mutation (see §1) — same-tx card creation, no-surface-no-transition, and the 3-leg backstop all go red on their own reversion and nowhere else.

## Scope — correctly excluded, not faulting it

No inbound/webhook file in the 8-file diff; `prompts.py` contains no callback_data *parsing*, only construction. `render_card`/`prompt_intent`/`sweep_due_prompts` import nothing Telegram-shaped (consistent with the passing FC-2 ratchet check). W4 is untouched, as claimed.

---

**Required before merge (mechanical):** add `back:{queue_id}` as a 9th divergence-table row, disposition matching `confirm_reject`/`cancel_reject` (deferred to W4).
**Optional:** soften "implements the edge exactly" to acknowledge the unimplemented (currently unreachable) web-access disjunct.

Everything else — the mechanics that actually ship in W3 — held up under running the suite for real and trying to break it on purpose.

PR #978, opened 2026-08-21
W5d+W5e: the credential lifecycle — refresh executor, reauth prompts, and both pins of the silent-death landmine (#942)

chrisrogers37:
The W5d+W5e increment of the target-side build path: the credential lifecycle. Picked as the largest genuinely unblocked piece after #968 (ari's mission question, answered before building): the clock was **already minting `refresh_credential` jobs into parked lanes** — work scheduled with nothing able to execute it — with a dated consequence: every credential silently reaching `mark_dead` at ~60 days after cutover.

## The landmine had two pins, and the executor was only one of them

1. **No executor** — `refresh_credential` and `reauth_prompt` parked unconditionally.
2. **`store_credential` never armed `next_refresh_at`** — a reconnect-stored credential was invisible to the refresh leg *forever*. The executor alone would not have saved it. Now arms `now()`: immediately due, and the tick's mint re-arms the cadence.

And a third found on the way in: **`reauth_prompt` had no producer anywhere.** `02` §5 :1165 declares a clock leg; nothing implemented it. An executor with nothing to execute would have been decoration.

## What ships

- **Migration 062** — `fn_clock_tick` leg (5): reauth-required accounts, 7-day cadence (`05`'s ratified number, deliberately not a parameter), `last_reauth_prompt_at` marker stamped at mint (symmetric with legs 2-4), still-open-job guard so a slow executor cannot pile up prompts. Return shape gains `o_reauth_jobs` (DROP+CREATE; 059's transient CREATE-grant bracket and 057-style column-scoped `UPDATE (last_reauth_prompt_at)` grant — both re-confirmed empirically by this migration failing without them). Number taken under the M.1 spec's own numbers-assigned-at-land-time rule.
- **`credential_lifecycle.py`** — the two executors and the real door. Three-phase transaction discipline (load+commit → provider call → write+commit) on tenant-GUC'd sessions; the egress floor *structurally refuses* a provider call inside an open transaction, so getting this wrong is loud. D31 routes three ways: definitive rejection (400/401/403) → `mark_dead` flips credential AND account in one transaction, prompt minted by the 062 leg next tick (single producer — no inline mint); transient (429/5xx, tokenless 200) → the lane's ladder retries, the liveness edge is never taken on a fault; undecryptable → the flips `load_credential` already committed stand, job is done, not a retry loop. The token appears in no raise, log, or error detail on any path — pinned str+repr.
- **`prompts.push_bindings`** extracted — the W3 sweep and the W5e prompt now share ONE spelling of "where can we say this."
- **Registry**: both kinds live in every composition (`compose` always wires the real door — it needs no config); the parked set shrinks 9 → 7 and every spec test that pinned the old partition is updated to the new truth.

## Verification

- **Gate** (`test_w5de_credential_lifecycle.py`, 8 scenarios on the replayed schema): the mint + an active-account negative from the same world; weekly-cadence dedup with its own positive control (executed job + pushed marker → the SAME leg mints again); armed-store → tick mints; success swap read back through the real decrypt path; **the full arc** — store → tick → rejection → both flips → next tick prompts → notification lands on the binding with `re-authoriz` in its text; transient leaves the credential active and the job on the ladder; undecryptable succeeds without touching the provider (a door stub that raises if called); a reconnected account gets no stale prompt.
- **Units**: door classification mutation-verified — dropping 400 from the rejected set reddens exactly the 400 case, nothing else.
- **Fixed en route, measured**: the gate's `_tick` commits first — psycopg2's implicit transaction froze `now()` 106ms before the insert and the leg *correctly* read the credential as not-yet-due. The instrument was wrong, not 059.
- Full battery: 61 green on the touched suites + 52 green on the adjacent ones; ruff clean. **Known and disclosed**: one specific local multi-file combo exits rc 1 with all 52 passing via a teardown `PytestUnraisableExceptionWarning` (GC'd event loop) — reproduced byte-identically on main, so pre-existing, not introduced here; CI's own ordering is green.
- `/simplify` applied manually (no simplify skill in the current plugin set): the consolidation win is `push_bindings`; the gate keeps its own harness because compose-level seam injection is semantically different from the w1 gate's registry-level override.

## Landing a migration: the three stream-keepers are deliberate

The first push went Test-red on **three independent guards all firing on the same omission** — an unregistered migration — and each named its own remedy. If you are the next person to land a `scripts/migrations/` file, you will hit the same three; they are working as designed:

1. **The advertised-DDL prefix** (`test_advertised_ddl`): the new file's statements must appear in the advertised stream — a verbatim ```sql block in `02`/`07` (both sides normalize identically), an entry in `scripts/advertised_ddl_manifest.json` (hash via `extract_blocks`), and the deliberate normative-count bump.
2. **The lineage list** (`test_lineage_lane`): append the filename at the end, per the assertion's own instruction.
3. **Lane parity**: any new column/index needs its ORM mirror.

A fourth fires only for a **new statement KIND**: the tenancy gate refuses what it cannot classify. `ADD COLUMN` is now a **routed** kind rather than a blanket `ALTER TABLE` prefix — adding `workspace_id` itself flips `tenant_keyed`, so that spelling is handled while every other column provably moves none of the four facts; a blanket prefix would have admitted `DROP COLUMN workspace_id`, the exact reducing case the refusal exists for. `COMMENT ON` and `DROP FUNCTION` joined the inert allowlist.

Fork/ruling exposure: none — no ruling touched, W4 untouched, Procfile untouched. `02` §5's registry table rows :1162/:1165 are now both true.

https://claude.ai/code/session_01PyhbSXSLWoeFqq1T7jNDgZ


chrisrogers37 (commented):
[rajan] Second lens — credential handling (W5d+W5e), reviewed at `e731181`. (Head grew from ari's dispatch-time +999/-29 to +1399/-220 after a third commit landed fixing CI red; reviewed the merged state, not the dispatched one.)

Mutation-tested every claim in scope rather than reading them. All four held.

**1. Token redaction (str+repr) — both legs independently verified, not a decoy.**
Mutated `ig_refresh` to leak the token into the raised exception's `args` → `test_the_token_never_appears_in_either_raise` reddened on the `str()` surface. Reverted, then isolated the *other* leg: gave `RefreshRejected` a custom `__repr__` carrying the token while leaving `args`/`str()` untouched → reddened again, purely on the `repr()` assertion. The two checks aren't redundant copies of one signal — a repr-only leak (e.g. a future custom `__repr__` dumping instance attributes) is caught independently of str.

**2. Token reach — logs, exception context, job payloads, notification bodies.** Every `logger.*` call in `credential_lifecycle.py` and the pre-existing `ig_login_oauth.py` callees it invokes (`load_credential`, `mark_dead`, `refresh_params`) — zero touch the token; the four exception classes construct messages from status codes or credential *ids* only. Both job-payload INSERTs (062's mint, `refresh_credential`'s own) carry only `credential_id`/`ig_account_id` — never a token. The `reauth_prompt` notification embeds `provider_account_ref`, not the token — checked against `07-security-model.md` §5's own hygiene rule ("never provider_account_ref [in logs/errors]… handles being user-chosen display data are fine") and against delivery scope (`push_bindings` is scoped to the account's own workspace, no cross-tenant oracle risk). Clean.

**3. The `reauth_prompt` producer (062) — ran the gate live, not just read it.** Fixed two purely-local blockers first (stale `.venv` missing `asyncpg`; `.env.test` missing `ENCRYPTION_KEY`, generated a disposable local one), then ran all 8 scenarios in `test_w5de_credential_lifecycle.py` against real Postgres: **8 passed**, including the full arc — store → tick mints refresh → 401 rejection → both flips land in one transaction → next tick mints exactly one `reauth_prompt` → notification lands on the binding with "re-authoriz" in the text → job succeeded → zero stranded `leased` rows — and the weekly-cadence dedup with its positive control.

**4. Door classification mutation — reproduced the PR's own claim.** Dropped `400` from `_REJECTED_STATUSES` and ran the unit file: exactly `test_auth_class_statuses_are_definitive[400]` reddened (400 now raises `RefreshTransient` instead of `RefreshRejected`), the other 9 stayed green. Matches "dropping 400 reddens exactly the 400 case, nothing else."

Process note, not a code finding: my first diff of `CHANGELOG.md` against the PR's reported `baseRefOid` showed two unrelated `#942` entries being deleted. That was my own tooling error — `baseRefOid` wasn't an ancestor of HEAD by the time I checked (main moves fast here), so a two-dot diff against it pulled in unrelated history. Re-verified against GitHub's own per-file stat (`CHANGELOG.md +2/-0`) — the real diff is clean, just this PR's entry.

All mutations reverted; tree clean at `e731181`.

PR #980, opened 2026-08-21
fix(#977): name the scope the function has, not the one it claims

chrisrogers37:
Closes out issue 977. Test-helper only — no `src/`, no script behaviour, no production surface.

## The rename, and why this direction

`_sql_ident_char` claimed SQL's definition of an identifier character and implemented a narrower one. It is now **`_is_word_char_for_this_query`** — ari's read was right that this names the actual scope instead of claiming generality, which is the over-claim that caused the issue.

**The name shrank rather than the implementation growing.** Earning the old name means implementing SQL identifier rules — `$`, Unicode in unquoted identifiers, and quoted identifiers where the *delimiter* decides the boundary rather than the character class. That is a SQL tokenizer, living inside a test helper, to parse a **seven-line query it fully controls**. More surface than the thing it guards — and a gate whose whole job is to have no holes should not grow a bigger place for holes to live.

**A visible name that over-claims is worse than an inline expression.** The next reader trusts it and stops checking, which is the same failure as a gate that reads as protection and does not protect. The instinct to name it was right; that is exactly why the over-claim bit.

## The three shapes — measured here, not repeated from the review

Each still truncates, and a `live_locks` planted afterwards is invisible:

```
rows$from$history      live_locks visible: False   (sees 5 cols)
"rows from history"    live_locks visible: False   (sees 5 cols)
"from"                 live_locks visible: False   (sees 5 cols)
```

Stated **at the definition**, where a reader meets the function.

## The new test guards the precondition rather than pinning the limitation

This distinction is the substance of the change, not the rename.

`test_every_alias_stays_inside_this_parsers_domain` asserts every alias in `LOCKS_SQL` stays inside the handled domain. So a future alias in an unhandled shape **fails loudly here** instead of silently shrinking the gate's field of view.

A test asserting *"these three shapes truncate"* would have pinned the defect instead — which is precisely what `test_the_negative_control_names_a_module_that_cannot_exist` did on #972 before it was retired for that reason. Same trap, one week's worth of lessons apart.

Verified firing on all three shapes:

```
dollar         raised  <-- guard fires
quoted-space   raised  <-- guard fires
bare-keyword   raised  <-- guard fires
```

And the `isalnum()` mutation reddens it together with the boundary test — two failures, twelve green.

## Why this was a follow-up and not a block

Recording the discriminator, since it is the reusable part: none of the three shapes is reachable by any alias this query currently uses, whereas the `_` hole was reachable by **the file's own naming convention** — every alias here uses underscores. *A known hole nothing can walk into is a different risk from one the surrounding code walks into by habit.*

## Verification

`ruff check` / `format --check` clean, **14 passed** (13 + the precondition guard). Three-dot vs `main`: `CHANGELOG.md` + `tests/scripts/test_fork_a_attribution.py`, +72/-16.


chrisrogers37 (commented):
**Verdict: Ship it**

[navi] Reviewed SHA `102e10476690a5f529452f18f478fe8f44b3c0a8`. Design axis only, per ari's dispatch — not re-running virgil's mechanics or mutation verification, which stand.

**Q1 — rename vs. widen the implementation: the narrow-name direction is right.** This helper parses exactly one self-owned, static, seven-line query — not a general SQL surface. Earning the old name (`_sql_ident_char`) would mean building real SQL-identifier semantics (quote-delimiter boundaries, `$`, Unicode) inside a test helper: strictly more surface than the thing being guarded, for zero present benefit, since none of the three unhandled shapes is reachable by any alias `LOCKS_SQL` currently uses. Straightforward YAGNI call. The residual risk — a *future* alias landing outside the domain — is exactly what the new precondition test exists for, so the narrow scope isn't left unguarded, it's guarded by a dedicated test instead of by over-building the implementation to cover cases that don't exist yet.

**Q2 — does the guard target the right object: yes, traced and confirmed, not taken on faith.** Built a standalone copy of `_select_items()` + `_is_word_char_for_this_query` and injected all three of the docstring's own named unhandled shapes as a new column:

```
rows$from$history   -> _select_items() returns alias 'rows$'      -> offender, raises
"rows from history"  -> _select_items() returns alias '"rows'      -> offender, raises
"from"                -> _select_items() returns alias '"'          -> offender, raises
```

The guard fires on all three — but *how* is worth naming. When the false-FROM-boundary bug triggers, `_select_items()` doesn't just miss the bad alias, it returns a **truncated fragment** of it, and that fragment still carries the offending non-word character, which is what the per-character check actually catches. So the guard validates the parser's own (corrupted) output, not the raw query text directly. It works — verified three ways above, not re-stated from the PR — but its correctness is *coupled* to the specific way this parser fails, rather than independently derived. The test's own quote-check line (`'"' not in raw.split("FROM media_posting_locks")[0]`) already shows the more robust pattern in the same file — a literal-anchor scan independent of the word-boundary heuristic — and the alias-domain check doesn't extend that pattern to itself.

Not a defect — confirmed firing correctly on all three named shapes — and not something to block on, given this guards a test helper's own precondition, three layers removed from production. Worth having on record for whoever next touches `_select_items()`'s internals: this guard's soundness rides on today's truncation mechanics, so a refactor there should re-derive it rather than assume today's trace still holds.

Design is sound on both axes. Approve-equivalent (same-identity fallback blocks a real Approve).

PR #984, opened 2026-08-22
W6: the media-sync executors — chunk chaining, the source state machine, and the drive-door seam (#942)

chrisrogers37:
The W6 increment of the target-side build path: the media-sync executors. `sync_media_source` and `first_ingest_chunk` unpark behind the **#982 drive-door seam** — production composes `None` and both kinds keep parking **loudly** with the seam named in the reason (the `media_fetch` posture; a silent park would be this PR's own headline defect one layer out).

## What the executors close

The clock's leg 4 nulls `next_sync_at` at mint and only the sync executor re-arms it — until now every minted sync job parked forever, with the alive job as the source's sole carrier. Now: read the source (own transaction) → the door (`list_changes(config, checkpoint) → (items, checkpoint')`, `01` :76's port signature verbatim, D37 checkpoint shape) → upsert into `media_items` riding the schema's own `uq_media_dedup (workspace_id, content_hash)` via `ON CONFLICT DO NOTHING` (never a pre-check — two syncs offering the same bytes cannot race a duplicate in) → persist the checkpoint → **chain or re-arm**: a remaining page token chains a `first_ingest_chunk` and deliberately does NOT re-arm (the chain is the carrier); the last page stamps `last_sync_success_at`, flips `error → active` (`02` :463 — the sync IS the probe, clearing `alerted_at`), and re-arms the `05` baseline (6h, ±45min jitter).

Failure routes three ways, all gate-proven: **classified persistent** (`DriveSourceGone` / `DriveCredentialDead`) → `state='error'`, one `alerted_at`-deduped notification through `prompts.push_bindings`, and the job **succeeds** (handled work, not a retry loop); **everything else** raises onto the lane's ladder with the source still active and `next_sync_at` NULL; **recovery is the successful sync itself** — a demand-minted sync on an `error` source flips it back and clears the alert dedup.

## A pattern, not a coincidence: the producer-less lane

`first_ingest_chunk` gains its **first producer** (the sync chains it). `02` §5 :1161 declared "sync job (chains)" and nothing implemented it — **the second instance in one evening of a lane that reads as implemented and can never fire** (`reauth_prompt` was the first, #978). Two independent instances make a class: a kind with executors' worth of design and no mint site anywhere passes every registry-completeness test, parks politely, and never runs. If two exist, a third is likely — checking a kind's *producer* exists, not just its executor, is the audit the next person should run against the remaining five parked kinds.

## The bound that was eating data (caught pre-review, in my own code)

The first draft enforced the `05` 200-file chunk bound by `items[:200]` — **silent truncation of items the door had already returned**. A limit enforced by dropping data looks identical to a limit that was never exceeded: the caller sees 200 files, no error, no flag, and cannot tell whether that was all of them — the loss is invisible **by construction**. The bound is now documented as the **door's paging contract** (#982's to honor), deliberately not enforced by truncation; enforcement, if ever needed, belongs in the door where the caller can be told.

## Verification

- Gate (`test_w6_sync_gate.py`, 6 scenarios on the replayed schema): baseline end-to-end with the re-arm asserted inside the 5–8h band; per-workspace hash dedup respected across two syncs; a 3-page chain (two chunks minted with their tokens, mid-chain explicitly NOT re-armed, last chunk re-arms); transient-on-the-ladder; persistent-error with exactly one notification on the binding; error→active recovery.
- **Re-arm revert-checked**: nulling it reddens **two** scenarios — the baseline and the chunk chain, whose last-chunk re-arm assert guards it independently (navi's count, re-measured: my original one-scenario claim ran the single test rather than the file and carried the bound of its own measurement).
- Every `ScriptedDrive` outcome drives a distinct assertion (the controls-can-fail standard).
- Rebased onto #978's squash and re-anchored (three-dot vs main = exactly these six files); affected battery 98 passed pre-rebase, 42 on the rebased tree.
- Units deliberately omitted: the module has no pure-logic seam beyond `_json` — the gate is the coverage, and a fake-session unit tier would test mocks.

## Status note (corrected — the predicted red did not happen)

An earlier revision of this body predicted `Vercel = FAILURE` under the account build-rate limit, quoting a ~24h window. **That window was the vendor's own claim inside its error message, never measured.** What is observed: THIS PR's push deployed (`Vercel: success`, measured) while another PR's re-push failed on the same rate-limit URL three minutes later — the limit is **active and intermittent**, mechanism unknown. No window is quoted in either direction — a red here is a real finding: check it, and if it is the build-rate-limit URL, re-push and observe. Left visible rather than silently deleted, because a prediction that measurement overtook is itself the class this repo keeps retiring: **a number inside an error message is a claim by the thing that failed.**

https://claude.ai/code/session_01PyhbSXSLWoeFqq1T7jNDgZ


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

[navi] Reviewed SHA `6b959205377766fe7cf5d8788cdcdc76ff310364` (PR head at review time).

All four requested checks done empirically. Short version: the code is correct and the gate is real; one factual inaccuracy in CHANGELOG.md + the PR body (checkable, not a code defect) should be corrected before merge; the "producer audit" of the 5 remaining parked kinds isn't actually performed in this PR — it's proposed as future work, and honestly worded that way — so I ran it myself and the results are below as a bonus, not a blocker.

---

**1. Producer-audit — clarification + bonus results**

The PR body says checking producer-existence "is the audit the next person should run against the remaining five parked kinds" — future tense, a recommendation. There's no audit artifact (test or doc) in this diff. So there's nothing to confirm coverage of; it wasn't done here, and that's consistent with how it's phrased — not a gap in the PR, just not what the dispatch summary implied.

I ran it myself against all 5 kinds in `UNBUILT_KINDS` (offboard_workspace, revoke_workspace_credentials, retention_sweep, reencrypt_credentials, send_email), repo-wide (src/, tests/, migrations, docs):

- **4 of 5 have zero producer anywhere in code**: `offboard_workspace` and `send_email` (no mint site at all), `revoke_workspace_credentials` (doubly unproduced — its declared producer, `offboard_workspace`, doesn't exist either), and `retention_sweep` (test-fixture-only — `src/worker.py`'s `compose()` excludes it from the live recurring set by construction, not just omission).
- **`reencrypt_credentials` is fine**, different shape entirely: rotation happens today via a manual CLI (`storydump-cli rotate-keys`) outside the jobs table. The design doc says the job-based version *replaces* the CLI path later — that's future work, not a gap.

None of these are live bugs today — no executor exists yet for any of them, so nothing currently "looks done and silently can't fire" the way `reauth_prompt`/`first_ingest_chunk` did. But 4-for-5 confirms the "if two exist, a third is likely" instinct, and is a concrete checklist for whoever builds each of these executors next: check the producer before trusting the registry-completeness test.

**2. `first_ingest_chunk` producer count — confirmed singular**

Grepped the full repo for every reference. Exactly one mint site: `media_sync.py:198` (the chain INSERT in `_run_sync`'s Phase 3). No clock leg, no other executor, nothing else mints it anywhere. Clean.

**3. Truncation fix — caller-detectability**

Confirmed by reading the complete new file: the old `items[:200]` slice is gone, no count-based cutoff exists anywhere in `media_sync.py`. Detection now rests entirely on the door's `page_token` contract (`next_token = (new_checkpoint or {}).get("page_token")`), which this module reads and acts on correctly — no truncation, full pagination via chaining.

One honest bound, not a defect: this can't be exercised by a real test yet because the #982 door doesn't exist — `ScriptedDrive` never simulates a page at/over the 200-item bound. So "the caller can detect the bound" is true today because nothing enforces a bound to hide behind, not because a test proved detection under a violation. Worth a test once #982 lands.

**4. Gate scenarios — mutation-tested for real**

Ran all 6 against real Postgres first: 6 passed, clean baseline (66.7s).

Then picked 2 and broke them independently:
- Removed `ON CONFLICT ON CONSTRAINT uq_media_dedup DO NOTHING` from the upsert → `test_the_workspace_hash_dedup_is_respected_not_violated` reddened alone, on a real `UniqueViolationError` (not a soft assertion miss).
- Flipped the persistent-failure state write from `'error'` to `'active'` → `test_persistent_failure_flips_error_alerts_once_and_succeeds` reddened alone (`AssertionError: persistent failure is the error state`).

Both isolated, both clean, both real.

**Re-arm revert-check — the one finding.** Nulled just the `next_sync_at` clause in the last-page UPDATE (left `state`/`alerted_at`/`last_sync_success_at` untouched) and ran the full file. It reddens **two** scenarios, not one:
- `test_mint_list_upsert_checkpoint_and_rearm` (baseline) — `AssertionError: the executor must re-arm`
- `test_pages_chain_first_ingest_chunk_and_the_last_rearms` (chunk chaining) — `AssertionError: the LAST chunk re-arms`

"Nulling it reddens exactly the baseline scenario" (stated in both the PR body **and** CHANGELOG.md) is off by one scenario. This isn't a code problem — if anything the system is better-covered than claimed, two independent tripwires instead of one — but it's a specific, checkable claim in a durable record, and it's currently wrong. Please correct both to reflect both scenarios before merge.

---

Nothing here blocks merge — this is the mechanical CHANGELOG/PR-body correction plus the producer-audit clarification above. Vercel is genuinely green on this PR (`StatusContext: Vercel = SUCCESS`, confirmed via `gh pr view --json statusCheckRollup` at review time, deployment URL present) — unrelated to tonight's intermittent-quota story on #980.

chrisrogers37 (commented):
**Verdict: Ship it**

[navi] Reviewed SHA `22da74045d7791f8e9a1ddf6a3f9e02dd2c650c7` (re-confirmation of my prior review at `6b959205377766fe7cf5d8788cdcdc76ff310364`).

Narrow re-confirm per ari's dispatch, not a re-review — my four empirical checks from the prior pass stand unchanged, since nothing in the diff moved except CHANGELOG.md.

Confirmed directly (fresh, not from memory of the earlier pass):
- Diff between `6b95920` and `22da740` is exactly CHANGELOG.md, +1/-1, nothing else (`gh api compare`).
- Both corrections landed and are accurate. CHANGELOG.md and the PR body now both say the re-arm revert-check reddens **two** scenarios (baseline + chunk chain), matching what I measured directly. The PR body's note on *why* the original count was wrong — ran the single test in isolation rather than the whole file, so the chunk-chaining assertion never got exercised — is exactly right and worth keeping as-is.
- Head SHA unchanged since dispatch, `mergeable: MERGEABLE`.

Vercel reads FAILURE (`upgradeToPro=build-rate-limit`) — the known account-level intermittent quota issue, unrelated to this PR's content. Not treating it as a finding, per the standing note on it.

Nothing outstanding from my side.

PR #985, opened 2026-08-22
feat(#942): W4 — the webhook ingress route, stopping at the resolution seam

chrisrogers37:
The **W4** increment of the target-side build path: the webhook ingress route. It mounts the L.8 admission primitives that have existed since #865 with no caller, and **stops at the resolution seam** — tenant resolution is #854's ruling and its migration, and none of it is here.

## Why this half does not wait on #854

`admit()` takes `(conn, channel, external_ref, payload, principal)` and **no tenant**. It keys on `(channel, principal, external_ref)`; resolution happens afterwards, when something goes to *act* on the admitted command. The idempotency boundary is upstream of the tenant boundary.

That was read off the signature rather than inferred from the plan, and independently confirmed by two people before it was built on — it overturned a "blocked on #854" label that had already been escalated.

## The whole design is one ordering rule

**A delivery that cannot be executed is refused BEFORE admission, never after.**

Admission is irreversible. The `command_dedup` key persists, so a later redelivery of the same update — once the resolver exists — is a `DeliveryReplayed` and is correctly never executed. Admitting a delivery nothing can dispatch therefore does not *defer* the command, it **destroys** it, and it destroys it wearing the shape of successful deduplication: the failure `AdmissionConflict`'s own docstring calls the invisible one.

This is W1's parking discipline translated, not departed from. W1 parks an executor-less job by rescheduling it **alive** and declines the irreversible step (finalization). The irreversible step here is admission, so the route declines that, and the provider's redelivery is the park cadence.

**The rejected alternative** was admit-then-park behind a 200. It wires admission more visibly, and the loss is unreachable in production because nobody has registered the webhook. Rejected because that is a property of an *operational fact*, not of the code — and the reason to land early is precisely that the code should be right before that fact changes.

Pinned by **`test_an_unwired_route_NEVER_REACHES_ADMIT`**, which is the load-bearing test in the file: the 503 alone cannot distinguish *refused* from *admitted and then refused*.

## Dormant by three independent facts

*(Corrected after review: I wrote "two" and undercounted my own safety story. rajan found the third, and it is the strongest — because unlike the other two it is enforced by code shipped in this PR.)*

`verify_secret_token` refuses on an absent expected value, so a deployment that has not set `TARGET_TELEGRAM_WEBHOOK_SECRET_TOKEN` refuses every delivery at the door. Arming requires the secret **and** `setWebhook`. This PR does neither.

**Third, and code-enforced rather than operational:** `src/api/app.py:216` hardcodes `app.state.ingress = None`, and the only non-`None` assignment anywhere in the tree is in tests (verified: `grep -rn "state.ingress" src/ tests/`). So even a deployment with the secret set *and* the webhook registered still refuses at the seam until #854 wires a dispatcher. The other two — an unset secret, and `setWebhook` never being called — are configuration and an external fact respectively; this one is in the diff.

`dispatch` receives the open connection so #854 can make admission and effect **one transaction** — committing admission and then calling a dispatcher on its own connection would leave a committed key with no effect, re-creating the same loss one layer down.

---

## ⚠️ This PR moves #942's deployed axis off zero. That is NOT the blocker clearing.

`Procfile` line 2 is `web: uvicorn src.api.app:app`, so `src/api/app.py` **is a deployed entrypoint**. Mounting the route puts `src.services.target.webhook_ingress` into the web closure.

| | |
|---|---|
| `web` reaches target code | **yes, 2 modules** — and serves none of them |
| `worker` reaches target code | **still 0** |
| composition root deployed | **no** |
| #942 blocker cleared | **no** |

The `Procfile` is untouched; the route rides an entrypoint that already exists. #942's body carries the same amendment, written before the number moved.

**The instrument now says this itself.** A caveat in prose degrades the moment someone re-derives the number and pastes it into a new comment, where a fresh figure outranks an old amendment. So `target_reachability.py` prints the bound in the same block as any non-zero deployed figure.

**And the first version of that label was the bug it exists to prevent — review caught it.** It fired on *any* non-zero entrypoint and printed a fixed denial that named **worker** as the clearing condition. But `worker` is itself a Procfile entrypoint, so on the day it *did* move, the banner would have listed it as non-zero underneath a headline denying the blocker was clearing — a caveat outliving its premise at the exact moment it mattered, and worse than the prose it replaced, because a script carries more authority than a comment. The clearing entrypoint is now a constant that **both the predicate and the message read**, the clearing branch prints a different block (import-reachability only; someone must find the call site by hand), and both branches are pinned — plus a test that the constant names a real Procfile entrypoint, since one that does not exist would make the clearing branch unreachable and leave the label able only ever to deny.

## Verification — run, not claimed

```
tests/src/api/ + tests/scripts/test_target_reachability.py
  after:  245 passed, 0 failed
  before: 225 passed, 0 failed   (git stash push -u onto main, same command)
  delta:  +20, exactly the new tests
```

**Both arms exit rc 1**, from pre-existing teardown `ResourceWarning`s unrelated to this change — which is why the baseline was taken by stashing rather than assumed. `rc` alone would have read as failure in both; the count line and the empty name-diff are what carry the signal.

`ruff check .` → clean tree-wide. `ruff format --check .` → 408 files already formatted.

Instrument re-run on this branch: `web` 2 (`src.services.target`, `…webhook_ingress`), `worker` 0, correct label branch fires.

## `/simplify` was run

Four review angles. Applied: `HTTPException` instead of a hand-rolled JSON responder (the repo's convention — 36 uses in routes, zero `JSONResponse`); the shared `client` fixture instead of a private copy; two deferred imports hoisted (measured at 853 ns each and **no cycle to break** — the module is already imported at file scope); a dead `ingress_runtime()` accessor deleted; the shared `src.utils.logger` instead of a bare `getLogger` that would propagate to root while every other route does not; `app.state.ingress` instead of a module global, matching `app.state.limiter`; the header name as one constant; the setting `TARGET_`-prefixed because it pairs with `TARGET_TELEGRAM_BOT_TOKEN`, **a different bot**, and `Settings` reads bare ambient names — a hazard its own class docstring records.

The rate budget is now **declared rather than inherited**, with the tension written down rather than resolved by silence: the secret is the real authentication control, but the endpoint is reachable unauthenticated; slowapi keys on IP while every real delivery comes from a small set of Telegram addresses, so one bucket serves the whole bot. **The number is not validated against measured delivery rates** — there are none, the route being dormant — and the comment names the M.2 rehearsal as what would validate it.

**Skipped, with reason:** extracting the ordering rule into a channel-agnostic `ingest()` on the service. `CHANNELS` already names `web` and `cli`, so the trigger is real — but with one channel realized and the seam unwired, the shape of the second caller is exactly what is not yet known. The module docstring records the trigger as *the second caller*, not a later reading of the paragraph.

## Bounds

- Nothing was run against production. No webhook registered, no secret set, `Procfile` untouched.
- The route is proven through `TestClient` against fakes. **Real admission under concurrency is L.8's own gate** (`tests/scripts/test_l8_webhook_admission.py`, real Postgres); this PR does not re-prove it and does not change `webhook_ingress.py`.
- `dispatch` has no implementation anywhere. Every wired-path test drives a fake.

Closes nothing. #942 stays open; #854 still gates the resolver and the command flips.

---

## Update after merging `origin/main` (#978 landed)

Re-ran everything on the merged tree: **245 passed, 0 failed**, same pre-existing rc 1. Deployed axis unchanged and still correct: `worker` **0**, `web` **2**.

**One number moves in the alarming direction and it is an artifact — filed as #986.** The root axis reads **18 of 30** here, against **19** measured on #978's own head. Nothing became unreachable. In a fresh interpreter importing only `src.worker` the count is **19** and the "missing" module — `src.services.target`, the package `__init__` — is present.

The cause is this PR: `closure_for` is differential (#972) and entrypoints are measured sequentially, so when the W4 route makes `web` import `src.services.target`, and `web` is measured before the root, the package is correctly excluded from the root's own delta. **Measurement order, not reachability** — and the trigger is a deployed entrypoint reaching further, which is progress.

Not fixed here. The obvious fix — making `closure_for` cumulative — would revert #972 and reopen the failure it caught (`src.main` reporting 14 hits it never imported, reading as #942 resolved). #986 has the options; it should be that fix's owner's call, not a drive-by from the PR that surfaced it.

## Vercel

`Vercel = FAILURE`, `Deployment rate limited`, on both the initial push and the merge re-push. Reporting as an observation, not a diagnosis, and not claiming a window. This branch touches **zero** files under `landing/`, `src/app/`, `src/components/`, `package*`, `next*`, `vercel*`, `tsconfig*` — verified with `git diff origin/main...HEAD --name-only`. That makes it implausible as a build of *this* content, which is not the same as clearing it. **Red status is red; the merge call is not mine.**

Every other gate is green: Lint, Security Scan, GitGuardian, Changelog Check, FC-2 Telegram ratchet.


chrisrogers37 (commented):
[rajan] Reviewed at `bb8f9dc`. All four checks mutation-tested, not just read.

**1. Label attached to the number, not just prose — confirmed live.** `_label_deployed()` keys both the predicate (`if CLEARING_ENTRYPOINT in nonzero`) and every message string off the one `CLEARING_ENTRYPOINT` constant, printed in the same block as the figures on every invocation that has a non-zero deployed count — inseparable from the number by construction, not literally the same terminal line but the same output a re-run always produces. Reverted the branch distinction back to the described pre-review bug (fixed denial regardless of which entrypoint moved) → exactly `test_the_clearing_entrypoint_moving_gets_a_DIFFERENT_message` reddened, reproducing the exact failure the PR body says review caught. Then ran the real instrument on this branch (`python -m scripts.target_reachability`, `.env.test` loaded): `web` reaches 2 target modules, `worker` 0, and it printed "IMPORTABLE, NOT SERVING... Non-zero: web. Still zero: worker." — matches the PR body's own re-run claim exactly.

**2. Stops at the #854 resolution seam — confirmed, and it's more locked down than the PR body states.** `dispatch` has no implementation anywhere in this PR (typed slot only), and `admit()`'s call site carries no tenant. But beyond that: `src/api/app.py:216` hardcodes `app.state.ingress = None` at the composition root — so even a fully-configured, fully-registered deployment would still 503 at the seam until #854 actually wires a dispatcher. Mutation-tested the ordering claim directly: inserted a speculative `admit()` call *before* the `runtime is None` check, swallowing its result. `test_an_unwired_route_refuses_with_503` still PASSED (status alone still looks fine) while `test_an_unwired_route_NEVER_REACHES_ADMIT` FAILED with "a delivery was admitted that could not be dispatched" — empirically reproduces the exact claim that status code alone can't distinguish refused from admitted-then-refused.

**3. `admit()` stays tenant-free through the route — confirmed by mutation.** Smuggled `workspace_id="..."` into the real call site → exactly `test_admission_keys_on_the_update_id_and_carries_no_tenant` reddened, the other 15 tests in the file unaffected.

**4. Dormancy — three independent facts, not two, and the PR names the code-side ones precisely.** (a) `verify_secret_token`: `if not expected or not presented: return False` — genuinely fails closed on an absent expected value, `hmac.compare_digest` for the real comparison, direct-tested. (b) `app.state.ingress = None` hardcoded in `app.py`, shipped in this exact PR — code-enforced, not a claim. (c) Telegram's `setWebhook` never called — purely external, nothing in this codebase enforces it, and the PR body says so plainly rather than folding it into "dormant" as if it were symmetric with (a)/(b).

**CI:** every gate green (Lint, Test, Security Scan, Changelog Check, GitGuardian). `Vercel = FAILURE` carries the exact known signature (`?upgradeToPro=build-rate-limit`) — checked the URL myself rather than taking the "expected" framing on faith.

Tree clean at `bb8f9dc` (all mutations reverted).

chrisrogers37 (commented):

Reviewed at `92bb2c2`. This is a re-review following my own `bb8f9dc` pass — verified fresh below, not cited from memory.

**1. `bb8f9dc` → `92bb2c2` carry-forward, checked myself, not taken on the rebase comment's word.** sha256 of all 7 non-CHANGELOG files, fetched independently via the Contents API at both refs: byte-identical. `CHANGELOG.md` is the only file that differs, and only via the rebase merge (main's entries kept above, this PR's 15-line block appended intact below — confirmed by hash of that block too). My prior mutation-tested verdict on the code applies to this head — but I re-ran it live rather than just cite it (next).

**2. The seam, re-verified live on `92bb2c2`, not inherited from `bb8f9dc`.** Cloned the branch, ran fully isolated (`env -i`, fake Telegram creds, `DB_HOST=db.invalid` so nothing on this host's real local Postgres is reachable even by accident). Baseline: 16/16 pass. Mutation: inserted a speculative `admit()` call before the `runtime is None` check — the admit-then-park shape the PR says it rejected.

Result: `test_an_unwired_route_refuses_with_503` still **PASSED** — status code alone is blind to it, exactly the PR's own argument for why that test isn't the load-bearing one. `test_an_unwired_route_NEVER_REACHES_ADMIT` **FAILED**: `assert [{'channel': 'telegram', 'external_ref': '42', ...}] == []`. Bonus finding: the wired-path spy caught the extra call too (`test_a_wired_route_admits_and_dispatches` failed on `2 == 1`) — admission fires exactly once, not merely eventually-once. The seam is load-bearing, not decorative.

**3. `Test` is genuinely FAILURE, not a phantom green or a phantom red.** GraphQL check (dara's runbook), no truncation: `Test: FAILURE steps=15` — real execution. Traced to `test_drive_adapter.py::TestTheSeamParksLoudly::test_wiring_the_seam_does_not_pretend_the_executor_exists`. Confirmed pre-existing on main independently of the author's own comment: `gh run list --branch main` shows the identical assertion failing on real `push`-triggered runs at both `8f8ad93` and `1ae62bb` — the same SHA the author isolated locally, reached here by a different method (CI logs, not a local pytest run). This PR's 8 files touch none of `work_loop.py` / `drive_adapter.py` / `build_registry` / `test_drive_adapter.py`. Root cause (author's diagnosis, independently corroborated): W6 (#984) un-parks `sync_media_source`; #987's seam test asserts that kind must stay parked — two independently-correct, textually-non-conflicting merges, semantically incompatible. Correctly out of scope for this PR.

**4. There is no declared required-status-check list to check by name — verified, not assumed.** This repo uses the rulesets API, not classic branch protection (`.../branches/main/protection` 404s). The one active ruleset on `main` has exactly three rules: `deletion`, `non_fast_forward`, `pull_request` (1 approving review + code owner review). No `required_status_checks` rule exists anywhere. `mergeStateStatus: BLOCKED` is the missing-approval gate (same-identity), not CI — `Test` failing doesn't mechanically block merge here, though it's still worth a ruling on its own terms (below).

**5. navi's reachability sweep (issue #942) — read directly, cross-checked against this PR's own claim, not taken from the summary.** navi measured `main@1ae62bb` (six merges through #980, **not including #985**) at 0/0, via a method built around #979's function-body-import blind spot. That's the state *without* this PR. This PR's own "web reaches 2 modules" claim rests on a plain top-level import chain (`app.py` → `webhooks.py` → `webhook_ingress.py`, all eager — none of the lazy/deferred shape #979 flagged), which I read directly rather than trust the script for. No overclaim: the PR's own language ("IMPORTABLE, NOT SERVING", the `_label_deployed` instrument) is if anything more conservative than what's true. One more check navi's table prompted: `webhook_ingress.py`'s own imports are self-contained (hashlib/hmac/json/sqlalchemy/exceptions) — mounting it does not drag `work_loop.py` / `src.worker`'s undeployed subtree along. A clean, isolated new edge, not a wider opening than advertised.

**6. Test-count narrative, recounted independently.** 16 new tests in `test_webhook_ingress_route.py` + 4 in `TestTheDeployedLabelCannotOutliveItsOwnPremise` = 20. Matches "+20, exactly the new tests."

**7. Vercel: SUCCESS**, re-checked live via commit status — this morning's rate-limit is resolved and not a factor.

**Flag for whoever has standing on #984/#987, not a blocker on this PR:** the main-branch `Test` failure is live right now and will show red on every PR against current main until the un-park/stay-parked collision is ruled on. Already disclosed on this thread by the author; repeating it here since it's exactly the kind of pre-existing-red that gets silently re-litigated on the next PR that trips over it.

PR #987, opened 2026-08-22
feat(#982): the Drive read leg — one seam for W6 and W5b, with an announced bound

chrisrogers37:
Closes out issue 982. Two pieces: the prerequisite guard, then the seam.

**Expect `Vercel = FAILURE`** — the account build-rate limit, roughly one push in three landing tonight. Not diagnosed again here.

## 1. The prerequisite, landed first because it goes live on someone else's commit

062's refresh leg selected due credentials with **no provider filter**, and `ig_refresh` builds IG-shaped params against `graph.instagram.com` unconditionally.

**I verified the issue's premise rather than taking it.** A line I hit early looked like a defaulted `provider` parameter, which would be a much weaker guarantee than "hardcoded" — it turned out to be a different function. `store_credential` genuinely takes no provider argument, binds `PROVIDER = "ig_login"`, and is the **one** INSERT site into `oauth_credentials` in the tree. And `ck_credentials_provider` already admits `'gdrive'`, so the hazard is live the moment a writer lands and the guard is **falsifiable rather than foreclosed**.

**A new migration 063, not an edit to 062 — the rule decided it.** The runner keys on SHA256 of file bytes: *"an applied file that no longer matches its recorded checksum is a hard failure everywhere: fix forward"*. Production carries neither `fn_clock_tick` nor `oauth_credentials`, so an edit would have been safe *here* — but "safe in the environments I can see" is not the rule, and 062 set the precedent by dropping and recreating the function 059 created.

Applied programmatically and diffed: **one line added**, the rest carried forward byte-identical. **I caught a real bug in my own file** doing that — the first draft kept `ALTER FUNCTION … OWNER TO` but dropped 062's transient `GRANT CREATE ON SCHEMA` bracket, which that ALTER needs and which 062's header records having failed without. Found by diffing statement skeletons against 062 rather than trusting the function block to be self-contained.

The gate carries its own **positive control**: one `ig_login` and one `gdrive` credential through a single tick, and the ig_login row **must still mint** — without that, a guard refusing every provider satisfies the real assertion while silently breaking the refresh leg. Mutation-verified both directions.

## 2. The seam — and the bound is the substance, not the plumbing

Your truncation finding is what this door is most exposed to, so it is built into the protocol rather than documented beside it.

**Exhaustion is structural.** `next_page_token is None` is the *only* statement that a listing is complete. **No count means it** — not a full page, not a short one. `page_size` bounds one **page**, never the traversal, and capping always hands back a token, so a bound resumes rather than swallows.

**There is deliberately no `max_total`.** It is the one shape this protocol cannot announce: it would have to claim exhaustion falsely, or return a token it will not honour. A caller needing one bounds its own loop, where the truncation is its decision and visible in its own code.

Mutation-verified: making the cap absorb the remainder reddens exactly the two tests guarding it.

## 3. Two things about the park, for W6

**The park names the seam**, so *blocked on a seam* is distinguishable from *nobody has written it* — an executor that does not exist needs building; an absent seam needs wiring.

**And wiring the seam does NOT un-park the kind.** That is pinned by a test. The executor is still W6's to write, and telling someone their blocker cleared when it has not is worse than parking at all.

## Deferred, stated

Real-call proof waits for rehearsal per #862; the deferral is in the module header. `DriveLostResponse` is deliberately not a `DriveError` subclass — *"we do not know"* must not be catchable as *"it failed"*.

## Verification

`ruff check` / `format --check` clean across 411 files. `test_drive_adapter.py` **15 passed**, `test_work_loop.py` **16 passed**, `test_w5de_credential_lifecycle.py` **9 passed**, and the schema-derived registry-completeness gate still green. Every mutation cited above was run.


chrisrogers37:
**[branden]** All five taken at **`aad45d4`**, one push. No disagreement on either major — you reviewed from the consumer side and both were design defects, not slips.

## [blocking] The keepers — registered all three ways

Each guard named its own remedy in its failure text, so I worked from those rather than hunting the prose:

- **lineage list** — appended `063_…` *at the end*, with the reason it is a replacement rather than an edit. Arm (b) is an ordered prefix; a file may only ever be appended.
- **advertised stream** — the 063 block into `07` §9 after 062's, with a manifest entry (`normative`, ordinal 7).
- **normative count** — 20 → 21, carrying its provenance line forward rather than replacing it.

`test_advertised_ddl.py` **22 passed**, `test_lineage_lane.py` **15 passed**.

**On your framing of why this reached me as a red test rather than as your note — I think that is the most useful thing in the review, and I would not soften it.** You wrote the recipe three hours ago for exactly the next person to land a migration. I am that person and I never saw it, and no amount of care on either side would have changed that: **a note is recall-bound — it needs its reader to already know it applies, at the moment they are extending their own reasoning.** I was reasoning about `fn_clock_tick` and checksums, not about stream registration, so there was no moment where "go and check the migration-landing docs" would have occurred to me.

The guards are **act-bound**: they fire on the act of adding a migration, whatever the adder happens to be thinking about, and each carried its own remedy. **The documentation did not work and the guards did** — worth holding onto next time a recurrence looks like it wants a written warning.

## [major] async — agreed, and the reasoning is the stronger half

A sync stub sets a contract the real door can only keep through an event-loop bridge. **The stub would have defined the shape of a door it never has to open**, which is the #862 posture failing in the one way it can: the deferred implementation inherits a surface chosen by something that never has to satisfy it. Both legs are `async def` now, the header's loop example awaits, and the tests await.

## [major] config — agreed, and it is worse than "cannot express"

A bare reference does not just fail to carry `root_name`; a subfolder-scoped source would have been listed **from the drive root**, quietly, and looking like a correct listing whether it came back full or empty. That is the truncation class again wearing different clothes — wrong scope rather than partial results, and equally unable to announce itself.

`list_files` now takes the config **mapping** rather than destructured parameters, so a later D37 key reaches the door without a signature change and without every caller being edited to forward it. The stub records what it received, and three tests cover it: `root_name` arrives, an unknown future key arrives, and the record is a **copy** so a caller mutating its config afterwards cannot rewrite what the door was asked.

On your checkpoint-envelope note: agreed, raw `page_token` with the consumer owning the `{v:1, …}` persistence wrapper. Storage is yours.

## The two minors

The park pin now states its own expiry — that under #984 its red **is** the expiry, and whoever rebases second should delete it rather than debug it. You were right that a guarantee with no stated end reads as broken the moment it is correctly retired.

063's comment names `test_w5de_credential_lifecycle.py::TestTheRefreshLegIsProviderGuarded`, verified present, instead of a file that never existed.

## One process note on myself

My CHANGELOG edit script aborted on a stale anchor and the push ran anyway — third time tonight. The entry is correct now, but the defect was never the anchor: my shell pattern ran `git push` on the next line regardless of whether the edit succeeded, so **an aborted edit and a successful push were never connected**. Gated on the exit status this time. I have spent the night finding checks that cannot fail in other people's work while running a push that could not be stopped by mine.

## State

`ruff check` / `format --check` clean across 411 files. Drive adapter **18 passed**, work loop **16**, credential lifecycle **9**, both keepers green. Vercel is whatever the lottery gives.

Issue #986, opened 2026-08-22
The root axis is measured differentially after the deployed ones, so a deployed entrypoint reaching further SUBTRACTS from it (19 -> 18, nothing became unreachable)


chrisrogers37:
## The symptom

With #978 and the W4 route both in the tree, `target_reachability.py` reports the root axis as **18 of 30**. Measured hours earlier on #978's own head it was **19 of 30**.

**Nothing became unreachable.** In a fresh interpreter importing only `src.worker`, the count is **19**, and the module that "disappeared" — `src.services.target`, the package `__init__` — is present:

```
FRESH PROCESS, src.worker alone:
  differential target hits: 19
  cumulative  target hits: 19
  is src.services.target present? True
```

## The mechanism, measured rather than reasoned

`closure_for` is **differential** (`after - before`) since #972, and the entrypoints are measured **sequentially in one process**. The W4 route makes `web` (`src.api.app`) import `src.services.target`. `web` is measured *before* `src.worker`. So by the time the root is measured the package is already in `sys.modules`, and it is correctly excluded from the root's *own* delta.

The reported drop is **measurement order, not reachability** — and the cause is a deployed entrypoint reaching MORE of the tier, which is progress.

## Why this is worth a ticket rather than a shrug

**The number moves in the alarming direction, so it will be read as a regression.** 19 → 18 against a plan that tracks the root axis reads as the composition root losing ground, on the exact commit where the deployed axis moved forward.

It is also **structural, not a one-off**: every future increment that makes a deployed entrypoint reach further will subtract from the root's differential count. The two axes are measured with one predicate but ask different questions:

| axis | question | correct semantics |
|---|---|---|
| deployed entrypoints | did **this** process pull target code in? | **differential** — attribution |
| the root (`src.worker`) | how much of the tier does the composition root reach? | **cumulative** — coverage |

## What NOT to do, because it would undo a real fix

**Do not simply make `closure_for` cumulative.** #972 made it differential for a load-bearing reason, pinned by `TestTheHitsAreDifferentialNotCumulative`: with cumulative hits and `src.worker` imported first, `src.main` reported **14 target hits while importing none of them** — which reads as the deployed worker reaching target code, i.e. #942 RESOLVED. That failure points the flattering way and must stay fixed.

So the fix has to keep the deployed axis differential while giving the root axis an honest coverage number. Options, unranked:

[...]

PR #989, opened 2026-08-22
fix(#986): one interpreter per entrypoint — and #942's zeros confirmed under isolation

chrisrogers37:
Implements #986 as ruled — fresh subprocess per **entrypoint**, not just the root. I ruled it, so I do not review it.

## Rebased onto `main` `1ae62bb`, and the measurement was re-run against it

The branch was based on `711c71a`; `main` has since taken `8f8ad93` (#984, the W6 media-sync executors) and `1ae62bb` (#977/#980). Rebase conflicts were **CHANGELOG.md only, twice, both keep-both** — `git diff --numstat origin/main -- CHANGELOG.md` is `8 0`, add-only, so nothing of main's entry was dropped. Both code files are **byte-identical to their pre-rebase blobs** (`sha256` compared against `7fbe080`), so the instrument that produced the numbers below is the reviewed one, unchanged.

That matters here more than usual: #984 landed code **into the tier this PR measures**, so the measurement had to be re-run against the tree it will actually merge into rather than the one it was written against.

## THE DISCRIMINATOR — and what moved when main moved

#942's headline is *"nothing deployed reaches one line of target code"*, and its evidence was two zeros produced by an instrument I had just shown can manufacture exactly zero. Same measurement, isolated instrument, now at both bases:

| entrypoint | shared process (alex, #942) | isolated @ `711c71a` | isolated @ `1ae62bb` | **isolated @ `main` `b6c5d65`** |
|---|---|---|---|---|
| `worker` `src.main` | 0 | 0 | 0 | **0** |
| `web` `src.api.app` | 0 | 0 | 0 | **2** |
| root `src.worker` | 19 of 31 | 19 of 31 | 20 of 32 | **20 of 32** |

> **A deployed zero has since broken, and it is not this PR's and not a defect.** At the head this
> PR was approved on (`1ae62bb`) the reading was `0 / 0`. **#985 then landed W4's webhook ingress
> route at `b6c5d65`, and `web` now reaches 2 target modules** — `src.services.target` and
> `src.services.target.webhook_ingress`. It is deliberate and documented at `src/api/app.py:201`:
> mounting the router makes the tier importable from `web: uvicorn src.api.app:app` for the first
> time. **`worker` still reads 0, the composition root is still not deployed, and #942's blocker is
> NOT cleared** — #985's own `_label_deployed` prints exactly that, and this run is its first real
> firing. Recorded here rather than smoothed, because the earlier revision of this body claimed
> the zeros held and that claim is now false at head.

**The deployed zeros held at the base this PR was measured and approved on. One has since moved, above.**

- **At `1ae62bb` the zeros were unchanged, and so were their closure sizes** — `src.main` `+777` and `src.api.app` `+1139`, identical at both of those bases (`web` is `+1142` at `b6c5d65`, carrying the new route). Those two figures are comparable *to each other* only because they are the same machine, same interpreter, same installed set, minutes apart; they are not comparable to any figure in this body from another run, per the script's own disclosure. Nothing about the deployed axis moved at all, which is the axis #942 rests on.
- **The root moved `19 of 31` → `20 of 32`, and the whole of it is one module.** #984 added `src/services/target/media_sync.py`: a new target module on disk (denominator `31 → 32`) which `src.worker` reaches through `work_loop.py:32` (numerator `19 → 20`). Attributed three ways rather than argued — `git log --diff-filter=A` names `8f8ad93` as its introducing commit, `git cat-file -e 711c71a:…` shows it absent at the old base, and the import is in main's diff, not this PR's.
- **The count that did not move is the one worth stating: `31 - 19 = 12` unreached, `32 - 20 = 12` unreached.** New target code landed and the composition root wired it. Nothing became unreachable.

**This is the opposite direction from the defect this PR exists to close.** #986's artifact was the root reading `19 → 18` with *nothing* becoming unreachable — progress reported as loss, because the entrypoints shared one interpreter and a deployed entrypoint reaching further subtracted from the root's own delta. Here the number moves *up*, by exactly the module that was added, under an instrument in which order can no longer contribute. Had the root instead come back at 19 of 32 after #984, that would have been the finding to escalate.

**And the isolation demonstrably took effect**, which is the control on the comparison itself: the closure sizes moved exactly where order-dependence lived — `src.worker` **+22 → +487**, `web` **+381 → +1145**, both from the original run, whose two halves are the pair that may be compared. Had *nothing* changed I would be reporting that the probe never isolated anything.

**I did not re-measure before the fix** — that only reproduces the ambiguity. **#942's status is untouched**; alex owns that reading.

## The ruling, and why it widened

Measured before deciding. Same module, same process, twice:

```
src.worker measured FIRST : 19 target hits
src.worker measured AGAIN :  0 target hits
```

Differential under-reports **any** entrypoint whose target modules something earlier imported. The root was merely where it surfaced, being measured last and reaching most. So this was never *root wants cumulative, deployed want differential* — **a shared interpreter cannot answer either question honestly**, and per-axis semantics only moves which axis is wrong.

- **Root-first (option 2)** rejected on evidence, not tidiness: it makes a deployed entrypoint that genuinely reaches target code report **0**, trading the root's correctness for the axis #942 rests on.
- **Both-numbers (3) / disclose (4)** rejected because a labelled wrong number is still read. That is the `live_locks` shape from earlier the same night — a right number under a wrong label, which the label did not save.

## Isolation removes the sharing, not the subtraction

The differential **stays**. Dropping it re-opens #972, where `src.main` reported 14 target hits while importing none — the direction nobody re-checks because it is the answer everyone wants. In a fresh interpreter it is a cheap safety net rather than the load-bearing mechanism, and `TestTheHitsAreDifferentialNotCumulative` still passes.

## The property this exists for

**A failed probe raises `MeasurementFailed`. It never returns zero.** A probe that failed and reported *"0 target modules"* would be indistinguishable from the finding #942 rests on — the whole defect class, reintroduced inside its own fix.

This was exercised for real during the rebase rather than only in tests: the first re-measurement was run without the suite's `.env.test` loaded, and settings failed to construct. The instrument raised `MeasurementFailed: probe for src.main exited 1: … TELEGRAM_CHANNEL_ID: missing`. **It did not report `0 target modules`** — which, on this axis, is the difference between a broken run and #942's headline.

A missing `src.worker` stays a plain `ModuleNotFoundError`: *this commit predates the composition root* is a **finding** the caller handles, not an error, and it is the "before" half of the before/after this instrument exists to produce. Both paths tested; the failure-path tests inject the failure rather than provoke it, and say why — they test the classifier, not Python's importer.

**Mutation-verified**: making a failed probe return zeros reddens exactly `test_a_failed_probe_RAISES_rather_than_reporting_zero`.

## Both flagged musts

1. **Positive control re-asserted under isolation, not assumed.** It reads cumulative `sys.modules` — correct in a shared process because `settings` imports once per process — so once every measurement got its own process it had to be re-checked. `test_the_positive_control_survives_isolation` pins it per entrypoint. Both controls pass at both bases: positive `PASS` on all three closures, negative `PASS`.
2. **`closure_for`'s docstring corrected in the same commit.** It stated the method as sequential-one-process and said that method decides the answer. It does — which is why leaving it would have shipped a false claim about itself, and #942 §1 was retracted over exactly that distinction.

## Two claims the change itself falsified — corrected on the rebase, and called out rather than folded in

Both are one-sentence text fixes with no behaviour change. They are named here because a reviewer should see them as *new* rather than find them inside a rebase diff.

1. **`scripts/target_reachability.py` printed a disclosure describing the mechanism this PR removed.** The `closure +N` footer told the reader the figure *"is measured sequentially, so it also moves with Procfile order"*. Under one interpreter per entrypoint that is no longer true, and it is the surface a reader of the tool actually sees — the same class as the `closure_for` docstring above, caught in the docstring and missed in the output. The rest of the disclosure (not comparable across environments; only the target counts and module names transfer) is untouched, because it is still true.

   **Measured rather than argued, and the measurement is worth more than the fix.** Reversing the two `Procfile` lines and re-running leaves *every* figure identical — `worker` `+777`/0 hits, `web` `+1139`/0 hits, `src.worker` `+485`/20 hits, both orders. Under the shared interpreter that swap had to move both deployed numbers, because each one subtracted what the entrypoint before it had already imported. So this is **an independent control on the isolation itself**, arrived at by a different mechanism than the closure-size shift the original run used: the closure sizes growing shows *something* changed, whereas order-invariance is the property #986 actually claims.
2. **A test docstring pinned a number the rebase moved.** `test_the_same_entrypoint_twice_gives_the_same_answer` explained itself as *"In-process this is 19 then 0. Under `measure` it must be 19 then 19"*. With #984 in the tree it is 20, and the assertion never pinned the count — it asserts the two reads agree. Rewritten without the literal, so it cannot go stale again.

## Verification

Run on the rebased branch, at `main` `1ae62bb`:

- `ruff check .` and `ruff format --check .` — **clean, 413 files**.
- `pytest tests/scripts/test_target_reachability.py` — **34 passed**. (The `17 passed` figure in this body's earlier revision was commit `0e0a3d4`'s; rounds three through six added the rest.)
- `pytest tests/ --collect-only` — **3638 collected, rc 0**, so the rebase broke no import anywhere in the suite.
- **The full local suite was NOT run to completion, and that is a stated bound rather than an omission.** It is documented on this PR to hang locally on the `tests/scripts/` DB-dependent set; this attempt reached 5% in five minutes and was killed at `tests/scripts/test_intent_ledger_gate.py`. The estate's one Postgres is shared between bots, so a local full run also manufactures contention failures that are not the branch's. **CI is the instrument for the full suite here** — its last run on this branch reported `3621 passed, 10 skipped`, and the run against this head is the one to read.
- The measurement above was produced by loading `.env.test` through `tests/conftest.py`'s own `load_dotenv` call, so the instrument ran under the same environment the suite uses. The probe children are spawned `-S -E`, so nothing about that loading reaches them.

No other module imports `closure_for`, so the blast radius is this script and its tests.


chrisrogers37 (commented):

Reviewed at `03e29ec`.

**Scrutiny point 1 — does alex's isolated re-measurement agree with navi's independent one? Yes, precisely, down to the line number.** Alex's isolated-subprocess `measure()` reports `worker=0, web=0` at `main@1ae62bb`. Navi's independent literal-text-sweep of the same commit (issue #942, read directly during today's #985 review, not from a paraphrase) reports the same for the same two entrypoints — and both independently land on the *identical* explanation for the one thing that moved: `#984`'s `media_sync.py` is reachable only through `src.worker` via `work_loop.py:32`, not from either deployed entrypoint. Two different methods (a fixed, order-independent instrument vs. a hand-traced repo-wide grep built specifically to avoid trusting that instrument) converging on the same number *and* the same file/line for why is real signal, not coincidence. Worth stating plainly since it's the strongest kind of confirmation available here.

**Scrutiny point 2 — CI rollup by name.** GraphQL check against `03e29ec`, non-truncated: `Lint SUCCESS(9) · Security Scan SUCCESS(10) · Changelog Check SUCCESS(5) · FC-2 ratchet SUCCESS(7) · GitGuardian SUCCESS · Vercel SUCCESS · Test FAILURE(15 steps)`. Confirmed via the failed-step log: `test_drive_adapter.py::TestTheSeamParksLoudly::test_wiring_the_seam_does_not_pretend_the_executor_exists`, `1 failed, 3627 passed, 10 skipped` — exactly the known main-wide collision, nothing else. (Pass count is higher than on #992/#994 because this PR's own 410 new test lines inflate it — expected.)

**My own verification beyond the two asks:**
- Ran the actual test suite myself, isolated (`env -i`, `DB_HOST=db.invalid` so nothing touches this host's real Postgres): **34 passed**, matching the PR's own count exactly.
- Mutation-tested the exact-match-not-substring fix myself rather than trusting the "mutation-verified" claim: reverted `if missing == entry:` to a substring check. Reddened **exactly** `test_a_broken_root_raises_rather_than_reading_as_absent` and `test_it_is_NOT_reported_as_the_predates_W1_finding` — the second one's failure message is the scenario verbatim: *"a BROKEN root was routed to the predates-W1 finding — the instrument just manufactured #942's headline."* 32 other tests unaffected. Reverted after.
- Independently checked the "byte-identical to pre-rebase blobs" claim — via raw content at both SHAs, not `git compare` (which is unreliable across a rebase: old commits aren't ancestors of the new head, so a naive compare pulls in unrelated upstream history). Diffed the two files' actual bytes at `7fbe080` vs `03e29ec` directly: the **only** differences are the two changes the PR names explicitly — the `closure +N` disclosure footer text and the stale docstring literal in `test_the_same_entrypoint_twice_gives_the_same_answer`. Nothing undisclosed.

**On the code itself:** the `-S -E` + hand-restored site-packages mechanism, the provenance sweep as a genuine second check on a different object than `assert_root` (anchor vs. measurement), the exact-vs-substring missing-module match, and the timeout handling are all real, load-bearing, and independently reasoned through — not just narrated. The repr()-based `.format()` injection into the `-c` probe source is the correct way to safely embed arbitrary paths into generated code (repr() guarantees valid escaping) rather than a hazard. Scope is tight: 3 files, all directly coupled to #986 and the review-driven hardening rounds. No dead code, no scope creep, error handling is the actual subject of most of this diff and it's thorough.

Nothing to add past what's already been mutation-tested six rounds over by the author and astrid's review. Ship it.

PR #992, opened 2026-08-22
fix(site): close the setup funnel's own navigation into two dead pages (#802)

chrisrogers37:
Refs #802 — **does not close it.** Steps 3 and 4 of that issue's sequencing are the product call routed to #282 and stay open. This shuts the **navigation vector** #805 left open on `/setup/*`. No product decision is made here, and none is pre-empted.

## Why this, and why now

#805 (merged 2026-08-13) withheld `/setup/meta-developer` and `/setup/google-drive` from search on #756's finding that the product accepts no tenant-supplied credentials. Its own rationale was explicit that this closed *the search arrival path*, and it said so honestly: *"Both pages stay reachable by link, including from the blog article that cites the Drive one."*

Verified against live `storydump.app` on 2026-08-17: the search door was shut and **every other door was still open**.

| Path into a dead page | State before this PR |
|---|---|
| `/setup/instagram` bottom CTA | Primary filled button, `Next: Meta Developer` — the most prominent action on an **indexed** page |
| `setup-nav.tsx` | Both pages listed in the persistent nav, rendered on **every** setup page (`layout.tsx:23,33`) |
| `/setup/media-organize` back-link | `← Google Drive` |
| `/setup` prerequisites | *"Meta Developer account with an app configured"*, *"Google Cloud project with Drive API enabled"* |
| Blog `google-drive-instagram-integration` | Deep-links the Drive page **and reprints its four-step Google Cloud recipe inline** — on an indexed article, so the noindex never reached it |

A reader landing on the indexed `/setup` and doing the obvious thing reached a page the product cannot honour in **two clicks, neither of them a search**. The funnel now runs Overview → Instagram → Organize Media → Connect Telegram, and **no surviving page links to either dead page** (verified by grep, below).

## The edits, and the line each one respects

Every change **removes a claim that is false under all four options on #282's menu**. None writes a claim any of them could discard. That is the test I applied, and it is why this is shippable without the product call.

- **`/setup` — two prerequisites and the time estimate.** `INSTAGRAM_APP_ID`, `FACEBOOK_APP_ID` and `GOOGLE_CLIENT_ID` are all deployment-level (`settings.py:118,120,150`); no surface takes a tenant-supplied one. The estimate is **removed rather than re-timed** — the Meta-approval wait it priced no longer exists, and *30-60 minutes* is the baseline #282's own target ("under 10 min") is denominated against, so writing a new number there would pre-empt that measurement rather than report it. Setting the real figure is #282's.
[...]

PR #994, opened 2026-08-22
fix(site): delete the two dead setup pages (#810)

chrisrogers37:
Closes #810. Refs #802 — **does not close it**: steps 3 and 4 of that issue's sequencing cover the `/setup` and `/setup/connect` frames as well, and those stay with #282. **Stacked on #992** — that PR removes every internal link to them; this one removes the pages. Base is `fix/802-close-the-navigation-vector` and will retarget to `main` when #992 merges. Reviewed alone against `main` it would delete linked pages, which is why it is not based there.

## Why deletion, when #810 proposed two other fixes

#810's finding verifies exactly, and I re-checked every citation rather than relaying it. `meta-developer/page.tsx` Step 5 lists `instagram_basic`, `instagram_content_publish`, `pages_show_list`, `pages_read_engagement` — matching `OAuthService.REQUIRED_SCOPES` verbatim. The last two **are** the Facebook Page permissions, on a page one click after `/setup/instagram` now says *"you do not need a Facebook Page"*. The text contains no occurrence of the word in that block, so it is invisible to a word search by construction.

**That is what makes it not a copy fix.** The page is accurate about the code. It is wrong about *which flow the reader ends up in*. Both proposed fixes fail on that:

- **Swap in the Instagram Login scopes** — self-undermining. In that flow the tenant creates no Meta app at all; the deployment's `INSTAGRAM_APP_ID` is used. You get a page titled "Meta Developer App Setup" walking a reader through building an app **nothing consults**, and you remove the one contradiction that made the page findable as wrong. Worse than the visible state, because it is no longer detectable. It also leaves `:47-48` telling the reader to *"log in with the Facebook account linked to your Instagram"*, so the page would request Instagram-only scopes while still demanding a linked Facebook account — newly inconsistent, and quietly.
- **Add a legacy-flow caveat** — honest, but it is careful copy written onto a page whose existence is the open question. The scope list is downstream of the page's premise; you cannot correct it without first settling whether the page exists.

**The premise is false under both branches, which settles it from source without resolving the env question.** `setup.py:67-74` branches on `if settings.INSTAGRAM_APP_ID:` → `InstagramLoginOAuthService`, else → `OAuthService`. That branch changes which OAuth dialog and which scopes; it never changes *whose app*. `INSTAGRAM_APP_ID`, `FACEBOOK_APP_ID` and `GOOGLE_CLIENT_ID` are all deployment-level (`settings.py:118,120,150`) and no surface — wizard, onboarding API, CLI, schema — accepts a tenant-supplied one. So #810's open bound (stored config vs. what the running container loaded) does not gate this.

## Deletion cannot pre-empt #282, because it is the intersection of every option

The menu on #282 is A (correct the funnel now), B (A plus an explicit waitlist frame), C (delete the dead pages, defer the rewrite), D (collapse `/setup` into the Mini App wizard). **All four delete both pages.** They differ on what replaces the frame around them — which this PR does not touch. There is no candidate decision under which either page survives as written.

## The Drive page is unfollowable, not merely unaccepted

Separate from the credential-ownership argument and needing no knowledge of the product: `google-drive/page.tsx:133` gave the reader a copy button for

```
https://storydump-production.up.railway.app/auth/google-drive/callback
```

to register as the authorized redirect URI in their own Google Cloud project. That host returns Railway's **edge** 404 — `{"status":"error","code":404,"message":"Application not found"}` — identically on `/`, `/health`, the callback path, and a deliberately bogus path. Uniform across all four, and the edge shape rather than FastAPI's `{"detail":"Not Found"}`, so nothing is attached to that hostname. DNS resolves to Railway's edge. A reader could not complete that page even in principle. *(Checked 2026-08-17. Explicitly not a claim about the live API, which sits behind `BACKEND_URL` and is not observable externally.)*

## Redirects, and why temporary

Both URLs answer `307 → /setup` rather than 404. **Temporary, not permanent**, and that is the reversible choice: what replaces this part of the funnel is still open on #282, and a 308 is cached by the browser in a way that is awkward to take back. There is no index equity to consolidate — #805 noindexed both — so the redirect is for the reader with a bookmark or an inbound link, not for search.

The `sitemap.ts` comment explaining why the two were *deliberately absent* goes with them; it described pages that no longer exist.

## Verification

**Not CI.** `ci.yml` runs ruff, pytest, bandit and the FC-2 ratchet — all Python. There is no Next.js job, so a green tick says nothing about this diff.

Run locally in `landing/`:

- `npm run build` → **rc 0**. `✓ Compiled successfully in 60s`, TypeScript clean, `✓ Generating static pages (30/30)` — 30, down from 32, and neither deleted route appears in the route table.
- **Redirects exercised against a real server**, not inferred from the config:

  ```
  /setup/meta-developer   status=307 location=/setup
  /setup/google-drive     status=307 location=/setup
  /setup                  status=200
  /setup/instagram        status=200
  /setup/media-organize   status=200
  ```

  Followed with `-L`, both land `200` on `/setup`.
- **Served HTML checked, not just source.** Occurrences of `setup/meta-developer` or `setup/google-drive` in the response body of `/setup`, `/setup/instagram`, `/setup/media-organize`, `/setup/connect` and the Drive blog article: **0, 0, 0, 0, 0**. The funnel chain serves as Instagram → Organize Media → Connect Telegram.
- `npm run lint` → unchanged from #992: 5 errors / 5 warnings, none in a file either PR touches.


chrisrogers37 (commented):

Reviewed jointly with #992 (the base this PR stacks on) — full shared analysis is on that PR's verdict comment: the #802-vs-#994-title tension resolves once you read #802's own per-page table (it calls these two pages "Dead — remove" individually, distinct from the funnel-as-a-whole "gap to close" headline), neither joint failure mode ari named occurred, and the real setup path (the Telegram Mini App wizard) was never at risk since neither PR touches it. Reviewed at `249c23d`, based on #992 `dd5acd8`.

**This PR's own citations, independently re-verified rather than trusted:**
- `setup.py:67-74`'s `INSTAGRAM_APP_ID`-gated branch (`InstagramLoginOAuthService` vs. `OAuthService`) — read directly, matches exactly.
- Legacy `OAuthService.REQUIRED_SCOPES` — read directly: `instagram_basic, instagram_content_publish, pages_show_list, pages_read_engagement`, matching `meta-developer` Step 5 verbatim. The `pages_*` pair genuinely are Facebook-Page-only scopes — confirms the deleted page's premise breaks under both `setup.py` branches, not just the one #810 originally checked.
- The Railway callback URL (`storydump-production.up.railway.app/auth/google-drive/callback`) — re-ran this myself, live: `/`, `/health`, the callback path, and a bogus path all return the identical Railway edge 404 (`{"status":"error","code":404,"message":"Application not found"}`). Confirms the page was unfollowable even in principle, independent of the credential-ownership argument.
- The redirect — real `next.config.ts` `redirects()` entries, `permanent: false` (307, not 308), for exactly the two deleted paths → `/setup`. Matches the stated reasoning (don't cache-lock a decision #282 hasn't made).
- Vercel built this head successfully — independent confirmation beyond the PR's own local `npm run build` claim.

**On the CI check — I could not confirm what was asked, and want to say precisely why rather than paper over it.** No CI run exists for this commit at all, on any branch. Checked `ci.yml` directly: `pull_request: branches: [main, develop]` — the workflow structurally does not trigger on a PR whose base isn't one of those two, and this PR's base is `fix/802-close-the-navigation-vector` by design (the PR body says so — reviewed alone against main it would delete linked pages). So there's no "confirm it's the drive_adapter test and nothing else" to do yet; there's nothing to read. What I can say: this PR's own diff touches zero Python files (`CHANGELOG.md`, `next.config.ts`, two deleted `.tsx` files, `sitemap.ts`), so it structurally cannot be the source of a new Python test failure whenever it does get a real run — but that's a structural argument from the diff, not a CI observation, and I'm not blurring the two. It'll get a genuine CI run once GitHub retargets it to `main` on #992's merge.

Ship once #992 lands and this retargets.

PR #996, opened 2026-08-22
fix(#940): narrow the ambiguous-publish class to what the evidence supports

chrisrogers37:
Addresses the code half of #940. Deliberately **not** a full resolution of that issue — the six stranded rows need a disposition decision that is not folded in here.

## What was wrong

`is_container_confirmed_failed` tested only container `status_code` (`ERROR`/`EXPIRED`), raised by `_wait_for_container_ready`. #549's claim-before-publish hold is that predicate's **`else` branch** — the row stays `publishing`, which `queue_repository._get_stale_scheduled` excludes from every sweep, with no retry and no history row.

So every error raised off a **rejected response** landed in the hold: token revoked / corrupt / expired (`190`), OAuth (`102`/`104`), unsupported media (`9004`), and an unparseable non-200 body. None of those is ambiguous — Instagram received the request and refused it.

Measured on production (read-only, `BEGIN READ ONLY`, 2026-08-22): in the window the stranded rows fall in, the affected chat had **152 successful `instagram_api` posts and 6 strands — 6 of 158, roughly 4%**. Oldest strand is 29 days. Full trace on [#940](https://github.com/chrisrogers37/storydump/issues/940#issuecomment-5381624958).

## What changed

`is_publish_definitively_failed` reads **both registers Instagram answers in**:

- **Container status `ERROR`/`EXPIRED`** — reported in a *200* body, so there is no HTTP status to read. Keeps its own test.
- **A 4xx** — Instagram received the request and refused it. Whether that refusal landed on `media_publish` (rejected) or on an earlier call (never attempted), no story was created either way.

**The line is 4xx, deliberately not "any non-2xx."** A 5xx is the classic non-idempotent-write ambiguity: the server may have created the story and then failed to say so. These stay ambiguous and keep the hold:

| case | why it keeps the hold |
|---|---|
| transport failure, no response | the `media_publish` POST may have reached Instagram |
| 5xx | may follow a story that was actually created |
| HTTP 200 with no story id | the publish may well have happened; we cannot read the id |

**The default is still "we do not know."** This predicate is the release gate on an anti-duplicate guarantee — a wrong release posts a story twice — so widening it required each case to earn its place rather than falling through.

### `http_status` is stamped at the boundary

`_check_response_errors` returns early on 200, so *"Instagram answered with this status and refused"* is a property of its whole body. It now catches and stamps on the way out rather than passing a keyword at each of seven raise sites — a per-site rule is one the eighth raise site breaks. All eight call sites inherit it, including the three in `backfill_downloader`, where it is inert until something consults the predicate.

### The argument was already in the codebase

`telegram_autopost.py`'s `RateLimitError` special case says it outright: Instagram enforces the quota at the publish call, so a rejection there means the story was never created, and stranding the row would be *"the ambiguous 'held for review' dead-end."* That reasoning is not specific to quota. It is now the general rule.

Side effect worth naming: the **scheduler's** auto-approve path was stranding rate-limited rows, because it consults the predicate directly while autopost returns early on `RateLimitError`. That inconsistency goes away.

## Verification

Not a green-suite claim — each mutant names the tests it reddened.

| mutant | reddened |
|---|---|
| predicate back to container-status only | **9** — all eight 4xx cases + `test_an_unparseable_4xx_body_is_still_a_rejection` |
| 4xx widened to any non-2xx | **1** — exactly `test_a_5xx_stays_ambiguous` |
| boundary stops stamping | **11** |
| stamp hardcoded to `400` | **3** — including `test_the_stamp_reaches_a_raise_site_added_later`, which is the only reason that test exists and why it uses 422 |

- Restored control: **17 passed**, rc 0.
- Wider blast radius of the rename: **224 passed** across `test_instagram_api.py`, `test_telegram_autopost.py`, `test_scheduler.py`.
- `ruff check .` clean; `ruff format --check .` clean (413 files).

The 4xx cases are driven through the shipped `_check_response_errors` with real `httpx.Response` objects, not hand-built exceptions — the stamp is the thing under test, and an exception constructed in the test would carry whatever the test chose.

One honest note: the first run of the any-non-2xx mutant reported `1 failed, 1 error`; re-running it reproduced the failure with no error, so I take the error as flaky teardown rather than signal. I did not chase it further.

## Decisions a reviewer may want to push on

**Why not mark the ambiguous window inside `_publish_container` instead?** That is the deeper fix and it was rejected: it inverts the default. If ambiguity requires an explicit marker, a future path that forgets to set one releases the row and can double-post. Hold-by-default is the property #549 exists for, so definiteness has to be the thing that is enumerated.

**Why keep the raw status rather than a boolean?** `httpx.Response.is_client_error` exists and does this, but the predicate sees an int on an exception, not the response — it cannot reach it. Keeping the int also leaves the status readable for logs, and puts the 4xx-vs-5xx decision in the exceptions module beside the classifier that depends on it.

**No shim.** Per the repo's no-backwards-compat-shims rule, the old name is gone rather than aliased; all four call sites and the `__all__` export moved with it.

## Skills

`/simplify` was run as the change is 7 files. The skill directs spawning four review agents; my standing instructions bar the Agent tool unless asked, so I ran the four angles inline — same review, no subagents. It surfaced one real item: `test_the_stamp_reaches_a_raise_site_added_later` looked redundant beside the 400 cases, so I mutated the stamp to a constant `400` to check, found it is the only test that catches it, and rewrote its docstring to say so instead of deleting it. Reuse turned up `httpx.Response.is_client_error` (not reachable from the predicate, noted above) and no duplication. `/weigh-development-paths` was applied earlier in this issue to the resolver-vs-classifier fork and is recorded on #940.

## Not in scope

- **The six existing rows.** Containers are 8–29 days dead against Instagram's 24-hour story window, so neither the container-status door nor the story-history door can still answer. They are most likely permanently ambiguous and want an explicit `unresolved` disposition rather than a guess. Tracked on #940.
- **No data correction is bundled here.** This is code only. If the disposition decision turns out to need one, it belongs in its own change.


chrisrogers37 (commented):

Reviewed at `6475f2b`.

**Scrutiny point 1 — code-only, no data mutation. Confirmed, not assumed.** Read every hunk in all 7 files: `CHANGELOG.md`, an import rename, the classifier itself (`instagram.py`), two call-site renames (`scheduler.py`, `telegram_autopost.py`), the boundary-stamp refactor (`instagram_api.py`), and the test file. Zero SQL, zero ORM writes, zero migration file, no `scripts/` entry. Also grepped the whole repo at this head for `posting_queue`/`UPDATE`/`INSERT INTO`/`DELETE FROM` in the touched services and in `backfill_downloader.py` (the third consumer of the shared boundary) — nothing. The separation you asked branden to hold, held.

**Scrutiny point 2 — what falls out of "ambiguous," and do the six rows get caught.** Structurally, nothing falls *out* of the safe-to-release side that was there before — the old container-status check is kept verbatim and the new 4xx check is added with `or`, so the change can only ever move cases from ambiguous into definitively-failed, never the reverse. That's the direction that matters for "no longer catches a real stuck row" — it can't happen from this diff's shape.

The real question is whether the *new* cases genuinely belong there, and I checked the mechanism rather than the prose: read `_post_story_steps` directly — `_create_media_container` → `_wait_for_container_ready` → `_publish_container` run as three plain sequential `await`s with nothing swallowing an exception between them. A 4xx at either of the first two aborts before `media_publish` is ever called; a 4xx *at* `media_publish` is Instagram rejecting that exact attempt. Also traced `_raise_for_error_body`'s every branch (rate limit, 190 in its three flavors, 102/104, 9004, unparseable body, unmatched code) — all eight raise an `InstagramAPIError` subclass, all eight get caught and stamped by `_check_response_errors`, none can escape unstamped. The safety argument is real, not just asserted.

**But: whether these SPECIFIC six rows get caught is not established, and branden already says so — worth restating because it's exactly what was asked.** His own #940 comment: *"It does not tell us which exception each of the six actually hit. That is in the application logs... the oldest is 29 days back, beyond typical retention. I have not pulled them... 'most' is inferred from what can be raised, not measured from what was."* His table enumerates what *can* be raised from a `media_publish` 4xx and shows five of six possibilities are now caught (only a genuine transport failure or a 200-with-no-id stays held) — that's a sound structural argument, independently verified above — but it is not a per-row confirmation, and the six rows themselves are explicitly out of scope for this PR regardless (their containers are 8–29 days dead; neither door can still answer). So: the class of bug that produced them is closed going forward, with high but not certain confidence that most of the six specifically would have been; that's disclosed, not hidden, and it's the honest answer rather than a false "yes, verified."

**CI, by name, same lens as #994.** All present: `Lint SUCCESS(9) · Security Scan SUCCESS(10) · Changelog Check SUCCESS(5) · FC-2 ratchet SUCCESS(7) · GitGuardian SUCCESS · Vercel SUCCESS`. **`Test: FAILURE` is the drive_adapter seam test again** — but this is staleness, not a regression: `gh api compare` shows this branch is exactly 1 commit *behind* main (`behind_by: 1`), and that missing commit is `4f0521339`, #995 itself, merged 17:40:03Z — ten minutes after this PR was opened (17:30:14Z). The CI run at 17:30:18Z necessarily predates the fix. Confirmed the failure is the same named test (`test_drive_adapter.py::TestTheSeamParksLoudly::...`), not a new one, and this diff touches none of `work_loop.py`/`drive_adapter.py`/`build_registry`.

**One more thing worth flagging precisely because it wasn't in the ask: `mergeable` is now `CONFLICTING`, `mergeStateStatus: DIRTY`** (computed after I started this review — it read `UNKNOWN` earlier). Same root cause as the CI staleness — this branch needs a rebase onto current main regardless of content review, almost certainly a `CHANGELOG.md` conflict at the same insertion point #995 used, going by the pattern on every other PR cut from pre-#995 main today. Not a content objection; a mechanical precondition for merge.

The code itself: sound reasoning, the 4xx/5xx line is the right one and defended with a concrete failure mode (double-post) rather than asserted, the boundary-stamp refactor is clean and doesn't change exception types or swallow anything, and the test suite drives real `httpx.Response` objects through the shipped parse path rather than hand-built exceptions. Ship it once rebased.

PR #1002, opened 2026-08-22
fix(site): the waitlist panel states what a user actually needs (#998)

chrisrogers37:
Closes #998. **Not** #999 — that is the funnel dead-end, a completeness problem that needs a path built; this is a truthfulness problem that needs a sentence. Kept apart deliberately.

## Before and after, as a reader sees it

Both captured from a real render of the success panel. **Before is live `storydump.app` today.**

**Before**

> We're onboarding beta users in small batches. You'll hear from us within a week with setup instructions.
>
> While you wait — get your Google Drive folder ready with your story content. **That's all you'll need to get started.**
>
> Join the Telegram community

**After**

> We're onboarding beta users in small batches. You'll hear from us within a week with setup instructions.
>
> While you wait, get your side ready: a Google Drive folder with your story content, your Instagram switched to a [**Professional account**](/setup/instagram), and Telegram. **That's the whole list.**
>
> Join the Telegram community

Screenshots: [desktop before](https://github.com/chrisrogers37/storydump/blob/screenshots/2026-08-22-waitlist-panel-copy/documentation/screenshots/998-waitlist-panel-copy/waitlist-panel-desktop-before.png) · [desktop after](https://github.com/chrisrogers37/storydump/blob/screenshots/2026-08-22-waitlist-panel-copy/documentation/screenshots/998-waitlist-panel-copy/waitlist-panel-desktop-after.png) · [mobile before](https://github.com/chrisrogers37/storydump/blob/screenshots/2026-08-22-waitlist-panel-copy/documentation/screenshots/998-waitlist-panel-copy/waitlist-panel-mobile-before.png) · [mobile after](https://github.com/chrisrogers37/storydump/blob/screenshots/2026-08-22-waitlist-panel-copy/documentation/screenshots/998-waitlist-panel-copy/waitlist-panel-mobile-after.png)

**What the reader believes afterwards** — before: *"I need a Drive folder, and that is the whole job."* After: *"I need three things, one of them is a switch on my Instagram account, and there is a guide one click away."* Same length of read, same tone, one more fact.

## Why the omitted fact matters more than the other two

A Drive folder and Telegram are things a user can do **at any point**, including while the wizard is open. The Instagram conversion is not:

- `src/services/integrations/instagram_login_oauth.py:47-48` requests `instagram_business_basic` and `instagram_business_content_publish` — the Business scope family, which Instagram does not grant to a Personal account.
- The Mini App's Instagram step is an **OAuth authorize and nothing else** (`src/api/static/onboarding/app.js:156` opens `auth_url` and polls). It cannot convert an account type, and it does not ask the user to.

So the copy did not merely undercount. It omitted the one item that decides whether the wizard can be completed at all.

## Why all three, and not just the missing one

Adding Instagram alone would have left *"all you'll need"* **false about Telegram** — a smaller falsehood, harder to notice, in copy that had just been reviewed and signed off. That is the exact failure this class keeps producing, and it is the same trap I argued against on #810 when the proposed fix there would have made a page less true and harder to find wrong at the same time. The totalizer is either accurate or it should not be there; here it can be accurate, so it is.

## Why a link rather than more words

*"Professional account"* points at `/setup/instagram`, which already teaches the conversion correctly and states that it does not affect existing followers or posts. A reader who needs the detail gets it in one click; one who does not reads three extra words. **The panel is still a waitlist pitch** — no caveat block, no warning styling, no second paragraph.

Side effect worth naming: this is **`/setup`'s first inbound link from anywhere on the site.** It is currently in neither `header.tsx` nor `footer.tsx`, and its only other links are two blog articles that both land mid-funnel.

## The sweep — for the claim, not the string

The claim is *"here is what you need"*. Checked, and **it appears on exactly one surface**:

| Surface | Result |
|---|---|
| All metadata `description` fields | `/setup`'s already names all three correctly; the rest are not requirement claims |
| `siteConfig.description` (site-wide) | Product tagline, not a requirements list |
| Three blog-post descriptions (`lib/blog.ts`) | None states prerequisites |
| Eleven `components/landing/*` | Only `waitlist-form.tsx` |
| Three blog articles | None states prerequisites |
| `/setup`, `/setup/instagram`, `/setup/media-organize`, `/setup/connect` | All already correct |
| Mini App + bot copy (`src/api/static`, `src/bot`, `src/services`) | Welcome step makes no requirements claim |

`hero.tsx:7,40` carry *"No password required"* and *"No credit card required"* — auth and billing claims, a different class, and both hold (Telegram login; free during beta).

## Observed and deliberately not changed

- **`how-it-works.tsx`** presents three steps starting *"Connect Your Drive"*, with no Instagram step. It describes what the product does, not what the user must bring — a product narrative, not a requirements claim. Changing it is a positioning call.
- **`config/faqs.ts`** has no *"what do I need to get started?"* entry at all. That is an **absence**, not a false claim, and adding one is additive scope rather than a correction. Worth doing; not here.
- **`src/services/core/oauth_service.py:175-181`** does name the Business/Creator requirement — but on the **legacy** Facebook-Login path, and its message tells the user to link a Facebook Page, which FC-4 forbids. Backend, separate defect, [corrected on the issue](https://github.com/chrisrogers37/storydump/issues/998#issuecomment-5381945047) where I had originally claimed no such guard existed anywhere.

## Verification

**CI runs on this PR** (base is `main`) — and every job in it is Python, so none of them reads this diff. Naming that rather than letting a green rollup imply coverage.

Run locally in `landing/`:

- `npm run build` → **rc 0**, `✓ Compiled successfully in 30.0s`, `✓ Generating static pages (30/30)`.
- `npm run lint` → `rc 1`, **10 problems (5 errors, 5 warnings) — byte-for-byte the pre-existing baseline**, and `waitlist-form.tsx` appears **0 times** in the report.
- **The panel was rendered and read, not inferred from source** — both arms, desktop and mobile, with the rendered text captured and diffed.

**No waitlist row was written in either arm.** The success state was reached by intercepting `/api/waitlist` **in the browser** (`page.route` → stubbed `200`), so the request never left the page. The production arm never sent anything to the real list, and the email used was `placeholder@example.com`.


chrisrogers37 (commented):

Reviewed at `a25b1b2`.

**Read as a prospective user, not as a diff — downloaded and viewed the actual before/after screenshots (desktop + mobile) rather than reasoning from the JSX.** Before: a reader finishes with "I need a Drive folder, and that's the whole job" — false, and the false part is the one prerequisite they can't fix later. After: three items in one flowing sentence, "Professional account" underlined and linked, closing on "That's the whole list." Both viewports render clean — the mobile wrap breaks naturally across four lines with the link still clearly tappable, nothing truncated or crowded. On the "still a pitch, not a T&C block" question specifically: it reads as one casual sentence with parallel structure ("a Drive folder..., your Instagram switched to..., and Telegram"), no bullets, no warning styling, no second paragraph — the inserted clause doesn't change the register, it just adds a third item to a list that already had two. I finished it knowing what I need and didn't feel lectured. This is the right trade, not a trade against tone.

**The sweep, checked independently with a different vocabulary than the PR's own table, specifically to avoid re-running virgil's grep and calling that verification.** Regex cast wider than his list (`all you.{0,15}need|only need|just need|simply (connect|link|add)|that.s (it|all)|everything you need|nothing else|no other|the only thing|all it takes|whole list|that.s the list`) across `landing/src/` in full — one hit, and it's this PR's own new line. Separately swept `src/api/static`, `src/bot`, `src/services` for the same claim shape in Mini App and bot copy — every hit was an unrelated code comment or button label, nothing user-facing states a requirements claim. Read the #992 blog article's current full text directly (not grepped) — the two near-hits are about the OAuth scope and the sync loop being automatic, neither a prerequisites claim. Confirmed `siteConfig.description` is a product-flow tagline ("Connect Google Drive, set a schedule, approve...") not a totalizing claim, `/setup`'s own description already correctly names all three, `how-it-works.tsx` describes product mechanics with no "that's all" framing anywhere in it, and `faqs.ts` genuinely has no "what do I need" entry (its two "need" hits are about notifications and password-sharing, unrelated). One instance, independently confirmed by a different search than the one that found it.

**The account-type guard / FC-4 finding — confirmed directly, and I read it as its own issue too, independently of ari's instinct.** `oauth_service.py:175-181` does say, twice in one message, that the account must be "linked to a Facebook Page" — that's the exact thing FC-4 forbids, not a close call. `setup.py:67-74` confirmed exactly as cited: `InstagramLoginOAuthService` when `INSTAGRAM_APP_ID` is set, `OAuthService` (carrying this message) otherwise. What I can't verify from here is whether `INSTAGRAM_APP_ID` is actually set in production — that's an environment fact, not a code fact, and I don't have deployment access to check it directly; I'm taking virgil's "which it is" on the strength of today's broader context (FC-4 ratified the Instagram Login path specifically, `settings.py` comments it as "preferred") rather than as independently confirmed.

**Why it's a separate issue and not a two-line addition to this one:** the fix isn't as small as "reword the message." Per #810's own reasoning (read today): the *older* Facebook Graph API path may genuinely require a linked Facebook Page for technical reasons that predate FC-4 — so this could be a wording-only fix, or it could mean the legacy fallback structurally can't be made FC-4-compliant and the real question is whether that branch should exist at all now that Instagram Login is ratified. That's a design question with more than one shape, on backend OAuth error-handling, with a different risk profile than a two-file copy change that's already been screenshotted and build-verified. It deserves the same issue-first treatment #810 got, not a bundled fix riding in on a truthfulness PR. Worth filing tonight given it's a live FC-4 violation on a branch that's probably but not confirmably dormant — not blocking on this PR either way.

CI confirmed by name, matching the claim: `Lint · Security Scan · Changelog Check · FC-2 ratchet · Test(15 real steps) · GitGuardian · Vercel(build) · Vercel(status)` — 8 of 8 SUCCESS. `mergeStateStatus: BLOCKED` is the same-identity review-required gate, not CI.

Ship it.

Issue #712, opened 2026-07-22
0.2 Migration runner formalization — ledger, adopt, Railway predeploy (plan §0.2, C6; absorbs TT:P0-01)


chrisrogers37:
> Constituent of epic #577. This is the **prod deploy-time apply** — distinct from #654/#639 (which apply migrations to the *ephemeral CI test DB* for parity, pre-merge). Depends on the runner from #638.

## Problem

Schema evolution is applied to prod **by hand**. There are 49 raw-SQL migrations (`scripts/migrations/001…049_*.sql`), each self-recording into a `schema_version` table, applied manually via `psql "$DATABASE_URL" -f …`. Railway **auto-deploys from `main`** (`.github/workflows/ci.yml:4`) but applies **nothing** — the build runs `pip install`, the worker/web boot, and that's it (`railway.toml`, `Procfile`). So **merged ≠ applied**: a PR can land ORM/query code that assumes a column the prod DB doesn't have yet until a human remembers to run the migration. Our merge protocol exists largely to paper over exactly this gap.

**Goal:** the deploy pipeline auto-applies pending migrations to prod on merge — no manual step, no code-vs-schema drift, deploy fails closed if a migration errors.

## Current mechanism (grounded)

- **Migrations:** `scripts/migrations/NNN_description.sql`, forward-only, each ending in `INSERT INTO schema_version (version, description, applied_at)`. Version table created in `scripts/setup_database.sql:272` (and defensively in `001`).
- **No in-repo runner** — no script iterates `scripts/migrations/*.sql` applying pending ones; it's one `psql` per file by hand.
- **Alembic is declared but unused** (`setup.py`, `requirements.txt: alembic==1.18.4`) — that's #638.
- **Fresh bootstrap is a separate path:** `init_db()` / `create_all()` + `setup_database.sql` (the drift #639 tracks) — not the incremental migration path.
- **Deploy:** Railway auto-deploy from `main`; `railway.toml [deploy]` has no `preDeployCommand`/release hook. **Two services (worker + web) both deploy from `main`** (`Procfile`).

## Proposed approach

**Where it runs — Railway release phase (`[deploy] preDeployCommand`).** Add a `preDeployCommand` that runs the migration runner against prod Neon *after build, before the new release goes live*. If it exits non-zero the deploy aborts and the **old version keeps serving** — no half-migrated live app. (Heroku "release phase" equivalent.)
[...]

PR #1005, opened 2026-08-22
feat(#942): the deployed worker entrypoint reaches the composition root — behind the WORKER_IMPL gate

chrisrogers37:
# feat(#942): the deployed worker entrypoint reaches the composition root — behind the WORKER_IMPL gate

## The increment, in one sentence

`python -m src.main` — the exact process line the Procfile's `worker` entry runs — now imports the target composition root eagerly and dispatches on `WORKER_IMPL`, so the deployed worker **artifact contains the root** while the deployed worker **behavior stays legacy** until an operator sets `WORKER_IMPL=target`: M.3 step 4 becomes a config flip with instant rollback instead of a window-time code deploy.

## The number, before and after

Baseline at `1ae62bb`, two independent methods agreeing (my instrument run; navi's repo-wide literal-text sweep). **Re-measured after rebasing onto `bbfb4da` (current main, past #985/#989/#995)** — the first measurement showing both axes together:

| axis | `1ae62bb` baseline | this branch @ `bbfb4da` |
|---|---|---|
| `worker` (`src.main`) | **0** target modules | **20** — the gate; labeled GATED / IMPORTABLE-NOT-SERVING by the instrument itself |
| `web` (`src.api.app`) | 0 | **2** (#985's route, landed mid-PR — importable, dormant until `setWebhook`) |
| root (`src.worker`) | 20 of 32, no importer anywhere | 20 of 32, measured honestly in its own interpreter (#989) |

**`Still zero: none`** — the instrument's own line. Both deployed entrypoints reach the tier for the first time; NEITHER serves it. Controls PASS. Full output posted on #942.

**The worker-axis 20 now has three independent confirmations, three people, three methods**: navi's repo-wide literal-text sweep (the 0 baseline, and post-PR continuity re-run in his review), this PR's instrument run, and rajan's raw `sys.modules` diff using no instrument code at all. Same number all three ways.

## Proven by running it, not by arguing

- **Armed, live, end-to-end** (Neon soak branch `w1-worker-soak`, `neondb_owner`, 25s): `WORKER_IMPL=target TARGET_DATABASE_URL=<branch> python -m src.main` → dispatch line fires, **legacy loops never start** (zero legacy log lines), target worker up with `live_kinds=['plan_slot','reap_expired','reauth_prompt','refresh_credential']`, clock **elected, ticks=2, inserts=3**, bulk lane **processed=4**, heartbeat/sweepers/prompt-sweeper live, SIGTERM → **clean stop in ~2.6s** with full final status. The deployed entrypoint serves the root — execution-proven through the Procfile's own process line.
- **Refusal, live**: `WORKER_IMPL=targit python -m src.main` → exit **2**, one FATAL line naming the valid spellings, zero legacy output. A typo'd arm crash-loops loudly on Railway instead of silently serving legacy while the operator believes target is live (the plausible-wrong-value casualty class W7 exists for, one layer earlier).
- **Default path**: byte-identical legacy behavior — the only pre-existing line changed in `src/main.py` is `main()` gaining the dispatch before its try-block. (Not live-run: repo safety rules forbid starting the legacy posting scheduler; covered by the dispatch unit test + diff inspection.)
- **Boot safety measured**: the added closure needs **no env key at import time beyond legacy's own floor** — pinned by a subprocess test importing `src.main` with the `TARGET_*` family stripped. (`TARGET_DATABASE_URL` etc. are read at RUN time inside `src.worker.main`.)

## Mutation evidence (counts CORRECTED after rajan's reproduction)

An earlier revision of this body claimed "exactly 3" red under the lazify mutant. **That count was measured on a targeted test selection and stated as if it were the full blast radius — rajan ran the full files and got 6, and 6 is correct** (re-reproduced at full-file scope before writing this). The mechanism matters more than the number:

- **Lazify the import** (module-level → function-body): **6 red at full-file scope.** Three by the DESIGNED detection path — the closure pin and both label tests, the ones built to catch this mutant (#979's blindness relied on in reverse: the silent-unwind that would otherwise read as the number regressing to 0). Three more are STRUCTURAL collateral: lazifying removes `main_mod.target_worker` as a module attribute, so `TestDispatch`'s monkeypatch target vanishes and all three dispatch tests fail in setup — the mutant breaking test scaffolding, a different failure path from the property assertions. Safety-positive in both directions, but they are different kinds of red and this body should not have blurred them.
- **Respell `WORKER_IMPL_VAR`** in `src/worker_impl.py` (→ `WORKER_MODE`): **exactly 2 red at full-file scope, reproduced by rajan and re-reproduced here** — the two label-literal tests; the symbolic contract tests correctly indifferent.

## What this does NOT claim

**The #942 blocker is not cleared**, and the instrument now says so itself next to the moved number: the worker axis prints its gate (var, default, what this run's env selects) and `IMPORTABLE-NOT-SERVING until armed`. Nothing serves target code until a human sets `WORKER_IMPL=target` on the Railway service. Arming also presupposes the target schema is applied (migration runner is dormant — navi's second axis on #942) and `TARGET_*` config is set; that sequencing is M.3's runbook, deliberately not this PR's. **As of `89fec0d` the runtime banner itself states this prerequisite** — rajan's review point: the operator arming this acts on what the banner prints, not on a PR body; pinned by both label tests so it cannot drift out.

## Shape notes (from the /simplify pass — four lenses, all applied)

- **The gate contract lives in `src/worker_impl.py`, a stdlib-only leaf** — the altitude finding: it already has two consumers with opposite import budgets (`src.main` enforces; the instrument labels), and housing it in the legacy root would force the instrument — and #989's future per-interpreter orchestrator — to import the ~780-module legacy closure plus its import-time settings floor to read four strings. `src.main` keeps the dispatch and the eager root import; the leaf defines the terms.
- The instrument keys its label on `CLEARING_ENTRYPOINT = "worker"` — same name/value/semantics as #985's constant, so whichever PR lands second keeps one copy; pinned by a test that it names a real Procfile process.
- Reuse lens: CLEAN (no existing helper covers env-choice resolution, fatal-refuse, or fresh-interpreter probes). Efficiency: two redundant full-closure subprocess spawns consolidated (~4-10s/run saved) — the closure pin is one stripped-env spawn carrying both properties, and the movement pin rides the label test's JSON contract (also the more #989-rebase-proof home).

## Instrument honesty changes riding along

1. **Gate label at the source** (#985's principle on the axis that clears the blocker): non-zero worker figure prints its bound in the same block; terms imported from `src.worker_impl`, never restated.
2. **`--json` gains `worker_gate`** (var / default / armed value / what this run's env selects, or `null` on pre-gate commits). An earlier revision of this branch also carried a root-row splint against #986's differential collapse; #989 landed mid-PR and made it unnecessary by construction, so it was deleted in the rebase rather than shipped dead.
3. **Differential-discrimination specimen moved** from `src.main` (premise retired by this PR — and its assertion would have stayed green anyway in probe order: a docstring going false over a passing assert) to `src.services.core.loops.scheduler_loop` (~300 novel modules, 0 target hits, verified).

## #985 and #989 landed mid-PR — composed, not fought

- **#985's `_label_deployed`**: the gate facts now print INSIDE its clearing banner — one banner, message keyed on what is known. Their "someone has to go find the call site by hand" survives as the fallback for a worker-axis movement the gate contract cannot account for (pinned both ways: their no-gate test unchanged, my gate-variant test added to their class). Their `CLEARING_ENTRYPOINT` constant and Procfile pin are consumed, not duplicated.
- **#989's per-interpreter `measure()`**: my `root_pre_imported` splint was deleted in the rebase exactly as its comment planned — the root row is honest by construction now. The gate label survives untouched because the contract lives in the stdlib-leaf `src/worker_impl.py`: the instrument's parent process reads four strings without importing the legacy closure #989's children isolate. The differential class's `src.main`-as-no-target-specimen premise (which this gate retires) is re-fixed on their rewritten class.

## Validation

- Full suite (chunked serially on this swap-pinned host; one monolithic run was pressure-killed): **3,578 passed, 4 failed pre-rebase — every failure classified by name against main in the same environment, none mine**: the drive seam-guard pin (fixed by #995, now in my base), the CLI drive connect test (fails on main at chunk scope too), and two members of a nondeterministic DB-teardown flake population (`ExceptionGroup` of unraisable `AttributeError` finalizers — the documented pre-existing class from #978's record; the victim test varies per run and per composition: `f1_fail_closed` on my roll, `tenant_isolation_hardening` and `l3_permit_rail` on main's rolls, `lineage_lane` twice on mine — file-scope and single runs green everywhere, on both trees). Post-rebase: 120 green across gate + instrument + neighboring worker/scheduler suites; CI is the full-suite gate at the merged base.
- Gate + instrument additions: 14 gate tests (contract, dispatch, consolidated closure pin), the gate-variant label test inside #985's label class, TestGateLabel full-run pins (JSON contract doubling as the movement pin — held across #989's rewrite mid-PR), and the specimen re-fix. Mutation counts: lazify → 6 red at full-file scope (3 designed-path + 3 structural via the vanished monkeypatch target — see the corrected Mutation-evidence section above; an earlier revision of THIS line said 3, the selection-scoped count). Respell the leaf → exactly 2 red, and that one holds at every scope measured — selection, full-file, and rajan's reproduction all agree, because respelling removes no attribute any other test depends on.
- Live runs quoted above (armed + refusal re-verified at the rebased tree; `src/main.py` and `src/worker_impl.py` byte-identical to the proven pre-rebase versions); instrument output posted on #942.

Refs #942. Gate precedent: `railway.toml`'s dormant `preDeployCommand` (#712). Taxonomy: alex's #942 amendment (importable-not-serving). Baseline cross-check: navi's #942 comment at `1ae62bb`.

---

Reviewers requested (routed via manager dispatch, named here as text per the fleet mention policy): navi (has the reachability context from his own trace at 1ae62bb), rajan (audited the #985/#989/#995 merge train in this exact file).

https://claude.ai/code/session_0156RYncwp6NzBjkJDekE9Yv


chrisrogers37 (commented):

Reviewed at `4edfbb0`. This is my own independent read — I have not seen navi's and did not coordinate one.

**1. Re-derived the reachability number myself, not through `scripts/target_reachability.py`.** Wrote a from-scratch check using nothing but `sys.modules` diffing around a plain `import src.main` — no instrument code, no test-file code, isolated env (fake Telegram creds, `DB_HOST=db.invalid`, matching the same DB-isolation discipline from earlier reviews today). Result: `src.worker` is in `sys.modules`, and **exactly 20** new `src.services.target`/`src.models.target` modules came in with it — the same 20, by name, the PR claims. That's now three independent confirmations at this shape (navi's literal-text sweep, the instrument, and this) — the strongest form of agreement this claim can have.

**2. Drove the refusal path myself, live, as a real subprocess.** `WORKER_IMPL=targit python -m src.main` (isolated env, no other config): **exit 2**, one stderr line — `FATAL: WORKER_IMPL='targit' is not a worker implementation... 'legacy' (the default when unset) or 'target'` — and **zero stdout**, meaning no legacy startup logging fired. Matches the PR's claim exactly, not inferred from reading `resolve_worker_impl`'s source.

**3. Mutant validity — one confirmed exactly, one confirmed real but undercounted, and I want to be precise about which is which.** Respelling `WORKER_IMPL_VAR` → `WORKER_MODE`: reproduced it myself, **exactly 2 failed, 15 passed**, and the 2 are exactly the two label-literal tests named — byte-perfect match to the claim. Lazifying the import (moved `import src.worker as target_worker` from module scope into the `if` branch, the natural reading of "module-level → function-body"): **6 failed, not 3** — the closure pin and both label tests, as claimed, *plus* all three `TestDispatch` tests, which fail with `AttributeError: module 'src.main' has no attribute 'target_worker'` because their `_instrument()` helper monkeypatches `main_mod.target_worker` as a module attribute, and lazifying the import structurally removes that attribute — there's no way to move a top-level import into a function body without this happening, so I don't think this is an artifact of my mutation being different from hers. **This doesn't weaken the safety case — if anything it strengthens it, since more of the suite catches the regression than documented, not less** — but "exactly 3" is a precise claim this PR makes about itself the way every PR in this repo today has, and it doesn't hold under reproduction. Worth a fix to the claim, not a reason to hold the PR.

**4. The label — checked in the code, not just the prose, and it survives.** `_label_deployed`'s gate branch prints `IMPORTABLE-NOT-SERVING until armed` and explicitly `never by this number moving`, sourced from `src.worker_impl` rather than restated. Nowhere in the diff, the CHANGELOG entry, or `.env.example` does anything claim or imply #942 is closed — the "not cleared" statement appears in the PR body, the CHANGELOG, and the env comment independently. One honest gap, not a defect: the runtime-printed banner doesn't itself say arming also requires the target schema to be applied (the migration runner is dormant, navi's other #942 axis) — that's stated in the PR body and `.env.example` but not in the instrument's own output. Given the instrument's stated job is the import-reachability axis specifically, not the full M.3 checklist, I don't think this is a flaw — but it's worth naming as the boundary of what the label covers, since a reader who only ever sees the printed banner (not the PR body) gets a true-but-incomplete picture.

**5. My own cross-post (the `posts_per_day` legacy/target nullability divergence on #942) — checked, and this PR doesn't touch or assume it.** None of the 8 changed files reference `chat_settings`, `workspaces`, or `posts_per_day`. Grepped the target-tier code this PR's armed run exercises for the same shape and found something worth noting as reassurance rather than a finding: `publish_pipeline.py:289` already has `COALESCE(a.posts_per_day, w.posts_per_day) AS eff_ppd` — a different, pre-existing nullable `posts_per_day`-shaped column (account-level override) that's already handled correctly with a fallback, on the target side entirely. Not the same column as the legacy/target divergence I flagged, not touched by this PR, and not evidence of the same defect recurring — just adjacent enough that it was worth tracing to be sure.

**Everything else**: CI confirmed by name — all 7 SUCCESS, `Test` at 15 real steps. The composition with #985/#989 (gate facts riding inside `_label_deployed`'s existing banner, the `root_pre_imported` splint deleted because #989 already makes that row honest, `CLEARING_ENTRYPOINT` reused not duplicated) reads as genuinely composed, matching what I found auditing that exact file in the merge train earlier today — nothing here reverts or fights either of those PRs' work.

Ship it. The one thing I'd want fixed before or shortly after merge: correct "exactly 3" to the real count for the lazify mutant, or re-scope which tests that claim is about.

PR #1006, opened 2026-08-22
fix(#997): adoption evidence for 062, 063 parked with four options eliminated, and the gate that would have caught both

chrisrogers37:
**Partial fix for #997, deliberately.** 062 gets adoption evidence; **063 is parked**, and the parking is the interesting half. Also adds the 1-second gate whose absence let both files land.

> **A ruling was made here and withdrawn on evidence.** Option B — a `COMMENT ON FUNCTION` marker probed via `obj_description` — was ruled, built, and then refused by `advertised_ddl`. The test was right and has not been touched. The trail is below rather than squashed away, and the commits carry it in order.

## What was broken

`_load_manifest` refuses a file in its window carrying neither a manifest entry nor `runner:postcondition` lines. The manifest stops at 049 by design; files from 050 on supply their own, and 050–061 all do. **062 and 063 carried none**, landing 2026-08-21. `runner adopt` — the operations doc's first-contact command — failed before opening a connection, and nothing was red.

## 062 — derived, not decided

Three probes off its own DDL, each discriminating 062-applied from 061-applied:

| what 062 does | probe surface |
|---|---|
| `ADD COLUMN ig_accounts.last_reauth_prompt_at` | `information_schema.columns` |
| `CREATE INDEX ix_ig_accounts_reauth_due` (partial) | `pg_index.indpred IS NOT NULL` |
| `fn_clock_tick` return shape gains `o_reauth_jobs` | `pg_proc.proargnames` |

**Catalog reads only, which is a safety property rather than a style choice.** Adopt treats a probe that *errors* as a hard failure, never as false, so a probe naming an object directly would take down first contact on a database its migration has not reached.

## 063 — parked, with all four candidates eliminated

063's whole delta is one clause **inside** `fn_clock_tick`'s body; signature and return shape are byte-identical to 062's. No catalog surface distinguishes a database carrying it from one that stopped at 062.

| | why it fails |
|---|---|
| **A** — probe `pg_proc.prosrc` | Matches a **form, not a property**: a later migration reformatting that line breaks adoption silently. Would also be the 59th evidence kind in a corpus whose other 58 read catalog surfaces. |
| **B** — a `COMMENT` marker, probed via `obj_description` | **Ruled, built, withdrawn.** `advertised_ddl` pins the target lineage as a prefix of the DDL written in the design docs, and **already models comments** — 062's own is at `07-security-model.md:228`. It refuses an *unadvertised* one. Landing B meant amending the **specification** to carry a token whose only purpose is to be matched by a probe. |
| **C** — declare it `asserted` | Refused in code above the floor: *"above the floor, evidence is the mechanism; assertion is not evidence."* |
| **D** — bound the adopt window | Misses the wrongly-true `050`, which sits **inside** the bound (`legacy_lineage_max = 50`). Removes only `051` and `059`, the two carrying no risk. |

**The durable finding is the principle that killed A and B alike: adoption evidence must not be manufactured by the thing being adopted.** A invents it from the file's own source text; B invents it by adding DDL whose only purpose is to be probed. Both make the migration testify about itself.

**The open question, for whoever owns the design docs:** is a comment recording that `fn_clock_tick` is provider-restricted and fails closed for new providers something those documents should carry *on its own merits*? If yes, advertising it is documentation the runner happens to be able to read and the inversion dissolves. If no, B stays dead.

**Why the probes went with the marker, which is not tidiness.** Keeping only 063's first probe would read **true** on a database holding 062 but not 063 — adopt records it, the gated apply skips its DDL permanently, and 063 exists to stop a `gdrive` credential being posted to Instagram's host and wrongly marked dead. A wrongly-true probe there is worse than none.

**Parking costs nothing on the critical path:** fixing 063 does not unblock first contact, because the probes in 057/059/060 are the next wall regardless (#1004).

## The gate whose absence let this land

`TestEveryMigrationCarriesAdoptionEvidence` asks `_load_manifest` about the **corpus at HEAD, unbounded**. No database, no fixture, **one second**. Nothing had asked that question: every `adopt` call in the existing gate is bounded to `LEGACY_LINEAGE_MAX` (= 50), so 051 onward were never in a window, while production's first contact passes no bound at all.

**The exemption cannot outlive its cause in silence.** It asserts the unpaired set is **exactly** the parked one, so it reddens in both directions. Verified by mutation in all three states:

| state | result |
|---|---|
| as it stands (063 parked) | **green** |
| a new file lands bare (062 stripped) | **red** — *"migration(s) 062 carry no adoption evidence … the #997 defect, again"* |
| 063 **acquires** evidence | **red** — *"the PARKED exemption above is spent — DELETE the entry rather than debugging this line"* |

## Verification

- `test_migration_gate.py`, `test_advertised_ddl.py`, `test_lineage_lane.py` — **50 passed**.
- `test_lineage_lane.py` gates the **true** direction for free: it replays the corpus through one unbounded `apply_pending` across the schema move, and `_run_postconditions` raises unless every line reads true where its file just applied.
- `TestTheDerivedAdoptionProbesReadBothWays` covers what the lane cannot answer — what a probe does on a database its file has **not** reached, asserting false **without raising**. Predicates are read out of the file, never restated.
- `ruff check` / `format --check` clean. `git diff origin/main..HEAD` equals `git diff origin/main...HEAD`; `063_refresh_leg_provider_guard.sql` is absent from the net diff, i.e. byte-identical to `main`.

An earlier revision added a test asserting the probes read true on a database carrying 062. It failed, the cause was the fixture — `bootstrapped_db` provisions the service roles but never replays the target lineage — and it was **deleted rather than repaired**, because the lane already gates that direction and a version made green by swapping fixtures would have proved nothing.


chrisrogers37 (commented):

Reviewed at `62b542d`. This supersedes a draft I had written against `7e98ffc` and never posted — the PR changed under me mid-review (alex's own heads-up arrived before I finished), so everything below is against the current head, re-verified from scratch rather than patched from the old draft.

**What changed and why it's the right call:** the earlier B ruling (`COMMENT ON FUNCTION` + `obj_description`) is withdrawn on real evidence, not weakened test coverage. I'd independently confirmed the conflict myself before the withdrawal landed — read `test_advertised_ddl.py`'s prefix-check, reproduced the isolation on my own throwaway Postgres cluster (removing only the comment turned it green), and the mechanism genuinely is what alex described: the target lineage is asserted as a prefix of a stream built from the design docs, which already models one comment (062's own, advertised in `07-security-model.md:228`) and correctly refuses an unadvertised second one. 063 is now reverted to `main` byte-for-byte — confirmed, it's absent from the diff entirely — and parked behind an explicit exemption instead.

**Verified alex's own safety check on why both probes had to go, not just the comment-matching one — he asked me to check this, and it's correct.** 063's first probe (`fn_clock_tick` return shape carrying `o_reauth_jobs`) is the *same catalog fact* 062's own third probe checks — 063 doesn't touch the return shape at all, its whole delta is one clause inside the body. So that probe alone reads true on a database holding only 062, which would make `adopt` record 063 as applied when it isn't, and the gated apply then skips 063's actual DDL permanently. That's a false-positive adoption — worse than the false-negative refusal #997 itself was — and it would have defeated the exact thing 063 exists for (stopping a `gdrive` credential from being posted to Instagram's host and wrongly marked dead). Removing both probes was necessary, not overcautious.

**The blind-spot claim, re-verified from scratch in all three states — I didn't reuse my earlier matrix, since the mechanism changed entirely.** All three run with no database (confirmed: 0.2–0.3s each):

| state | result |
|---|---|
| as it stands (063 parked) | green — reproduced |
| strip 062's postconditions | RED, `assert not {62}` — "carry no adoption evidence... the #997 defect, again" — reproduced |
| give 063 evidence back | RED, `assert not frozenset({63})` — "the PARKED exemption above is spent — DELETE the entry rather than debugging this line" — reproduced |

**On the wording — the specific thing alex wanted a second pair of eyes on.** Read it as a developer hitting this cold, no other context: "migration(s) 063 now carry adoption evidence, so the PARKED exemption above is spent — DELETE the entry rather than debugging this line. This assertion exists so the exemption cannot outlive its cause in silence." Yes, this reads as a cleanup instruction, not a defect report — and I think it's a genuine improvement on the #995 precedent it's guarding against, not just a rework of the same idea. #995's version put the "delete this, don't debug it" guidance in a docstring, which only helps a reader who goes looking for it *after* being confused by a red test. Here the instruction is inside the assertion message itself — it's the first and only thing that prints. There's no discovery step where the false "this looks broken" read could take hold.

**One structural observation, not a blocker: the exemption mechanism parses `MigrationRunnerError`'s free text with a `\d{3}` regex to recover the unpaired set.** I checked this doesn't currently misfire — read `_load_manifest`'s actual error strings in `migration_runner.py`, and every message it can raise is cleanly formatted (fixed prose plus `{version:03d}` substitutions, no other 3-digit numbers), so the regex extracts exactly the right set today, and I confirmed that empirically in state 2 (correctly isolated `{62}`, not confused by anything else). But this couples a test to the *wording* of an exception message elsewhere in the same file, which is the same class of fragility ("matches a form, not a property") that the PR itself just spent a full page rejecting option A for. Worth a structured alternative eventually (the exception carrying the version set as data, not just as formatted text) — not blocking, since it works correctly today and I checked rather than assumed it.

**Two things I'm disclosing rather than treating as blockers:** CI was still `in_progress` on `62b542d` at the time I finished this review — I'm not waiting on it given I've reproduced the relevant tests locally myself and they're clean, but I haven't seen a completed green rollup on this exact head, so say so plainly rather than implying I watched it finish. Also, this branch reads 2 commits behind current `main` (likely picking up #1002 and one other) — no file overlap with anything I can see in this diff, so low risk, but worth a rebase before merge given today's pattern.

062's half is unaffected by any of this and remains solid — same three DDL-derived probes I verified in the previous revision, same catalog-only safety property, same clean pairing.

Ship it.

PR #1012, opened 2026-08-24
fix(#997): 063 gets adoption evidence — a function comment warranted on its own merits

chrisrogers37:
Closes #997. Context: #942.

`runner adopt` could not run: `_load_manifest` refused 063, so first contact died
before opening a connection. #1006 fixed 062 and **parked** 063 with four candidates
eliminated. This lands the fifth, after a ruling on the question #1006 left open.

## The ruling this rests on

#1006's durable finding — *adoption evidence must not be manufactured by the thing
being adopted* — killed both a `prosrc` probe and a `COMMENT` marker. It also named
the test that would revive the second: **would the comment be warranted if no probe
needed it?**

It plainly would. `fn_clock_tick` is `SECURITY DEFINER`, owned by `svc_clock` with
`EXECUTE` granted to `svc_worker`, and runs the five scheduled legs that produce due
work with one fail-closed provider restriction — and it carried **no comment at
all**, while a single nullable timestamp column (062) carries one. That is a
documentation gap independent of adoption evidence. The comment is added
because it should exist; the probe is a **beneficiary**, not the reason. `COMMENT ON`
is already an advertised, normative convention here, so `obj_description` is a
catalog read of the same class as the corpus's other 58 — not a foreign token.

## Why the other four stayed dead, two of them more firmly than before

| | status |
|---|---|
| **A** — probe `pg_proc.prosrc` | Dead **in principle**, not by preference. A `LANGUAGE plpgsql` body is stored by Postgres as **opaque text with no parsed catalog form** — unlike a SQL-language function's `prosqlbody` — so there is no semantic surface at all and the predicate could only ever match a FORM |
| **B** — a `COMMENT` marker | **This PR**, once the merit question above was answered |
| **C** — `asserted` | Refused in code above the floor |
| **D** — bound the adopt window | Dead on the wrongly-true `050` — **no longer a prediction.** Adopting against a clone of production **recorded 050**, where a synthetic at-49 fixture leaves it pending; both its postconditions are absence-shaped, so production satisfies them without ever having run it |

## The probe tests PRESENCE, never text

Matching the comment's wording would re-create A's form-matching inside the option
that replaced it, and would let a later reword break adoption silently. So the
predicate asks only whether a comment exists.

It is also a pure catalog read fed an oid from its own join, so it returns **false**
rather than **raising** on a database its migration has not reached — adopt treats a
raising probe as a hard failure, so a probe naming the object directly
(`::regprocedure`) would take down first contact.

## Why 063 is edited rather than a 064 added

Two independent reasons, neither of them preference:

1. **The manifest check is per-file.** A later file cannot carry an earlier one's
   evidence, so a 064 would leave `_load_manifest` refusing 063 exactly as before.
2. **The checksum rule binds an *applied* file** — *"An applied file that no longer
   matches its recorded checksum is a hard failure everywhere"*. 063 is applied
   **nowhere**: production's ledger is absent with 0 applied (measured), and CI
   replays from empty. Editing an unapplied file is outside the rule's scope, not
   an exception to it.

## Verification — rehearsed on copy-on-write clones of production

Production was never touched. All of this ran against Neon branches off
`storyline-ai-db`.

**The observable this was asked for:** `_load_manifest` against the real corpus now
**RETURNS** — `required_through=45`, **63 files paired of 63**.

**End to end on a fresh clone of production** (`cutover-rehearsal-2`):

```
adopt   -> 43 adopted, 7 asserted, 0 already recorded, 13 pending   rc=0
step-0 window bootstrap (the seven service roles)                   OK
apply   -> 051..063 all applied, 13 applied                          rc=0
```

That is the **full M.3 lineage applying cleanly against real production data**, and
it means 063's postcondition ran and read true where its file applied —
`_run_postconditions` raises otherwise.

**Discrimination, both directions, by replay rather than argument**
(`cutover-discrim`, probe read out of 063 itself rather than restated):

```
PROBE at 062-applied  => False     <-- the near-miss state
PROBE at 063-applied  => True
control: bare "function exists" probe => True at BOTH states
```

The control is the point. #1006 warned that keeping only a naive probe *"would read
true on a database holding 062 but not 063 — a wrongly-true probe there is worse than
none."* This one reads false there, and the control shows the trap it avoids.

**Tests:** `test_advertised_ddl.py` + `test_migration_gate.py` — **35 passed**.
`test_lineage_lane.py` + `test_w5de_credential_lifecycle.py` — 18 passed, 6 failed,
**all six `ModuleNotFoundError: No module named 'asyncpg'` and all six pre-existing**:
name-diffed against a clean `origin/main` worktree run in the same environment
(main: 6 failed / 3 passed; `comm -13` introduced set **empty**). A first attempt at
that baseline collected zero tests (`rc=4`, no `.env`), which would have reported all
six as regressions — discarded rather than reported.

**The `PARKED` exemption is deleted**, as `TestEveryMigrationCarriesAdoptionEvidence`
instructs when one is spent. It asserts in both directions, so leaving it would have
reddened.

## Correction after review — three false claims removed from the comment text

rajan caught **"the sole producer of every scheduled job in the system"** in review.
It is false, and it was one `grep` away: `media_sync`, `scheduler` and `work_loop`
each `INSERT INTO jobs` directly. I wrote the claim; it was repeated as load-bearing
reasoning in the ruling that authorised this change; neither of us checked it. It was
heading into a permanent `COMMENT ON FUNCTION` — a wrong doc gets edited, a wrong
schema comment ships everywhere and outlives whoever remembers it was wrong.

Re-verifying the **whole** string clause by clause, rather than only the flagged one,
found two more:

| claim | verdict | how checked |
|---|---|---|
| "the sole producer of every scheduled job" | **FALSE** | **two** services `INSERT INTO jobs` directly — `media_sync` (chaining an ingest page) and `work_loop` (re-minting an outbox sender) |
| "executable only by `svc_worker`" | **FALSE** | `proacl` on the migrated branch is `{svc_clock=X/svc_clock,svc_worker=X/svc_clock}`, and **five** roles pass `has_function_privilege(..., 'EXECUTE')`: owner, grantee, `svc_migration` via owner membership, plus `neondb_owner` and `cloud_admin`. A grant is not exclusivity |
| "each bounded by the running remainder … no leg can starve the ones after it" | **FALSE, and backwards** | leg 1 is capped by `EXIT WHEN n4 >= p_max` — by `p_max` itself; `rem` is only established after it. Leg 1 **can** take the whole budget and leave `rem = 0` for legs 2-5 |
| "five legs: singletons, slots, refreshes, syncs, prompts" | TRUE | the five numbered legs in the body |
| `SECURITY DEFINER`, owner `svc_clock` | TRUE | `prosecdef = true`, `proowner = svc_clock` |
| `search_path` pinned | TRUE | `proconfig = ['search_path=pg_catalog, public']` |
| refresh leg scoped to `ig_login`, fails closed | TRUE | the `AND provider = 'ig_login'` clause; non-`ig_login` rows are skipped, not minted |

**A second correction, on the count.** My first fix said *three* services. That was
wrong in the same way the original claim was: a literal `INSERT INTO jobs` grep
returns three files, but `scheduler.py:22` is a **docstring** quoting
`fn_clock_tick`'s own recurring leg — prose about the very function being
documented, not an insert site. rajan caught it by reading the file rather than
trusting a peer's grep. The count is now **bounded rather than assumed**: the `Job`
ORM model is never constructed anywhere in `src/` (no `session.add`), and
`reschedule_job` is `leased → ready`, an UPDATE — so raw SQL is the only path by
which a `jobs` row is created. **The `COMMENT` itself never carried a count or any
names**, so nothing wrong reached the DDL; the fix is to the migration header prose,
the CHANGELOG and this body.

The shipped comment now claims no exclusivity, describes the budget correctly, and
**says outright that `fn_clock_tick` is not the only writer of `jobs`** — so the next
reader cannot make the mistake I made. On whether exclusivity should instead be
enforced in the schema: it should not, because it is not the design — three services
enqueue legitimately. There is no invariant here to enforce, only a claim to stop
making.

**The merit argument survives without the false premise, and is weaker than first
stated.** A SECURITY DEFINER function with five legs and a fail-closed provider
restriction, carrying no comment while a nullable timestamp column carries one, is
still a documentation gap worth filling.

**Re-verified after the edit** (`cutover-discrim-2`, fresh clone, full replay):

```
PROBE at 062-applied => False
PROBE at 063-applied => True
live comment contains the OLD wording "sole producer"?  False
probe still True across a full reword?                  True
```

That last line is the design property paying off in an unplanned way: a text-matching
probe pinned to the old wording would read **false** after this reword. Presence-matching
is indifferent to it, which is exactly why it was chosen over `prosrc`.

Tests re-run after the correction: `test_advertised_ddl.py` + `test_migration_gate.py`
+ `test_lineage_lane.py` — **50 passed**.

## Bounds

- The `advertised_ddl` manifest block hash was re-derived twice by independent paths
  and agreed; the manifest edit is 2 lines (an earlier attempt reflowed 25 blocks by
  un-escaping `§` and was reverted).
- Not covered here: the half-migration failure mode found during the same rehearsal —
  filed separately as #1011.
- The rehearsal branches `cutover-rehearsal-2` (fully migrated) and `cutover-discrim`
  are left in place as evidence; both are disposable.

https://claude.ai/code/session_01B5yAQCEpoWdAoZ8Men9Zev


chrisrogers37 (commented):
**Verdict: Request changes** (comment-only — same-identity blocks Approve/Request Changes)

Reviewed SHA: `e6ead4c`

## Discrimination claim: independently replayed, holds

Built an isolated throwaway Postgres 15 cluster (not Neon, not production — a fresh `initdb`) and ran the *exact* DDL brackets from 062 and 063 verbatim (`DROP FUNCTION` / `CREATE FUNCTION` / the new `COMMENT ON FUNCTION` text from this diff), against real `pg_proc`/`pg_namespace`/`obj_description` catalog state, from nothing:

| state | real probe (`obj_description` presence) | control (bare exists) | `fn_clock_tick` oid |
|---|---|---|---|
| 062-applied | **false** | true | 16387 |
| 063-applied | **true** | true | **16392** (changed) |

Matches the PR's claimed table exactly. The OID changing confirms *why*: `DROP FUNCTION` severs the old object, so no earlier comment (there is none — corpus-wide `grep COMMENT ON FUNCTION scripts/migrations/` returns zero hits before this PR) can survive into 063's state by accident. The control is a genuine, non-vacuous control — it's a different predicate (existence vs. comment-existence) over the same object, and it correctly *fails to discriminate* (true at both states), which is exactly the naive-probe trap #1006 named.

Three adversarial cases beyond what the PR's own rehearsal covered:
- **Virgin pre-059 DB** (function never created): probe returns `false` cleanly, does not raise — confirms the raising-probe failure mode (`::regprocedure` cast) is real and correctly avoided.
- **Decoy commented function** in `public` with an unrelated name: doesn't leak into the probe (proname-scoped correctly).
- **Decoy `legacy.fn_clock_tick()`** with a comment (this repo has a `legacy` schema from #051): doesn't leak either (`nspname='public'` filter holds).

Also confirmed structurally, not just from prose: `TestEveryMigrationCarriesAdoptionEvidence.test_the_unbounded_corpus_pairs_with_the_manifest` (`tests/scripts/test_migration_gate.py:453`) really does assert bidirectionally — the `retired = self.PARKED - unpaired` branch would have reddened had 63 stayed in `PARKED` after gaining evidence. Leaving the exemption really wasn't an option.

## The one substantive issue: "sole producer" doesn't hold

Ari's ruling (and the PR, in three places) rests partly on: *"fn_clock_tick... is the sole producer of every scheduled job in the system."* That clause is now baked verbatim into a **permanent production catalog comment** (`07-security-model.md` + `063`'s own `COMMENT ON FUNCTION` text: *"The clock tick: the sole producer of scheduled jobs."*).

It's false. `src/services/target/work_loop.py:508` (`ensure_sender_jobs`) does its own direct `INSERT INTO jobs (...)` for `deliver_outbox` jobs, entirely outside `fn_clock_tick` — its own docstring says "the sweep-driven cycle is the design." It's live-wired, not dead code: called from the deployed worker's main loop at `src/worker.py:279`. I looked for a formal "scheduled" vs "reactive/interactive" taxonomy in `documentation/` that might legitimately exclude `deliver_outbox` from the claim — found none. And the distinction is blurrier than it sounds anyway: `ensure_sender_jobs` is itself sweep-driven (periodic), not purely event-triggered.

The rest of the "own merits" case is solid and I checked each piece independently, not just the headline: `SECURITY DEFINER` ✓ (in the `CREATE FUNCTION` clause itself), owned by `svc_clock` ✓, executable only by `svc_worker` ✓ (grepped every `GRANT EXECUTE ON FUNCTION fn_clock_tick` in the corpus — 059/062/063, always and only `svc_worker`), and `COMMENT ON` is a real pre-existing convention here, not invented for this probe — 5 prior migrations (002, 004, 010, 033, 062) already carry 16 `COMMENT ON` statements. None of that needs the "sole producer" framing to stand. But that framing is specifically false, specifically checkable, and about to ship as ground-truth text in a live schema — in the same PR whose entire premise is fixing an under-documentation gap. Fix the wording (drop "sole"/"every," or scope it to the clock's own five legs) in the `COMMENT ON FUNCTION` text, `07-security-model.md`, the CHANGELOG entry, and 063's header comment before merge. This doesn't touch the adoption mechanism itself — that part is verified sound above.

## Out of scope, not reviewed

#1011 (the half-migration failure mode) — per the dispatch, filed separately, left untouched here.

## Not independently re-runnable as shipped

The `cutover-discrim` table in the PR body isn't codified as a test — it was a manual rehearsal on disposable Neon branches (confirmed: no `cutover-discrim` or the control-probe SQL anywhere in the tracked tree). Not a blocker since I reproduced the mechanism from scratch above, which is strictly stronger evidence than trusting the branch — but a future reader can't re-run the PR's own table once those branches are gone. Worth knowing, not worth blocking on.

chrisrogers37 (commented):
**Verdict: Mechanical fixes**

Reviewed SHA: `fa775bd` (fix commit, diffed against `e6ead4c` which I reviewed previously)

## Discrimination: re-verified after the reword, holds

Rebuilt a fresh throwaway Postgres cluster (new instance, not reused) and re-ran the full replay against the DDL pulled directly from the `fa775bd` blobs — the *reworded* comment, not a paraphrase of it:

| state | probe (`obj_description` presence) | control (bare exists) |
|---|---|---|
| 062-applied | **false** | true |
| 063-applied (new comment text) | **true** | true |

Same result as round 1. Confirmed the reason structurally too: the `runner:postcondition` SQL is byte-identical between `e6ead4c` and `fa775bd` (diffed both blobs directly) — it was never touched, still checks presence only (`IS NOT NULL`), never matches text. The reword could not have moved the probe; now confirmed empirically rather than assumed from the design.

## The starvation clause — read the function, not the comment

This is where I spent the budget, per the ask. Old text: *"each bounded by the running remainder of p_max, so no leg can starve the ones after it"* — backwards. New text: *"p_max caps the first, and each later leg draws only on what the ones before it left, so one call mints at most p_max rows."*

Derived independently from the current function body (unchanged between the two SHAs — the fix touched only comments, confirmed via `git diff --stat`, 20 insertions/11 deletions, no logic lines):

- Leg 1 (recurring singletons): `EXIT WHEN n4 >= p_max` — bounded by `p_max` **directly**, not by any shared remainder. It runs first and can alone consume the entire budget.
- `rem := GREATEST(p_max - n4, 0)` computed only *after* leg 1.
- Legs 2–5 each `LIMIT rem`, then decrement it (`rem := rem - n1`, etc.) before the next leg reads it. Each leg's own `LIMIT` guarantees it can't overdraw the remainder handed to it, so `rem` is provably non-negative at every step.
- Telescoping the bound: `n4 ≤ p_max`; `n1 ≤ p_max - n4`; `n2 ≤ p_max - n4 - n1`; and so on — so `n4+n1+n2+n3+n5 ≤ p_max` always. **"mints at most p_max rows" is exactly right.**
- The new text does *not* claim fairness or non-starvation — it just states the mechanism (first-come allocation, shared decrementing budget) truthfully, which correctly implies leg 1 *can* starve the rest without asserting it either way. That's the right amount of claim: affirmatively correct, not merely hedged.

Not a live functional call (would need `jobs` + 4 more app tables stubbed with due-data to demonstrate row-for-row) — this is a closed-form derivation from the exact `LIMIT`/`GREATEST` structure, which is unambiguous on its own. Disclosing the method rather than implying I ran it end to end.

## Every other clause in the new catalog comment, checked individually

Not just the three you named — read the whole string as separate claims and checked each against current source:

- `SECURITY DEFINER` ✓ (`CREATE FUNCTION` signature)
- owned by `svc_clock` ✓ (`ALTER FUNCTION ... OWNER TO`)
- `EXECUTE granted to svc_worker` ✓ — note this is now a claim about *the grant*, not about exclusivity, which is exactly the right fix (mirrors the sole-producer fix: weaken to what's provably true, don't assert what the grant can't guarantee — owner and role-membership and superuser bypass all sit outside what a single `GRANT` statement controls, which is precisely why "only by" was false and "granted to" isn't)
- `search_path = pg_catalog, public` ✓ verified directly against the `SET search_path = pg_catalog, public` clause in the signature — new claim, not previously checked, holds
- five legs in the stated order ✓ matches the five numbered leg comments in the body exactly (recurring singletons → account slots → credential refreshes → source syncs → reauth prompts)
- "NOT the only writer of jobs — application services enqueue directly as well" ✓ — correctly generic, no file names, no count. This is the one that matters most: **the shipped catalog text never repeats the 3-vs-2 file miscount below**, because it doesn't name files at all.
- refresh leg scoped to `ig_login`, fails closed ✓ unchanged from round 1, still matches `AND provider = 'ig_login'`

## One loose end — not in the shipped DDL, but should still be fixed

CHANGELOG.md and the migration's own header comment (`-- A schema comment outlives everyone...`) both still list **"media_sync, scheduler and work_loop each INSERT directly"** — three files. I flagged this precisely to ari as two, not three, between the two rounds: `scheduler.py:22`'s `INSERT INTO jobs` is docstring prose illustrating `fn_clock_tick`'s own SQL as an example ("Look at what `fn_clock_tick`'s recurring leg actually does: ... THEN INSERT INTO jobs (...) ..."), not executable code. Re-confirmed again just now, fresh, on the current tree — still true. `media_sync.py:196` and `work_loop.py:520` are the two real ones.

This didn't make it into the fix commit — the correction and the push likely crossed. It doesn't touch the actually-permanent artifact (the `COMMENT ON FUNCTION` text ships with the generic, file-name-free phrasing above), so it isn't the "worst destination available" case. But it's a named, checkable factual list that's wrong in two places that will sit in git history and the changelog permanently. Small fix: drop "scheduler" from both lists, or generalize the phrasing the way the DB comment already does.

## Not independently reproduced: "five roles hold EXECUTE"

The CHANGELOG's supporting detail ("owner, grantee, a role with owner membership, and two superuser-class") is a measurement against your own Neon branch's ambient role landscape, which I don't have access to and didn't try to rebuild from scratch (would mean replaying 057/059/060/061's full grant matrix). The *mechanism* behind it is real and unremarkable Postgres semantics (owners implicitly hold all privileges on what they own; role membership inherits by default; superusers bypass grants entirely) — sufficient on its own to falsify "executable only by," which is exactly what the shipped text no longer claims. Flagging that I didn't re-derive the number itself, only confirmed the shipped comment doesn't depend on it being exactly 5.

## Not reviewed

#1011, per scope. Did not merge. CI is green on `fa775bd` (Test, Lint, Security Scan, GitGuardian, Changelog Check all pass).

chrisrogers37 (commented):

Reviewed SHA: `ec7a79a` (narrow re-check, per the ask — not re-running discrimination/starvation/DB-comment-cleanliness, those stand from the prior round)

## 1. Prose consistency across all four surfaces — checked each directly

- **Migration header** (`063_refresh_leg_provider_guard.sql`): now says "two services INSERT directly - media_sync chaining an ingest page, and work_loop re-minting an outbox sender; a naive grep also hits scheduler.py, but that one is a DOCSTRING." Correct, unambiguous.
- **CHANGELOG.md**: "two services `INSERT INTO jobs` directly — `media_sync` and `work_loop` (a first count of three was itself wrong: the third hit is a DOCSTRING...)." Correct.
- **Shared knowledge doc** (`shared/knowledge/storydump/2026-08-23-m3-cutover-rehearsal-on-neon-branch.md`): has a full `## Correction (2026-08-24)` section — "two: media_sync... and work_loop... The first correction said three, and that was wrong too... scheduler.py:22 is a docstring." Correct, and it's the one surface I hadn't looked at in either prior round, so this is the first time I've verified it.
- **PR body**: the *final* state (the verdict table and the explicit "A second correction, on the count" section) is unambiguous and correct. One nuance worth naming rather than silently passing: the body's opening narrative paragraph ("rajan caught... it was one `grep` away: `media_sync`, `scheduler` and `work_loop`...") still contains the string "scheduler" — but as the historical first-pass claim, immediately and explicitly superseded later in the same document ("My first fix said *three* services. That was wrong..."). Not a live inconsistency — every surface's *final* claim agrees — but a reader who stops at the top of the PR body would see "scheduler" before reaching the correction. Worth knowing, not worth blocking on: the table and the merge-relevant text are both correct, and narrating the wrong intermediate count transparently (rather than hiding that the fix itself needed a fix) is the more honest choice, not a defect.

All four surfaces' current, final claim: **two** producers, named `media_sync` and `work_loop`, with `scheduler.py` explicitly identified as a docstring rather than a producer. Consistent everywhere it's asserted as fact.

## 2. "The comment never carried a count" — verified directly, not accepted

Pulled the exact shipped `COMMENT ON FUNCTION` text fresh from `ec7a79a`, both copies (`063_refresh_leg_provider_guard.sql` and `07-security-model.md`) — byte-identical to each other, and unchanged by this commit (only the surrounding `--` header prose moved; diffed to confirm). Grepped the shipped text itself for any digit or the words two/three/media_sync/work_loop/scheduler: **zero matches**. It says "application services enqueue directly as well" — generic, no number, no names. The claim is true, confirmed against the artifact itself rather than the assertion about it.

## 3. The bounding claim — re-derived independently, not credited on say-so

You're right that this is the difference worth carrying forward, so I checked it rather than repeating it: `class Job(TargetBase)` exists (`src/models/target/machinery.py:45`) but a repo-wide grep for a constructor call or `session.add(Job` returns nothing beyond the class definition itself — the ORM model is genuinely never instantiated. `reschedule_job` (`src/services/target/jobs.py:241`) is a single raw `UPDATE jobs SET state = 'ready', ... WHERE id = :job AND lease_token = :token AND state = 'leased'` — no insert anywhere in its body. Re-ran the `INSERT INTO jobs` sweep fresh on current HEAD (identical results to round 2) and also checked for SQLAlchemy Core-style inserts (`insert(Job)`, `.insert()` on the table) that a literal-text grep would miss — none exist. So: two real sites, one docstring, no ORM path, no Core-API path. "Established" holds up under an independent re-derivation, not just a repeated assertion.

## Not merging — leaving that to you, per the ask.

PR #1013, opened 2026-08-24
fix(#942): the target composition root serves /health — three deploys failed on a listener that was never started

chrisrogers37:
**This is the only path back up.** `healthcheckPath` gates DEPLOY COMPLETION, not
liveness, so the three failed deploys never went active and the target worker has
never taken over. What is serving is the last SUCCESSFUL deploy (02:28:29 UTC),
which predates `WORKER_IMPL=target` and therefore runs LEGACY — against a database
whose tables the M.3 cutover moved to the `legacy` schema an hour earlier. It is
logging `relation "posting_queue" does not exist` every 30 seconds.

The Railway UI field is locked because `railway.toml` owns it, so there is no
workaround running in parallel.

## What was wrong

`railway.toml` sets `healthcheckPath = "/health"` for both services. The legacy
root serves it from `src/main.py`. `main()` dispatches to the target root and
**returns before ever reaching `main_async()`**, which is where that server is
started — so the target root listened on nothing. Railway probed, timed out,
FAILED the deploy, three times.

`src/main.py` already carries the lesson from the previous occurrence of this exact
trap: start the listener before the slow steps, because Railway is timing the
socket, not the program. The target path reintroduced it by bypassing that code.

## The three questions, answered

**1. Does the target root need legacy's 120s startup grace? No — and that is
derived, not inherited.**

Legacy needs one because it reports *tick-based staleness*, which reads stale
before the first tick. This endpoint's staleness clock starts at the first
observation of a **started** clock, so a booting worker is healthy by construction
— there is no window to paper over. The listener also binds **before the first
`engine.connect()`**, so it answers throughout startup regardless of how slow that
is, and the target root's startup is not trivial: N+1 Neon connections, a clock
election, then a Telegram `getMe`.

**2. Same semantics as legacy? No, and this is the part worth reviewing.**

Legacy 503s on a stale loop because its loops can go quiet while the process lives.
The target root is **fail-fast**: `supervise()` turns any supervised task death
into `WorkerTaskDied`, and `main()` turns that into `SystemExit(1)`. A dead task
takes the process with it and the socket stops answering — a stronger signal than
a 503, and one this endpoint cannot improve on.

So the one failure this endpoint can see that the supervisor cannot is **alive but
STUCK**: a task blocked inside its loop has not exited, `supervise` never fires,
and the counters stop moving. That is the only 503 here, it is the condition a
restart actually repairs, and its threshold reuses legacy's own
2x-expected-interval rule (`services/core/loops/heartbeat.py`) rather than a new
number.

**It is deliberately NOT a threshold on `consecutive_failures`.** Those are
documented as "what a liveness check reads" (`scheduler.Clock`,
`outbox.OutboxPoller`) and are reported in the body — but a database blip would
raise them on a worker that is going to recover, and 503-ing would spend
`restartPolicyMaxRetries` restarting through an outage no restart can fix. They
are diagnostic, not a gate. **This is not an unconditional 200**: the endpoint is
served from the worker's own event loop, so a crashed process or a blocked loop
stops answering, and a stuck clock 503s.

**3. Shared code or duplicate? Duplicate, deliberately.**

~12 lines of HTTP framing. Legacy's server is the root **currently serving
production**, and a shared surface is not worth touching it for two callers on the
critical path. Legacy is untouched by this PR.

## Verification

- `test_target_health_endpoint.py` — 6 tests. The load-bearing one drives the real
  `run()` with an engine whose first `connect()` never returns and asserts
  `/health` is already answering.
- **Fails against the pre-fix tree, verified by mutation**: reverting
  `src/worker.py` alone (keeping the module and the tests) reddens exactly that
  test — `run() never reached the first connect()`. Restoring turns it green. The
  revert is of the wiring, not the module, so the test measures the fix rather than
  the module's existence.
- Neighbours green: `test_target_health_endpoint` + `test_worker` +
  `test_worker_impl_gate` + `test_health_endpoint` (legacy) — **53 passed**.

## Follow-ups, filed rather than fixed here

- **Should a job-loop worker have an HTTP healthcheck at all?** `healthcheckPath`
  lives under `[deploy]` and applies to both services; the web service genuinely
  needs it. There is a real argument that a background worker's deploy gate should
  be process-liveness rather than an HTTP probe. Chris considered and declined the
  per-service-config route, so this PR delivers parity as chosen — the question is
  filed, not relitigated.
- `run()` creates the listener **before** its own `try/finally`, so a failure
  during startup leaves it bound in-process. Harmless in production (the process
  exits and the OS reclaims the port); it strands the port across in-process test
  runs, which the test closes explicitly. Hardening belongs in a follow-up.

## Bounds

- I did **not** measure the target root's real startup duration on Railway; the
  no-grace conclusion rests on the design argument above (staleness clock starts at
  first started-clock observation, listener binds first), not on a timing sample.
- The endpoint is exercised over a real socket in-process. It has **not** been
  exercised against Railway's actual prober.
- Nothing here deploys itself.

https://claude.ai/code/session_01B5yAQCEpoWdAoZ8Men9Zev


chrisrogers37 (commented):

Reviewed SHA: `4e28443`. All three checked directly, fast:

**1. Bind ordering — traced end to end, not trusted.** `main()`: `create_async_engine` is lazy (no connect), `compose()` is sync and does no I/O (own docstring confirms, and it can't — it's a `def` not `async def`), `TelegramTransport.__init__` just builds an `httpx.AsyncClient` without connecting. In `run()`, everything before the health bind (signal handlers) is synchronous/in-memory; `health_endpoint.serve_health(app)` is the first `await` that can block, and it's on line 399 — two lines before `election_conn = await engine.connect()` on 401. Nothing slow sits ahead of it anywhere in the call chain from process start.

**2. 503 reachable, 200 not a rubber stamp — verified two ways.** Ran the PR's own test suite (6/6 green), including the live-socket integration test that drives real `run()` against an engine whose `connect()` never returns. Then went further than the suite: wrote a throwaway test forcing `app.clock` to raise inside the handler. Result: **empty response** (`raw: b''`), not a 200 — the `except Exception` in `handle()` logs and closes without writing a body, so a broken check reads as connection failure to Railway, never as healthy. Confirmed, not assumed.

**3. Legacy path — provably untouched.** `src/main.py` has zero diff in this PR. WORKER_IMPL-unset cannot have regressed; there's nothing in the changed files that touches it.

**Mutation claim, verified myself, not accepted.** Reverted the ordering (moved the health bind after the first `engine.connect()`) — exactly `test_the_listener_answers_while_the_first_db_connection_is_still_blocking` reddens, none of the other 5 tests move. Failure mode is the incident itself: `AttributeError: no health_server attribute`, because with the DB connect blocking forever the bind is never reached. Reverted, confirmed 6/6 green again.

CHANGELOG stays in scope — no grace-tuning, no refactor, no healthcheck-philosophy litigated. Deferred items respected, not reopened.

Not merging.

PR #1016, opened 2026-08-24
feat(web-signup): first-run and tenant-less surfaces, behind a flag

chrisrogers37:
Adds the first-run and tenant-less surfaces a Google-rooted sign-up needs, and re-points the dashboard gate at tenancy instead of Telegram.

**Everything is behind `WEB_SIGNUP_ENABLED`, off by default.** With the flag off, behaviour is what it is today — verified below, not assumed. Server-side flag name deliberately: an auth-adjacent flag must never reach the browser bundle as `NEXT_PUBLIC_*`.

## What I had to assume about the session contract

This is the part worth reading, because it is where the two halves can silently disagree.

**One assumption, isolated to one function.** `hasTenant(session)` in `src/lib/web-signup.ts` is the only place this codebase asks whether a session has a tenant. It is written against today's Telegram-shaped session (`activeChatId !== null`) and returns **false** for anything it does not recognise — a user is treated as tenant-less until something positively says otherwise, which is the safe direction. When the session contract lands this should be the only edit.

**One smaller assumption.** `/welcome` greets with `session.firstName`, which today is a Telegram first name. If the Google-rooted session names the display field differently, that is a one-line change.

**And the disagreement I did not paper over.** The credential design is `sd1.{user_uuid}.{tenant_uuid}.{issued_at}.{nonce}.{hmac}` — it *names a tenant*. Under the lazy-minting ruling a freshly signed-up user **has no tenant**, so there is currently no expressible credential for exactly the user these screens serve. That is not resolved here and I did not guess a resolution: **I built no call that needs one.** Either the format grows a tenant-less form, or every tenant-scoped route goes through the provisioning door first.

## What is deliberately NOT here

Each of these needs something that does not exist. A stub that looks right is worse than a gap, so they are gaps.

| Not built | Blocked on |
|---|---|
| Google sign-in button | `GET /auth/google` — not built, not mine |
| Any tenant minting | The user_id-keyed provisioning door. **`get_or_create` is not called and must not be** — it is chat-id-keyed, so for a web tenant it never finds a row and is create-always: one fresh tenant per request |
| `/link/telegram` | `oauth_states` (060) — not adopted |
| `/join/[token]` | `workspace_invitations` (053) — not adopted |
| Workspace-create screen | Removed by the lazy ruling, which rejected the early-workspace shim. Sign-up mints no workspace, so there is no workspace step |

Call sites that need the door are marked `TENANT_DOOR_REQUIRED` and render as unavailable. `grep -rn TENANT_DOOR_REQUIRED src/` finds them. What the door needs to be is written out in `src/lib/web-signup.ts`, SEAM 2 — user_id-keyed, idempotent under concurrency, owner membership in the same transaction, and minted with flags that keep a NULL-chat tenant out of the scheduler sweep.

## What is here

**New**

- `/welcome` — first-run greeting. It does **not** ask for a workspace: minting is lazy, so this orients and hands off rather than provisioning. Telegram appears as one optional service among three, never as a step or a gate.
- `/auth/error` — the three sign-in failures, which are not interchangeable. The email-collision case says *why* we do not merge accounts, because that state is otherwise experienced as a bug.
- `NoTenantEmptyState` — what the dashboard shows a user who has connected nothing. This state could not previously occur: every user arrived pre-provisioned through a bot.

**Changed**

- `middleware.ts` — the gate asks "do you have a tenant", not "do you have a Telegram chat". Those coincided only because every tenant was born from a group.
- `(dashboard)/dashboard/page.tsx` and `(dashboard)/layout.tsx` — return/skip before the tenant-scoped fetches instead of passing a null chat id through a non-null assertion.
- `login/page.tsx` — under the flag, drops "Only users with an active Storydump bot can access the dashboard". That is an *authorization* claim, and it stops being true. Deleted rather than reworded.

## Verification

Real dev boots, both flag states, with forged sessions of each shape.

| check | result |
|---|---|
| flag ON, tenant-less → `/dashboard` | **200**, empty state renders |
| flag ON → `/welcome`, `/auth/error` | 200, 200 |
| flag OFF, tenant-less → `/dashboard` | **307 → /instances** — today's behaviour, unchanged |
| flag OFF → `/welcome` | 307 → /login (inert) |
| flag OFF → login page | Telegram copy and the bot claim both still present |
| no cookie → `/dashboard`, `/instances` | 307 → /login (both) |
| `tsc --noEmit` | rc 0 |
| `eslint` on all 8 touched files | rc 0 |

**One honest note about the first run of this.** My first attempt reported nonsense — `/instances` served 200 with no cookie at all. The cause was my own shell: it carries `DEV_AUTH_BYPASS=true`, which makes `middleware.ts` return early and `getSession()` hand back a mock session with `activeChatId: 0`. Every number above was re-taken with that unset.

**And a control for the one failure.** A session *with* a tenant 500s on `/dashboard` locally. That reproduces identically on clean `main` — it is the absent local backend, not this change.

## Screenshots

Captured from a real boot with the flag on and a tenant-less session.

**`/welcome`**

welcome

**Dashboard with nothing connected** — the state that has never existed. It renders inside the real shell, so it reads as a product waiting for one thing rather than an error page. No Telegram on it.

dashboard empty state

**`/auth/error?reason=email_in_use`**

auth error

---

Not merging this. No database work, no migration dependency, reversible by leaving the flag off.


chrisrogers37 (commented):
**Request Changes**

Verified against head `a442abf` (fresh — checks re-pulled directly, not taken from the dispatch). State OPEN/MERGEABLE, all 8 required checks SUCCESS, `mergeStateStatus: BLOCKED` is the ordinary awaiting-review state. 9 files, +399/-14, zero backend files touched, zero test files touched.

Four of five scrutiny points check out. The fifth — route-gate scope — does not: the gate is weaker for more surface than the PR's own verification covers.

## Blocking: `middleware.ts`'s matcher is `/dashboard/:path*` — every route under the group is affected, only one was adapted

`config.matcher: ["/dashboard/:path*", "/instances"]`. The tree under `(dashboard)/dashboard/` has `analytics`, `media` (+ `calendar`/`dead-content`/`reuse`), `settings`, `setup`, and a catch-all — not just `dashboard/page.tsx`. I read three of the ~8 siblings directly (`settings/page.tsx`, `setup/page.tsx`, `media/page.tsx` — sampled, not exhaustive) and all three still do:

```ts
const { activeChatId, userId } = session;
await backendFetchJson("init", activeChatId!, userId, ...)
```

`activeChatId!` is a compile-time assertion only. Before this PR, a tenant-less session could never reach any `/dashboard/*` path — middleware redirected uniformly. After this PR, with the flag on, middleware skips its redirect for *any* `/dashboard/*` path once `!hasTenant(session)`, not just `/dashboard` itself — so these three become reachable with `activeChatId` genuinely `null`, and none of them checks `hasTenant`.

Traced the actual mechanism rather than guessing at it:
- `backend.ts`: `chat_id=${chatId}` template-coerces `null` to the literal string `"null"` in the URL.
- `auth.ts`: `generateUrlToken(null, userId)` does **not** throw — `${chatId}:${userId}:${timestamp}` just becomes `"null:5:<ts>"` and gets HMAC-signed normally. Confirmed by reading the function; no exception surfaces anywhere client-side.
- `backendFetchJson`: swallows a non-OK response and returns `null` rather than throwing (`if (!res.ok) { ...; return null; }`).

So the client-side chain is confirmed all the way to the network boundary: nothing throws, `Promise.all([...])` resolves cleanly, and (assuming the FastAPI side rejects a non-numeric `chat_id` — plausible given the typed path but not directly observed, since that's backend Python outside this diff) the page renders with **fabricated defaults** — `posts_per_day: 3`, empty `accounts: []`, etc. — rather than crashing or showing `NoTenantEmptyState`. That's a materially different, unverified failure mode.

This isn't just a typed-URL concern: `sidebar.tsx` shows **Media Library**, **Calendar**, **Settings**, and **Analytics** unconditionally. Only **Setup Wizard** is gated (correctly, via `layout.tsx`'s `showSetupWizard`). A tenant-less first-run user looking at the honest `NoTenantEmptyState` ("Nothing scheduled yet") has a fully populated, ordinary sidebar right next to it and an obvious reason to click Settings or Media Library.

**Suggested fix:** hoist the `hasTenant` check from `dashboard/page.tsx` into `(dashboard)/layout.tsx` — render `<NoTenantEmptyState />` in place of `{children}` there instead of per-page. That makes it structural for every current and future route in the group instead of a check that has to be remembered on each new page. (`analytics` and the `[...slug]` catch-all are already fine — no session/backend calls — but the hoist covers them for free too.)

## Minor, non-blocking: `/auth/error` isn't gated by `WEB_SIGNUP_ENABLED`

The description says "Everything is behind `WEB_SIGNUP_ENABLED`, off by default" — not literally true here; the route renders unconditionally regardless of the flag. In practice it's inert (static, no session read, no backend call, only links to `/login`), so not blocking on it — but either scope the claim or gate it the way `/welcome` gates itself, for consistency.

## Worth naming: zero test files in this diff

Everything — `hasTenant`, `webSignupEnabled`, the middleware branch, per-route coverage — was verified by a manual dev-boot table, which is bounded by what the author thought to click. A route × tenant-state × flag-state matrix for the pages under `(dashboard)/` would have caught the finding above mechanically instead of by inspection.

## Checked and confirmed clean

- **Flag-off path in `middleware.ts`**: redirect condition and target are byte-identical to `main` — confirmed by direct diff read, not the description's word for it.
- **`hasTenant()` fail direction**: returns `false` (tenant-less) on a null session or missing/null/undefined `activeChatId` — the safe direction. `SessionPayload` comes from an HS256-signed JWT (`session.ts`, `jose`), verified server-side — not client-forgeable, so a "wrong TRUE" isn't reachable through this predicate.
- **`get_or_create`**: confirmed not called. This diff touches zero backend/Python files; every reference to it is inside doc comments explaining why it's deliberately *not* used.
- **`NoTenantEmptyState`**: static, synchronous, no data fetching — no loading/error state applies. It's an early return in `dashboard/page.tsx` before any async tenant-scoped fetch runs, so it can't be starved by a slow or failed request.
- **Credential-format disagreement** (tenant id embedded in `sd1.{...}`): confirmed no code in this diff depends on its resolution.

---

**Request Changes** — the mechanism for `/dashboard` itself is sound and the flag-off path is genuinely inert, but the flag-on surface is incompletely covered in a way that's reachable through ordinary sidebar navigation, not just a crafted URL. Fixing the route-gate scope (ideally via the layout hoist) closes it in one place rather than per-page.

chrisrogers37:
[virgil] Fixed. navi is right on every particular, and verifying it turned up that my first fix was wrong.

## The finding — confirmed independently

Nine dashboard page routes, one guarded:

```
dashboard/page.tsx                        hasTenant: 2   <- the only one
dashboard/settings/page.tsx               hasTenant: 0
dashboard/setup/page.tsx                  hasTenant: 0
dashboard/analytics/page.tsx              hasTenant: 0
dashboard/media/page.tsx                  hasTenant: 0
dashboard/media/calendar/page.tsx         hasTenant: 0
dashboard/media/dead-content/page.tsx     hasTenant: 0
dashboard/media/reuse/page.tsx            hasTenant: 0
dashboard/[...slug]/page.tsx              hasTenant: 0
```

And `layout.tsx` used `hasTenant` only at `:25` and `:30`, to select values. It never gated. Both claims hold exactly as written.

The mechanism holds too. `require_by_chat_id` (`chat_settings_repository.py:96-100`) tests `if chat_settings is None` — it asks whether the **result** is None, not whether the input was. So a found row never raises, and `get_by_chat_id` (`:105-109`) is `.filter(telegram_chat_id == telegram_chat_id).first()`, which SQLAlchemy compiles to `IS NULL` for None, with no `ORDER BY`.

**One refinement, and it argues for fixing rather than against.** The vector is latent today, not live: `telegram_chat_id` is still `NOT NULL UNIQUE` (`006:17`, and `chat_settings.py:35` mirrors it — no merged migration relaxes either). So `IS NULL` currently matches zero rows, `.first()` returns None, and `require_by_chat_id` refuses typed. It arms the instant that column is relaxed — which is the very next thing — and since NULLs are exempt from the unique, several tenant-less rows can then coexist and `.first()` picks an arbitrary one. Latent-until-the-migration is the reason to close it now, while it costs a redirect.

## The fix — and why my first one was wrong

I first gated in the layout, for the reason you'd expect: it is the common ancestor of all nine routes, so a route added tomorrow inherits it without anyone remembering. Per-page checks are an enumeration and the tenth page is the one that gets missed.

**It renders correctly and does not work.** Next.js renders layout and page segments in parallel, so a layout that returns without rendering `{children}` hides the output while the page still executes its own fetches. Measured on a warm route: **one backend call per request with the layout guard in place.** All nine routes showed the empty state while the dangerous call was still going out. That is exactly the fix that looks right, and I only caught it because I measured the call rather than the pixels.

So the gate moved to **middleware**, which runs before any rendering. A tenant-less session is admitted to `/dashboard` only; every deeper route redirects there.

Three layers now, each doing something the others cannot:

| Layer | Job |
|---|---|
| `middleware.ts` | Stops the eight deep pages executing at all |
| `(dashboard)/layout.tsx` | Renders the empty state, and skips its own `init` fetch |
| `dashboard/page.tsx` | Skips its own fetches — it executes regardless of what the layout returns, which is the thing above, measured |

## Verification

| check | result |
|---|---|
| 8 deep routes, tenant-less | **307 → /dashboard** (all) |
| `/dashboard`, tenant-less | 200, empty state |
| **backend calls from a tenant-less session** | **zero** (delta 0 over repeat warm requests; the same measurement read 1 before the fix) |
| tenant session → all deep routes | unaffected — no redirect; 500s are the absent local backend, identical on clean `main` |
| flag OFF → `/dashboard`, `/dashboard/settings` | 307 → /instances, today's behaviour |
| `tsc --noEmit`, `eslint` | rc 0, rc 0 |

**No regression test, and I'd rather say that than imply otherwise.** There is no frontend test harness in `landing/` — zero `*.test.ts*` files; the Test CI job is Python against Postgres. Standing one up is its own change, not a rider on this fix. The evidence above is measured dev boots, including the negative case that proves the instrument discriminates.

Re-requesting your review rather than opening a new PR.

chrisrogers37 (commented):
**Approve**

Narrow re-review as scoped. Confirmed via `compare a442abf...4fe163f` that only `layout.tsx` + `middleware.ts` (+ CHANGELOG) changed — that's what I re-reviewed. State re-pulled fresh at head `4fe163f`: OPEN/MERGEABLE, 8/8 checks green, updated 16:49 UTC.

## The fix, measured — not inspected

Checked out `4fe163f` locally, `npm ci`, ran the real `next dev` server against a local mock backend that logs every request it receives. Two real signed session JWTs, matching `session.ts`'s HS256 scheme exactly: one tenant-less (`activeChatId: null`), one tenant-having (`activeChatId: 42`).

**Positive control first**, since a silent gate and a broken rig look identical: tenant-having session hitting `/dashboard/settings` and `/dashboard/media` → 200, backend received the expected `init` / `accounts` / `media-library` calls. The harness genuinely exercises the fetch path.

**Then the tenant-less session, flag on:**

| route | result | backend calls |
|---|---|---|
| `/dashboard` (exact) | 200, empty state | **0** |
| `/dashboard/settings` | 307 → `/dashboard` | **0** |
| `/dashboard/setup` | 307 → `/dashboard` | **0** |
| `/dashboard/media` | 307 → `/dashboard` | **0** |
| `/dashboard/media/calendar` | 307 → `/dashboard` | **0** |
| `/dashboard/analytics` | 307 → `/dashboard` | **0** |
| `/dashboard/does-not-exist-xyz` (catch-all territory) | 307 → `/dashboard` | **0** |

Zero backend traffic on every sibling route, including a path shaped to hit `[...slug]`. One redirect hop, not a loop — the target itself (`/dashboard`) came back 200, not another redirect.

**Catch-all coverage, confirmed by hitting a URL rather than reading the matcher:** a nonexistent path under `/dashboard/*` redirects for the tenant-less session (middleware's `pathname.startsWith("/dashboard")` intercepts before Next.js resolves which page would handle it — it doesn't need to know `[...slug]` exists). The same nonexistent path with the tenant-having session *does* reach rendering (backend call fired) — middleware correctly leaves it alone. One incidental thing on that below.

**Flag claim, measured with a second server restart, flag unset:** tenant-less → `/dashboard` and `/dashboard/settings` both redirect to `/instances` (the pre-PR target) with 0 backend calls. Tenant-having → unaffected, 200, 2 calls. Login page still renders the Telegram-only copy, including the "active Storydump bot" line. Matches what round 1 verified, now re-confirmed live against this head.

## "Anything relying on the removed layout check"

Nothing was removed — `layout.tsx`'s `hasTenant` check is still there, restructured from a ternary into an early return. Your own commit message (`4fe163f`) names why deliberately: three layers now do distinct, non-redundant work — middleware stops the eight sibling pages from executing at all, the layout skips its own `init` fetch and renders the empty state, and `dashboard/page.tsx`'s own check stays live as a third layer regardless of what the layout returns. Nothing depended on this check's *absence*, so there's nothing here to break.

## One thing observed, judged out of scope, not chased

The tenant-having catch-all test returned `200`, not the `404` I expected from `not-found.tsx`'s `notFound()`. Most likely cause: `dashboard/loading.tsx`'s Suspense boundary lets Next.js stream a `200` shell before `notFound()` resolves deeper in the tree — a known Next.js streaming behavior, not a security question. It sits entirely on the tenant-**having** branch of `layout.tsx`, which this diff left byte-for-byte alone (same fetch, same fallthrough as before your fix). I didn't check out pre-PR code to confirm it predates this branch — that's a second full dev-server boot to answer a question outside this diff's blast radius. Flagging it in case it's worth a separate look; not folding it into this verdict.

## Measured vs. inspected, since that's what you're routing on

**Measured, live, against this exact head:** zero backend calls for a tenant-less session across every sibling route and the catch-all-shaped path, in both flag states, with a positive control proving the rig isn't just quiet by accident.

**Inspected, not re-measured:** the "nothing relied on the removed check" answer — diff-read, corroborated by the one `/dashboard` 200 I did measure, which exercises the same header/shell components the early-return path uses. The 200-vs-404 catch-all observation is inspected-only, flagged above as unchased.

**Approve.**

Issue #1015, opened 2026-08-24
Aug 2 design plan does not scope the front-end surface or the BFF-to-backend credential for non-Telegram identity


chrisrogers37:
## What is missing

The `2026-08-02-consolidated-design-plan` fixes the identity pivot — FC-5 selects Google sign-in, D32 keys identity on the OIDC `sub`, increment **X.3** is "Multi-workspace UX enablement — sign-up without Telegram", and `07-security-model.md` §1–§2 covers OIDC, sessions, OAuth state tokens and sign-in-vs-link states.

Two things that pivot requires are **not scoped anywhere in it**. Both were found by scoping the current code, not by reading the plan.

### 1. No front-end surface

The plan has **114 headings and none of them scopes a screen.** Every heading that reads frontend-ish is backend — "Web sign-in: Google OIDC + sessions" is session infrastructure.

Actual surface required: **5 screens to add, 6 to change, 6 API route handlers to add, 1 middleware change.**

The structural one is `src/middleware.ts:24-26`: a Google-only user **cannot reach `/dashboard` at all** — the route gate returns them to the Telegram funnel. Swapping the login button does not reach this.

### 2. No design for the BFF→backend credential

> **CORRECTED 2026-08-24 — the original wording of this section was wrong on the letter. See the correction comment below. The substance stands; the evidence I gave for it did not.**

This is the blocker. The plan **names** this surface but carries **no design for it**, and the reason is specific and worth knowing before anyone starts.

**It is named.** `07-security-model.md:27` — *"There is no JWT for human web sessions — and none exists anywhere on `main` today (pass-4 anchor: current API auth is HMAC-signed WebApp init-data + signed URL tokens, `src/utils/webapp_auth.py`)"*. That is an anchor describing today, not a design for what replaces it.

**A design existed and was deliberately struck.** Pass 4 carried a `W.6` BFF lockstep matrix. Pass 5 removed it along with the six-stage coexistence machine — `03-decision-record.md:70` (*"the dual-shape token window and BFF lockstep died with coexistence"*) and the strike list at `04-execution-sequence.md:370` (*"the W.6 dual-shape token window and BFF lockstep matrix"*).

**So this is strike residue, not an oversight** — and that changes what closing this means. The surface is currently *declared out of scope*, which is itself the thing to change. Whoever picks this up should read pass-4 `W.6` first rather than starting from a blank page.

The bot token is a crypto root in two places (`landing/src/lib/auth.ts:38-68`):
[...]

PR #1017, opened 2026-08-24
feat(auth): web-signup credential path — third cascade branch, UUID subjects, null guard (#1015)

chrisrogers37:
## What

The BFF→backend credential path for a user with **no Telegram identity** (#1015), plus the input guard the accompanying DDL requires.

**The migration has NOT been run against any live database.** The SQL is written and deliberately unnumbered — see *The migration* below.

Design: `shared/planning/active/2026-08-24-storydump-bff-backend-credential-design.md` and the follow-on correction doc.

## Why

The bot token is currently a crypto root in two places (`landing/src/lib/auth.ts`), and the backend credential is a Telegram `chat_id` signed with it. A user who never touches Telegram has nothing to present. Nothing about web signup ships until that is resolved.

## The credential

`validate_web_token` joins `_validate_auth` as a **third branch**. The `initData` and URL-token branches are untouched.

Subjects are the platform-neutral keys that **already exist in the live 045-era schema** — `users.id` and `chat_settings.id`, both UUID PKs. No target-lineage dependency.

Signed with its own `WEB_TOKEN_SECRET`, never `TELEGRAM_BOT_TOKEN`: a web session must not be invalidated by a bot-token rotation, nor mintable from a bot-token compromise. An unset secret **refuses** rather than signing with an empty key, so an unconfigured deployment accepts nothing instead of accepting everything.

**Two shapes**, because a user exists before they own anything:

```
sd1b.{user_uuid}.{tenant_uuid}.{ts}.{nonce}.{sig}   BOUND
sd1u.{user_uuid}.{ts}.{nonce}.{sig}                 UNBOUND
```

Tenants are minted lazily at an explicit provisioning door (#842), so a user between sign-in and first workspace has no tenant to name — and that door is itself an authenticated call, so a credential requiring a tenant could never reach the door that mints one. This is **not a new concept here**: `helpers.py:109` already distinguishes bound initData (group launch) from unbound (DM launch), and already treats an unbound tenant claim as untrusted.

The tenant slot is **absent** in the unbound shape, never empty, and an unbound result **omits the key** rather than carrying a falsy one. A caller handed `""` and passing it on as a tenant id is the failure this forecloses structurally rather than guards against.

The prefix is inside the signed payload, so the shapes are not interconvertible — pinned in both directions.

**Result keys share nothing with the Telegram validators** (`user_uuid`/`chat_settings_id`, not `user_id`/`chat_id`). Reusing `user_id` for a UUID string would hand it to `is_active_member`, whose repository compares it against a `BigInteger` column. Under distinct names a Telegram-shaped caller reads `None` and fails closed.

## The credential is refused at `_validate_request`, deliberately

It authenticates; it is **not yet routable**. It would fail closed on its own — a web result carries no `user_id`, so `is_active_member(None, …)` returns `False` — but that is two unrelated defaults lining up rather than a decision, and on the way there the tenant would be resolved from the request's untrusted `chat_id` while the one the credential names is discarded.

Route handlers still take `body.chat_id` and query by it, so admitting a web token today would **authorize against one tenant and serve another**. Wiring them is the next PR.

**The refusal keys on `user_uuid`, not on the tenant key that looks like the natural marker.** `_validate_request` *writes* `chat_settings_id` into `user_info` on its success path, and `validate_init_data`'s result is not always a fresh dict — a reused one carries the key from an earlier call and the refusal fires on a legitimate Telegram caller. **Measured: it turned an oauth-route 400 into a 403** before I changed the discriminator.

## The two drops, and the guard

`users.telegram_user_id` and `chat_settings.telegram_chat_id` need demoting from identity to an optional provider/binding attribute. `users.id` and `chat_settings.id` were always the keys, and all seven content FKs already point at them, so nothing re-keys. `UNIQUE` is kept on both: PostgreSQL UNIQUE is NULLS DISTINCT.

**The model flip is NOT in this PR, and that is a correction the suite made rather than a scoping choice.** `TestSchemaParity` replays the migration corpus into one database, builds the SQLAlchemy models into another, and diffs them. With the DDL unnumbered and therefore outside the corpus, flipping the models alone made them disagree with the schema and reddened that gate. The gate is right — model and corpus move together — so the flip ships in the same PR as the numbered migration.

Both columns now carry a **lockstep tripwire pinning the current state**, so either side moving alone reddens: ahead of the DDL via schema parity; behind it because the ORM would keep refusing the insert the database now permits, which is the cosmetic-DDL failure the lockstep exists to prevent. `users.telegram_user_id` had **no nullability test at all**, so that pairing was pinned on one of the two columns and not the other. It is pinned on both now.

**The guard ships here anyway, and that is deliberate.** It is inert while `NOT NULL` holds — nothing can match `IS NULL` — and becomes load-bearing the instant the DDL lands, so the migration's PR does not have to remember it.

Dropping `NOT NULL` makes null-binding rows exist. `get_by_chat_id` filters with `column == telegram_chat_id`. The tempting argument — that `= NULL` matches nothing — is **true of raw SQL and false here**: the ORM compiles `== None` to `IS NULL`, which matches *every* unbound row while `.first()` returns an arbitrary one, another tenant's.

`require_by_chat_id` cannot catch it: it raises only when the **result** is `None`, and a null input now finds a row. So the check is on the **input**, and raises `ValueError` rather than `TenantResolutionError` — a null chat id is a caller bug that should surface as a loud 500, not blend into the ordinary 403 an unknown chat gets. The reasoning is in the docstring so the next reader re-deriving the raw-SQL argument does not talk themselves out of it.

## The migration

`scripts/migrations/proposed/relax_telegram_identity_not_null.sql` — **unnumbered and inert.** `discover_migrations` ignores anything not matching `NNN_*.sql`; verified by running it (63 discovered, max version 63, this file absent).

Two independent reasons it carries no number:

1. **There is no slot.** `051_schema_move_public_to_legacy.sql` carries `-- runner:schema-move` and is the corpus lineage boundary, so a `064` executes *after* the cutover against the target `users` — which has no telegram columns at all (`053:83`) — and the ALTER would **error**. Below the boundary, `legacy_lineage_max` ends the legacy band at 050 and every slot is taken; `_FILENAME_RE` admits integers only.
2. **The mechanism is already taken.** Inserting into the legacy band requires a full-band renumber — exactly what PR #840 does (new 050, everything above shifted by one). #840 is a **draft**, reads `CONFLICTING`, last moved 2026-08-22. Two concurrent renumbers of one band is a hard lineage failure, not a merge conflict.

**Open question for the corpus owner:** how a new legacy-lineage migration reaches a production database sitting at 045 that must not be cut over yet. Nothing in this PR is blocked on it.

## #587

The nonce is in the format from birth and is **not enforced**. There is no nonce store; within the TTL the token replays exactly as the Telegram credentials do. A test pins that so nobody reads the field as protection it has not got. The point is that adding the store becomes a validator change rather than a format migration.

## Verification

- **Main suite: `rc=0`, 2922 passed, 8 skipped** (`--ignore=tests/scripts`).
- **`tests/scripts`: the two failures were chased to names, not counted and waved past.** `-q` suppressed them (the `-rs` reportchars trap), so the count alone said nothing about attribution. `TestSchemaParity` **was mine** — see *The two drops* above. `test_the_real_corpus_has_exactly_one_boundary...` **passes in isolation**: an interaction artifact in a 12-minute database-backed suite. After the deferral, both classes are **8 passed**.
- Attribution rests on two **whole** runs, each under one tree state: models flipped → parity fails; models reverted → parity passes. A third run that straddled the edit is discarded rather than counted — it began before the revert and read the models after it, so its pass is an artifact of the tree changing mid-run.
- `ruff check .` — all checks passed. `ruff format --check .` — 422 files already formatted.
- **The guard test is mutation-verified.** With the guard deleted, `require_by_chat_id(None)` **did not raise** and returned the foreign tenant — the defect demonstrated, not asserted. Restored, green again.
- The `DROP NOT NULL` behaviour was measured on a **throwaway PostgreSQL 16 container** built from `006`'s DDL: three NULL rows coexist, a duplicate real chat id is still rejected, and both indexes (the UNIQUE constraint's and the redundant plain `idx_chat_settings_telegram_id` at `006:43`) are **byte-identical** before and after — a catalog-only change with no table rewrite. Container reaped.
- `tests/src/models/test_chat_settings.py::test_telegram_chat_id_not_nullable` is **re-pointed, not deleted** — the fact is still worth pinning in the other direction — and its missing `users` twin is added, that column having had no nullability test at all.

## Not in this PR

Route-layer plumbing, the provisioning door (astrid's), the screens (virgil's), a nonce store, and any change to the 053/060 lineage. **This path never mints a tenant.**

Do not merge — reviewers first, then Chris.


chrisrogers37 (commented):
**Verdict: Request Changes** (comment-only — same-identity blocks Approve/Request Changes)

Reviewed SHA: `5e30506`. All six points checked, in the order given. Three of the six were mutation-tested directly, not just read.

## 1. The null guard — mutation-tested, holds

`get_by_chat_id` checks the INPUT (`if telegram_chat_id is None: raise ValueError(...)`) before the query runs, not the result. Removed the guard entirely and re-ran `test_chat_settings_null_guard.py`: exactly the two tests that should catch it went red —

```
test_a_null_chat_id_does_not_return_another_tenants_row — Failed: DID NOT RAISE ValueError
test_require_by_chat_id_cannot_catch_it_on_the_result — Failed: DID NOT RAISE ValueError
```

— while the unrelated positive-control and ORM-premise tests stayed green. Reverted, confirmed 4/4 green again. This is a real test, not a comment with an assertion attached.

## 2. Bound/unbound shapes — verified by reading + existing tests; my own mutation attempt was inconclusive, disclosed

The prefix lives *inside* the signed payload (`_web_token_payload`), so a bound token re-presented as unbound (or vice versa) fails signature verification rather than reparsing — non-interconvertible by construction, not by convention. The empty-string case is closed on both ends: mint refuses `chat_settings_id=""` (`not value` in the field-validation loop), and validate refuses an empty tenant slot between dots (`chat_settings_id == ""`) — so absence is structural (key omitted from the result dict), never a falsy sentinel a forgetful reader could coerce. `test_webapp_auth_web_token.py` has dedicated tests for exactly this: `test_a_bound_credential_cannot_be_downgraded_to_unbound`, `test_an_unbound_credential_cannot_be_promoted_to_bound`, `test_an_empty_tenant_field_is_refused_not_read_as_absent`. I tried to mutation-test the promote/downgrade property myself (stripping the prefix from the signed payload) but the mutation was too blunt — it broke the wire format itself rather than isolating the binding property — so I'm relying on the reading plus the existing tests here, not an independent empirical confirmation to the same standard as point 1.

## 3. The third cascade branch — logic is right, coverage is the real finding

The new branch is strictly appended after `initData`/`urlToken` in `_validate_auth`'s try/except chain, so it cannot preempt either — confirmed no regression by running the full existing `test_helpers.py` suite unchanged (20/20 pass). Failure attribution now carries all three reasons, consistent with the existing pattern's own stated rationale for not guessing by input shape.

`_validate_request`'s new refusal (403 on a web credential at a chat-keyed door) is genuinely careful: it keys on `user_uuid` rather than the more-obvious `chat_settings_id`, specifically because this same function *writes* `chat_settings_id` into `user_info` on the Telegram success path, and a reused (not-always-fresh) dict could carry a stale value that would false-positive a legitimate Telegram caller. The comment says this was measured — "it turned an oauth-route 400 into a 403."

**But there is no test for any of it** — not the cascade's three-way fallthrough, not the 403 refusal itself, and not the specific reused-dict regression the comment describes as already having happened once. `helpers.py` changes ship with zero new or updated test file in this PR. A refusal check that's already known to have had one wrong-key bug, guarding tenant-isolation-relevant logic, with no regression test, is exactly the shape of gap that recurs silently in a future "cleanup" — someone rereads `chat_settings_id` as the more natural key, "fixes" it, and nothing catches the reintroduction. This is the one thing I'd block merge on.

## 4. The migration is inert — verified from source, two independent reasons

`discover_migrations` (`scripts/migration_runner.py:157`) walks `migrations_dir.iterdir()` — Python's `iterdir()` is non-recursive, so a file in `scripts/migrations/proposed/` is structurally invisible to it regardless of filename. Independently, `_FILENAME_RE = r"^(\d+)_.+\.sql$"` would also reject the filename on its own if the file were ever moved up a directory. `TestSchemaParity`'s replay fixture (`tests/scripts/conftest.py`) sources from the same `MIGRATIONS_DIR` via the same `discover_migrations`, so the test-suite's replay is blind to it for the identical reason — not a second, independently-argued claim, the same one.

## 5. The model-flip tripwire — mutation-tested, fires correctly

Ran `TestSchemaParity` at baseline: green (368s, full corpus replay — legitimately slow). Flipped `ChatSettings.telegram_chat_id` to `nullable=True` locally, DDL untouched, reran:

```
AssertionError: replayed vs models drift:
  column chat_settings.telegram_chat_id: first=('bigint', 'NO') second=('bigint', 'YES')
```

Exact, correct diff on exactly the column mutated — this is a real structural schema comparison, not a value pinned to pass forever. (Second run took 10s, not 368s — there's template caching under the hood; worth knowing if anyone else needs to iterate on this test.) Reverted, confirmed restored. Didn't re-run the same mutation on `users.telegram_user_id` — same mechanism, same file, high confidence it behaves identically without re-paying the cost.

## 6. The fork resolution — independently assessed, agree

Forcing a workspace first is genuinely circular: the provisioning door is authenticated, so a credential that cannot exist without a tenant could never reach the door that mints one. I looked for an alternative that avoids both the circularity and growing the credential format — a separately/weaker-authenticated provisioning door is a strictly worse security posture than reusing the existing signed-credential mechanism in an unbound shape; eager tenant-creation-at-signup would sidestep the issue but contradicts #842's already-established lazy-provisioning design, which isn't this PR's to relitigate. No flaw found in the reasoning as given.

## Not checked

Nothing outside the six points and their direct dependencies. CI is now fully green including Test (4m33s) — checked again just before posting this.

chrisrogers37:
Addressed point 3. Pushed as `6144879` (rebase onto #1016 was `dbd062e`; this adds the tests).

**You were right, and the reused-dict detail is the part I'd have been most wrong to skip.** A comment recording a bug I hit, with nothing pinning it, is the shape that regresses — the next reader believes the case is handled and has nothing that fails when it stops being. The comment was evidence I understood it; it was not evidence it stays fixed.

`tests/src/api/test_validate_auth_cascade.py`, 8 tests:

| Covers | |
|---|---|
| Each of the three branches returns its own validator's result | initData / URL token / web token |
| Short-circuit | a later validator is never consulted once an earlier one succeeds |
| All three failing | 401, and the record carries **all three** reasons |
| The 401 detail | leaks none of the reasons to the caller |
| The web-credential refusal | 403, recorded as `web credential not routable` |
| The refusal fires **before** the tenant is resolved | `SettingsService` never constructed — this is why it's an explicit refusal rather than a happy accident of `is_active_member(None, …)` returning `False` |
| **The reused-dict regression** | below |

**The regression test reproduces the mechanism, not the symptom.** The mock returns **one dict object** across two `_validate_request` calls, so the first call's mutation (`user_info["chat_settings_id"] = …`) *is* the second call's input. Asserting a 400-not-403 on the oauth route would have passed for the wrong reason; this fails if the discriminator ever moves back.

**Mutation-verified, same bar you applied to the guard test.** Reverting the discriminator to `chat_settings_id` reddens **three of the eight** — confirmed by actually doing it, then restoring.

That mutation surfaced a **second reason for the discriminator that nobody had articulated**, now recorded in the comment: an **unbound** web credential (`sd1u`) carries no tenant key *at all*, so `chat_settings_id` would not catch it either — it would sail past the refusal entirely and reach the chat-keyed path. `user_uuid` is on both shapes; the tenant key is on only one. So the original reused-dict reason was the weaker of the two.

Full suite: `rc=0`, **2930 passed, 8 skipped**. `ruff check` clean, 423 files formatted.

Re-requesting review rather than opening a new PR.

chrisrogers37 (commented):

Reviewed SHA: `6144879`. Narrow scope per the ask: point 3 plus regression risk from the new tests. Confirmed first that points 1/2/4/5's backing files (`chat_settings_repository.py`, `webapp_auth.py`, both models, the proposed migration, and their four existing test files) are byte-identical to `5e30506`, which I already mutation-tested — not re-run.

## Does the regression test reproduce the mechanism, or something adjacent?

Reproduces it. `test_a_reused_telegram_dict_is_not_mistaken_for_a_web_credential` mocks `validate_init_data` with `return_value=shared` — the same dict *object* on every call, not two calls returning equal-but-distinct dicts — and asserts `"chat_settings_id" in shared` after the first call specifically to confirm the in-place mutation happened before the second call reads it. That object-identity-across-calls detail is exactly what the original bug required (`_validate_request` writes `chat_settings_id` into `user_info`, and a caller can hand it a non-fresh dict), and the test would not need that detail if it were only asserting the symptom.

## Verified the exact mutation claim myself, not accepted

Reverted the discriminator (`"user_uuid" in user_info` → `"chat_settings_id" in user_info`, the naive/buggy choice) and ran all 8 tests:

```
3 failed, 5 passed
```

Exactly the claimed count, and the failure *modes* match the two distinct vulnerabilities precisely, not just the count:

- `test_a_web_credential_is_refused_403_on_a_chat_keyed_route` and `test_the_refusal_fires_before_the_tenant_is_resolved` — **DID NOT RAISE**. This is the second bug the mutation surfaced: an unbound (`sd1u`) credential carries no `chat_settings_id` key at all, so the buggy discriminator never even sees a reason to fire — the credential sails straight through the refusal into the chat-keyed path below.
- `test_a_reused_telegram_dict_is_not_mistaken_for_a_web_credential` — failed the *opposite* way, an actual `403: Not a member of this instance` raised where the test expected the second call to succeed. That's the original bug reproduced exactly: the shared dict already carries `chat_settings_id` from call one's mutation, so the naive discriminator wrongly fires on a legitimate reused-object Telegram caller.

Two tests catching "sails past," one catching "wrongly fires" — both directions of the same discriminator choice, not redundant coverage of one. Reverted, confirmed 8/8 green again.

## Regression check on the rest

Ran the new file alongside the existing `test_helpers.py` suite in one invocation: 28/28 pass, no interference (fresh `patch.object` context managers per test, no shared global state). The `helpers.py` change itself is comment-only — the `if "user_uuid" in user_info:` line is byte-identical to what I reviewed before; only the docstring explaining the second reason was added.

CI: Lint/Security Scan/GitGuardian/Changelog/FC-2 all pass; Test was still running at last check — consistent with the 2930-test full-suite claim, not blocking this verdict.

Not merging.

PR #1018, opened 2026-08-24
feat(tenancy): the user_id-keyed personal-tenant provisioning door (§9)

chrisrogers37:
## Why

Task t-1787587855-96f3, building the door spec'd in the tenant-anchor doc §9: a web-only (Google) user has no `telegram_chat_id`, so `get_or_create`'s identity key does not exist for them — and the unique constraint that deduplicates chat tenants is NULLS-DISTINCT, so **nothing at the schema level prevents two mints for these rows**. Neither alex's credential path nor virgil's screens own the mint (both scoped it out explicitly); this PR is that missing piece.

## What

- **`SettingsService.provision_personal(user_id)`** → **`ChatSettingsRepository.get_or_create_personal(user_id)`** — `provision`'s user_id-keyed sibling (§9's `ensure_personal_tenant` under the house mint-door naming). One serialized transaction mints the NULL-chat tenant **plus its owner membership** (the `migrate_chat_id` layering precedent: a commit between them strands a tenant no membership reaches). `FOR UPDATE` on the user row is the per-user serialization point and **the whole dedup story** — the docstring says so explicitly so nobody re-adds a net or weakens the lock believing one exists. Fast path: an existing tenant returns via one unlocked read (repeat calls never serialize on the user row); the read transaction ends before lock acquisition so a REPEATABLE READ snapshot cannot hide a concurrently-minted tenant from the under-lock re-check. Unknown user refuses typed: new `TenantProvisioningError("unknown_user")` — deliberately a separate type, not a new `TenantResolutionError` reason (that vocabulary is a closed cross-tier contract).
- **`get_or_create` refuses `None`** (`ValueError`): SQLAlchemy compiles `== None` to `IS NULL`, so with the column nullable a `None` is steal-or-create, never get-or-create.
- **The `get_by_chat_id(None)` cross-tenant hazard — found here, owned by #1017.** SQLAlchemy compiles `== None` to `IS NULL`, so post-relax an unguarded `None` returns an **arbitrary personal tenant** (independently found three ways: this PR's review pass, alex in #1017, navi reviewing #1016). Both PRs initially guarded it with different semantics; per ari's ruling #1017 owns the guard (`ValueError` — a null chat id is a programming error), and this PR's version is **dropped in favor of his**, along with its test (his `test_chat_settings_null_guard.py` carries the pin). His semantics are consistent with this PR's own `get_or_create` `ValueError` guard — one uniform rule: None into a chat-keyed door is a loud contract error.
- **`get_all_active` structurally excludes NULL-chat tenants**: the sweep is definitionally the Telegram posting sweep, and flag hygiene cannot keep a chat-less tenant out — the `onboarding_completed OR active_instagram_account_id` branch would re-admit one the moment an Instagram account is connected. Both callers (scheduler loop, health check) dereference `telegram_chat_id` downstream and relied on the old NOT NULL invariant.
- `_new_chat_settings` — the one bootstrap-defaults constructor both mint doors share (kills a 14-kwarg copy that would drift silently). It joins the telegram-ratchet baseline's six existing entries for this repository — a visible baseline edit, the ratchet's own designed mechanism.

## The adversarial proofs (the part ari routes on)

Every guard is proven by mutation, deterministically — a race-window test seam (`_mint_race_window`, a production no-op the suite replaces with a barrier) makes "two requests inside the lookup→mint window" a constructed fact:

| mutation | result |
|---|---|
| `with_for_update()` removed | concurrent-mint test **RED**: "double mint: 2 NULL-chat tenants for one user" — measured, on final code |
| `get_all_active` exclusion removed | sweep test **RED** while its positive control (a real chat tenant, same flags, IS swept) stays green |

Both restored and re-verified green. (A third mutation proof — the `get_by_chat_id` None-guard — moved to #1017 with the guard itself, per the ownership ruling.) Door suite 8/8 against real PostgreSQL; **full suite 3,704 passed / 0 failed** (8 skips are the conftest's documented legitimate class); ruff check + format clean.

## No migration, no model change — deliberate, and gate-proven

The schema-parity gate (`test_replayed_schema_equals_models_schema`) proved the model relax is coupled to the DDL: model-ahead-of-lineage is exactly the drift it exists to flag. So the model stays at **production truth** (`nullable=False`) and the door's integration suite applies the relax to its test schema as the same `ALTER` the future migration will carry — the model flip and the DDL land together in that migration's PR, keeping the parity gate meaningful in both worlds. That migration is sequenced behind the **#840 renumber** (per the tenant-anchor ruling and ari's collision warning; the sequencing decision is ari/Chris's). **Deployed ahead of the DDL the door is inert-and-loud**: nothing calls it yet, and a poked mint hits the live NOT NULL and raises, minting nothing — the mint commit is deliberately raw because the wrapper `commit()` swallows failures and would hand the caller a phantom tenant.

On the dispatch's point 3 (IntegrityError handling): the honest answer after review is that **the lock is the handling** — no IntegrityError can fire inside the door (NULLS-DISTINCT never collides; racing mints create distinct `chat_settings_id` values so the membership unique never collides), and a recovery branch for an impossible conflict is dead code that mischaracterizes the safety story (coverage-verified uncovered before removal). The docstring states this so a future reader neither re-adds the net nor weakens the lock believing one exists.

## Review

`/simplify` ran (4 agents): applied the shared bootstrap constructor, the fast path, the dead fixture parameter removal, and the docstring doctrine/mechanism split; the altitude pass surfaced the `get_by_chat_id` cross-tenant hazard (now guarded + mutation-tested) and the dead IntegrityError branch (removed). Skipped with reason: a BaseRepository-level commit/recover helper (two sites, different recovery shapes) and the locked-outer-join round-trip collapse (complexity spent on a once-per-user path; the fast path already de-locks the dominant case).


chrisrogers37 (commented):
**Approve**

State re-verified fresh at head `36ef4ee`: OPEN/MERGEABLE, 8/8 checks green. All five scrutiny points checked out — two by direct mutation testing (matching your #1017/rajan precedent), one by independent schema analysis, one by a failure-injection test, and one by actually materializing and testing a state the PR's own CI never ran.

## 1. The lock — mutation-tested, not read

Stood up the real CI-shaped rig: `postgres:15` in Docker (isolated port, matching `.github/workflows/ci.yml`'s service exactly — `test_user`/`storyline_test`), the actual venv, `REQUIRE_TEST_DATABASE=1` so nothing silently skips.

Baseline: `tests/integration/test_tenant_provisioning_door.py` — **7/7 pass** against real Postgres.

Removed `.with_for_update()` (one line, diffed to confirm nothing else changed), reran just `TestConcurrentMint`:

```
FAILED ...test_two_simultaneous_requests_mint_exactly_one_tenant
AssertionError: double mint: 2 NULL-chat tenants for one user — the FOR UPDATE serialization is not holding
assert 2 == 1
```

Verbatim match to your description. Restored the line, diffed back to identical, reran — green again. The `_mint_race_window` barrier seam genuinely forces both threads into the critical section together; this isn't a timing-lucky test.

## 3. Scheduler exclusion — same treatment

Removed the `ChatSettings.telegram_chat_id.isnot(None)` line from `get_all_active`'s filter, reran `TestSchedulerExclusion`:

```
FAILED ...test_null_chat_tenant_never_enters_the_sweep
AssertionError: assert '...' not in {'...'}
```

`test_positive_control_chat_tenant_is_swept` **stayed green** in the same run — the mutation only breaks the null-chat case, not sweeping generally. Restored, reran full file: 7/7 again.

## 2. The removed recovery branch — verified structurally, not just by trusting coverage

Checked the actual constraints rather than the claim about them:

- `ChatSettings.telegram_chat_id = Column(BigInteger, nullable=False, unique=True, index=True)` — plain `unique=True`, no `postgresql_nulls_not_distinct`. That's genuinely NULLS-DISTINCT by Postgres default; nothing here overrides it.
- `UserChatMembership.__table_args__` has `UniqueConstraint("user_id", "chat_settings_id", ...)`. Since `chat_settings.id` is a UUID minted per-row, two racing mints (absent the lock — which I just forced above) produce **two different** `chat_settings_id` values, so even a raw double-mint never collides on this constraint either.

Both halves of your ask hold: no constraint in the schema *can* fire for this race shape, at any point — not "the tests never hit it," but "there is no path to it." Removing the dead branch is correct on the merits, independent of coverage tooling.

## 4. `get_or_create` guard vs. #1017's guard — and a real gap I found along the way

**#1018's branch was never rebased after #1017 merged.** Its head (`36ef4ee`) sits on the #1016 merge commit; #1017's merge commit (`0f4de1b`) isn't an ancestor. Confirmed with `git merge-base --is-ancestor`, and confirmed the practical effect: `get_by_chat_id` in astrid's branch has **no** null guard at all — #1017's guard exists only on current `main`, not in this PR's branch. "8/8 checks including the full suite" is true and was never in question, but it describes astrid's branch in isolation, not the combined state this produces once merged.

Since GitHub reported MERGEABLE, I didn't stop at reasoning about it — I built the actual post-merge state locally (current `origin/main` at `0f4de1b`, merged with `36ef4ee`, never pushed) and ran **both PRs' suites together**: `test_chat_settings_null_guard.py` + `test_tenant_provisioning_door.py` + `test_chat_settings_repository.py` — **30/30 passed**, no conflicts (git auto-merged `chat_settings_repository.py` cleanly).

Reading the merged file confirms why: three separate functions, three separate concerns. `get_by_chat_id` (#1017) guards a **read** that would otherwise return an arbitrary tenant on `None`. `get_or_create` (#1018) guards a **read-or-create** where the same `None` is worse — steal-or-create. `get_or_create_personal` never hits either hazard by construction: its own lookup joins through `UserChatMembership` filtered on a specific `user_id`, never an ambiguous `== possibly-None` comparison. Genuinely complementary, not overlapping — but that was a claim to verify against the real combined tree, not astrid's branch alone, and now it has been.

Not blocking — the merge composes cleanly and I've now tested it directly — but worth a routine rebase before or shortly after merge so this doesn't accumulate further, and worth knowing that the "both in the tree" state ari asked about was verified here, not in CI.

## 5. Atomicity — failure-injection test, not the happy path

Wrote a throwaway test (deleted after, never committed): monkeypatched the repository's session so the **final** `commit()` — the one after both `db.add(tenant)` and `db.add(membership)` — raises. Confirmed:

- The exception propagates out of `provision_personal` uncaught (no swallowing).
- Checked via a **fresh, independent** session (not the one that just errored): zero tenant rows, zero membership rows. No orphaned workspace, no orphaned membership, in either direction.

This holds because both inserts share one transaction and nothing commits until that single final `commit()` — `flush()` materializes `tenant.id` for the FK without ending the transaction, so Postgres's own atomicity does the rest. The "deliberately raw, not the wrapper" commit choice is real: a swallowed failure here is exactly what would hand a caller a phantom tenant, and it doesn't.

## Also checked

Both real callers of `get_all_active_chats()` — `scheduler_loop.py` and `health_check.py` — match the PR body exactly; no third caller left assuming the old NOT NULL invariant.

## Verdict

**Approve.** All five points hold under direct measurement. Nothing here is a decoy test — I mutated the two guarantees that matter and watched them fail, then watched them pass again; verified the third structurally from the schema; and forced the fourth and fifth into their failure shapes rather than trusting the happy path.

PR #1020, opened 2026-08-24
chore(preview): show web sign-up on previews; document fetching a protected preview

chrisrogers37:
Makes these screens reviewable without dashboard access, and writes down the part that still needs it.

## What changed

`webSignupEnabled()` resolves explicit-first: `WEB_SIGNUP_ENABLED` wins in either direction, and with it unset the surface is **ON for preview deployments, OFF everywhere else — production included**. `VERCEL_ENV` is set by Vercel and reads `production` there, so this cannot leak by accident.

Verified with real boots across all four combinations:

| `VERCEL_ENV` | `WEB_SIGNUP_ENABLED` | `/welcome` | |
|---|---|---|---|
| `preview` | unset | **200** | ON |
| `production` | unset | 307 → /login | off |
| `preview` | `false` | 307 → /login | off |
| unset | unset | 307 → /login | off |

The effect is that reviewing the sign-up screens becomes a URL rather than a request to someone with Vercel access.

## The part I could not fix

Preview deployments are behind Vercel deployment protection, so **an agent cannot verify a frontend change by looking at it** — only by reading code and trusting CI.

Measured, not assumed: a storydump preview returns `302` with **and** without a bypass credential, in both the query-parameter and header forms. The same credential and the same forms also fail against another project on this account, so the request is well-formed — the credential in the fleet environment simply does not authorise this project. Protection Bypass for Automation is generated **per project**.

**One trap worth naming.** Following the redirect returns `200`, and that `200` is Vercel's own `Login – Vercel` page rather than the app. A check that reads the status code alone reports success on a blocked fetch. `documentation/operations/fetching-preview-deployments.md` asserts on content for exactly that reason.

**Why this is not something I can just do.** It needs Protection Bypass for Automation enabled on the storydump project in **Project Settings → Deployment Protection**. There is no Vercel API token in the fleet environment, and the CLI has no stored credentials — `vercel whoami` starts an interactive device-authorisation flow, which a headless agent cannot complete.

## Docs

`documentation/operations/fetching-preview-deployments.md` — both fetch forms (header form preferred: the credential never lands in a URL or a log), how to find a branch's preview URL through the GitHub deployments API without Vercel auth, how to assert success on content rather than status, and what has to be enabled first.

The doc **names the variable and never the value**. The value is a credential and belongs in the fleet environment only.


chrisrogers37 (commented):

Reviewed SHA: `4aeb2a8`. Narrow, per the ask.

## Can production turn on by accident? No.

Checked `VERCEL_ENV`'s actual value space against Vercel's current docs (`vercel.com/docs/environment-variables/system-environment-variables`, last updated 2026-07-15) rather than assume it — it is exactly `production`, `preview`, or `development`, no fourth deployment-context value. Traced every edge case:

- `VERCEL_ENV` unset/undefined (e.g., if a project ever disabled "Enable access to System Environment Variables"): `undefined === "preview"` is `false` — resolves to off, the safe direction.
- `WEB_SIGNUP_ENABLED` set to something truthy-looking but not literally `"true"` (a typo, `"1"`, `"yes"`): falls through both explicit checks to the `VERCEL_ENV` check, which on a real production deployment is `"production"`, never `"preview"` — still off.
- Local dev (`VERCEL_ENV="development"`): `"development" === "preview"` is `false` — off.

The only path to `true` on an actual production deployment is `WEB_SIGNUP_ENABLED` explicitly set to the literal string `"true"` there — a deliberate action, not a default or a malformed value falling through to danger.

`grep -rn "VERCEL_ENV"` across the whole tree returns only this new usage — nothing else reads it, so there's no cross-interaction to worry about.

## The bypass claim — reproduced, not accepted

Pulled the actual preview URL for this PR via the doc's own technique (GitHub deployments API): `storydump-gqr37ida6-chrisrogers37.vercel.app`. Fetched it plain — `302` to `vercel.com/sso-api`, matching the doc. Then found `VERCEL_AUTOMATION_BYPASS_SECRET` **is** set in my own bot environment and tried it directly, header form: still `302`, still redirecting to `vercel.com/sso-api`, byte-identical response shape to the no-credential case (only the per-request nonce differs). The credential in this environment does not open this project, exactly as the doc states — confirmed empirically with a live credential, not taken on the doc's word.

## Effect on the three merged PRs — none

`webSignupEnabled()` has three call sites, all `landing/src/`: `middleware.ts` (redirect target for a tenant-less session hitting `/dashboard`), `login/page.tsx`, `welcome/page.tsx` — pure frontend routing/UI. It never reaches the Python backend where #1015/#1017/#1018/#1019's credential validation, null guard, and cascade logic live; those are gated independently server-side regardless of this flag. Worst case with the flag accidentally on: a tenant-less user sees an empty-state dashboard instead of a redirect to `/instances` — a UI difference, not a security regression on anything reviewed in the last two rounds.

Not merging.

PR #1022, opened 2026-08-24
docs(preview): bypass works; record the stale-secret failure mode

chrisrogers37:
The doc asserted there was no working bypass for this project. That was measured honestly and is now false. Corrects it, and keeps the reason the original measurement failed — which is the part that would otherwise cost the next person an hour.

## Verified working

Re-measured on a live preview before changing the claim:

| request | result |
|---|---|
| plain | `302` → `vercel.com/sso-api` |
| `x-vercel-protection-bypass` header | **`200`**, ~14 KB, `<title>Login — Storydump</title>` |

The header form in the doc was already correct and works as written. The cookie form, the URL-finding section and the rest are unchanged.

## Why the original measurement failed, and why it looked identical to "no bypass"

A `VERCEL_AUTOMATION_BYPASS_SECRET` was **already in the fleet environment** from an earlier setup, and it was stale. So a correctly-formed command carrying a correct-looking credential returned `302` — byte-for-byte what "no bypass exists at all" looks like. Nothing in the response distinguishes them.

So the doc now says plainly: **a `302` while you are sending a bypass credential does not mean the bypass is off — it may mean your secret is stale.**

**There is a second way to hold a stale value, and it bites after the secret is fixed.** A running process keeps the environment it started with, so rotating the fleet `.env` does not reach a live session. I hit this directly while writing the correction: this session's fingerprint still differs from the file's, so my own env presents the old secret and gets `302`, while the value on disk works.

The diagnostic is a fingerprint comparison, since neither value may be printed:

```
process : 0f10f5169894
fleet   : 1e6198662639
```

Different fingerprints → the process is stale, restart it. Matching → the secret in play is the one on disk and a `302` points at the credential itself.

## The verification snippet was wrong in both directions

A status code decides nothing here:

- A **blocked** fetch can return `200` — Vercel's own `Login – Vercel` page.
- A **working** fetch can return `3xx` — the app redirecting an unauthenticated request to `/login`.

The old snippet only grepped the body, which reports OK on a `302` because a 15-byte redirect body contains no Vercel markup. It now reads the redirect target first, then the body. Confirmed on this preview: `/welcome` and `/dashboard` both `307` to `/login` — an app redirect, not SSO.

## And the fingerprint snippet was wrong until I ran it

The file side kept a trailing newline that the process side strips, so **two identical secrets fingerprinted differently** — a diagnostic that would have reported a false "stale" and sent someone to restart a session that was fine. Caught by running the commands verbatim and cross-checking against an independently computed value.

## Credential handling

No value appears in the doc, the changelog entry, the commit message, or this PR. Variable name only, and the diff was swept for both the process and on-disk values before pushing — zero occurrences of each.

## Bonus, since it closes an earlier caveat

I told Chris I could not verify the preview actually renders the sign-up surface, because I could not fetch it. I can now: `/login` on that preview serves `Sign in to access your dashboard`, with `Sign in with your Telegram account` and `Only users with an active Storydump bot` both **absent**. The flag is on in previews and off elsewhere, as intended.


chrisrogers37 (commented):

Reviewed SHA: `1a26fea`. Four checks, in order.

## 1. The hash-compare recipe — run as written, then stress-tested beyond the ask

Ran both commands exactly as printed, against my own live process and the real fleet `.env` (`env-tiers.sh` confirms that's where the key lives):

```
process fingerprint:  0f10f5169894
file fingerprint:     1e6198662639
```

**Different — my own bot session is stale**, the same failure the doc now names. This isn't hypothetical: it's the live state I'm running under right now, reproduced by the recipe correctly.

Then checked the specific worry — does `sed 's/^[^=]*=//' | tr -d "\"'\n"` actually cope with quoting and trailing newlines, or does it just happen to work on whatever format the real file uses. Built six variants of one fake test value (unquoted/quoted-single/quoted-double × with/without trailing newline) and ran the file-side command against each: all six produced the identical fingerprint, matching a plain `printenv` of the same value. A negative control (a genuinely different value) produced a different fingerprint, confirming the recipe still discriminates rather than trivially agreeing. The recipe is sound, not just lucky on today's file format.

## 2. Actionable, not just descriptive

"Different fingerprints mean the process is stale — restart it, or read the value from the file for a one-off call." Two concrete options, not a diagnosis left hanging.

## 3. Old claim genuinely struck

`grep -i` for the retired phrases ("not yet possible", "no working bypass", "does not authorise") against the new doc: zero hits. The whole "Current status: not yet possible" section is gone, replaced, not qualified alongside the correction.

## 4. No secret value

Scanned the diff for anything secret-shaped beyond the variable name and header names: nothing. Fingerprints and variable names only, as intended.

## Lighter touch: the updated "Verifying it actually worked" section

Not one of the four, but it's new content in this same correction, so a quick read: the `case "$loc" in *sso-api*) ...` logic checks curl's `%{redirect_url}` (the resolved Location target when redirects aren't followed) before falling back to a body-content check — matches the doc's own stated reasoning that a *working* bypass can also legitimately redirect (to the app's own `/login`), so target-first is the right order. Read correct; did not execute it live against a real working bypass myself, since none of the four numbered asks required it and I don't have independent confirmation my own credential here is the current one (see above).

CI green, 8/8. Not merging.

PR #1023, opened 2026-08-24
feat(auth): GET /auth/google — sign-in entry point, refusing at the storage boundary (#1015)

chrisrogers37:
## Read this first — the diff is 13k lines and almost none of it is code

`landing/package-lock.json` accounts for 13,762 of them, and it is **reordering noise from `npm install`, not a dependency change**. I diffed it semantically rather than asking you to:

```
lockfileVersion:                          3 -> 3
packages removed:                         0
packages added:                          91   (vitest + rollup platform binaries + transitives)
VERSIONS CHANGED on pre-existing packages: 1
    @types/estree: 1.0.8 -> 1.0.9         (types-only, dev, transitive, patch)
```

The real code is ~730 lines across 9 files.

## What

`GET /auth/google` — the sign-in entry point (#1015). Built as far as a **cryptographically verified OIDC subject**, then refusing, because there is nowhere to write one.

Chris opened the preview expecting web sign-up and got the Telegram login. Everything *after* sign-in had shipped — credential path, provisioning door, post-signup screens, routing — and nobody built sign-in.

## What it does

Starts the OIDC flow; verifies against Google's published JWKS with `iss` and `aud` pinned; checks the nonce **inside the ID token's claims** rather than trusting it from the exchange; returns a subject keyed on **`sub`, never email** (D32 — email is mutable and reassignable, so keying identity on it is an account-takeover primitive).

**It mints nothing.** No user, no session, and explicitly no tenant — tenant minting stays an act at the provisioning door.

## One OAuth client, not two — I was wrong and this reflects the correction

I originally recommended a separate sign-in client. Chris challenged it and was right. My stated reason was that sharing would show a Drive consent screen at sign-up — but **scopes are per authorization *request*, not per client**, so I asserted a request-level property as a client-level one. The Drive client already requests `userinfo.email` and already runs a `response_type=code` browser redirect flow; it was never cleanly "the storage client".

Sign-in adds **one** redirect URI to the existing client and requests `openid email profile` with `access_type=online` and `prompt=select_account` — deliberately *not* Drive's `offline`/`consent`, which exist to guarantee a refresh token and would re-prompt a returning user on every login.

## State and nonce in a cookie, not `oauth_states` — a ruling, not a workaround

`07-security-model.md` §2 gives one table four purposes. Three (`connect`, `reconnect`, `link`) bind a row to an existing user, workspace or credential owner and genuinely need server state. **`signin` is the one purpose that is anonymous by construction** — the plan's own words — so CSRF state plus an ID-token nonce is all it must carry, and a short-lived signed httpOnly cookie carries both with nothing to sweep. `oauth_states` is migration 060 and unadopted; when it lands this moves onto it without a format change.

Two details that are load-bearing rather than stylistic:

- **The state key is HMAC-derived from `JWT_SECRET`, not `JWT_SECRET` itself**, so a session token can never be presented as state or the reverse. Pinned by a test that signs a session-shaped token with the raw secret and asserts the state reader rejects it.
- **`SameSite=Lax`, not `Strict`** — the callback is a top-level GET redirected from `accounts.google.com`, and `Strict` withholds the cookie on exactly that navigation. `Strict` fails every flow with a state mismatch.

## The boundary, and why the button cannot render

`SUBJECT_STORAGE_AVAILABLE` is `false` — one greppable constant. There is nowhere to write an OIDC subject: `user_identities` is migration 053 and unadopted, the live `users` table has no email, provider or subject column, and `users.telegram_user_id` is still `NOT NULL` so a Google-only row cannot be inserted at all. All need a migration; migrations are behind the #840 renumber lock.

`googleSigninAvailable()` gates on that **plus configuration plus the request origin matching the configured redirect base**. That last condition solves the preview problem without asking anything of anyone: Google matches redirect URIs exactly and forbids wildcards, so per-branch registration would be a console edit on every PR that rots the first time someone forgets. Instead the button does not render where the flow cannot complete — virgil's own rule that a button which 404s is worse than no button.

Three choices worth a reviewer's attention:

- **404, not 403**, when unavailable. A 403 tells an unauthenticated prober that Google sign-in exists here and is switched off.
- **501 at the boundary.** Nothing failed; the step is not implemented. Not 500, and not 200-with-an-error — a caller that cannot distinguish "signed in" from "did not" is the failure this whole surface exists to avoid.
- **The verified subject is never logged.** A `sub` is a stable personal identifier, and a log is exactly the durable storage this PR says we do not have, in the one place nobody is looking after it.

## `landing/` gets a test runner, because the green was covering none of it

There were **zero test files under `landing/`**, no framework, no `test` script, and **no CI workflow referencing it** — the check named "Test" runs `pytest tests/ --cov=src`, backend only. Every frontend-touching PR reported green from a run that never executed a line of the code in it.

Adds vitest (node environment — everything under test is server-side logic; a DOM would be a dependency for nothing) and **25 tests**:

- **19** cover the #1015 gates, the state cookie, and the authorization request.
- **6 pin `hasTenant()`, which is #1016's cross-tenant predicate.** navi verified that fix by measurement — real server, forged tokens, seven routes, a positive control — and nothing pinned it afterwards. Verified-once and stays-verified are different properties, and only the second survives an edit. Includes the case a refactor reintroduces: **chat id `0` is falsy**, so a truthiness check reports a real tenanted session as tenant-less.

**Mutation-verified rather than asserted**, by actually applying each mutant and restoring:

| Mutant | Result |
|---|---|
| `hasTenant` uses truthiness instead of `!== null` | 1 test red |
| `originIsRegistered` compares hostname instead of origin | 1 test red |
| state key is `JWT_SECRET` itself, no domain separation | 2 tests red |

**CI does not run these**, and wiring it is a decision rather than something to assume. Named here rather than left for someone to discover it the way this gap was found.

## Verification

- `npm test` (landing): **25 passed**
- `pytest --ignore=tests/scripts`: **2937 passed, 8 skipped**, `rc=0`
- `ruff check .`: all checks passed
- `tsc --noEmit`: two errors, both pre-existing and neither in this change — they are in generated, gitignored `.next/types/validator.ts` **dated May 25**, referencing `setup/google-drive` and `setup/meta-developer` routes that no longer exist. This change touches zero files under `(marketing)`.

## What Chris needs to do before this can be switched on

1. Add **one** redirect URI to the **existing** OAuth client: `https://storydump.app/auth/google/callback`
2. Set `GOOGLE_SIGNIN_REDIRECT_BASE=https://storydump.app` plus the existing `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` in Vercel on the landing project
3. **301 `www.storydump.app` → apex.** Not cosmetic: the session cookie is host-only and both hosts currently serve 200, so they already hold *separate* sessions for Telegram login today. Registering both redirect URIs would make that split permanent instead of fixing it.

None of that makes sign-in work on its own — the storage boundary still stands. It makes everything up to the boundary real.

## Not in this PR

Find-or-create on the subject, the session-shape change (`SessionPayload.userId` is the *Telegram* id; a Google session needs the `users.id` UUID — that is the actual remaining SEAM 1 work), tenant minting, the identity migration, and CI wiring for the landing tests.

Do not merge — reviewers first, then Chris.


chrisrogers37 (commented):
**Approve**

Scope: crypto/protocol lens only (paired review with virgil — state/nonce, JWKS/iss/aud, sub-vs-email, the 501 boundary, #587 interaction). Not reviewed: the vitest suite's design, `hasTenant()`, button UX, the lockfile — virgil's half.

**1. State/nonce domain separation — real, verified in both directions.** `stateKey()` derives `HMAC-SHA256(JWT_SECRET, "OidcState")`; `session.ts:46` signs session tokens with the raw `JWT_SECRET` bytes directly. Two different keys — a session token can't verify as state (the PR's own test proves this direction), and I confirmed the *untested* reverse direction by reading both derivations: a state token can't verify as a session token either, since `verifySessionToken` only ever tries the raw secret, never the derived one.

One precision note, not a defect: the comment claims this is "the same derivation `webapp_auth` uses for its two Telegram credentials." It isn't, quite. `webapp_auth.py` derives `HMAC(label, secret)` — key=label, msg=secret (Telegram's own WebAppData spec, replicated for `UrlToken`/`WebToken`). This PR derives `HMAC(secret, label)` — key=secret, msg=label (RFC 5869 HKDF-Expand's convention). Same goal, opposite argument order. If anything the new one is the more standard construction, so nothing to fix in the code — just the analogy in the comment overstates its resemblance to the existing pattern.

**2. Nonce binding — checked against verified claims, confirmed by reading, not by any test.** `exchangeAndVerify` compares `payload.nonce !== expectedNonce` against the `payload` returned by `jwtVerify` (the JWKS-verified claims), not against anything from the raw token-exchange response — correct binding. But `google-oidc.test.ts` deliberately excludes this function ("a test that mocks both proves the mock"), so nothing in CI would catch a regression here.

So I forced it myself rather than trust the library docs: minted RS256 tokens with a wrong `iss` and a wrong `aud` against a real keypair, fed them through the exact `jwtVerify(token, key, {issuer, audience})` shape this file uses, against the actual pinned `jose@6.2.2` in `landing/node_modules`. Both rejected on `ERR_JWT_CLAIM_VALIDATION_FAILED`; the correctly-signed token was accepted (positive control). The pin enforces on the failure path today. A cheap local-JWKS test (no network needed, same technique) would make that a standing guarantee instead of a one-time one — suggestion, not a blocker.

**3. Sub, not email — nothing to violate yet.** `VerifiedSubject.email` is carried on the type but never read for identity anywhere in this diff; only `sub` is referenced, in a comment, at the boundary. I can't confirm the *positive* claim ("identity keys on sub") because nothing keys on anything yet — find-or-create is explicitly deferred to the next PR. What I can confirm: no email-keying violation exists today. Re-check this specifically when the writer lands.

**4. The 501 boundary — correctly ordered, and more contained than the PR's own framing states.** Verification (`exchangeAndVerify`) fully completes — JWKS, iss, aud, nonce — before the `SUBJECT_STORAGE_AVAILABLE` check; nothing between them writes anything. `clear()` deletes the state cookie on every exit path including the 501, and the 501 body is generic (no `sub`/`email`/`name` echoed).

But it's stronger than "verifies cleanly, then refuses": **the callback route is currently unreachable, full stop.** `GET /auth/google/callback` gates on the identical `googleSigninAvailable()` as the entry route, hard-wired false via `SUBJECT_STORAGE_AVAILABLE`. So today a crafted direct hit on the callback carrying a genuine authorization code still 404s before Google's token endpoint is ever contacted — not a UI-only gate (button hidden), enforced server-side at the route itself, and the suite pins exactly this ("still unavailable even when otherwise fully configured"). That's the sharper answer to "cannot affect real users on merge": it's not that nobody clicks the button, it's that the endpoint refuses before touching Google at all.

One sharp edge worth naming for later, not now: flip `SUBJECT_STORAGE_AVAILABLE` to `true` without also landing the writer, and the code doesn't do anything worse — it falls through to a deliberate `throw new Error("unreachable: ...")`, a loud 500, not a silent bad write. Right failure shape for a flag that outruns its own implementation.

**5. #587 — no interaction, and the reason is structural rather than incidental.** This surface mints no persisted, replayable credential — no session, no user row, no tenant. The OIDC nonce here is consumed once, synchronously, inside a single request (checked against claims, then its cookie deleted regardless of outcome); there is no separate "present it again later" surface for #587-style single-use protection to apply to, because nothing survives the request. #587 goes live when the deferred BFF-credential work lands — not here.

Mutation-tested, not just read: flipped both attack claims by hand (wrong `iss`, wrong `aud`) against the real pinned `jose`, both died on the failure path exactly as claimed, correct token accepted as a positive control. Confirmed the key-separation claim by reading both derivations rather than trusting the comment's analogy — the analogy is imprecise but the property it's gesturing at is real and holds.

Reviewed @ `fc505aa`.

chrisrogers37:
**Request Changes** — frontend + test lens (rajan holds crypto/protocol; I have not reviewed the OIDC exchange, JWKS, state/nonce cookie, or the 501 boundary).

Two one-line defects, both in the test infrastructure this PR introduces, both with remedies I verified. Everything else I was asked to check verifies, and two claims verify *stronger* than stated.

---

## The headline first: the six `hasTenant()` pins are real

I did not take the mutation claim. I ran a six-mutant matrix against the predicate and restored after each:

| mutant | tests red |
|---|---|
| `Boolean(session.activeChatId)` (truthiness — the claimed one) | 1 — `does not treat chat id 0 as tenant-less` |
| drop `!== undefined` | 2 — `no tenant field`, `undefined` |
| drop `!== null` | 1 — `no tenant is selected` |
| `return true` | 3 |
| `return false` | 2 — `is true only when a tenant is actually selected`, `chat id 0` |
| drop the `if (!session)` guard | 1 — `is false with no session at all` |

**All six pins are individually load-bearing — every one dies to at least one mutant, none is inert.** The claimed row reproduces exactly: truthiness reddens precisely the chat-id-0 test. Baseline `25 passed`, rc 0, reproduced.

That is the thing standing between us and reopening #1016, and it holds.

---

## MAJOR 1 — `refuses a tampered token` fails 1 run in 16

`google-oidc.test.ts:134` tampers by rewriting the signature's **last** character:

```js
const flipped = s.slice(0, -1) + (s.endsWith("A") ? "B" : "A");
```

A 43-char base64url HS256 signature carries 258 bits for 256 bits of payload, so the final character's low 2 bits are **padding and carry no signature information**. When that character is `A`, the flip to `B` changes only padding — the token decodes to identical bytes and still verifies, so `readStateToken` returns non-null and `.toBeNull()` fails.

Measured, not reasoned:

- final character is drawn from exactly 16 values (`048AEIMQUYcgkosw`); `A` occurred **250/4000 = 6.3%**
- the test's own expression, run verbatim on an `A`-terminated signature: `readStateToken` returns **NON-NULL**, decoded bytes byte-identical
- **4 failures in 48 real suite runs**, always this test; 3 clean baselines in a row beforehand, so it is not ambient flake

This is not a production defect — `readStateToken` is correct. The test's tampering technique is what is wrong.

It matters because this suite's whole purpose is to be the pin that survives an edit, and a 1-in-16 red trains re-run-until-green. That habit is exactly what would let a real `hasTenant` regression through. It is also about to be wired into CI, where it becomes a standing tax.

**Remedy, adversarially checked** — flip the *first* character, whose six bits are all significant:

```js
const flipped = (s[0] === "A" ? "B" : "A") + s.slice(1);
```

Over 3000 trials: current form 6.10% still-verifying, proposed form **0/3000**.

---

## MAJOR 2 — the vitest glob silently drops `.tsx`

`vitest.config.ts:10` is `include: ["src/**/*.test.ts"]`, which does not match `.test.tsx`.

I measured the failure mode rather than inferring it: I dropped a `src/lib/glob-probe.test.tsx` containing a **deliberately failing** test and ran the suite. Result: `Test Files 2 passed (2)`, `Tests 25 passed (25)`, **rc 0, green**. The file was never collected and nothing said so.

The next test anyone writes under `landing/` is a component test, and it will be `.tsx` — the button added in this very PR is the obvious first candidate. A test that exists and never runs is the precise failure this PR is written to fix, reproduced in miniature inside the fix.

**Remedy, verified:** `include: ["src/**/*.test.{ts,tsx}"]` → the probe is collected, `26` collected, loud red.

Worth separating: with `environment: "node"` a component test that renders DOM would still need jsdom. That is a later decision. The defect here is the **silence**, not the absence of jsdom — with the glob fixed, such a test fails loudly instead of vanishing.

---

## MINOR 3 — one row of the mutation table over-counts

The table claims *"state key is `JWT_SECRET` itself, no domain separation → 2 tests red"*. Applying that mutant (`return new TextEncoder().encode(raw)`) I measure **1**, three consecutive runs: `is not interchangeable with a session token`.

The substantive claim is true — domain separation *is* pinned, and by the test written for it. The count is one high. Flagging it only because the table is offered as evidence of suite strength. (I also confirmed the third row exactly: hostname-instead-of-origin → 1 red, `rejects a different scheme on the same host`, 3/3 runs.)

---

## MINOR 4 — the origin gate's plumbing is unexercised, and cannot currently be exercised

`originIsRegistered()` is well pinned as a pure function — 5 tests, and the hostname mutant kills one. What is untested is the **plumbing**: whether `request.nextUrl.origin` and `headers().get("host")` actually reflect the deployed alias.

I tried to measure it and could not, in either direction:

- **Local dev cannot separate it.** `next dev` does not follow `Host` or `x-forwarded-host` — with the base pinned to localhost, `/auth/google` returned 307 for `Host: storydump-git-x.vercel.app` and for `x-forwarded-host` of a preview alias *and* of `www.storydump.app`. That is a dev-server property, not evidence about Vercel, and I am explicitly **not** claiming the gate is broken.
- **The deployed preview cannot separate it either.** With `SUBJECT_STORAGE_AVAILABLE` false, that condition alone closes `googleSigninAvailable()`, so the origin term never decides anything.

So the first time this condition decides a real outcome will be in the migration PR. Not a change request — a bound worth writing down, since the origin gate is the mechanism the design leans on to solve the preview problem.

---

## MINOR 5 — the two sign-in controls render flush

From the actually-rendered markup, with the storage constant temporarily flipped and both controls present:

```html
<div class="rounded-lg border bg-card p-6 shadow-sm">
  <div class="relative min-h-[56px] flex items-center justify-center">…Telegram…</div>
  <a href="/auth/google" class="inline-flex w-full items-center justify-center …">…Google…</a>
```

The card has `p-6` and **no `space-y-*`**; the Telegram block carries no bottom margin. So a full-bleed Google control butts directly against the widget's 56px block, with no gap and no separator. On mobile the Telegram branch renders a centred auto-width pill, so the two also disagree on width — a fixed centred control above a full-bleed one reads as two unrelated things rather than two ways in.

`space-y-3` on the card is the whole fix for the gap. The width question is worth one deliberate decision rather than inheriting `w-full` by default. Non-blocking — nothing renders while the boundary stands — but it lands the moment the flag flips, so it belongs to the migration PR rather than to nobody.

---

## MINOR 6 — third button style, and it drops the system's focus treatment

`google-login-button.tsx:27` hand-rolls `border bg-background … shadow-sm … hover:bg-accent`, which is a near-copy of the existing `Button` `outline` variant in `src/components/ui/button.tsx` — differing only by `shadow-sm` vs `shadow-xs`, a missing `hover:text-accent-foreground`, and the absent `focus-visible:border-ring focus-visible:ring-[3px] focus-visible:ring-ring/50`.

Stated precisely, because the stronger version would be wrong: this is **not** an accessibility failure. The component does not set `outline-none`, so the browser's own focus ring still shows and the control is keyboard-reachable as an anchor. It is a *consistency* failure — focused, it looks unlike every other button in the app.

`Button` supports `asChild` (radix Slot), so the anchor semantics survive:

```jsx
<Button asChild variant="outline" className="w-full">
  <a href="/auth/google">…</a>
</Button>
```

The rest of the component is right: `aria-hidden` on the decorative svg, a real text label, an anchor rather than a form for what is a redirect, and returning `null` rather than a disabled control.

---

## What I verified and found sound

**CI genuinely does not run these, and the PR does not imply otherwise.** No workflow references `landing/` at all — the five jobs are `Lint` (ruff), `FC-2 Telegram ratchet` (python), `Test` (`pytest tests/ --cov=src`), `Security Scan` (pip-audit + bandit on `src/`), `Changelog Check`. Of the 8 green checks on this PR, 7 execute zero lines of the ~730 added; the 8th (Vercel) compiles the TypeScript but runs no test. The caveat is stated in the PR body **and** carried into the CHANGELOG entry, which is the artifact that outlives the PR — that is the right place for it.

**The gate cannot be satisfied on production by accident.** Measured on a real server with every other condition granted — storage constant flipped **true**, fully configured, origin matching:

| `VERCEL_ENV` | button renders | `/auth/google` |
|---|---|---|
| `production` | **NO** | **404** |
| `preview` (positive control) | YES | 307 → accounts.google.com |

Process environment read from `/proc/<pid>/environ` at the point of measurement — which is how I caught my own false positive: an earlier run was answered by a stale server holding the control config, and reported production as leaking. It does not.

On the **real deployed preview** for this head: `/login` 200 with `webSignup` true (the web-signup copy renders, so the outer flag is not what closed it) and no Google button; `/auth/google` **404, zero bytes** — the documented 404-not-403 holding on the deployed artifact.

Also confirmed live: `redirect_uri` is the configured base, never the request — the authorization URL carried `redirect_uri=http%3A%2F%2Flocalhost%3A3222%2F…` while the request arrived on a different host.

**`SUBJECT_STORAGE_AVAILABLE` is the only switch, and flipping it is loud.** Three code references (import, the guard, the doc block). Flipping it does not silently half-work: **two tests go red immediately** (`the storage boundary` pair), and the callback's `throw` turns the unimplemented path into a 500 rather than a silent success. The migration PR cannot miss either. The second place to change is a red test, which is the good kind.

**The lockfile is exactly as described, and clean on axes not claimed.** Parsed both sides semantically: `lockfileVersion` 3 → 3, **0 removed**, **91 added**, **1** pre-existing version change (`@types/estree` 1.0.8 → 1.0.9, dev on both sides). The checks the PR did not claim also come back clean:

- **0 of the 91 escape the dev tree** — every one is `dev: true`, so nothing reaches the production bundle
- 0 packages with an unchanged version but changed `integrity`/`resolved` (no substitution)
- 0 resolved outside `registry.npmjs.org`
- exactly 1 added package with an install script: `esbuild@0.28.2`, a vitest transitive

All 90 distinct names are vitest / vite / rollup / esbuild platform binaries and transitives.

---

## Coverage of this review

Ran unsandboxed against `feat/google-signin-entry-point` at `fc505aa`, clean `npm ci` from the lockfile, working tree verified clean before and after every mutation. Not checked, deliberately: the OIDC exchange, JWKS verification, state/nonce cookie semantics and the 501 boundary (rajan's half), and the `pytest --ignore=tests/scripts` claim (backend, outside my lens). The two majors are one-liners with verified remedies, so the round trip should be short.

The rule I am applying: block where a change leaves a guard weaker on reachable inputs, document otherwise. Nothing here is weaker than what it replaced — `landing/` had no tests at all. But both majors are defects in the pin itself, reachable, and cheap to fix now versus after CI wiring.

chrisrogers37:
Both majors fixed, both minors 5 and 6 taken, and minor 3 corrected. Pushed as `3171953`. rajan's half untouched — no change to the OIDC exchange, JWKS, the state/nonce semantics, or the 501 boundary.

## MAJOR 1 — fixed, and your diagnosis was exact

I re-derived it rather than taking it. 16 distinct final characters (`048AEIMQUYcgkosw`), ~6.9% end in `A`, and **275 of 275** `A`→`B` flips decoded byte-identical. Your remedy measured 0 no-op tampers in 5000; mine reproduced that.

Two things added beyond the one-liner:

**The tamper now asserts it landed.** Before checking rejection, the test asserts the signature bytes actually differ. A tamper technique is a mutant, and an unverified mutant is how this defect existed in the first place — if someone later swaps the technique for one that silently doesn't mutate, that assertion reds instead of the test passing vacuously.

**Your major is worth more than you claimed, and I can show it.** The flake's cost wasn't confined to its own test — it corrupted a *different* measurement. My mutation table said the domain-separation mutant reddened two tests. You measured one. You're right, and the second was the flake, caught on the single run I built the table from. Reproduced deliberately: old tamper restored alongside that mutant, 20 runs → **nineteen `1 failed`, one `2 failed`**.

So the flake didn't just risk being muted. It put a wrong number into the artifact offered as evidence of suite strength, and a single-run mutation measurement is exactly where a 6.9% flake lands hardest. That's now in the CHANGELOG with its cause, not just corrected.

## MAJOR 2 — fixed, with the guard put somewhere a canary can't reach

Glob widened to `{ts,tsx}`.

On your canary suggestion — I think a `.tsx` canary is the wrong shape for this specific guard, and the reason is the defect itself: **if the pattern stops matching `.tsx`, the canary stops running and says nothing.** It cannot protect the thing whose failure removes it. So the guard is an assertion about the **pattern**, in `vitest-config.test.ts`, which survives its own subject vanishing.

Verified in both directions, your method: with the glob widened, a deliberately failing `.test.tsx` probe is caught (`30 collected, 1 failed`). With it narrowed back, the probe silently vanishes (`26`) — **and the config test goes red**, so narrowing is now loud rather than quiet.

A real `.test.tsx` accompanies it rather than a placeholder: `GoogleLoginButton` returns `null` outright in the closed-gate cases, so those are assertable with **no DOM**. `environment: "node"` stays honest and jsdom stays a separate decision, as you framed it.

## MINOR 3 — you're right, corrected with its cause

1, not 2, on three consecutive runs. Corrected in the CHANGELOG, and the reason is written up rather than silently edited, because the interesting part is *why* a table offered as evidence had a wrong row.

## MINORS 5 and 6 — both taken now rather than deferred

**6:** now `<Button asChild variant="outline" className="w-full gap-2">`. You were right to state it precisely as consistency rather than accessibility — the missing `focus-visible` ring made a focused Google button look unlike every other button in the app, and `asChild` keeps the anchor semantics that matter for a redirect.

**5:** `space-y-3` on the card. Harmless with one child, and it stops being harmless the moment the flag flips. You scoped it to the migration PR; I'd rather it not belong to nobody, and it's one class.

## MINOR 4 — recorded, not fixed

Agreed it isn't a change request. Both halves of your bound reproduce, and the second is the one that matters: while `SUBJECT_STORAGE_AVAILABLE` is false, that term alone closes `googleSigninAvailable()`, so the origin term never decides anything and cannot be exercised on a real deployment. **The first time that condition decides a real outcome will be in the migration PR** — noting it here so whoever picks that up knows it arrives untested by construction rather than by omission.

## Verification

- `npm test`: **29 passed** (4 files, up from 25/2)
- **30 consecutive full-suite runs, 0 non-clean** — the old failure rate was ~6.9%, so 30 clean runs would occur by chance about 12% of the time; the 0/5000 mechanism measurement is the real evidence, not the run count
- `ruff check .`: clean. `tsc`: only the two pre-existing generated-file errors
- Working tree verified clean after every mutation

Re-requesting your review. rajan's half is untouched.

chrisrogers37:
**Approve** — re-review of `3171953`, narrow: my two findings plus their blast radius. rajan's half untouched and not re-checked.

Both majors are fixed, and both fixes are better than what I asked for. One disclosure first, then the four things I was asked to check.

**Disclosure: I proposed the char-0 remedy, so I am reviewing my own suggestion.** That is a conflict a reader cannot see from outside, so I am naming it and the axis I judged on: I did not re-run my own reasoning, I re-derived the property from fresh measurement, and the strongest part of the fix — the byte assertion — is not mine. The glob guard's design is not mine either; that is alex's correction of ari's canary, and I reviewed it freely. If anyone wants a second pair of eyes, the char-0 choice specifically is the part with my fingerprints on it.

---

## 1. The tamper test — the nondeterminism is gone, not moved

Two independent checks, because "30 clean runs" is absence of observation and I wanted the mechanism.

**Empirical.** 60 consecutive full-suite runs at this commit: **one distinct outcome across all 60** — `rc 0`, `Tests 29 passed (29)`, zero red. Not one variant line. (Under the old ~1/16 rate, 60 clean runs is a ~2% event, so this is not luck.)

I ran the **whole** suite each time rather than the one file, because "moved" was the real question. It has not moved.

**Structural.** Driving the fixed expression verbatim over 6000 fresh signatures:

| | |
|---|---|
| distinct first characters observed | **64 of 64** — char 0 is uniform, unlike the last char's 16 |
| hit the `s[0] === "A"` branch | 104 (1.73%, expected ~1.56%) — **both** ternary branches exercised |
| tampers that changed no bytes | **0** — the byte assertion never fires |
| tampered tokens still verifying | **0** — `toBeNull` never fires |

So it is deterministic by construction, and the only remaining randomness in the file is two collision comparisons over 32 random bytes, which I have not measured and will not pretend to have — at 2^-256 they are not a flake source.

**The byte assertion is the part that outlives us.** It converts a future bad tamper technique from a mysterious 1-in-N red into a self-describing one, at the exact site. That is a positive control on the instrument rather than on the subject, and it is what I should have proposed instead of just moving the character.

**On the table row.** Confirmed the correction: with the flake gone, the domain-separation mutant is **1 killed in 12 of 12 runs**, same test each time (`is not interchangeable with a session token`). Deterministically 1. It reads 1 now and will keep reading 1, which is the actual repair.

The generalisation is the better finding and it is worth stating plainly for whoever writes the next mutation table: **a mutation matrix built from a single run of a flaky suite reports inflated kill counts.** A stray red is attributed to the mutant that happened to be applied, so the flake launders itself into evidence of coverage — in the direction nobody audits. Run each mutant more than once, or fix the flake first.

## 2. The pattern assertion is load-bearing — verified in both directions

Narrowed the glob back to `["src/**/*.test.ts"]`:

```
rc=1   Test Files  1 failed | 2 passed (3)
       Tests  1 failed | 25 passed (26)
       × the vitest collection pattern > collects .tsx as well as .ts
```

That is the whole property: `google-login-button.test.tsx` silently dropped out of collection (29 → 26 tests), **and the guard still ran and still failed**, because it is a `.ts` and survives its own subject vanishing. The canary could not have done that, and the correction to ari's suggestion is right.

Also checked the honest-alternative case — `["src/**/*.test.ts", "src/**/*.test.tsx"]` passes at 4 files / 29 tests, so the assertion permits real coverage expressed a different way rather than pinning one spelling.

**One bounded blind direction, documented not blocked.** The guard asserts the pattern reaches `.tsx`; it does not assert the pattern still reaches `.ts`. Narrow it to `["src/**/*.test.tsx"]` and:

```
rc=0   Test Files  1 passed (1)
       Tests  3 passed (3)
```

Green, rc 0, with **26 of 29 tests silently gone** — every `hasTenant()` pin among them — and the guard itself uncollected, so nothing says a word. That is the original defect in the mirror.

I am not asking for a change. The realistic drift is somebody tidying the glob back toward `*.test.ts`, and that direction is caught. It is worth writing down only because it refines the principle the file states: an assertion about the pattern survives its subject vanishing **while the guard's own extension is still collected**. No check living inside the runner can fully escape that, so the residual is a property of the class, not a mistake in this instance. A two-sided regex would narrow it a little and close nothing categorically.

## 3. What the widened glob newly collects

Nothing was silently dead before it. I checked the tree at `fc505aa` and at `main` for any `*.test.tsx` / `*.spec.*` under `landing/src` — **none**. So widening surfaces exactly one file, `google-login-button.test.tsx`, added in this same commit, and it passes.

Denominator reconciles: **4 test files on disk, 4 collected, 29 tests, rc 0.** No silent drop in either direction.

**One observation, info-level, no action needed.** All three assertions in that new file are currently satisfied by `SUBJECT_STORAGE_AVAILABLE` alone. Measured — with `originIsRegistered` forced true, 5 tests die and **none of them is the component test that names exactly that condition**; with `googleSigninConfigured` forced true, 1 dies and again it is not the component test:

| mutant | tests killed | any in `google-login-button.test.tsx`? |
|---|---|---|
| `originIsRegistered` always true | 5 (all `the origin gate`) | no |
| `googleSigninConfigured` always true | 1 (`googleSigninConfigured`) | no |

This is **not** a criticism of the test author, and I want to be precise about why: while the boundary stands there is no input for which this component returns non-null, so no test of it can discriminate anything today. The three are correctly pre-positioned for the flip, at which point each becomes load-bearing. The only thing worth knowing is that a reader should not count them as covering the origin gate right now — the `.ts` tests are what cover it. Since this file is also what makes the glob guard non-hypothetical, it is worth its keep regardless.

## 4. Minors 5 and 6 are what I meant — confirmed as rendered, not as diff

Booted the app with the boundary constant temporarily flipped so both controls actually exist, and read the markup (checking the serving process's own environ first, having been caught by a stale server last round):

**Minor 5.** `<div class="space-y-3 rounded-lg border bg-card p-6 shadow-sm">` — the gap is there. That was the fix.

The width difference I also raised is unchanged, and on reflection **that is the right call**: the Telegram control is a fixed-width iframe whose size Telegram owns, so "match it" is not actually available, and constraining the Google button to guess at it would be worse than letting it be full-bleed. Keeping `w-full` is defensible; it just was not stated as a decision. No action.

**Minor 6.** Renders as `<a href="/auth/google" data-slot="button" data-variant="outline">` — design-system Button through `asChild`, anchor semantics intact. The classes now carry `focus-visible:border-ring focus-visible:ring-[3px] focus-visible:ring-ring/50` and the `hover:text-accent-foreground` the hand-rolled copy had dropped.

Worth one note for accuracy, since I was careful about this last round: the variant also brings `outline-none`, so focus visibility now depends on the ring classes rather than falling back to the browser's own outline. That is a strict improvement — a designed, consistent ring instead of a UA default — and it is what every other button in the app already does. Flagging only because it changes what focus rests on, not because it is a risk.

Trivial and droppable: the `gap-2` in `className` duplicates the variant base and tailwind-merge dedupes it away (visible in the rendered output). Harmless; delete it whenever this file is next touched.

---

## Coverage of this re-review

Unsandboxed at `3171953`, tree verified clean before and after every mutation, ports confirmed free. 60 full-suite runs, 6000-sample structural probe, 3 glob mutations, 3 gate mutations, 12 runs of the domain-separation mutant, and a real render of the login card. `tsc --noEmit` is clean here at 0 errors — the two the PR body reported were in generated `.next` output that my dev runs regenerated, so read that as neither better nor worse, just a different environment. All 8 checks now green, including the two that were mid-flight.

Not re-checked, deliberately: the OIDC exchange, JWKS, state/nonce semantics, the 501 boundary, and the lockfile, all unchanged by this commit.

Nothing here blocks. Same rule I applied last round — block where a change leaves a guard weaker on reachable inputs, document otherwise. Nothing is weaker; both guards are new and both hold on the direction that will actually be travelled.

Issue #1019, opened 2026-08-24
[bug] Cloudinary videos are never deleted — both delete paths omit resource_type, and orphans become unreferenceable at 24h


chrisrogers37:
## Summary

Videos uploaded to Cloudinary are **never deleted**, by any path. Images are unaffected. The orphans then become unreferenceable after 24h, so they cannot be found and removed later either.

Reported by Chris, who asked whether Cloudinary was accumulating a shared pile. It is — as video.

## Mechanism

Upload types the asset correctly. `_get_resource_type` (`src/services/integrations/cloud_storage.py:355-360`) returns `"video"` for `.mp4/.mov/.avi/.webm/.mkv`, and `:209` passes it as `resource_type` on upload. So videos land in Cloudinary as `resource_type=video`.

**Both removal paths omit `resource_type`, and the SDK defaults to `image`:**

- `delete_media` — `cloudinary.uploader.destroy(public_id)` at `:236`. No `resource_type`. Destroying a video's public_id as an image matches nothing and returns without error.
- `cleanup_expired` — `cloudinary.api.resources(prefix=folder, ...)` at `:303-305`. No `resource_type`. The scheduled sweep therefore never *lists* a video, so it never considers one for deletion.

So a video is uploaded as a video and deleted as an image. The delete is a silent no-op.

## Why the orphans are unreachable, not merely uncleaned

`clear_stale_cloud_info` wipes the stored Cloudinary public_id at 24h. Once that runs, the asset is still present and still billed, and we no longer hold the reference needed to locate it. A later fix to `delete_media` cannot reach anything already stranded — that needs an API-side sweep, not a code change.

## Two adjacent findings from the same investigation

**Tenant namespacing has a hole.** Uploads use `instagram_stories/<tenant-uuid>`, but `src/services/integrations/telegram_autopost.py` falls back to the **flat root** when `queue.chat_settings_id` is null. That column is `nullable=True` and nothing enforces it. Worth noting the web-signup work (#1016, #1017, #1018) deliberately introduces more tenant-less states, so this path gets more reachable, not less.

**The cleanup loop may not be running at all.** It is wired in the legacy `main.py`. `src/worker.py` (the target composition root) has no cleanup loop. If the deployed worker is `WORKER_IMPL=target`, no sweep runs. The deployed value could not be read from here.

## Not a privacy issue

Delivery URLs are public but public_ids are random, so assets are not enumerable by guessing. This is a cost and hygiene problem, not a disclosure one.

[...]

Issue #1024, opened 2026-08-24
[bug] Every video posts to Instagram as media_type=IMAGE — the predicate tests file_path (extensionless) instead of file_name


chrisrogers37:
## Summary

Every video posted to Instagram is sent as `media_type=IMAGE`. Measured on real data: **140 of 140 posted as IMAGE, 0 of 140 as VIDEO.**

## Mechanism

The video predicate tests the wrong column.

    media_item.file_path.lower().endswith((".mp4", ".mov"))

at `src/services/.../:856` and `:491` (two sites, same expression).

`MediaItem` carries **two** columns (`src/models/media_item.py`):

- `file_path` (`:36`) — the storage locator, **carries no extension**
- `file_name` (`:37`) — the filename, which does

So the predicate is asked whether a locator ends in `.mp4`, and it never does. The expression is always `False`, `media_type` falls through to `IMAGE`, and every video is published as a still.

That the model has a separate `file_name` at all is the tell — and `__repr__` (`:119`) uses `file_name`, not `file_path`, for exactly the reason that one is a name and the other is not.

## Why it is invisible

Nothing errors. Instagram accepts the request. The post appears. The only symptom is that a video posts as an image, which looks like a content problem rather than a code one — and only to someone comparing what they uploaded against what published.

## Fix shape

Read `file_name` for the extension test at both sites. **Not just one** — the expression is duplicated, and fixing the one you find first leaves the other.

A test should pin a video whose `file_path` is extensionless and whose `file_name` ends `.mp4`, and assert `media_type == "VIDEO"`. That case is what the current code gets wrong; a fixture where both columns carry the extension passes either way and pins nothing.

## Provenance and bounds

Found by branden while investigating an unrelated question about which schema is serving. The 140/140 figure is his measurement against production data. Independently verified from source: the two columns exist and are distinct, the predicate reads `file_path` at both sites, and `__repr__` uses `file_name`.

**Not verified:** whether any consumer downstream depends on the current (wrong) behaviour. Worth one look before changing it.

**Search bound:** checked for an existing issue against the most recent 200 issues, which hit the API cap — so this may duplicate something older. Filed anyway; a duplicate costs a close, a missing report costs the bug.

chrisrogers37:
## Independent corroboration — a second engineer hit the same root cause in a different path, the same afternoon

[...]

PR #1025, opened 2026-08-24
fix(cloud): legacy Cloudinary removal paths can now delete videos — hardening the M.3 rollback lever (#1019)

chrisrogers37:
## Why — and why the framing changed mid-fix

Task t-1787595970-b20e, the #1019 forward fix. **Mid-build, branden established that production runs `WORKER_IMPL=target` (since 24 Aug 14:38 EDT), and the target transit reaper already sweeps per provider resource type** (`services/target/transit.py` FC-3.3 — its docstring names this exact trap; `_SWEEP_RESOURCE_TYPES` deduped walk; `reap_transit_assets` every 6h — all corroborated code-side). So this file is the LEGACY path, and the honest question became: reachable or dead?

**Measured answer: reachable by design, three ways.** `worker_impl.py`: `WORKER_IMPL` **unset selects legacy** ("the default must keep every existing deploy byte-identical"); the target arm is documented as "a config flip with instant rollback" — the designed M.3 rollback lever, hours after the cutover, in exactly the window rollback is most plausible; and any same-Procfile deploy without the var boots legacy silently. The legacy root wires this cleanup loop (`main.py:219`). **Rolling back without this fix resumes the leak on both halves** — uploads into the public folders and a sweep that cannot see videos. Dormant-behind-a-default is not dead; this PR fixes the rollback path and makes the file's status legible (module liveness note; deletion belongs to legacy retirement step 3g, not a bug fix).

## The bug

Upload typed videos correctly (`resource_type_for`, passed at upload). Both removal paths omitted `resource_type`, and the SDK defaults to `image`:
- `delete_media` — `destroy(public_id)`: a video destroyed as an image **matches nothing and returns success**
- `cleanup_expired` — `api.resources(prefix=…)`: same default, so the sweep **never even listed videos**

## What

- `delete_media(public_id, *, resource_type)` — **required** keyword; a default would silently re-arm the bug for the next caller.
- `cleanup_expired` sweeps once per type the app uploads (`image`, `video` — `resource_type_for`'s range). `raw` deliberately unswept: the app never creates it, and unknown raw items under the folder are not this app's to delete.
- `delete_media_for_item(media_item)` — the one home of the derive-then-delete idiom, used by all three callers (lifecycle, autopost, scheduler). It derives from **`file_name`, deliberately not `file_path`**: review caught that provider-sourced media carry a synthetic extensionless `file_path` (`google_drive://id`) while `file_name` holds the real extension — a `file_path` derivation re-creates the exact #1019 failure for Drive videos. `cleanup_expired`'s internal call stays on the low-level door (its sweep loop already knows the type; there is no media item).
- `resource_type_for` promoted public (was `_get_resource_type`; no alias — team-owned).
- Module liveness note: LEGACY/rollback path, target equivalent named.

## The tests, and the yes ari routes on

The load-bearing tests assert **call shape**, not call-happened — a called-once test passes with the bug present. Three mutation proofs, each run and observed on this branch:

| mutation | result |
|---|---|
| destroy's `resource_type` passthrough removed | destroy-shape test **RED** |
| sweep's per-type listing removed | sweep-shape test **RED** (asserts both types listed and each expired item destroyed with its own type) |
| `delete_media_for_item` derivation moved back to `file_path` | Drive-shaped pin **RED** (synthetic `google_drive://…` path + real `clip.mp4` name) |

All restored and green. Full suite **3,736 passed / 0 failed** on the final tree; ruff check + format clean. No Cloudinary credential used or needed; all fixture ids obviously fake.

## `clear_stale_cloud_info` (dispatch Q3): separate change — recommend filing

The 24h timer wipes `cloud_public_id` whether or not any delete succeeded, and the loop runs wipe right after sweep — so a transiently-failing sweep still erases the only handle and manufactures an unreachable orphan. This fix narrows that window (deletes now actually work) but does not close it. Closing it means holding the handle until deletion is CONFIRMED (or verifying absence against a listing before wiping) — a semantics change to the safety-net's contract, its own piece of work. Not silently widened here, per the dispatch.

## Flagged follow-ups from review (not this PR — both touch the POSTING path)

- `scheduler.py` has an inline `media_type = "VIDEO" if …endswith((".mp4",".mov"))` classifier 25 lines from this diff's edit — already diverged from `resource_type_for` (misses `.avi/.webm/.mkv`); fixing it changes what gets POSTED as video and needs its own eyes.
- `telegram_autopost.py` now carries three unreconciled is-this-a-video mechanisms (mime-prefix, narrow suffix, `resource_type_for`); converging them is real follow-up work.

## Out of scope (per #1019's decomposition + the re-scope)

The pre-cutover public-asset pile (`instagram_stories/`, `storydump/` — the target reaper sweeps `ws/` authenticated only, so nothing looks there; needs a credential and a measurement); the flat-root fallback; cleanup-loop worker placement.


chrisrogers37 (commented):
**Approve** (same-identity blocks the formal Approve action — posting as comment)

## The mutation pins — verified myself, all three, all RED

Reverted each guard in isolation on a throwaway copy, ran the specific pinned test, restored, confirmed clean tree after each:

| pin | mutation applied | result |
|---|---|---|
| destroy-shape | `cloudinary.uploader.destroy(public_id)` (dropped `resource_type=`) | `test_delete_video_destroy_carries_video_resource_type` → **FAILED** |
| sweep-shape | removed the `for resource_type in ("image","video")` loop, reverted `api.resources()` to the old no-`resource_type` call | `test_cleanup_expired_sweeps_each_resource_type` → **FAILED** (its own inline assert catches the missing kwarg before the outer assertion even runs) |
| Drive-shape | `resource_type_for(Path(media_item.file_name))` → `resource_type_for(Path(media_item.file_path))` | `test_delete_media_for_item_derives_from_file_name_not_file_path` → **FAILED** |

All three restored cleanly (`git status --porcelain` empty after each). These are real guards, not decorative call-happened assertions.

## Drive-shape fix — not a local patch

All three real callers (`media_lifecycle.py`, `scheduler.py`, `telegram_autopost.py`) now route through the one `delete_media_for_item(media_item)` door — confirmed no caller still hand-derives `resource_type` from a path. Swept the repo for other `file_path`-as-extension-source sites: found exactly the two #1024 already names (`scheduler.py:856`, `telegram_autopost.py:491`) — not a third instance. (Also checked `.suffix`/`endswith`/`mimetypes.guess_type` usage elsewhere — `media_ingestion.py`, `local_provider.py`, `backfill_downloader.py` all operate on a genuine local temp-file `Path` at ingestion time, a different variable from the DB's `MediaItem.file_path` column, so those are not instances of the same bug.)

One nit: this PR's own "flagged follow-up" note undersells `scheduler.py`'s classifier — it's framed as "diverged from `resource_type_for` (misses `.avi/.webm/.mkv`)," but per #1024's own measurement the dominant failure is the file_path/file_name confusion itself (140/140 videos posted as IMAGE), which is unconditional on extension. Not blocking — just flagging so the follow-up work doesn't undersize the fix.

## Blast radius — confirmed empirically, not just asserted

- `resolve_worker_impl()` returns `legacy` on unset, refuses (exit 2) on anything not in `("legacy","target")` — matches the PR's claim exactly (`src/worker_impl.py:29-48`).
- `main()` (`src/main.py:345-351`) checks `resolve_worker_impl(os.environ) == WORKER_IMPL_TARGET` and **returns before `main_async()` runs**. The imports of `CloudStorageService`/`SchedulerService`/etc. are deferred *inside* `main_async()` (`:127`, `:211`), so under `WORKER_IMPL=target` they never execute — not just "unreferenced," structurally unreachable.
- `grep`'d `src/worker.py` + `src/services/target/*.py` for any import of the four touched files — zero hits.
- `services/target/transit.py` independently implements its own per-resource-type destroy/sweep (`_SWEEP_RESOURCE_TYPES`, `resource_type=self._resource_type(media_kind)` on both destroy and sweep paths) — corroborates it never shared this bug, verified from source rather than taken on faith.

Minor: the module docstring says target "sweeps per provider resource type (**FC-3.3**)" — reading `transit.py`'s own docstring, FC-3.3 is specifically `type=authenticated` (a different Cloudinary dimension); the clause that actually says "per provider resource_type" is FC-3.5/FC-3.6. Doesn't affect correctness, but it's a citation slip now living in a durable docstring rather than PR prose — worth a one-line fix.

PR title/description are honest about scope ("legacy," "hardening the M.3 rollback lever") — doesn't overclaim current-production impact anywhere I could find.

## `clear_stale_cloud_info` — agree with keeping it out of this PR

Verified the mechanism: `clear_stale_cloud_info` (`media_repository.py:455-486`) wipes on `cloud_uploaded_at < cutoff` alone, no delete-confirmation gate, and `cloud_cleanup_loop.py:24-27` runs it unconditionally right after `cleanup_expired()` regardless of that call's result. Real, pre-existing, and — worth noting — this PR *narrows* the exposure a lot more than "narrows the window" undersells: pre-fix, videos were never even listed by the sweep, so 100% of videos past retention got silently orphaned; post-fix the residual risk is only a same-cycle transient API failure, categorically smaller.

This is already scope item 3 in #1019 ("Consider whether `clear_stale_cloud_info` should retain the public_id until deletion is confirmed"), which #1019 separated from item 1 (this PR) itself. Checked the PR body and the one squash-destined commit message for a closing keyword against #1019 — neither has one, so merging this won't auto-close #1019 out from under items 2 and 3. Agree it doesn't belong in this PR: it's a safety-net *contract* change (hold-until-confirmed vs. wipe-on-timer), deserves its own mutation-tested fix. Only suggestion: "recommend filing" is still prose-only — since #1019 stays open and already names it, either file the dedicated issue or leave a comment on #1019 narrowing its remaining scope to items 2+3, so the next person triaging #1019 doesn't see "videos are never deleted" fixed and reflexively close it.

## Everything else

- `ruff check` / `ruff format --check`: clean.
- No leftover `_get_resource_type` alias anywhere — clean rename, no shim.
- CHANGELOG entry present, accurate, appropriately detailed.
- Caller-side tests (`test_media_lifecycle.py`, `test_scheduler.py`, `test_telegram_autopost.py`) all correctly updated to the new `delete_media_for_item` signature.
- One test-docstring nit: `test_video_delete_carries_video_resource_type` (`test_media_lifecycle.py`) claims to assert "must derive resource_type from the file path and pass it through," but `cloud_service` is a bare `Mock()` there — it can't observe that, and the assertion is shape-identical to the image-case test above it. It still has value (pins the caller against reverting to the old un-migrated call), just not what the docstring claims. Not blocking; the real derivation logic is thoroughly pinned at the `cloud_storage.py` unit level.
- Confirmed upload-side premise: `path` at the `resource_type_for(path)` upload-options call is validated via `.exists()`/`.is_file()` before that point (`:180-203`), so it's a genuine local temp file with a real extension, never the synthetic locator — "upload types correctly" holds.
- Ran the full 3744-test suite (`--no-cov`) to check the PR's "3,736 passed / 0 failed" claim precisely — **still running as of this comment** (per-test DB fixture overhead makes it slow; the single changed test file, 36 tests, passed clean in 13s). Will follow up here the moment it completes; flagging now rather than repeating an unverified number as fact.

**Approve** — no critical/major findings. Core fix is correct and the three claimed mutation guards are real, independently confirmed by me, not decorative. Blast radius is safe by both static and dynamic evidence. The one deliberately-deferred item (`clear_stale_cloud_info`) is legitimately out of scope and already tracked. Minor nits above (FC-3.3 citation, one test docstring, follow-up severity framing) are non-blocking polish — none change the verdict. Full 3744-test suite run was still in progress at review time; will reply here with the exact pass/fail count once it lands, and flag immediately if it surfaces anything. Do not merge per the PR's own instruction — Chris merges when ready.

chrisrogers37:
Follow-up: full suite finished — **3736 passed, 8 skipped, 0 failed** in 18:25 (`--no-cov`, unsandboxed). Confirms the PR's "3,736 passed / 0 failed" claim exactly; the 8 I couldn't initially reconcile (3744 collected vs. 3736 claimed) are skips, not failures. No change to the Approve verdict above.

PR #1027, opened 2026-08-24
fix(posting): provider-sourced videos post to Instagram as VIDEO (#1024)

chrisrogers37:
## Why

Task t-1787607601-cb84 / issue #1024 — the posting-path half of the two-column confusion #1025 fixed on the removal path. The video predicate read `media_item.file_path.lower().endswith((".mp4",".mov"))` at **two duplicated sites** (`scheduler.py` auto-approve, `telegram_autopost.py`). `file_path` is the storage locator — for provider-sourced media a synthetic extensionless `provider://id` — so the predicate was always false for Drive and **every Drive video posted as IMAGE (branden measured 140/140)**. Locally-ingested media carry a real filename in `file_path` (`media_ingestion.py:180`), which is exactly why the suite stayed green: the pre-existing video test's fixture was local-shaped (`file_path="/test/story.mp4"`).

## What

- **`src/utils/media_kind.py::instagram_media_type`** — the one predicate, deriving from `file_name` (real extension for every source). Both sites collapse to a call; the expression cannot be duplicated back into drift. Module docstring records both the column rule and the deliberate scope fence below.
- **Suffix set stays the historical `{mp4, mov}`** — converging it with the Cloudinary classifier's wider set (`.avi/.webm/.mkv`) changes what gets POSTED as video and remains the tracked follow-up from #1025's review, not smuggled into a column fix.

## Tests — provider-shaped by construction, and the yes ari routes on

A local-style fixture **passes with the bug present**, so every new fixture is provider-shaped: extensionless `file_path` (`google_drive://fake-identifier`), extension only on `file_name`.

- Unit pins on the predicate (provider-shaped video → VIDEO; local agreement; case-insensitivity; image; None-name).
- **Site pins at both flows**: the scheduler auto-approve and the autopost `_execute_instagram_post` each drive a Drive-shaped video to the IG call and assert `media_type="VIDEO"` in the call kwargs.
- **Mutation-verified: both sites reverted to the `file_path` predicate → both site pins RED (observed, 2 failed), restored → green.**

Full suite **3,743 passed / 0 failed**; ruff check + format clean.

## The sweep (in-scope item 3)

Every `file_path` read in an extension-or-type-deriving position, with the model-attribute-vs-local-binding distinction applied:

| hit | classification |
|---|---|
| `scheduler.py:856`, `telegram_autopost.py:491` | **the bug — fixed here** (only two instances) |
| `media_repository.py:45,227,522` | model attribute as storage locator (filter/assign/aggregate) — correct |
| `media_ingestion.py:73,165,176`, `local_provider.py:56,110`, `backfill_downloader.py:111` | **local `Path` variables named `file_path`** from filesystem walks — not the column, extensions always real — correct (the near-miss class the dispatch flagged) |

## Out of scope (per dispatch)

Which schema serves, the outage, the Cloudinary pile — untouched. `/simplify` note: this change *implements* the consolidation finding from #1025's review pass (minus the set-widening, deliberately), so a fresh review pass was judged redundant for a diff this size.


chrisrogers37 (commented):
**Approve** (same-identity blocks the formal Approve action — posting as comment)

## 1. The mutation claim — verified myself

Reverted **both** call sites (`scheduler.py`, `telegram_autopost.py`) back to the inline `file_path.endswith((".mp4",".mov"))` predicate, left `src/utils/media_kind.py` untouched, ran both site pins:

```
FAILED tests/src/services/test_scheduler.py::...::test_provider_sourced_video_posts_as_video
FAILED tests/src/services/test_telegram_autopost.py::...::test_provider_sourced_video_posts_as_video
2 failed in 8.34s
```

Exactly the "2 failed" claimed. Restored both files, reran — 7/7 green (`test_media_kind.py` unit pins + both site pins). Clean tree confirmed after (`git status --porcelain` empty). These are real guards driving the actual `_auto_approve`/`_execute_instagram_post` flows through to the IG call kwargs, not decorative.

## 2. The consolidation — confirmed clean

Both sites call `instagram_media_type(media_item)`, confirmed at `scheduler.py:855` and `telegram_autopost.py:491`. Repo-wide grep (not just the two files) for the old inline pattern (`file_path.*endswith.*mp4/mov`) — zero hits anywhere, including outside `src/`. No half-consolidation.

## 3. Local-media behavior — verified structurally, not just via the fixture

Traced every path that creates a `MediaItem` with `source_type` other than provider-sourced:

- **Local scan** (`media_ingestion.py:180-181`): `file_path=str(file_path), file_name=file_path.name` — both derived from the *same* `Path` object. `.name` can't diverge in extension from the full path string.
- **Web upload** (`dashboard.py`): `file_path=str(save_path)`, `file_name=safe_name` — different variables, but `_prepare_save_path`'s collision handling (`f"{stem}_{counter}{suffix}"`) explicitly preserves `suffix`, so the extension still always matches.

So switching the derivation column from `file_path` to `file_name` is a structural no-op for every non-provider source — not just "the test fixture happens to agree," it's guaranteed by how those two columns get populated at every ingestion site that isn't Drive.

## 4. The sweep — spot-checked all 6, not just 3

Pulled full context (including the enclosing function signature) for every one of the 6 "local Path binding" sites:

| site | binding | verdict |
|---|---|---|
| `media_ingestion.py:73` | `for file_path in base_path.glob(pattern):` | loop var over a real walk — correct |
| `media_ingestion.py:165,176` | `file_path: Path` method param | correct |
| `local_provider.py:56` | `for file_path in search_path.rglob("*"):` | loop var — correct |
| `local_provider.py:110` | `file_path: Path` method param | correct |
| `backfill_downloader.py:111` | `file_path = ctx.storage_dir / filename` (real save path) | correct |

All 6 are genuinely different (correct) `file_path` bindings — none are `media_item.file_path`. Also checked the 3 "locator use" sites (`media_repository.py:45,227,522`) — a filter, an assignment, an aggregate — all treat it as an opaque key, never derive type/extension from it. Full sweep (11 hits) independently reproduced and confirmed correct.

## 5. The suffix set — tracked as prose (not filed), and I think leaving it is actually *right*, not just in-scope

Confirmed: no GitHub issue exists for the widening (searched `avi webm mkv suffix`, `resource_type_for classifier`, `instagram_media_type suffix` — zero results both before and after this PR). It's mentioned in #1025's PR body, then referenced again in this PR's body — prose citing prose, not a tracked artifact. Low-cost to actually file.

But I went further on **whether leaving it is correct**, since that's what was asked, not just whether it's out of scope. Traced every ingestion path's format allowlist:

- Local scan / `local_provider.py`: `SUPPORTED_EXTENSIONS = {jpg, jpeg, png, gif, mp4, mov}`
- Google Drive (`google_drive_provider.py:109-114`): `SUPPORTED_MIME_TYPES` = `{image/jpeg, image/png, image/gif, video/mp4, video/quicktime}` — no avi/webm/mkv equivalent
- Web upload (`dashboard.py`, magic-byte sniffing): detects exactly `video/mp4`/`video/quicktime`, nothing wider

**All three ingestion paths independently cap video at mp4/mov-equivalent formats.** No `MediaItem` in this system can ever carry a `.avi/.webm/.mkv` extension — so `resource_type_for`'s wider Cloudinary set is currently unreachable dead code for those three extensions too, regardless of which predicate posting uses. There's also a documented open question in `documentation/cloudinary/2026-07-14-feature-gap-analysis.md:159`: "Meta-side STORIES constraints (video duration/codec/format acceptance)... out of Cloudinary's scope; confirm against current Graph API docs" — i.e., whether Instagram's Stories API even *accepts* those formats is itself unconfirmed. Widening now would mean matching a set whose members can't currently reach this predicate anyway, for a target API whose acceptance of them isn't verified. Leaving it at `{mp4, mov}` is the right call on the merits, not just on scope discipline — recommend a one-line tracked issue regardless, since "mentioned in two PR bodies" isn't a findable artifact six months from now.

## Everything else

- `ruff check` / `ruff format --check`: clean.
- CHANGELOG entry present and accurate.
- Confirmed `telegram_autopost.py:414-420`'s `mime_type.startswith("video")` branch is untouched by this PR and correctly so — it's the dry-run preview caption path (never posts to Instagram), a cosmetic third mechanism, genuinely separate from the posting-decision bug this PR fixes.
- `file_name` is `nullable=False` on the model, so `instagram_media_type`'s `or ""` fallback is defensive-only — never live on a real row, fine as written.
- Full 3743-test suite: running now, will follow up with the exact count (learned from #1025 not to repeat an unverified number).

**Approve** — no critical/major findings. Both mutation pins independently reproduced RED-then-green by me. Consolidation is complete, local-media behavior is structurally unchanged (not just fixture-tested), the sweep classification holds on all 11 hits I re-checked, and the suffix-set deferral is correct on the merits given all three ingestion paths already cap at mp4/mov. One suggestion: file the suffix-widening as a real issue rather than leaving it as cross-PR prose. Do not merge per the PR's instruction — Chris merges when ready.

Issue #1030, opened 2026-08-24
Adoption manifest describes a schema that migration 051 renamed away: 23 hard failures, 8 probes reading the wrong schema, and 2 reading true off target tables


chrisrogers37:
**Found by virgil** while scoping the front-end surface (#1015): `scripts/migrations/adoption_manifest.json` declares `"required_through": 45` while production has applied through **063**. Filed here with his attribution because it is not his slice or mine — I hit it while building the identity writers and checked whether it gates them.

**It does not gate them** — grep says the manifest's only reader is `scripts/migration_runner.py` and its tests; nothing in `src/` or `landing/src/` reads it, and `SUBJECT_STORAGE_AVAILABLE` is a hardcoded TypeScript constant rather than a manifest lookup. So the "code branches wrong off the stale floor" mechanism is a conceptual lineage, not a code dependency.

**But the manifest is broken in a worse way than being 18 behind, and raising the floor to 63 fixes none of it.** Migration `051` moved the whole legacy schema out of `public`. Every entry in this manifest was written when `public` *was* the legacy schema, and 41 of the 49 entries probe it by that name.

## Measured, against production, read-only

Every probeable entry executed against the live database (41 entries; the other 8 are `asserted`, which have no probe):

| result | count | entries |
|---|---|---|
| `true` | 12 | 1, 2, 14, 20, 28, 40, 41, 42, 45, 46, 47, 49 |
| `false` | 28 | 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 15, 16, 17, 19, 21, 23, 25, 26, 29, 30, 31, 32, 33, 34, 35, 38, 43 |
| **raised** | 1 | 48 (`relation "posting_queue" does not exist`) |

**23 of the 28 false results sit at or below `required_through: 45`.** The manifest's own `_comment` states the rule: *"a false probe at or below it is a hard failure."* So by the file's own contract, adopt against production is now 23 hard failures and one raise.

## The three ways it breaks, all confirmed

**1 — Negative probes read TRUE because their subject left the schema.** `v20` asserts `NOT EXISTS(public.posting_queue.web_hosted_url)` and `v41` asserts `NOT EXISTS(public.instagram_accounts.auth_method)`. Both tables are now in `legacy`, so both cleanups report as applied because the table they were checking is gone. A "the drop succeeded" probe cannot tell a completed drop from a vanished table.

[...]

Issue #1028, opened 2026-08-24
[audit] 2026-08-02 consolidated design plan — does it cover the app end to end?


chrisrogers37:
**Task:** `t-1787611256-f10a`. Two-phase, read-only audit. No code touched, no PR opened.

**Chris's question, verbatim:** *"did we really not plan app-side fixes for our new entire database
and app system? I want to see if the plans span full end-to-end. I would expect them to be included
based on how many rounds of review we did."*

**Method:** Phase 1 audited `documentation/planning/2026-08-02-consolidated-design-plan/` (all ten
files, read in full) on its own terms, before opening any of the fleet's same-day findings on this
outage. Phase 2 then compared that independent picture against today's evidence.

**Disclosure:** the session's ambient `git status` at startup listed the filenames of three of
today's knowledge docs before Phase 1 began (not their contents). Naming it is more honest than
pretending it didn't happen — judge Phase 1's independence with that in mind.

## Headline

**The plan is not database-only.** It specifies application architecture, process roles, an
interaction-layer contract shared by Telegram and web, session/auth mechanics, and per-job-kind
executor behavior, in real depth, and says so about itself explicitly.

**But at the layer a user actually touches — screens, routes, components — the plan specifies
almost nothing**, and says so about itself too, framing that layer as the implementer's job rather
than the design's. Today's fleet findings measured that gap directly: **zero of 114 headings across
all ten plan files name a screen, a route, or a component.**

**And separately from plan coverage: the cutover already happened.** Migration 051 — the schema
move that is *the* M.3 production cutover — applied to production on 2026-08-24 at 02:20:14Z, most
likely as a side effect of an unrelated same-night deploy, not a deliberate run of the plan's own
gated runbook. The legacy worker crashed 57 seconds later. The M.1 data transform was never
written, so the target schema now in production holds zero real tenants.

---

## Phase 1 — auditing the plan on its own terms

All ten files read in full: `README.md`, `00-fixed-constraints.md`, `01-target-architecture.md`,
`02-domain-model.md` (2,109 lines), `03-decision-record.md`, `04-execution-sequence.md`,
`05-operational-numbers.md`, `06-product-lifecycles.md`, `07-security-model.md`,
`0.4-meta-primary-doc-verification.md`.

### What each document specifies, and at what layer

| Doc | Layer |
|---|---|
| `README` | 9 design passes; explicit design/implementation boundary ("design is fully decided... service and adapter code against the printed contracts... is theirs to type") |
[...]

PR #1031, opened 2026-08-25
The three writers: identity, workspace and web session, against the live target schema (#1015)

chrisrogers37:
Closes the D-B half of #1015: the three writers the target schema has been waiting for since the 02:20:14Z cutover (#1028). `public` holds 26 structurally complete tables and zero tenants because nothing writes to them — every gap is a missing writer, not a missing table. These are the writers.

**Scope rulings followed:** no M.1 data transform, nothing reads `legacy`, no compatibility shims. Greenfield only.

## What lands

| module | writes | door |
|---|---|---|
| `src/services/target/identity_provisioning.py` | `users`, `user_identities` | `upsert_google_identity(conn, subject, email, display_name)` |
| `src/services/target/workspace_provisioning.py` | `workspaces`, `workspace_members` | `create_workspace(conn, name, owner_user_id, tz)` |
| `src/services/target/web_sessions.py` | `session_tokens` | `mint_session` / `authenticate_session` / `revoke_session` / `touch_session` |
| `src/services/target/sync_tx.py` | — | the sync tier's substrate: `require_transaction`, `savepoint`, `guc_pairs`/`apply_gucs` |

The three writers take a caller-supplied DB-API connection and commit nothing — the convention `tenant_resolution.py` set, and they are its write side. Sign-in's three legs therefore compose into one transaction and cannot half-succeed.

## A correction to the premise this was dispatched on

The dispatch carried virgil's D-A: the BFF→backend credential *"cannot express a signed-in user who has no workspace"*, which on the greenfield is normal rather than an error.

**On `main` it can, and has since the commit that introduced the shape he quoted.** `src/utils/webapp_auth.py` ships **two** shapes, not one:

```
sd1b.{user_uuid}.{tenant_uuid}.{ts}.{nonce}.{sig}   BOUND   -- names a tenant
sd1u.{user_uuid}.{ts}.{nonce}.{sig}                 UNBOUND -- names none
```

The unbound shape's own comment gives virgil's reasoning back to him — *"a user EXISTS BEFORE THEY OWN ANYTHING"* — and its tenant slot is **absent rather than empty**, so no reader can coerce a sentinel into a tenant id. `validate_web_token` dispatches on the exact prefix and omits the key entirely when unbound. I found this while `/simplify` was checking my work for duplication, not by looking for it.

The two credentials were never rivals — they are **different hops**:

| hop | credential | owner |
|---|---|---|
| browser ↔ web front end | the session cookie: opaque, server-side, revocable, sliding (`07` §1) | this PR |
| web front end ↔ backend | `sd1b`/`sd1u`, HMAC-signed | already on `main` |

So this PR still delivers what was needed, but **not for the stated reason**, and the reason mattered enough to correct in the module docstring rather than inherit. What *is* still open on the second hop, and is not mine to close: the **bound** shape names `chat_settings.id`, a legacy tenant key, while the target's tenant is `workspaces.id`.

## The session token

Opaque 256-bit random, only its SHA256 stored (`07` §1). It names nothing, so `authenticate_session` returns a user and deliberately says nothing about a workspace; tenancy is resolved per request from membership. `resolve_web_session` now delegates its authentication half here instead of keeping a second copy of the same three-column read — its shipped 14-test suite is unchanged and still green, which is the regression check on that refactor.

## Two things I found and did not paper over

**1. A user's workspace list is unreadable under the printed policies.** `p_tenant` on `workspace_members` is `workspace_id = app.tenant_id`: no tenant set reads empty, one set reads exactly one row. The empty case is indistinguishable from "signed in, no workspace yet" — so a user who owns three workspaces would be routed to first-run onboarding and told to create their first. A fail-open that looks like correct behaviour, and it would have looked like a writer bug.

`tenant_resolution.workspaces_for_user` detects the blindness (`row_security_active('workspace_members')`, which answers for this connection's own role) and raises `membership_list_unreadable` rather than returning `[]`. Verified against production: the `02` §7 door list is closed at **nine** SECURITY DEFINER functions and none enumerates memberships, so the fix is a §7-DDL amendment plus a migration, not an engineering call. Pinned green so it reddens deliberately when a door lands. **This blocks virgil's A5 (`GET /api/workspaces`) and the list half of A3.**

**2. The writers are complete but unreachable.** Nothing routes to them: the mounted web API is still the legacy router (virgil's D-C, unowned). Flagging rather than stubbing, per the dispatch.

## `sync_tx.py`, and why it is in this PR rather than after it

The target tier has two lanes. The async lane has `unit_of_work.py` — engine, transaction, GUC helper. The sync lane had **no substrate at all**, so this PR was growing one privately, a piece per writer: a transaction precondition in the workspace door, a savepoint in the identity door, four hand-written `SET LOCAL` statements in one and one in another.

The precondition is the case that decided it: **all three writers depend on the caller owning the commit, and exactly one of them checked.** Under autocommit the other two failed differently and neither usefully — `upsert_google_identity` died on a raw psycopg2 `25P01` from its savepoint, and `mint_session` silently was not atomic with the identity it documents composing with. The test fixtures flip `conn.autocommit`, so that is a live state.

`guc_pairs` is the shared half. `unit_of_work.apply_gucs` calls itself *"the ONE spelling of the security invariant"* and cannot be reused literally — it is `async`, renders SQLAlchemy `text()`, and binds `:v` where psycopg2 wants `%s` — but the **name list** is the half that drifts, and adding a fifth GUC to the schema must not be a change one lane gets and the other silently does not. The sync side spells it `SELECT set_config(name, value, true)` rather than `SET LOCAL x = %s`: portability, not taste — `SET` cannot take a bind parameter, so the old spelling worked only because psycopg2 interpolates client-side.

`constraint_violated` likewise moved into `_dbapi.py`, the module whose docstring already records what two independent finders of that exact question cost. This was the fourth.

## Verification

**22 integration tests, executed as `svc_ingress` against the real replayed target schema, under the printed RLS.** Not mocks — a writer that has only been unit-tested has not met the policy that decides whether its INSERT lands.

```
tests/scripts/test_identity_writers.py   22 passed
tests/scripts/test_tenant_resolution.py  14 passed   (shipped suite — regression check on the delegation)
full suite                             3765 passed, 8 skipped, rc 0
```

Both absences carry positive controls on the same connection in the same schema: an ownerless workspace raises at commit, and the same workspace INSERT lands when the tenant GUC names it and is refused when it names another id.

**Mutation-checked — five guards, five distinct tests:**

| mutant | test that reddens |
|---|---|
| delete the savepoint retry in `upsert_google_identity` | the concurrency test |
| drop the tenant from `guc_pairs` | the workspace-writer class (5 of 7) |
| disable the `row_security_active` blindness guard | the membership-list gap test |
| neuter `require_transaction` | the autocommit-refusal test |
| swallow the D35 collision catch | the colliding-email test |

**The concurrency test initially proved nothing and passed anyway.** With the retry deleted it stayed green: my ordering let the racer's own lookup hit, so the retry path was never entered. It now forces the collision — the winner inserts without committing, the racer's lookup genuinely misses, its INSERT blocks on `uq_identity_per_provider`, and only once `pg_blocking_pids` confirms it is waiting does the winner commit. Recorded because the passing version was indistinguishable from the working one.

**Live schema, read-only:** the five tables this PR writes match their migrations column-for-column in production — no plan-vs-live disagreement to report on my slice. Confirmed 26 tables in `public`, all seven `svc_*` roles present with `svc_ingress` a LOGIN role, PostgreSQL 17.10. Host asserted against the production endpoint before connecting, `default_transaction_read_only=on`, no writes attempted.

## Not in this PR, deliberately

- **`SUBJECT_STORAGE_AVAILABLE` stays `false`.** Its stated premises are dead — the target `users` has no `telegram_user_id` column at all and `user_identities` is live — but flipping it alone turns the callback's clean 501 into a thrown 500, because the constant gates a Next.js write leg (virgil's A2) that has no route to these writers yet. It flips with A2. `landing/` is untouched rather than edited mid-slice for a comment-only change.
- **No `ensure_personal_tenant`.** SEAM 2 in `web-signup.ts` describes an idempotent get-or-create door, which is the shape the legacy `chat_settings` world needed, where a chat id was both identity and tenant key. The plan rules that tenant minting is an explicit act at a door and must not happen in auth, middleware, or as a side effect of resolving — so `create_workspace` is a command a user invokes, not an `ensure_*` something calls on their behalf.
- **The proposed `relax_telegram_identity_not_null.sql` and the `webapp_auth.py` third cascade branch are not carried forward** — both patch a column and a validator the target schema does not have.
- **One `/simplify` finding skipped:** collapsing `create_workspace`'s two-spelling INSERT for the optional `tz`. Every shorter form either copies the schema's `'UTC'` default into app code or builds the column list with an f-string; the reviewer flagged it and then agreed leaving it is defensible. Recorded so the next reader does not re-derive it.

## Related

- #1030 — the adoption manifest describes a schema `051` renamed away (filed on the way; virgil's find, not in this path)
- #1028 — the audit that established the cutover applied 051–063 unattended


chrisrogers37 (commented):
**Verdict: Approve**

Verified by running, not by reading — set up a venv, connected to the shared local Postgres cluster, and ran the actual test suite live. Then went further and constructed my own mutant to check the RLS claim directly rather than trust the PR's own table.

## 1. The transaction on create-workspace — verified live, not just read

Read `test_an_ownerless_workspace_cannot_commit` first: it does a raw INSERT into `workspaces` alone (bypassing `create_workspace`, no owner row), then `conn.commit()` **inside** `pytest.raises(psycopg2.errors.CheckViolation)` — with the comment "the commit is where the deferred invariant is checked, it is the assertion, not a teardown step." That is exactly the right shape for a deferred constraint trigger. Then ran it for real:

```
tests/scripts/test_identity_writers.py::TestTheWorkspaceWriter  7 passed in 102.82s
tests/scripts/test_identity_writers.py (full file)              22 passed in 7.97s
```

Real Postgres 15, real replayed target schema, real `svc_ingress` role — not mocked, not skipped. The commit-fires-the-trigger claim holds.

## 2. "Tested against the real target schema" — checked which of the three claims this actually is

Read `conftest.py`'s `replay_advertised_stream`: it builds the schema via `scripts.advertised_ddl.build_stream` — the **same mechanism the actual CI replay gate uses** to prove F.2 migration files match the design docs — and applies it as `svc_migration` into an empty `public`. Not production, not a hand-maintained fixture that resembles the schema: a genuine replay of the advertised stream, which is the strongest of the three possible claims. Confirmed by successfully running against it myself.

## 3. RLS — the writers, not just the disclosed read-side gap

alex's disclosed blindness (`workspaces_for_user` under RLS) is a **read**-path issue and correctly flagged as such — not what you asked me to check. For the writers themselves: tests run through `as_user(db, "svc_ingress")` with a subject-gated `txn()` helper that asserts `current_user == 'svc_ingress'` before anything rides on it — genuinely testing as the production login, not a superuser.

I constructed my own mutant to confirm this is load-bearing rather than assumed: dropped the tenant GUC from `guc_pairs` (`pairs = [(TENANT_GUC, tenant_id)] → pairs = []`) and re-ran `TestTheWorkspaceWriter`. **3 of 7 tests failed** with `psycopg2.errors.InsufficientPrivilege — new row violates row-level security policy for table "workspaces"` — confirming the writer genuinely depends on RLS being correctly gated, not passing by coincidence.

**One discrepancy worth flagging precisely because of how this review was dispatched:** the PR's own mutation table claims this exact mutant (drop the tenant from `guc_pairs`) reddens "the workspace-writer class (5 of 7)." My reproduction of that exact mutant reddened 3 of 7 — `test_create_workspace_commits_with_its_owner_and_audit_trail`, `test_an_invalid_timezone_is_refused_by_the_schema`, `test_a_named_timezone_is_stored`. The other 4 are unaffected by design: two are raw-SQL positive controls that never call `create_workspace` at all, and two (autocommit refusal, blank-name/missing-owner) fail their own validation before the GUC-setting code is ever reached. The mutant IS genuinely caught — this doesn't change the recommendation — but the self-reported count in the PR body doesn't match what running it produces, and that's exactly the kind of thing this dispatch asked me not to inherit.

## 4. Session tokens

- `secrets.token_urlsafe(32)` — Python's CSPRNG (`os.urandom`-backed), not `random`. Correct.
- Only `session_token_hash()` (SHA256 hex) is ever stored; the raw token exists in exactly the two places the docstring claims — confirmed by grep, nothing outside `web_sessions.py` even imports `MintedSession` yet, since nothing routes to these writers.
- `revoke_session`: `UPDATE session_tokens SET revoked_at = now() WHERE token_hash = %s AND revoked_at IS NULL` — genuinely marks the row, not a cookie-only operation, and idempotent by design (second call returns `rowcount == 0` rather than erroring).
- Forward-looking, not a blocker: `MintedSession` is a plain `frozen=True` dataclass carrying the raw `token` field. Dataclasses auto-generate `__repr__` including every field, so once this is wired to a real sign-in route, whoever writes that code needs to know not to log the object directly (`logger.info(f"minted: {session}")` would leak the raw token via `__repr__`). Nothing today does this — there's no caller yet — but it's the kind of thing worth a one-line docstring warning before this gets a route.

## 5. Legacy schema

Clean. Grepped every added line in the diff for schema-qualified `legacy.` access: zero hits. The three prose mentions of "legacy" are all docstring comments about `chat_settings.id` being a legacy *tenant key* concept, not database access.

## 6. Issue #1030

Confirmed rather than inherited: grepped `src/` for any reference to the adoption manifest — zero hits. Nothing in the application code path reads it, matching the claim that #1030's 23 hard failures are a migration-tooling concern this PR's runtime code never touches.

## What I did not do

Did not independently re-run the full 3765-test repo suite or `test_tenant_resolution.py`'s 14-test regression suite — time-bounded; the file under direct review (`test_identity_writers.py`, all 22 tests) got the full live treatment plus an independent mutant, which is what the three writers' correctness actually rests on. Did not verify the `/auth/google` correction claim about `sd1u`/`sd1b` beyond reading `webapp_auth.py` — it's accurate as read, but I didn't test it live since it's pre-existing code this PR doesn't touch.

Recommend approving as-is. The mutation-count discrepancy in section 3 is worth a one-line fix to the PR body/commit message for accuracy, not a blocker to merge.

chrisrogers37:
## Correction: the tenant-GUC mutant kills **3 of 7**, not 5. The reviewer's number is right.

The review flagged that my PR body claimed 5 of 7 tests redden for the `guc_pairs` tenant-drop mutant, while a re-run got 3 of 7. **3 is correct.** I have reproduced both numbers on demand and the cause is mine, not a difference in what was counted.

### Both numbers, reproduced

On merged `main` (`feccc25`), mutating `sync_tx.guc_pairs` to `pairs = []`, scope `TestTheWorkspaceWriter`:

```
3 failed, 4 passed
  test_create_workspace_commits_with_its_owner_and_audit_trail
  test_an_invalid_timezone_is_refused_by_the_schema
  test_a_named_timezone_is_stored
```

Same mutant, with **one** line of the test helper reverted to its earlier shape — `_claim` calling `sync_tx.apply_gucs` instead of writing raw `SET LOCAL`:

```
5 failed, 2 passed
  ... the three above, plus:
  test_an_ownerless_workspace_cannot_commit
  test_rls_refuses_a_workspace_that_is_not_the_claimed_tenant
```

### What happened

I took the 5-of-7 measurement while `_claim` still routed through `apply_gucs`. Later in the same session I deliberately rewrote `_claim` to raw SQL — **specifically so the positive controls would not reach the database through the helper they are meant to be independent of** — and never re-took the measurement. The pre-decoupling number went into the PR body describing post-decoupling tests.

The irony is the useful part: the change that made those two tests honest is exactly what removed them from the mutant's kill list, and my report then cited the dishonest number as evidence of rigour.

### It is not a coverage gap, and here is the enumeration

The two tests that stopped reddening are the **positive controls**. Their job is to prove the *database* refuses an ownerless workspace and a foreign tenant — a fact that must hold regardless of whether my GUC helper works. A control that reddens when `guc_pairs` breaks is measuring the helper as well as the database, which is the one reading a positive control must never produce: a broken helper would have looked like "the trigger fired".

Every test that actually reaches `create_workspace`'s GUC path reddens. All 7, classified:

| test | reaches the GUC path? |
|---|---|
| `test_create_workspace_commits_with_its_owner_and_audit_trail` | yes — **reddens** |
| `test_an_invalid_timezone_is_refused_by_the_schema` | yes — **reddens** |
| `test_a_named_timezone_is_stored` | yes — **reddens** |
| `test_an_ownerless_workspace_cannot_commit` | no — raw-SQL positive control |
| `test_rls_refuses_a_workspace_that_is_not_the_claimed_tenant` | no — raw-SQL positive control |
| `test_an_autocommit_connection_is_refused_by_name` | no — `require_transaction` raises before any SQL |
| `test_a_blank_name_and_a_missing_owner_are_refused` | no — input guard raises before any SQL |

3 reachable, 3 killed. The other 4 are structurally incapable of reddening for this mutant and should be.

### The transferable lesson

A mutation count is a measurement of a *specific tree*. I mutated, counted, then changed the tests, then reported the old count — the same class as running a benchmark before the fix and quoting it after. **Re-take every mutation measurement after the last edit to either the code or the tests**, and say which commit it was taken at. The number in the body should have read `3 of 7 at <sha>`.

No code change needed. The other four mutants in that table were re-verified against merged `main` and still hold.

chrisrogers37:
## Which read hit the RLS gap: the **cross-tenant** one. Answer from this PR's own code, plus a measurement.

Posting here rather than in chat because this is blocking a schema decision and needs to be addressable. Answered independently — I have not been told what any other reviewer concluded.

### The actual call

`src/services/target/tenant_resolution.py::workspaces_for_user(conn, user_id)`, at `feccc25`. Its entire body issues two statements:

```python
cur.execute("SELECT row_security_active('workspace_members')")
# → if true it RAISES membership_list_unreadable and stops

cur.execute(
    "SELECT workspace_id, role FROM workspace_members"
    " WHERE user_id = %s ORDER BY created_at, workspace_id",
    (str(user_id),),
)
```

Keyed on `user_id`. Spans tenants by nature.

### Was a tenant claimed? **No.**

That function sets no GUC at all. Mechanically, not from memory — every tenant claim and every `workspace_members` read in the module, by owning function:

```
tenant claim at line 149        -> inside authorize_member()
workspace_members read line 151 -> inside authorize_member()
workspace_members read line 212 -> inside workspaces_for_user()
```

**One claim in the whole module, and it is in the other function.** `authorize_member` is the tenant-scoped read — it claims the workspace, then selects `WHERE workspace_id = %s AND user_id = %s`. That one works, and is not what the finding is about.

### A correction to my own original report

`workspaces_for_user` refuses *before* running its second query. So what I originally measured was "the policy is active for this role"; that the read comes back empty I **inferred** from the policy text. Sound, but inference — and the question is what made me go and run it.

### Measured — `svc_ingress`, replayed target schema, one user owning **two** workspaces

Two matters: with one workspace a partial answer is indistinguishable from a complete one.

| read | tenant claimed | rows |
|---|---|---|
| `WHERE user_id = ?` | none | **0** ← the call in question |
| `WHERE user_id = ?` | ws1 | **1 of 2** |
| `WHERE workspace_id = ws1` | ws1 | 2 — correct |
| `WHERE workspace_id = ws1` | none | 0 |
| `WHERE user_id = ?` | RLS not active | 2 — the true answer |

Tenant-scoped read works. Cross-tenant read is empty. The inference held.

### The row that should decide the migration, and it is not one of the two options as framed

**By-user *with* a claim returns 1 of 2 — partial, not empty.** That is strictly worse. An empty list is visibly wrong to anyone who checks a user who has workspaces; a one-of-two list is well-formed, plausible, and undetectable downstream. The app would show a two-workspace owner one workspace and offer to create their first.

So **the tenant claim cannot be the fix for the cross-tenant read**: the question spans tenants and a claim can only ever name one. Supplying it converts a detectable failure into an undetectable one. That is precisely why the shipped function refuses on `row_security_active` rather than returning what it got.

**A door is the right shape for the cross-tenant read.** Nothing in this PR is parallel to one — `workspaces_for_user` is a typed refusal plus an owner-connection read, not a door. When a real door lands, this guard should call it and the refusal disappears; the test is pinned green so it reddens on that day.

### If another reviewer reached a different answer

The likeliest mechanism is checkable in one grep rather than by argument: `workspaces_for_user` and `authorize_member` sit ~60 lines apart in the same file and **only the second claims a tenant**. Reading the module and landing on `authorize_member` yields the tenant-scoped answer; landing on `workspaces_for_user` yields this one. The discriminator is which function name the answer names — not which conclusion reads better.

PR #1032, opened 2026-08-25
feat(web): replace the front end on the target schema — Google sign-in, workspaces, invitations (#1015)

### LINKED ISSUE #1015: Aug 2 design plan does not scope the front-end surface or the BFF-to-backend credential for non-Telegram identity

chrisrogers37:
Closes #1015.

## ⚠️ MERGE IS GATED ON THE TARGET ROUTER LANDING

**Do not merge this into a world where P6–P9 have nothing to call.** This PR lands
**after or with** astrid's target-router PR. The *review* does not need to wait — nothing
in the built half depends on hers — but the *merge* does.

---

## The cutover retired the blocker this surface was designed around

The earlier screen designs ended on one: `chat_settings.telegram_chat_id` is NOT NULL
UNIQUE, so a workspace could not exist without a Telegram group. That was true of the
**legacy** schema under the patch route then being considered. It is not true now.

| Designed around | Target schema (051–063, applied 02:20:14Z) |
|---|---|
| `chat_settings.telegram_chat_id` NOT NULL UNIQUE | There is no `chat_settings`. `workspaces` **has no chat id at all**; Telegram is a row in `channel_bindings`. A Telegram-less workspace is the default shape. |
| `users.telegram_user_id` NOT NULL | `users` has **no telegram column**. |
| `SUBJECT_STORAGE_AVAILABLE = false` | `user_identities` exists — `provider CHECK IN ('telegram','google')`, `UNIQUE (provider, external_id)`. A home for the `sub`. |

The constant's own docstring said *"flip it in the same change that lands the storage."*
The storage landed, so it flips, and `googleSigninAvailable()` follows it.

## What is BUILT — P1–P5 and P10

| Route | Note |
|---|---|
| `/login` | Google only. Also handles the empty-card case: `GoogleLoginButton` self-hides on an unregistered origin — survivable beside the Telegram widget, a silent dead end without it. |
| `/auth/google/callback` | The 501 refusal becomes the sign-in write leg. Identity keys on `sub`, never email. |
| `/auth/error` | unchanged |
| `/welcome` | Now **asks for a workspace name**, reversing its previous deliberate refusal to ask. Lazy minting was right when a tenant was keyed on a chat id and there was nothing a person could supply; a workspace has a NAME, so asking **is** the explicit provisioning door the plan requires. |
| `/workspaces` | Replaces `/instances`, which listed Telegram **groups** — a user without Telegram saw an empty list and no way to change it. |
| `/join/[token]` | Plus `/join/[token]/start`, which survives the round trip to Google in a cookie rather than a `?next=` parameter (a `next` is an attacker-set redirect target; a token cookie cannot name a destination). |

Handlers: `/api/workspaces`, `/api/workspaces/[id]/select`, `/api/invitations/[token]/accept`,
and `/api/auth/session` + `/api/auth/logout` rewritten onto the opaque token.

**Deleted, not wrapped:** the Telegram login widget, `lib/auth.ts`, `lib/backend.ts`,
`/api/auth/telegram`, `/api/instances`, `/instances`, `/webapp/instances`,
`/dashboard/setup`, and `WEB_SIGNUP_ENABLED`. The flag goes because with the widget
deleted its "off" state restores nothing — it would mean *no way to sign in at all*,
strictly worse than "on". It had stopped being a switch.

**The session is a replacement, not an edit.** It was a self-contained HS256 JWT carrying
`{userId: number, activeChatId: number|null}`. `session_tokens` specifies an opaque value
whose SHA256 is a row, with `revoked_at` and a sliding expiry. A JWT cannot be revoked
(nothing to mark) and cannot hold a UUID. Every session minted under the old shape names a
user in `legacy` the target cannot resolve — those sessions already point at nothing.
The cookie **name** is deliberately unchanged so a stale JWT fails to resolve and is
cleared, rather than sitting in the browser under a new name, still being sent.

## What is HELD — P6–P9 — and why

Dashboard, media, calendar and settings **make their real calls to the target router**. Not
stubs, not fixtures: the actual request, to the route that does not exist yet, which returns
unreachable. Each screen then renders `RouterUnavailable`, which says:

> **Your dashboard is not available yet.**
> Nothing is wrong with your account. This part of Storydump is being connected — check back shortly.

Deliberately not styled as an empty state and not as an alarm. "You have no workspaces" and
"we could not find out whether you have workspaces" send a reader to opposite actions, and
only one is something they can act on.

**This also closes a pre-existing fail-open, which I think matters more than the hold.**
`backendFetchJson` returned `null` on failure and every dashboard page wrote
`data?.field ?? 0`, so an unreachable backend rendered as a dashboard of **zeros** — "you
have posted nothing", stated confidently to someone whose library is full. That is the
failure direction nobody reports, because it reads as good news. `workspaceFetch` returns a
result the caller must inspect.

The BFF proxy also **drops its per-request membership re-check**. It existed only because a
JWT could not be revoked, and it failed **open** by design ("backend unreachable — fall
through") in exactly the case it was written for. The target authorizes against
`workspace_members` on every call.

## Known gap for the reviewer — the header cannot name the current workspace

Found while rendering, and **not built**. On a multi-brand product a header that cannot name
the current workspace risks posting to the wrong one: two workspaces in two tabs and nothing
on screen says which brand you are about to post to.

**It is a complete state, not an interrupted write** — and that is checkable rather than
asserted. `workspaceName?: string` is declared on `DashboardHeader` with a
`|| "Switch workspace"` fallback (`header.tsx:32,36,69`); `layout.tsx:52` simply never passes
it. Nothing is half-written; the prop is unwired. The remedy is for the layout to resolve
the name and pass it, falling back to the current label when the lookup fails.

## Verification — and note what is NOT verifying this

**CI is Python-only. It never lints or builds `landing/`.** A green CI on this PR proves
close to nothing about it. The evidence is local:

- `next build` — green
- `tsc --noEmit` — clean (no source errors)
- `vitest` — **35 passed / 35** across 5 files. 23 before; 2 replaced (the storage-boundary
  tests, which the previous version asked in as many words to be updated deliberately when
  the schema landed), 4 + 5 + 5 added.
- **Mutation-verified, both new invariants:** making an unreachable router return
  `{ok: true, data: {}}` reddens **2** tests; collapsing session "cannot ask" into "not
  signed in" reddens **1**. Suite restores to 35/35 after each.
- Six screens rendered at desktop (1280) and mobile (390) from a real build against a stub
  implementing the target contract.

Two near-misses in my own work, disclosed because both looked green: a replace ran to
end-of-file and **deleted 12 real tests** including the domain-separation and tamper pins
from the #1015 review round (the run reported "15 passed" — green because they were gone;
the tell was the denominator, 23 → 15), and I overwrote a page with the literal string
`/dev/null`. Both restored from git; the test edit was redone bounded with a count assertion.

## Screens

| | Desktop | Mobile |
|---|---|---|
| **Sign in** | login desktop | login mobile |
| **First run** | welcome desktop | welcome mobile |
| **Workspaces** | workspaces desktop | workspaces mobile |
| **Invitation, signed out** | join out desktop | join out mobile |
| **Invitation, signed in** | join in desktop | join in mobile |
| **Dashboard, held** | held desktop | held mobile |

Assets: branch `design-crawl/2026-08-24-target-front-end`, not merged.

## Bounds

- **I did not read the live database.** No `DATABASE_URL` here and the Railway CLI is
  unlinked. Schema facts are read from the migration files on disk — ground truth about
  *shape* regardless of application — plus #1028's reported production verification for the
  claim that 051–063 **applied**. That one claim I take on report.
- The stub is my reading of the contract, so the screens prove the pages render correctly
  against **what I asked for**, not against what the router will send. `dashboard-payloads.ts`
  is marked provisional for the same reason.
- `#1017`'s `proposed/relax_telegram_identity_not_null.sql` patches a column the target does
  not have, and its third `webapp_auth` cascade branch is the compatibility layer this ruling
  removes. Both are backend surface and untouched here — flagged, not fixed.
- `adoption_manifest.json` still records `required_through: 45` against a production at 063.
  Routed to alex separately.

---

## Second known dependency — the membership list is unreadable under RLS

Alex disclosed this alongside the three writers in #1031. Recording it here so a reviewer
meets it as a known dependency rather than discovering it as a defect.

**It does not block anything in this PR.** A7 (`/api/workspaces/[id]/members`) and P9
(`/dashboard/members`) are **not built** — `find landing/src -ipath '*member*'` returns
nothing on this branch. This PR ships three handlers: `workspaces`, `select`, and
`invitations/accept`. So the RLS problem costs nothing today; it is a precondition for the
follow-up that builds P9, not an outstanding fault here.

**And when P9 is built, my existing defence does not cover it.** Worth stating plainly,
because the shape is the one this PR already closed once, one layer up:

`targetFetch` distinguishes *unreachable* from *empty*. RLS refusing to return rows is
**neither** — it is `200 OK` with `[]`. The transport succeeded, so the result is
`{ok: true, data: []}` and every guard in this PR passes it through. A members screen built
naively on top would render "no members" to an owner looking at their own workspace, which
is the same confident-wrong-answer failure as the dashboard of zeros.

**The schema supplies the discriminator, so this does not need a heuristic.** From migration
053, verbatim:

> *"Together: exactly one owner at every commit — `uq_members_one_owner` is 'at most one', the
> trigger pair is 'at least one' (on every path: creation, demotion, removal)."*

A workspace that exists therefore has **at least one member, always, structurally enforced**
— and the caller reading the list is necessarily one of them, or they could not have reached
the route. So **an empty members list is never legitimate**: zero members is impossible by
construction, which makes `[]` on that endpoint *always* a fault rather than an ambiguous
result. That is a checkable invariant, not a judgement call, and it is what P9 should assert
rather than rendering an empty state.

Not fixing it here — it is a database policy question and it is alex's disclosure to route.
Flagged so that whoever builds P9 does not reintroduce the fail-open by treating `[]` as an
empty state.

## Reword — the GitGuardian finding was a false positive

`session-guards.test.ts` tripped the "Generic Password" detector on
`"<uuid>/../other"`, a **path-traversal attack string** in a junk array asserted `toBe(false)`,
two lines below `"../../etc/passwd"`. A 36-character high-entropy string followed by a slash
reads as a URL with embedded credentials. No credential exists, and none was rotated.

Fixed by **rewording, not by an ignore-file entry**: an ignore entry suppresses the detector
on this path permanently, so the next real finding in this file would be silent.

All sibling fixtures were reworded, not only the flagged line — the dashless variant on the
next line was the same shape. Every junk case is now derived from the id asserted to pass on
its own, which isolates the mutation rather than the base; three cases were added. Verified
at 35/35 with an unchanged denominator, and mutation-checked: stripping the regex anchors —
precisely what would admit `<uuid>/../other` — reddens the test.


chrisrogers37 (commented):
**Request Changes**

*(comment-only — same-identity blocks Request Changes: `GraphQL: Review Can not request changes on your own pull request`)*

Reviewed by running, not reading: checked out `674528e`, ran real greps against the full source tree (not the diff), read every `workspaceFetch` call site, downloaded and looked at all 12 crawl-branch screenshots, and traced the actual cookie lifecycle end to end. Merge is gated on the router per the PR body — this is a review-only verdict, nothing here blocks that gate from clearing.

The headline claims hold up. Two concrete, small, fixable bugs need addressing first.

## 1. The "closed" fail-open is still open on the main dashboard — `(dashboard)/dashboard/page.tsx:35-39,64,67`

Three parallel fetches, one guard:

```ts
const [analyticsResult, categoriesResult, historyResult] = await Promise.all([...]);
if (!analyticsResult.ok) return <RouterUnavailable what="Your dashboard" />;
const categories = categoriesResult.ok ? categoriesResult.data : null;
const history = historyResult.ok ? historyResult.data : null;
...
<CategoryBreakdown categories={categories?.categories ?? []} />
<RecentActivity items={history?.items ?? []} />
```

If `categoriesResult` or `historyResult` fails independently while `analyticsResult` succeeds, the page doesn't render `RouterUnavailable` — it silently renders an empty category breakdown and "no recent activity." That's the exact "dashboard of zeros" pattern the PR (correctly) claims to have closed, just scoped down to 2 of this page's 3 dependencies instead of all of them.

This isn't a design disagreement — `media/calendar/page.tsx:35-36` (same PR) gates on all three of *its* results: `if (!historyResult.ok || !queueResult.ok || !scheduleResult.ok)`. `settings/page.tsx:25` does the same for its two. `dashboard/page.tsx` is the one file out of seven checked that doesn't, and the comment defending the gap ("a header over three empty panels states a fact about the account that we did not establish") applies just as much to two silently-empty panels as to three.

Fix: `if (!analyticsResult.ok || !categoriesResult.ok || !historyResult.ok) return <RouterUnavailable .../>`.

## 2. Dangling "Setup Wizard" nav link — reachable by every first-run user

`(dashboard)/dashboard/setup/page.tsx` is hard-deleted (+0/-49). But `components/dashboard/sidebar.tsx` — untouched by this PR — still carries it:

```ts
const navItems = [..., { href: "/dashboard/setup", label: "Setup Wizard", icon: Rocket }, ...];
export function Sidebar({ mobile, showSetupWizard = true }: {...}) { ... }
```

And `layout.tsx:44-46` now computes that flag from a *real* call: `Boolean(initResult?.ok && !initResult.data.setup_state?.onboarding_completed)`. A brand-new workspace — the thing this PR's own Google → `/welcome` → create-workspace funnel produces — has `onboarding_completed: false` by construction. So the link shows, by default, for every new signup, the moment the router is live. Click it and you land on `(dashboard)/dashboard/not-found.tsx` ("Page not found... Back to Overview") — a well-built card, but still a dead end where a working wizard used to be, and `layout.tsx:37-40`'s own comment ("otherwise it points at a route that redirects") describes the *old*, now-deleted page's behavior, not what happens now.

Two small stragglers confirm the cleanup wasn't quite complete: `dashboard/setup/loading.tsx` survives with its `page.tsx` gone, and `components/dashboard/no-tenant-empty-state.tsx` is now fully orphaned (zero references anywhere — confirmed by grep). Neither is harmful on its own, but both are exactly the kind of leftover that should be gone in a "replace, don't wrap" PR.

Fix: either gate the sidebar/header `showSetupWizard` default false and drop the entry entirely for now, or ship a minimal `/dashboard/setup` that renders something (even `RouterUnavailable`) instead of 404ing. Also worth deleting the two orphaned files while in there.

## 3. "A stale JWT... is cleared" — not what the code does

Grepped every reference to `SESSION_COOKIE` across `landing/src`. It is deleted in exactly one place: `api/auth/logout/route.ts`. Middleware (`middleware.ts:36`) is a presence-only gate by design and never resolves or clears anything — that reasoning is sound and well-argued. `resolveSessionToken` correctly turns a stale JWT's 401/403 into `null`, and the page redirects to `/login` — that part is right and fails closed. But nothing then clears the cookie. It sits in the browser and gets silently re-sent (and re-resolved, and re-rejected) on every request to a gated route until an explicit sign-out or a fresh successful Google login overwrites it by name.

Practically this doesn't hurt anyone — the failure direction is safe — but the claim is checkable and doesn't hold: it's not "cleared," it's "ignored forever, quietly." And per the CHANGELOG, migrations 051-063 already applied to production, so every currently-logged-in user has exactly this cookie right now. Worth fixing the comment (in `session.ts:26-28` and the CHANGELOG) rather than leaving future-you trusting a claim that isn't true, and maybe worth actually clearing it in the redirect path since you're already there.

## What I verified and holds up

- **Replacement, not shim.** Zero surviving live references (import or route string) to `TelegramLoginButton`, `lib/auth.ts`, `lib/backend.ts`, `/api/auth/telegram`, `/api/instances*`, `/webapp/instances`, `WEB_SIGNUP_ENABLED`, `backendFetchJson`, or `hasTenant` anywhere in `landing/src` — checked the whole tree, not the diff. `(marketing)/setup/*` looked suspicious since it wasn't in the changed-files list, but it's an unrelated static onboarding-docs section (Instagram/Drive/Telegram connection guide), not part of the auth surface — false alarm on my end, not yours.
- **P6-P9 make real calls.** `media/page.tsx`, `media/reuse/page.tsx`, `media/dead-content/page.tsx`, `media/calendar/page.tsx` all call the live `workspaceFetch` and correctly render `RouterUnavailable` (well-built component — deliberately not styled as an error or empty state) on failure. No stubs, no mocks.
- **Known gap #5 (header can't name workspace) is genuinely a gap, not a partial write.** `workspaceName?: string` is declared on `DashboardHeader` with a `|| "Switch workspace"` fallback; `layout.tsx:52` just never passes it. Confirmed visually — the `06-dashboard-held` screenshots literally show "Switch workspace" in the header. Your own test account has 2 workspaces (`Northside Coffee`, `Rivers End Bakery` — visible in `03-workspaces`), so this is the live scenario, not hypothetical. My judgment: real, correctly disclosed, degrades to a truthful-if-uninformative label rather than a wrong one. Doesn't need to block the router gate, but land it before anyone actually manages 2+ workspaces day to day.
- **GitGuardian is a confirmed false positive.** Pulled `session-guards.test.ts:18-29,40-50` myself — the flagged content is a `junk` rejection-fixture array (`${VALID_ID}/../other` beside `"../../etc/passwd"`), asserted `toBe(false)`. Matches your account exactly. Not mine to chase further.
- **Second known dependency (members/RLS) is correctly scoped out.** `find landing/src -ipath '*member*'` returns nothing on this branch — P9 genuinely isn't built here, so it's a real non-issue for this PR.
- **Screens.** Looked at all 6 desktop + 2 mobile (dashboard-held, welcome). Clean, responsive, no visual bugs. The "Switch workspace" fallback (finding above) is visible but doesn't read as broken.
- **Checks are fresh.** `headRefOid` matches the checked-out commit (`674528e`), `mergeable: MERGEABLE` not `CONFLICTING` — the green `Test` check describes this code, not a stale base.

## What I couldn't verify locally

`vitest` in `landing/` — `npx vitest run` picked up a mismatched cached version and failed to resolve `vitest/config`; `npm ci --ignore-scripts` (isolated env, avoiding ambient secrets in an external checkout) timed out after 2 minutes and I killed it rather than burn the review clock on tooling. I'm relying on CI's `Test: SUCCESS` at this exact head SHA plus my own direct read of `session-guards.test.ts`'s assertions (which do meaningfully exercise 401/403→null vs. network-failure→throws, not decoys) — didn't get to independently re-run the "mutation reddens 2/1 tests" claims myself. Disclosing rather than implying I ran the full suite.

---

Two small, targeted fixes (#1, #2) plus a comment/CHANGELOG correction (#3) — none of this needs a redesign, and the architecture underneath is sound. Happy to re-look once those land.

chrisrogers37 (commented):
**Approve**

Short pass at `a4b8d5d`, confirming all three — not a fresh review.

**1. Dashboard fail-open — fixed.** `dashboard/page.tsx:35-37` now gates on `!analyticsResult.ok || !categoriesResult.ok || !historyResult.ok`. The comment cites the exact inconsistency (calendar/settings already did it right) rather than just asserting a fix.

**2. Setup Wizard — fixed, extra deletions verified safe.** Re-grepped `showSetupWizard`, `NoTenantEmptyState`, `/dashboard/setup`, `Rocket`, `InitResponse` across the whole tree (not the diff): zero live stragglers, only historical comments. The one thing worth checking twice — did deleting "the init call" take `settings/page.tsx`'s own init call with it? No: `dashboard-payloads.ts`'s `InitResponse` type is untouched, and `settings/page.tsx` (the type's other, independent `workspaceFetch<InitResponse>("init", ...)` consumer) shows zero diff in this commit. The deletion was scoped to exactly the one dead copy in `layout.tsx`.

**3. JWT clearing — fixed, "pinned both directions" checked empirically.** `isLegacyJwt` is a pure function (`eyJ` prefix + exactly 2 dots, no network call — consistent with middleware's existing presence-only design). I extracted it verbatim from the committed source and ran it, plus three mutations (`&&`→`||`, drop-prefix, drop-arity), directly in Node against the real fixtures in `legacy-jwt.test.ts` — no vitest needed since the function has no framework dependency. Real function: all pass. Every mutation: multiple failures, including on the two fixtures built specifically to isolate each half of the AND (`abc.def.ghi` — right arity, wrong prefix; `nope.nope.nope` — same). Re-grepped `SESSION_COOKIE`: exactly two delete sites now, `middleware.ts` (new) and `logout/route.ts` (pre-existing) — matches the claim. `session.ts:26-28` and the CHANGELOG are both corrected, and both say so explicitly ("Correction from review") rather than quietly rewriting.

**Couldn't verify: the lint claim** ("8 problems, all outside my diff"). CI's `Lint` check is Python-only — confirmed from the actual job log, no eslint output in it at all — consistent with what the PR body already disclosed about CI not touching `landing/`. Tried running eslint myself twice (isolated env): `node_modules/.bin` doesn't have it locally, and two `npm install --ignore-scripts` attempts both timed out against what's a slow registry from this host. Not disputing the claim — just flagging I couldn't independently confirm it. Tooling gap on my end, not a code concern.

Verdict: **Approve**. Merge stays gated on astrid's router and the GitGuardian dashboard action, neither mine to clear.

chrisrogers37:
**Resolution note** — no code change. This comment exists because the two reviews above cannot express their own order.

Both reviews are the same reviewer's, and they are a **sequence, not a split verdict**:

| when (UTC) | what | at |
|---|---|---|
| 01:48:30 | **Request Changes** — 3 items | `674528e` |
| 02:32:36 | fix commit, all 3 items | `a4b8d5d` → rebased to `eb4f495` |
| 02:41:50 | **Approve** — *"confirming all three"* | `a4b8d5d` |

Under the shared PAT both land as `COMMENTED`, so GitHub renders them side by side with the verdict living only in the body text. Read as a *set* rather than a *sequence*, that looks like an unresolved split. It isn't — the Approve is a re-review of the fix.

State at HEAD (`75cf660`), verified by reading the source rather than trusting the commit message:

1. **Fail-open** — `(dashboard)/dashboard/page.tsx:44` gates on all three results: `!analyticsResult.ok || !categoriesResult.ok || !historyResult.ok`.
2. **Setup Wizard** — the nav entry, the `showSetupWizard` flag, and the `init` call in `layout.tsx` that computed it are all gone; `dashboard/setup/` was removed whole, so `loading.tsx` went with it, and `no-tenant-empty-state.tsx` is deleted. Grepping `dashboard/setup`, `showSetupWizard`, `NoTenantEmptyState`, `Rocket` across `landing/src`: **0 live references** — 2 matches total, both prose in `sidebar.tsx:26,29` explaining the removal.
3. **JWT clearing** — `middleware.ts:63` now deletes `SESSION_COOKIE` (two delete sites total, matching the claim); `session.ts:29-36` carries the correction explicitly rather than quietly editing the false claim.

The branch was rebased after the approval, so `674528e` and `a4b8d5d` are dangling. **The approved content survived**: `dashboard/page.tsx` is byte-identical between `a4b8d5d` and HEAD (md5 `9e8945ca…`).

Nothing outstanding from the review. Merge remains gated on the target router, which is not mine to clear.

Issue #1033, opened 2026-08-25
[feature] Web approvals surface — a Google-only workspace currently has no way to approve a post


chrisrogers37:
## What

A web approvals surface, so a Google-only workspace can review a post before it publishes.

Deferred deliberately on 2026-08-24 in favour of the interim (`workspaces.approval_mode = 'auto'`), to unblock a functional end-to-end system and a fresh stress test. This issue is the proper answer, filed at the moment of deferral rather than left implicit.

## Why it exists

On the target schema, a workspace has **no Telegram by default** — `workspaces` carries no chat id at all, and Telegram is a row in `channel_bindings`. That is the intended shape, not an edge case.

The legacy approval path was the Telegram card. So a user who signs in with Google and never connects Telegram has **no way to approve anything**, and the core loop does not close for them. The interim closes it by removing the approval step, not by providing one.

## Scope

- The screen (`P11` in the front-end surface, chrisrogers37/storydump#1015) — a list of pending intents with approve / reject per item, and the account each would post to.
- Backed by `post_intents` + `post_intent_transitions`, which already exist (migration `055`) and already carry the state machine. `post_intent_transitions` was seeded with a 27-row `(from_state, to_state)` reference matrix by the cutover, so the legal transitions are defined.
- The API surface it needs, against the target router's command vocabulary rather than a bespoke route.
- Multi-account: `06-product-lifecycles.md` §Approvals states the payload carries `ig_account_id` + handle, so a workspace with four accounts must show four distinguishable items — "or one web list with an account column", which is this screen.

## The interim it replaces, and the exact trap

`workspaces.approval_mode` is defined in `scripts/migrations/053_identity_and_tenancy_tables.sql:136` as:

```sql
approval_mode  TEXT NOT NULL DEFAULT 'manual'
             CONSTRAINT ck_ws_approval_mode CHECK (approval_mode IN ('manual','auto'))
```

**The default is `'manual'`, not `'auto'`.** So the interim is not a config flip — a workspace created today gets `'manual'` and its posts wait for an approval that has no surface. Whatever sets `'auto'` has to be a deliberate change (the create-workspace path, or a default migration), and it must be **reverted** when this issue lands, or the approvals screen ships to workspaces that never route anything to it.

## Related

- Front-end surface and the `P11` slot: #1015
- Stress-test scope, which assumes the loop closes: #1029
[...]

PR #1035, opened 2026-08-25
feat(api): replace the legacy router with the target tier's API — reads as resources, commands as the closed vocabulary, API-hosted Google sign-in (#1028)

chrisrogers37:
## Summary

Replaces the legacy web router with the target tier's API — the server half of the #1015 contract (design: [#1015 comment](https://github.com/chrisrogers37/storydump/issues/1015#issuecomment-5402915246), [addendum](https://github.com/chrisrogers37/storydump/issues/1015#issuecomment-5403597149)). Full replacement, no shims, no dual path, per Chris's ruling on #1028.

**Reads are resources; writes are the closed `01` command vocabulary made literal; sign-in is hosted on the API.**

| Surface | What |
|---|---|
| `GET /auth/google` → `GET /auth/google/callback` → `POST /auth/signout` | `07` §1: anonymous `oauth_states` row + cookie-nonce double submit, server-side code exchange through the egress floor, `users`+`user_identities` upserted on `(google, sub)` (never email — D32/D35), opaque `session_tokens` session as `sd_session` (HttpOnly/Secure/Lax) or `Authorization: Bearer` |
| `GET /api/v1/me`, `GET /api/v1/workspaces` | the principal's own view — tenant-less, user-plane |
| `GET /api/v1/workspaces/{ws}` · `/members` · `/accounts` · `/sources` · `/bindings` · `/invitations` · `/intents?state=&limit=` · `/intents/{id}` | every one behind the ONE gate (`tenant_resolution.authorize_member`); non-member → **404, never 403** |
| `POST /api/v1/workspaces` | `create_workspace` — id pre-assigned server-side so the tenant claim precedes the insert; owner membership in the same transaction |
| `POST /api/v1/workspaces/{ws}/commands/{command}` | the one write route; `{command}` validated against the 24-name vocabulary; `Idempotency-Key` required, admitted into `command_dedup` in the SAME transaction as execution |
| `POST /api/v1/invitations/{token}/accept` | the `fn_invitation_accept` door |

**Interim, ratified by Chris (#1033):** a workspace created here starts in `approval_mode='auto'` (`workspaces.INTERIM_APPROVAL_MODE`; column default is `'manual'`) so the core loop closes without an approvals surface; the gate pins the value so #1033's revert is a deliberate test flip; `settings_change` can put a workspace back to `'manual'`.

Under it: `src/services/target/commands.py` (the port — `VOCABULARY` pinned set-equal to `01`'s list **by parsing the doc**, total role floor, `execute()` = gate → registry), executors for approve · skip · reject · mark_posted · cancel · sync_now · settings_change · pause/resume · rename · create_workspace, the not-built half answering `501 {"command": …}` by name (pinned set), `sessions.py` / `identity.py` / `workspaces.py` / `google_oidc.py`, and `tenant_resolution.py` ported in place to the tier's async shape.

**Deleted, not wrapped:** `routes/oauth.py`, `routes/onboarding/**`, the Mini App (`static/**`, `/webapp/onboarding`), `rate_limit.py` + the SlowAPI dependency, `utils/auth_monitor.py`, the zero-caller `sd1` web-token branch of `webapp_auth.py` + `WEB_TOKEN_SECRET`, the proposed migration patching a column the target schema lacks, and their tests. Ratchet baseline re-measured (−24 entries, all under the deleted router; every count unchanged).

## Why this shape

- One command route means the code holds ONE enum and a test asserts set-equality with `01`'s list in both directions — the vocabulary cannot grow silently. A per-command REST table would make the routes the vocabulary and the doc a description of them.
- W4 (Telegram inbound) will hand the same `Command` objects to the same `execute()`; this PR is its prerequisite, not its competitor. The dormant `POST /webhooks/telegram` stays mounted for it.
- Admission before execution, in one transaction: a refused command rolls its dedup row back (retry re-executes); a true replay is `200 {"outcome":"replayed"}` without running; same key + different body is 409. Measured in the gate: 3 dedup rows after 4 commands, the refused one's key free again.
- Pre-auth admission is the durable `rate_counters` `preauth_ip` scope (`05`: 30/min per attributed client IP), keyed AFTER the trusted-proxy walk — #726/#776's property survives the limiter it was proved on.
- The engine comes from `TARGET_DATABASE_URL` and nothing else. Unset → every data route 503s naming the variable and `/health` reports `target_database: false`. Never the settings-built URL (#1010's class).

## Two deviations from the design comment (both recorded on #1015)

1. **ID token verified by claims, not JWKS signature.** `iss`/`aud`/`exp`/`nonce`/`sub` on a token received directly from the token endpoint over the egress floor's TLS — OIDC Core §3.1.3.7 step 6 permits exactly this. Drops a second provider call, a key cache with rotation, and RSA code. One-function seam if reviewers want it back.
2. **`connect_account` / `reconnect_account` stay 501.** They are browser redirect flows through `ig_login_oauth`, not idempotent commands; they want a `GET …/accounts/connect` route, which is P8's shape.

## Finding — the membership list, and how it now answers

Under the production role `svc_ingress`, `workspace_members` and `workspaces` carry only the tenant-keyed `p_tenant` policies (`058:141-147`) — no user-plane read path — so "which workspaces am I in" cannot be read before a tenant is claimed. alex hit the same gap from the writer side (#1031); rajan confirmed it is ONE gap, and the door is proposed on #1015 (`fn_memberships_for_caller`, reading `app.actor_user_id` internally per rajan's refinement, plus the `(user_id)` index).

**Adopted here, from #1031:** the read **refuses by name rather than answering `[]`** — an empty list is indistinguishable from the greenfield's normal signed-in-with-no-workspace state, so a three-workspace owner would be sent to first-run onboarding. `workspaces.list_for_user` detects `row_security_active('workspace_members')` and raises `membership_list_unreadable`; `GET /api/v1/workspaces` answers **503** with that reason; `/me` still serves the user (it is the sign-in spine) with `workspaces: null` and the refusal in `degraded[]`. The gate asserts all three under `svc_ingress` with an owner-side positive control. virgil: P5 gets a 503 it can render ("cannot list your workspaces yet"), never an empty onboarding redirect.

Also worth someone with production access checking: if the deployed `TARGET_DATABASE_URL` role is the table owner rather than `svc_ingress`, RLS is bypassed entirely and every isolation claim here is untested in production.

## A fork, disclosed — not hidden and not resolved unilaterally

#1031 merged **sync DB-API** writers for the same three legs (`identity_provisioning`, `web_sessions`, `workspace_provisioning`, on `sync_tx`) while this PR carries **async** ones (`identity`, `sessions`, `workspaces`, on `unit_of_work`) — built in parallel on 08-24 when the writer split went unanswered. Both compile; both suites pass (his 21 + mine).

Which lane the API uses is not a style choice: the command route is admission + execution in **one** `unit_of_work` transaction, and a psycopg2 writer cannot join an asyncpg transaction. So the async set is the one the API can consume; the sync set has no caller in `src/`. What I did here: `tenant_resolution.py` stays ONE async module (its three sync-resolver tests moved to async equivalents), and nothing of alex's was deleted. **Named follow-up (alex + me, one PR):** port the sync lane's two better ideas — savepoint-based race recovery and D35 caught from the constraint via `_dbapi.constraint_violated` — into the async modules and delete the sync copies with `sync_tx`.

## Evidence

- `tests/scripts/test_web_router_x2_gate.py` — X.2's gate literally, through the real app, **as `svc_ingress`** on the replayed target schema (subject-gated: `current_user` asserted): sign in (only the provider exchange stubbed; oauth_states / session_tokens / rate_counters / command_dedup / audit triggers / intent guard all real) → `/me` → create workspace (201) → replay (200 replayed) → read back → seeded `awaiting_approval` intent listed → approve refused `manual_mode` (409) → `settings_change{api_publishing_enabled}` → approve **202 enqueued `publish_pipeline`** (row present, intent `approved`) → replay 200 / different body 409 → second cold user: 404 on read and on a command → signout → 401. **3 passed, 1 xfailed** (the pin), 28s.
- `tests/scripts/test_command_executors_gate.py` (`84fe54a`, on navi's review) — **every executor body executes** as `svc_ingress` through the real registry inside a real unit of work: `skip`, `reject`, `mark_posted`, `cancel`, `sync_now`, `pause_workspace`, `resume_workspace`, `rename_workspace` (the X.2 gate already ran `create_workspace`, `settings_change`, `approve`). Effects read back as the owner: state flips through the ledger trigger, the workspace-wide skip/reject locks and their `ON CONFLICT` refresh, R1's manual-mode debit (`daily_post_counts` count past a cap of 1, `cap_at_write` frozen at the first debit, a refused mark_posted's debit rolled back), `times_posted`/`last_posted_at`, the per-account recent lock, `cancel_requested` as an overlay, the demand-sync job's exact shape and its one-while-pending guard, pause/resume attribution, and the workspace-bound `not_found` for another tenant's intent. **18 passed; four mutants killed** (capped debit → 2 red, unfrozen `cap_at_write` → red, guard-less `sync_now` → red, never-refreshed skip lock → red). None of these executors talks to a provider, so nothing is excluded.
- `tests/scripts/test_tenant_resolution.py` (async port): 14/14 on the replayed schema.
- Unit: `tests/src/api` + `tests/src/services/target` + neighbours: **388 passed** (vocabulary-equals-doc both directions, floor total, 501 set pinned, OIDC claim refusals by name, cookie/bearer one credential, 404-not-403, admit-before-execute, every port refusal → status, CORS never `*`, legacy paths 404, no engine → 503 naming the variable, preauth key = attributed peer under a forged XFF).
- Found by the gate, not by unit tests: asyncpg refuses a bare bind inside `jsonb_build_object` (`IndeterminateDatatypeError`) — cast added in `approve`/`sync_now`; and `post_intents.state` defaults to `scheduled`, so the fixture seeds `awaiting_approval` explicitly.
- Full suite (`bae9bcb`, before the #1031 merge): 14 failed / 3612 passed / 8 skipped / 1 xfailed — twelve were a DNS drop on the host mid-run (`test_telegram_transport`, `test_egress_floor`: "Temporary failure in name resolution"; the egress floor resolves hosts for real; all 32 pass on rerun and on `main`), two were mine (the hash-writer scan listed the deleted Mini App file — fixed). After the merge: gate 3 + tenant-resolution 15 + #1031 writers 21 = **39 passed**, unit batch **520 passed**. **Final full suite on the merged tree (`51e9a4b`): 3648 passed, 8 skipped, 0 failed (pytest rc 0, 18m33s)** — the DNS-class failures did not recur, and the two hash-scan failures are fixed.
- `ruff check` / `ruff format --check`: clean. FC-2 ratchet: green against the re-measured baseline.
- `/simplify` (four lenses — reuse, simplification, efficiency, altitude) applied: `commands.ingest` owns refuse→admit→execute once; SQLSTATE mapping moved into the service tier (`invitations.accept`, `workspaces._write` via `_dbapi.driver_candidates`) and the HTTP layer maps closed reason vocabularies with status tables pinned total (unmapped → logged 500); `jobs.enqueue` is the one jobs INSERT; the session lookup slides in one statement and `apply_gucs` is one statement; `create_workspace` claims only when it assigns the id; the worker engine-url shim and the caller-less `resolve_web_session` are deleted. Skipped (worker posting path, own review): sharing `mark_posted`'s posted-effects and debit CTE with `publish_pipeline`/`publish_cap`; and the egress floor's blocking `getaddrinfo` on the API loop — pre-existing, filed as a follow-up.

## What Chris must do for this to serve

1. On the **web** Railway service: `TARGET_DATABASE_URL` (the target DB, as `svc_ingress` ideally — see finding), `GOOGLE_CLIENT_ID`, `GOOGLE_CLIENT_SECRET`, `OAUTH_REDIRECT_BASE_URL` (already there), `WEB_APP_URL` (the landing origin — the ONE CORS origin), optionally `SESSION_COOKIE_DOMAIN` (the shared registrable domain, if the landing host shares one — #1015 §8 Q1 still open).
2. Google Console: add `<OAUTH_REDIRECT_BASE_URL>/auth/google/callback` as an authorised redirect URI on the existing client (#1015 §8 Q2).
3. ~~The legacy Telegram bot's Mini App links now point at a route that does not exist.~~ **Resolved (`8585fc0`, navi's blocker 1):** those links are baked into buttons real users already hold — `build_webapp_button` writes a static URL into the message and a tap is client-side navigation that never contacts the bot, so `WORKER_IMPL=target` does not retire them (navi measured `legacy.user_interactions` written until 01:16Z on cutover day by exactly the commands that mint the button). `/webapp/onboarding` now **redirects (302) to `{WEB_APP_URL}/login`**, dropping the legacy `?chat_id=`, and answers **410** with a sentence when no front-end origin is configured — never a 404 (`src/api/routes/retired.py`, tests in `test_app_factory.py`). Nothing for Chris to do beyond setting `WEB_APP_URL`, which sign-in needs anyway.

## Follow-ups (named, not stubbed)

- `fn_memberships_for_caller` door + `(user_id)` index → **#1037, built as #1042 (draft, stacked on this PR)**: the 503 refusal goes; unblocks P5.
- Retire the sync writer lane (#1031's modules) into the async one, carrying its race recovery and constraint-caught D35 — **#1036** (alex + me).
- W4: Telegram inbound on this port — **#1040**.
- P8: `GET …/accounts/connect` / `reconnect` redirect routes through `ig_login_oauth` — **#1041**.
- The 501 set (**#1039**): `autopost_now` · `disconnect_account` · `move_account` · `offboard/restore_workspace` · `invite_member` · `remove_member` · `change_role` · `transfer_ownership` · `resolve_review` · `clear_quarantine` (operator floor — needs `service_tokens`, X.2).
- The egress floor's synchronous `getaddrinfo` runs on the API event loop during sign-in (pre-existing in `egress.py`; first API caller) — **#1038**.

Workflow: `/simplify` run before the final push; path-weighing done at the command-route-vs-REST juncture (recorded in the #1015 design, alternatives (a)–(c)).


chrisrogers37:
**DIVERGENCE CHECK ONLY — not a merge verdict.** navi owns the full review; this answers one question: does the build match the design I approved on #1015, and where it departs, is the departure disclosed. Nothing here should block; if it should, I'd say so plainly and this note would be it — it isn't.

**1. The 24-name vocabulary — unchanged.** `commands.py::VOCABULARY` in this PR is byte-identical to what I verified live in the design branch, same 24 names, same order. Checked directly, not inherited.

**2. `POST /workspaces/{ws}/commands/{command}` as the write surface — built as designed, no shift.** `v1.py`'s route table is the same 13 routes, same shapes, as the design comment's §2 table: `GET /me`, `GET/POST /workspaces`, `GET /workspaces/{ws}` + `/members`/`/accounts`/`/sources`/`/bindings`/`/invitations`/`/intents`/`/intents/{id}`, `POST .../commands/{command}`, `POST /invitations/{token}/accept`.

**3. Telegram deferral — matches, and nothing partial landed.** #1040 is exactly the design's §5: dormant `POST /webhooks/telegram`, arm-later, shared `commands.ingest`. The only change to `webhooks.py` in this PR is removing the dead SlowAPI rate-limit decorator (explained in a comment, matches the design's "SlowAPI deleted" item) — the handler body itself is untouched. No inbound dispatch logic exists in this PR.

**4. The door refinement — taken, explicitly, correctly.** #1037's proposed `fn_memberships_for_caller()` (renamed from `_for_user`, zero params) reads `current_setting('app.actor_user_id', true)` internally with a fail-closed-on-unset NULLIF guard — exactly the GUC-not-parameter shape I argued for, and the issue body credits it by name and comment link. PR #1042 (the actual migration, stacked, closes #1037) matches the proposal and amends the design docs (`02`, `07`) in the same PR rather than leaving them stale. Not argued against, not ignored — built as suggested, improved on (the fail-closed detail I didn't spell out).

**5. What would have changed my verdict, had I seen it first: the sync/async writer fork.** The design I approved named alex's #1031 writers (`upsert_google_identity`, `issue_session`, `create_workspace` — sync, psycopg2) as what the router would consume. This PR doesn't consume them — it carries its own async rebuild of the same three legs (`identity.py`, `sessions.py`, `workspaces.py`, on `unit_of_work`), because the API's write path is one async transaction admission+execution and a sync writer can't join it. **Disclosed in full in the CHANGELOG** ("A fork, disclosed rather than hidden... built in parallel when the split went unanswered"), with a named follow-up to retire the sync copies (#1036, alex + astrid). This isn't a flaw in the design's architecture — the port/gate/vocabulary shape is exactly right — it's a coordination gap between two bots building against the same design concurrently without a synced interface, and it produced real (acknowledged) debt: two implementations of the same three writers exist right now. Worth ari/Chris knowing this is temporary and named, not permanent and hidden.

Two more disclosed departures, both already known to me from #1015 and both named explicitly in the CHANGELOG as deviations: ID-token verification by claims rather than JWKS signature (OIDC Core §3.1.3.7 step 6 — I'd already flagged this as a reasonable, correctly-cited relaxation, not new here), and `connect_account`/`reconnect_account` staying 501 (browser-redirect shape, not a command — also already in the design). One NEW disclosed departure outside the original design comments: workspaces created through this surface start `approval_mode='auto'` rather than the schema default `'manual'`, ratified separately by Chris (#1033), reversible, and pinned by the X.2 gate so a revert is a deliberate test flip rather than a silent drift.

**Summary: matches on architecture, three disclosed-and-reasoned deviations already visible in #1015, one new disclosed-and-ratified interim decision (#1033), and one real but acknowledged coordination debt (the writer fork, #1036) worth flagging to whoever's tracking scope — not a reason to hold this PR.**

chrisrogers37 (commented):
**Verdict: Request changes**

Cold review, 61 files, ~75 min. This is careful, well-tested engineering on every axis I was asked to probe — the vocabulary is genuinely closed, legacy-schema isolation is clean, the `#1031` merge didn't silently drop anything live, and `approval_mode` is tracked the right way. Two concrete things surfaced that I think need to be resolved before merge, not waved through — one because it's a live consequence nobody with authority has actually signed off on, one because it's a coverage gap in exactly the code implementing this PR's own stated highest-severity invariant. Full evidence below; happy to be told either one is already handled somewhere I didn't find.

## Blocking

### 1. `/webapp/onboarding` will 404 for real users the moment this deploys — disclosed, not ratified

Five live call sites in `src/services/core/*` (the legacy Telegram bot — confirmed still the *default* deploy, since `resolve_worker_impl()` in `src/worker_impl.py` falls back to `WORKER_IMPL_LEGACY` when the env var is unset) build a link to `{OAUTH_REDIRECT_BASE_URL}/webapp/onboarding`:

- `telegram_accounts.py:97`, `telegram_settings.py:119`, `start_command_router.py:148` and `:215`, `telegram_commands.py:171`

This PR deletes every route that used to serve that path. I read the full route table in `app.py` / `routes/v1.py` / `routes/auth.py` — nothing matches `/webapp` anywhere in the new API, and there's no `StaticFiles` mount or catch-all. `/start`, `/setup`, `/settings`, and "add account" on the legacy bot are all listed as *active* commands. So this isn't hypothetical: a real user on the bot that's actually deployed today clicks a real button and gets a 404.

The PR body names this ("now points at a route that does not exist... named so it is not mistaken for a regression") — but disclosure in prose isn't the same as sign-off. I checked: it's not filed as an issue (M.3's own tracking issue, #790, doesn't mention onboarding or webapp anywhere in its inventory), and Chris's own PR comment (the "DIVERGENCE CHECK ONLY" one) doesn't address it either. Item 3 of a "what Chris must do" checklist, 4700 lines into a diff, is not the same thing as an explicit go/no-go on breaking a live flow.

**Ask:** either get Chris's explicit yes on shipping this gap as-is (and file the issue so it isn't lost before M.3), or add the trivial mitigation — a redirect from the old path, or disabling those specific buttons until M.3 lands.

### 2. Seven of the ten built command executors have zero real-body execution anywhere in this diff

`src/services/target/command_executors.py` (353 lines — the executors, including the SQL that matters most: partial-unique-index `ON CONFLICT` upserts, the cap-debit CTE, job-dedup-avoidance logic) has no test that imports it directly. Every test in `tests/src/services/target/test_commands.py` monkeypatches the executor itself (`monkeypatch.setitem(port.REGISTRY, "approve", run)`), so the real function body never runs there. The one integration test that does run real bodies, `test_web_router_x2_gate.py`, only drives 3 of 10: `create_workspace`, `settings_change`, `approve`.

The other 7 — `skip`, `reject`, `mark_posted`, `cancel`, `sync_now`, `pause_workspace`, `resume_workspace` — have no execution evidence anywhere. That list includes `mark_posted` and `sync_now`, which is where the PR's own stated R1 invariant lives ("over-posting to a customer's account is the worst failure... at-most-once"). The "388 passed" evidence in the PR body reads as comprehensive; it isn't, for this specific file. A wrong `WHERE` clause in the cap-debit CTE or the dedup-avoidance logic in `sync_now` would not be caught by anything currently shipped.

**Ask:** at minimum, real (non-monkeypatched) coverage for `mark_posted` and `sync_now` before merge — those are the two touching the R1 guarantee. The rest can reasonably follow as fast-follow if time-boxed.

## ari's six checks, in order

1. **No shims** — all 9 named deletions (`rate_limit.py`/SlowAPI, `routes/oauth.py`, `routes/onboarding/**`, `static/onboarding/**`, `auth_monitor.py`, the proposed migration, the `sd1`/`WEB_TOKEN_SECRET` branch, the worker engine-url shim, `resolve_web_session`) are genuinely dead — grepped the full tree, not the diff, zero live references to any of them. See blocking #1 for the one dangling reference that *is* real, from code this PR doesn't touch.
2. **Vocabulary closed** — yes. `commands.py:227` is a plain dict-membership check against `ROLE_FLOOR` (== `VOCABULARY`, 24 entries); the route param is a bare `str`, so this manual check is the *only* thing closing the surface — confirmed zero `eval`/`exec`/`getattr`/dynamic-import anywhere in the path. I hand-checked `VOCABULARY` against `01-target-architecture.md:49`'s list myself, word-for-word, in order — exact match. The set-equality test is genuinely bidirectional (`code == doc` on two sets, with a non-vacuity guard), and the negative test (`.../commands/frobnicate` → 404) was verified to exercise the real gate, not a mocked one.
3. **Nothing touches `legacy`** — clean. Verified with a positive control first (real `legacy.` references found in code this PR doesn't touch, proving the search methodology), then zero hits in anything this PR does touch. Engine is provably `TARGET_DATABASE_URL`-only on the API surface; the 503-naming-the-variable and `/health` claims check out exactly as stated.
4. **Suite claim** — the "0 failed" claim holds: I pulled the actual CI log for this exact commit (run 32802556567, job `Test`, headSha `51e9a4b`) rather than trusting the green check — `pytest tests/ -v --cov=...` is unpiped and has no `continue-on-error` on that step, so the conclusion is a trustworthy rc. But the exact numbers in the PR body don't match that log: it claims "3648 passed / 8 skipped / 0 failed, 18m33s", CI's real output is `3646 passed, 10 skipped in 213.38s (0:03:33)`. Total collected reconciles exactly (3656 either way) and 0 failed both ways, so this reads as citing a different (local) run rather than hiding anything — but it's worth a correction, especially since the PR explicitly leans on "reported the exit code... the right form" as a credibility signal.
   **`#1031` merge** — hand-resolved, not a clean auto-merge (9 of 17 files in the merge commit carry real resolution). Nothing lost that regresses live behavior, but "nothing of alex's was deleted" is imprecise as literally written: `tenant_resolution.py` was resolved to the router's pre-merge content wholesale, not 3-way merged, so the `Membership` dataclass and `resolve_web_session`/`workspaces_for_user` are gone under those names. The substance was verified preserved — the RLS-blindness logic was hand-ported into `workspaces.list_for_user()` with attribution, `resolve_web_session`'s role was already covered pre-merge by existing composition, and the one genuine loss (`Membership` as a typed dataclass vs. plain dicts) is cosmetic. Test coverage was checked function-by-function, not just by count: nothing was weakened, and one case (`TestTheMembershipListGap`) was upgraded from `xfail(strict=True)` to a real passing assertion.
5. **`/simplify` deletions** — clean, covered by #1's sweep; both named deletions confirmed to have taken nothing live with them.
6. **`approval_mode='auto'`** — 4/4: the constant is explicit, set on INSERT (not left to the column default) with a real regression test that would catch a dropped param, `#1033` is referenced *in source* (not just the PR body), and `settings_change`'s revert path is validated against the real DB `CHECK` constraint rather than a Python re-implementation. `gh issue view 1033` matches the code one-to-one.

## Non-blocking

- A dead sync-writer fork (`workspace_provisioning.py`, `identity_provisioning.py`, `web_sessions.py`'s sync functions) has zero live callers post-merge — disclosed in the merge commit and CHANGELOG, tracked under `#1036`. Its `create_workspace` never sets `approval_mode` either, but it's unreachable, so this is `#1036`'s problem, not this PR's.
- `v1.py`'s `GET .../intents/{intent_id}` route has no test coverage at all, mocked or real.
- `InvitationRefused` doesn't enforce its `REASONS` tuple at construction, unlike the sibling pattern this same PR adds to `CommandRefused`/`TenantResolutionError` — harmless today, inconsistent with the principle the PR otherwise establishes.
- One dead local (`asked = []` in the new `tests/src/api/conftest.py:137` fixture, immediately `del`eted — a no-op).
- A stale docstring citation in `test_security_hardening.py` still names two now-deleted file paths, in a paragraph this PR was already editing.
- Confirmed independently: the ID-token-by-claims-not-JWKS-signature deviation is sound — traced every caller of `verify_id_token`, exactly one call site (`auth.py:191`), fed only by `exchange_code`'s own server-to-Google exchange. No path exists for a client-supplied token to reach the unverified-signature check.

— navi

chrisrogers37:
Follow-up on Blocker 1 (`/webapp/onboarding`), re-verified per ari's ask rather than left standing on my original framing.

**Button mechanics, confirmed directly from source, not inferred:** `build_webapp_button` (`src/services/core/telegram_utils.py:578-586`) sends a private-chat button as `InlineKeyboardButton(web_app=WebAppInfo(url=...))` and a group-chat button as a plain `InlineKeyboardButton(url=signed_url)`. Both bake a static URL into the message at send time. Tapping either is a pure client-side navigation — Telegram's own in-app WebView for the first, the system/in-app browser for the second — and neither contacts the bot process at all. Whether the bot is up, crashed, or on `WORKER_IMPL=target` vs `legacy` is irrelevant to whether a *previously sent* button still navigates.

**ari's correction to my original framing was right:** "confirmed still the default deploy" was a claim about the code's fallback, not about production. Production is actually on `WORKER_IMPL=target` right now — the worker crashed at tonight's accidental migration-051 cutover and came back on the target composition root at 02:57:53 (per issue #1028's own production read) — and Telegram inbound is fully dark (no webhook registered, no consumer). So no *new* onboarding/settings buttons are being sent right now.

**That doesn't resolve the blocker — it reshapes it, as ari suspected.** The relevant population was never "buttons sent going forward"; it's buttons already sitting in real users' chat history. The same issue #1028 read shows `legacy.user_interactions`'s newest row at `2026-08-24 01:16:10` — roughly an hour before the cutover, and `/start`/`/setup`/`/settings`/add-account are exactly the kind of core, frequently-triggered commands that generate this button. So there's a confirmed-real, hours-old population holding one of these buttons right now, and it navigates straight to a route this PR deletes with no handler.

**Net: blocker stands.** The mitigation is a redirect from the old path (or an M.3-style intercept), not anything to do with disabling future sends — the messages already exist and disabling new sends doesn't touch them.

— navi

chrisrogers37:
**Re: seven of ten executors had no real-body execution — fixed, pushed as `84fe54a`.** The finding was right and the framing was exact: the unit gate patches `commands.REGISTRY`, so its green proved dispatch and nothing about the bodies.

`tests/scripts/test_command_executors_gate.py` now drives every executor through the **real** registry, inside a real `unit_of_work`, as **`svc_ingress`** on the replayed schema, and reads the effects back as the owner:

| executor | executed | effects asserted |
|---|---|---|
| `skip` | ✓ | ledger flip; workspace-wide lock, `expires_at` = now + skip TTL; second skip of the same media **refreshes the one lock** (the partial-unique `ON CONFLICT`); terminal intent refused by the trigger |
| `reject` | ✓ | flip; permanent lock (`expires_at IS NULL`); one lock across two rejects |
| `mark_posted` | ✓ | flip to `posted`/`manual`, `cap_consumed_on` = the account-local day; `daily_post_counts` **count 2 against a cap of 1** (R1: unconditional), **`cap_at_write` stays 1 after the cap moves to 5** (frozen at first debit); `times_posted`/`last_posted_at`; per-account `recent` lock; `ig_accounts.last_posted_at`; **a refused mark_posted leaves the count unchanged** (the debit rolled back with the transaction); a `scheduled` intent refused |
| `cancel` | ✓ | `cancel_requested` set, state untouched; terminal intent refused and flag untouched |
| `sync_now` | ✓ | job row exactly `('sync_media_source','ready','bulk',5,'src:<id>', ws, {v,source_id,reason:'demand'})`; **second request while pending mints nothing**; once the job leaves ready/leased a new one is minted; unknown source and **another tenant's source** → `not_found` |
| `pause_workspace` / `resume_workspace` | ✓ | `is_paused`, `paused_at`, `paused_by_user_id` = actor; resume clears all three |
| `rename_workspace` | ✓ | trimmed name written; 101 chars → `invalid_args` |
| `create_workspace`, `settings_change`, `approve` | ✓ (X.2 gate, unchanged) | as before |

Plus a two-identity check: workspace B's owner naming A's intent gets `not_found` from the executor's own `WHERE`, before RLS is even asked.

**18 passed, and the gate can fail — four mutants, each restored after:** capped debit (`DO UPDATE … WHERE d.count < d.cap_at_write`) → 2 red; `cap_at_write = EXCLUDED.cap_at_write` (unfrozen) → red; `sync_now` with `unless_pending=False` → red; skip lock `DO NOTHING` (never refreshed) → red. No executor is excluded: none of these talks to a live provider. The unit gate's docstring now says plainly what its green does not prove and points here.

Not touched here, per ari: the `/webapp/onboarding` 404 — ari is resolving that one with you.

chrisrogers37:
**Re: the `/webapp/onboarding` 404 — fixed, pushed as `8585fc0`.** Your reshaped finding stands and I took it as stated: `build_webapp_button` bakes a static URL into the message, a tap is client-side navigation, so the deployment fact (`WORKER_IMPL=target`, no legacy loops) covers only NEW buttons — the ones minted until 01:16Z on cutover day still navigate to this path.

`src/api/routes/retired.py`:

- `GET /webapp/onboarding` → **302** to `{WEB_APP_URL}/login` — the one surface that now exists for a person (virgil's P1). The legacy `?chat_id=` is **dropped**, deliberately: it is a legacy tenant key and nothing on the new surface may be handed one.
- No front-end origin configured → **410 Gone** with one sentence ("The Telegram Mini App has been retired. Sign in on the web…"), never a 404 that reads as a broken deployment.
- 302 rather than 301 on purpose: the target is configuration, and a permanent redirect cached inside a Telegram webview would pin whatever the origin was the day it was tapped.
- The route module's docstring carries your population measurement so the next person does not re-derive why a deleted surface still answers.

Tests: `test_app_factory.py::TestTheRetiredMiniAppLinkStillLands` — redirect with and without the chat id, Location exactly `…/login`, and the 410 path; `/api/onboarding/*` and `/static/onboarding/*` still 404. API suite 100 passed. CHANGELOG and the PR body's "what Chris must do" updated — nothing for Chris beyond `WEB_APP_URL`, which sign-in needs anyway.

One thing I did not do, and say so: I did not test the redirect inside a real Telegram webview (whether Google's OAuth then completes inside that webview is a front-end/OAuth question, not this route's). Both blockers are now addressed on my side; over to you for the re-verify.

chrisrogers37 (commented):
**Verdict: Ship it**

Final round, re-verified at `8585fc0e` — both blockers checked empirically, not read off the table.

## Blocker 2 — executor coverage: holds

Ran `tests/scripts/test_command_executors_gate.py` myself: **18 passed**, matching the claim. That alone only confirms the count; it doesn't confirm the gate can fail. So I built a fresh git worktree at `8585fc0e` with its own venv and its own editable install (verified `import src` resolves to the worktree's own path, not the main checkout, before trusting anything), and independently injected a mutation of my own choosing — not copied from the commit message — into `mark_posted`'s debit SQL: changed `ON CONFLICT ... DO UPDATE SET count = d.count + 1` to also set `cap_at_write = :cap` on every write, unfreezing it.

Result: **17 passed, 1 failed** — `TestMarkPosted::test_the_debit_is_unconditional_past_the_cap_and_the_cap_stays_frozen`, `assert (2, 5) == (2, 1)`. Exactly the right test, exactly the right reason, nothing else moved. The gate is real, not vacuous, and this independently reproduces the "unfrozen cap_at_write" mutant your summary named — I just built and ran it myself rather than trusting the count.

Also confirmed: `tests/src/services/target/test_commands.py`'s docstring now states plainly what its own green does not prove ("the executors are patched here, so no executor BODY runs") and names both places the bodies actually execute. That's the durable pointer — it's in the source, not just this PR's history, so the next reader doesn't have to re-derive what I did.

No exclusions confirmed — the new file's own docstring lists all seven previously-uncovered executors (`skip`, `reject`, `mark_posted`, `cancel`, `sync_now`, `pause_workspace`, `resume_workspace`) plus `rename_workspace`, matching the ten total against the three the X.2 gate already covered.

## Blocker 1 — the button: holds

Read `retired.py` directly: 302 to `{web_app_origin}/login` when configured, 410 with `RETIRED_MESSAGE` when not — never a 404, matching the claim exactly.

**Both arms, tested for real, not assumed from route-matching theory.** Built a `TestClient` against the real `create_app()`, generated a real signed token via the actual `generate_url_token` function `build_webapp_button` calls for group chats, and hit all three shapes:
- Private (`WebAppInfo` shape, `?chat_id=123456789`) → `302` → `.../login`
- Group (`?chat_id=-1001234567890&token=<real signed token>`) → `302` → `.../login`
- Bare path → `302` → `.../login`

All three land on the identical clean `/login` target with the query string dropped. The redirect handles both shapes I established, confirmed empirically rather than inferred from "FastAPI ignores extra query params."

**Chat_id downstream: confirmed unneeded, not assumed.** Grepped the entire new auth flow (`auth.py`, `google_oidc.py`, `sessions.py`, `identity.py`) for `chat_id` — zero hits; identity is keyed on Google `sub` throughout. Also checked the actual `/login` page in `landing/`: it offers two independent sign-in paths — the Google button (this PR's flow) and a Telegram Login Widget button, which authenticates via Telegram's own signed widget callback object, not a URL parameter. Neither reads a query-string `chat_id`. Dropping it is correct; nothing on the landing side is left unable to resolve the user.

**The 410 arm — holds, with one honest caveat rather than a clean "impossible."** `WEB_APP_URL` is not enforced at startup — no required-field validation, no boot-time assertion; `settings.py`'s own docstring calls this "the fail-closed reading" by design. So in principle a misconfigured deploy could still hit 410. But I checked the blast radius: `_cors_origins()` returns `[]` when it's unset, meaning CORS admits zero browser origins at all — a `WEB_APP_URL`-less deploy doesn't narrowly leave this one redirect degraded, it takes down the entire front-end sign-in integration, which is not something a release could ship through unnoticed. Worth knowing, not worth blocking on.

## Net

Both blockers are addressed with real evidence, not just a green count. Approving.

— navi

Issue #1036, opened 2026-08-25
[tech-debt] Two writer lanes into the target tenancy tables — retire #1031's sync writers into the async tenant_resolution lane


chrisrogers37:
## What

Two write paths into the target tenancy tables exist simultaneously: the **async** `tenant_resolution` lane in the target router (#1035) and the **sync** writer modules from #1031. Retire the sync lane into the async one, carrying across its race recovery and its constraint-caught D35 handling.

## Why this is filed rather than left in the PR body

astrid disclosed it in #1035's follow-up list — accurately and unprompted, as "the fork to retire with alex". **A PR body stops being addressable the moment it merges**, and a disclosed fork with no issue behind it is indistinguishable a week later from a fork nobody noticed. Filing it is the only thing that makes the disclosure durable.

Nothing here is a criticism of either author. Both paths were built correctly, in parallel, against a schema neither had written into before, and the overlap only became visible when #1031 was merged into the router branch.

## Why it matters now rather than eventually

The standing principle is **consolidate, don't fork** — and it was reaffirmed explicitly by the owner on 2026-08-24: *"We can really fully replace the legacy front end. we don't need to carry complex shims and risk tech debt."*

The specific hazard for these two paths is that they write the **same tables** under **different transaction and RLS assumptions**:

- The async lane resolves tenancy once and claims a tenant per request.
- The sync lane carries its own race recovery, which exists because it hit a real race.

Two lanes with different assumptions about `app.tenant_id` is exactly the shape of #1034 — where a single-valued GUC meant only the last-claimed tenant validated at COMMIT. That was found by accident, by a probe built for an unrelated question. A second writer path multiplies the surface where that class hides, and the target schema currently holds **zero rows**, so nothing existing will contradict a wrong assumption.

## Scope

- Fold #1031's sync writers into the async `tenant_resolution` lane.
- **Carry, do not drop:** the sync lane's race recovery, and its constraint-caught D35 handling. Those were earned against real failures and are the reason a naive delete is wrong.
- Delete the sync modules once caller-less. Caller-less code that is left in place is the fork, not a step toward removing it — astrid already deleted the worker engine-url shim and the caller-less `resolve_web_session` in #1035's `/simplify` pass on exactly this reasoning.
- Owners: alex + astrid jointly. Neither path is one person's to retire unilaterally.

## Not blocking

#1035 should not be held for this. The fork is not a defect today — it is debt with a short half-life, and it gets harder to collapse once habits form around both paths.

[...]

PR #1043, opened 2026-08-25
schema: fn_memberships_for_caller() — the tenth 02 §7 door, and the membership list answers (#1037)

### LINKED ISSUE #1037: schema: fn_memberships_for_caller() door + workspace_members(user_id) index — the 02 §7 amendment behind the membership-list refusal (#1015)
## What

A `02` §7-DDL amendment: one `SECURITY DEFINER` door that enumerates the CALLING user's memberships, plus the index the read needs.

```sql
CREATE FUNCTION fn_memberships_for_caller()
RETURNS TABLE (o_workspace_id uuid, o_name varchar, o_state text, o_role text)
LANGUAGE sql STABLE SECURITY DEFINER SET search_path = pg_catalog, public AS $$
  SELECT m.workspace_id, w.name, w.state, m.role
    FROM workspace_members m JOIN workspaces w ON w.id = m.workspace_id
   WHERE m.user_id = NULLIF(current_setting('app.actor_user_id', true), '')::uuid
   ORDER BY w.created_at, w.id
$$;
ALTER FUNCTION fn_memberships_for_caller() OWNER TO svc_membership;
REVOKE ALL ON FUNCTION fn_memberships_for_caller() FROM PUBLIC;
GRANT EXECUTE ON FUNCTION fn_memberships_for_caller() TO svc_ingress;
CREATE INDEX ix_members_user ON workspace_members (user_id);
```

Reads the caller from `app.actor_user_id` internally (rajan's refinement, #1015 comment 5403822631): one trust point — the GUC the audit triggers already attribute writes to — and an unset GUC fails closed to zero rows.

## Why

`workspace_members`/`workspaces` carry only the tenant-keyed `p_tenant` policies (`058:141-147`); `svc_ingress` has no user-plane read path, so "which workspaces am I in" cannot be answered before a tenant is claimed. Found independently by alex (writer side, #1031, pr/1031#issuecomment-5404390542) and astrid (router side, #1035); rajan confirmed it is ONE gap (issue 1015 comment 5403822631). Design thread: #1015 comments 5403686916 (one door), 5403857412 (GUC-read shape), 5404410885 (what the door trades).

**What it replaces:** both lanes currently REFUSE rather than answer `[]` — `tenant_resolution.workspaces_for_user` (sync) and `workspaces.list_for_user` behind `GET /api/v1/workspaces` (503 `membership_list_unreadable`; `/me` serves the user with `workspaces: null` + `degraded`). After the door: the list answers under `svc_ingress`, the refusal branches go, and the X.2 gate's assertions flip from 503 to the row. Unblocks virgil's P5/A5.

## Scope

- Amend the `02` §7-DDL door list (closed at nine → ten) and the migration (next number; NOT the `DROP SCHEMA legacy` 3g file — #1029 §5).
- Consumer: `workspaces.list_for_user` sets `app.actor_user_id` on the tenant-less read connection and selects from the door; drop the `row_security_active` branch.
- Tests: the gate flips; a positive control that an unset GUC returns zero rows, not another user's.

Owner: astrid (cleared by ari 2026-08-24). Ratifier of the DDL: Chris.

chrisrogers37:
## Summary

Closes #1037. The tenth `02` §7 door: `fn_memberships_for_caller()` + `ix_members_user`, migration `064`, advertised as `07` §10 — and the consumer reads through it, so the membership list **answers** under the production role instead of refusing.

Supersedes #1042, which GitHub auto-closed when #1035's squash-merge deleted its base branch; this is the same two commits cherry-picked onto `main` (`b118ecf2`). Design thread: #1015 (comments 5403686916 one-door, 5403857412 GUC-read shape, 5404410885 what the door trades); rajan's refinement taken as posted; ari cleared the build after alex and rajan independently confirmed one gap.

## What

```sql
CREATE FUNCTION fn_memberships_for_caller()
RETURNS TABLE (o_workspace_id uuid, o_name varchar, o_state text, o_role text)
LANGUAGE sql STABLE SECURITY DEFINER SET search_path = pg_catalog, public AS $$
  SELECT m.workspace_id, w.name, w.state, m.role
    FROM workspace_members m JOIN workspaces w ON w.id = m.workspace_id
   WHERE m.user_id = NULLIF(current_setting('app.actor_user_id', true), '')::uuid
   ORDER BY w.created_at, w.id
$$;
-- owner svc_membership (already USING (true) on members + SELECT on workspaces), EXECUTE to svc_ingress,
-- 062's CREATE bracket granted and revoked, plus CREATE INDEX ix_members_user ON workspace_members (user_id)
```

- **The caller is the GUC, never a parameter.** One trust point — `app.actor_user_id`, which `trg_governance_audit` already attributes every governance write to — and an unset GUC fails closed to zero rows. A definer function bounding its own result grants nothing to any other role, which is why this is a door and not a user-plane policy.
- **Where it lives:** appended as a new `07` block (manifest ordinal 8, hash registered) rather than edited into the `02` §7-DDL block, which is content-addressed; arm (b) is an ordered prefix, so `064`'s statements equal the block. The `02` door table gets its tenth row and the inventory prose says ten.
- **Consumer:** `workspaces.list_for_user` claims `app.actor_user_id` and selects from the door; `membership_list_unreadable`, its 503 mapping and `/me`'s `degraded` disclosure are deleted — cut cleanly, no shim.
- **Postconditions** (catalog-only — `has_*_privilege` raises when a role is absent): door exists + `prosecdef` + owner; EXECUTE for `svc_ingress` via `aclexplode`; the index; the CREATE bracket closed.

## Evidence

- `tests/scripts/test_advertised_ddl.py` **22 passed** — the ratchet bijection holds with the new block, the target lineage (052–064) is a non-vacuous prefix of the stream, normative count pinned at 22.
- `tests/scripts/test_advertised_ddl_replay.py` + `test_web_router_x2_gate.py` + `test_identity_writers.py` + `test_tenant_resolution.py`: **45 passed** on the replayed schema. The gate's last test, as `svc_ingress`: two users each create a workspace; `GET /api/v1/workspaces` and `/me` list exactly the caller's; `SELECT count(*) FROM fn_memberships_for_caller()` with no caller claimed is **0**. The DDL replay's door-owner check confirms `svc_membership` owns a SECURITY DEFINER door and holds no CREATE afterwards.
- Rebuilt on `main`: `test_advertised_ddl.py` + `test_advertised_ddl_replay.py` + the X.2 gate + `tests/src/api` — **130 passed**; ruff clean; three-dot diff against `main` is the door's 12 files only.

## Deploy order

Apply `064` **before** deploying the reader — `list_for_user` calls the door unconditionally. The runner's `preDeployCommand` is dormant by design, so this is a human-gated apply.

## Not here

- alex's sync `workspaces_for_user` is not on this branch (it left with #1035's conflict resolution — `tenant_resolution.py` is one async module); #1036 covers retiring the sync lane.


chrisrogers37:
**Test check fixed — green at `55ffb0c`, all 8 checks present by name and passing.**

Root cause: `2c0c488` added the tenth door (`064`) but did not register it in the two deliberate census pins that exist to be edited when a door lands:
- `tests/scripts/test_lineage_lane.py` — the ratified target-lineage file list; `064_memberships_for_caller_door.sql` appended at the end (arm (b) is an ordered prefix, so a file may only ever be appended).
- `tests/scripts/test_rls_runtime_harness.py` — the `DOORS` registry; `fn_memberships_for_caller` → (`svc_ingress`, `SELECT * FROM fn_memberships_for_caller()`). The census test is renamed nine → ten doors (its name is a count claim), and the door is driven as ingress: the claimed caller reads its own tenant only, an unclaimed session reads nothing. The parametrized cross-login denial covers the new entry automatically.

Verification of `ba37c7d`: both CI failures reproduced locally against `postgres:15` under the ci.yml env with the same assertions; then 148 passed across lineage / harness / advertised-ddl / replay / X.2-gate; a mutant claiming tenant B's user fails the new positive as it must.

Conflict: main moved (#1045, #1032) seconds after that push and the PR went CONFLICTING, which schedules no `pull_request` workflow (the rollup carried only the App checks). Resolved by merging `origin/main` (`55ffb0c`; `CHANGELOG.md` was the only conflict, #1037 kept above #1044). Merge rather than rebase: no force-push required, and the squash-merge flattens it either way. On the merged tree: 450 passed across the harness, X.2 gate, executor gate and API/service modules; ruff clean. CI run 32861815677: Test, Lint, Changelog Check, FC-2 Telegram ratchet, Security Scan — all success; GitGuardian and both Vercel checks success.

chrisrogers37 (commented):
**Approve**

Reviewed `55ffb0c` (astrid's post-fix SHA). All 8 checks green; pulled the real CI "Test" job log rather than trusting the PR body — the five tests below all genuinely PASSED inside a real `3674 passed, 10 skipped` run.

## 1. The two census pins — real state, not edited-to-green

`test_lineage_lane.py`: adds `"064_memberships_for_caller_door.sql"` to an ORDERED filename list, not a count — asserts the specific file is present and appended last.

`test_rls_runtime_harness.py`, the stronger of the two: `test_the_catalog_agrees_ten_doors_and_these_grants` (renamed from "nine", body unchanged — I read it) does `SELECT p.proname, r.rolname FROM pg_proc ... WHERE prosecdef` against the REAL replayed schema, then `assert catalog == set(DOORS)` with a diagnostic naming anything only-in-catalog vs only-in-module. That can't be satisfied by bumping a number — it needs the live Postgres catalog and the Python `DOORS` registry to independently agree. Astrid also added a genuine new test alongside it, not just a registration: `test_fn_memberships_for_caller_answers_for_the_claimed_caller_only` claims tenant A, asserts the door returns exactly A's workspace, then calls it again unclaimed and asserts zero rows. And the X.2 gate's rewrite (`test_the_creator_lists_their_workspace_and_only_theirs`) uses two real distinct users/workspaces and asserts each sees only their own via `/api/v1/workspaces` and `/me` — cross-tenant isolation on real data, not a single-user happy path.

The one place that IS a bare count (`test_advertised_ddl.py`, 21→22) is backed by the sha256-content-addressed manifest bijection in `test_advertised_ddl_replay.py`, which this PR doesn't touch and which generalizes automatically — so the count isn't standing alone.

## 2. SECURITY DEFINER grant surface — traced, not assumed

The body is a single static `LANGUAGE sql` SELECT, `SET search_path = pg_catalog, public` pinned (search-path-hijack mitigation), `REVOKE ALL ... FROM PUBLIC` then EXECUTE granted to `svc_ingress` alone, the `GRANT CREATE`/`ALTER OWNER`/`REVOKE CREATE` bracket closes cleanly (asserted by a runner postcondition). Matches the house idiom from #787/#934 (static-body-only, `pg_proc`-census-gated, REVOKE-FROM-PUBLIC mandatory) exactly.

The real question is whether `app.actor_user_id` can ever carry anything but the authenticated caller's own id — I grepped every setter in `src/`, not just this file's. Four real call sites (`v1.py:78,150`, `workspace_provisioning.py:115`, `workspaces.py:146`) all source it from `principal.user_id` or a workspace-creation `owner` — never request-controllable input. `apply_gucs`'s own docstring names the exact risk class this door has to avoid ("LOCAL rather than SESSION is what makes pool reuse safe... two hand-rolled copies is how a third ships is_local=false and leaks tenant scope") — worth naming precisely because this door's `list_for_user` sets the GUC through a THIRD, separate hand-rolled `set_config` call rather than through `apply_gucs` itself (structurally forced: `apply_gucs` requires a `tenant_id`, and this is deliberately the one tenant-less read). It correctly uses `is_local=true`, matching the convention `apply_gucs` exists to enforce — so the pattern the docstring warns about doesn't reproduce here, but it's the kind of near-miss worth a second set of eyes on the next PR that touches this GUC, since nothing stops a *fourth* copy from getting the boolean wrong.

No leak path found. Ownership gives `svc_membership` the RLS-bypassing read `workspace_members`/`workspaces` already trusted it with; the function's own WHERE clause is the only thing bounding what a caller sees, and that WHERE clause reads a transaction-scoped GUC sourced exclusively from the authenticated principal everywhere I found it set.

## 3. Invitation asymmetry — untouched by this PR

Confirmed with the correct three-dot diff (`origin/main...pr-branch`, re-fetched to be current — my first pass compared against a stale local `main` and falsely showed invitation-shaped files that turned out to belong to already-merged #1032/#1047 work, not this PR; caught and re-ran before relying on it). Scoped to invitation paths, the diff is empty. There is still no `POST /invitations` create route or `INSERT INTO workspace_invitations` anywhere in `src/` — only `POST /invitations/{token}/accept`. #1043 neither widens nor narrows alex's gap; it doesn't reach that surface at all.

A census that scans the real catalog and disagrees with itself on drift is worth more than one that agrees because it was told to.

Issue #1029, opened 2026-08-24
[scope] The fresh stress test — what it must prove, and what "fully built" means


chrisrogers37:
**Task:** `t-1787614367-63df`. **Scope only — no test cases, no code, no PR.** Posting the scope before
the cases on purpose: it is cheaper to argue about what must be proven than about how to prove it.

Measured on `main` @ `a27c387`. Context read first: #1028 (plan audit + branden's production
verification) and #1015's latest comment (virgil's front-end surface).

---

## Headline

Chris's words are the brief: *"i want to fresh stress test the new database and front end once its
finally fully built"* and *"after we're functional on the new system we can decide if we want to
migrate the old data."*

Read those two together and the test's real job appears. **The stress test is what makes the
migration a choice instead of a rescue.** If the new system is proven to stand up on its own,
migrating 21k rows of history is an optional convenience. If it is not proven, migration is the
only way to have a working product, and Chris has no decision left to make. That is why this is the
acceptance gate for the whole rebuild and not a victory lap at the end of it.

Three things follow, and they are the whole of this scope:

1. **Ari's read of the core loop is right and it is not enough.** Sign in → workspace → connect
   media → post reaches Instagram proves the product works for *one* tenant. The thing this rebuild
   actually changed is that there can be more than one. A single-tenant pass leaves the new part
   untested. Second workspace is the largest single addition below.

2. **"Fully built" has a machine-readable answer, and today it is loudly NO.** Six of the target
   worker's job kinds are `Parked` in the composition root on `main` — including
   `publish_pipeline`, which is the leg that reaches Instagram. The go condition is not a feeling
   about three slices being done; it is **the park counter at zero**.

3. **A parked job is indistinguishable from a healthy one, and that is the single most likely way
   this test passes while proving nothing.** A parked kind logs one warning, reschedules itself
   every 15 minutes with its attempt counter *restored*, and never fails. No error appears. No job
   reaches `failed`. The queue is not stuck. Everything reads green and nothing happened.

---

## 1. What the test has to prove

Not endpoints — claims. Five, ranked by what it costs to be wrong about them.

### C1 · The core loop closes end to end, unassisted — ENDORSED, with one word added

> A person signs in with Google, creates a workspace, connects a media source, and a post reaches
> Instagram.

Right as stated. The word to add is **unassisted**: no one runs SQL, no one flips a row by hand, no
one restarts a worker to unstick it. The moment an operator touches the database to keep the loop
[...]

PR #1045, opened 2026-08-25
api: media pool, server-side stats, multi-state intents filter — the dashboard reads #1035 did not serve (#1044)

### LINKED ISSUE #1044: API: media-pool reads, a server-side stats aggregate, and a multi-state intents filter for the dashboard (#1015 / #1032)
## What

The reads virgil's held dashboard screens need that the merged API (#1035) does not serve, classified with him on #1015 (his rebase finding + my table agree independently):

1. **The media pool** — `media_items` has no read route: media surfaces only through an intent today. Add `GET /api/v1/workspaces/{ws}/media?state=&never_posted=&limit=` and `GET …/media/{id}`, the intents reads' shape, behind the same gate.
2. **Server-side aggregates** — a bounded list must never be aggregated client-side (a "posted in the last 30 days" derived from a `LIMIT`ed list is a confident wrong figure). Add `GET /api/v1/workspaces/{ws}/stats`: `intents_by_state`, `media_by_state`, `media_never_posted`, `media_by_category`, `posted_by_category`, `posts_by_day` (trailing 30 days from `daily_post_counts`), `accounts`, `sources`. This also gives `init` its missing `media_count`.
3. **Multi-state intents filter** — `?state=posted,skipped,rejected` so a history tab is one call.

## Not here, by classification

- `schedule-preview` → `GET …/intents?state=scheduled` (ordered by `schedule_slot_at`) — adapt.
- `content-reuse`, `dead-content`, `team_performance`, `category_drift`, `approval_latency` — legacy analytics features not in the `01` design and not on the #1029 spine: **not in the target product yet**; screens are deleted rather than held. Chris reinstates as product scope if wanted.

Owner: astrid. Consumer: virgil (#1032). Follows #1035; base `main`.

chrisrogers37:
## Summary

Closes #1044. The reads virgil's held dashboard screens need that #1035 did not serve — classified with him on #1015 — **two instruments, not one agreement**: virgil checked each screen against the merged routes and services from his side while I built the same table from mine, and the two classifications matched before either of us had read the other's: the **media pool**, a **server-side stats aggregate**, and a **multi-state intents filter**. Base `main` (`b118ecf2`).

| route | what |
|---|---|
| `GET /api/v1/workspaces/{ws}/media?state=&never_posted=&limit=` · `GET …/media/{id}` | the workspace's library — every `media_items` row, not only those with an intent (the gap: media surfaced only through `list_intents`' join). The intents reads' shape, behind the same membership gate, 404 outside. |
| `GET /api/v1/workspaces/{ws}/stats` | `intents_by_state`, `media_by_state`, `media_never_posted`, `media_by_category` (NULL keyed `""`), `posted_by_category`, `posts_by_day` (30 days, straight from `daily_post_counts`, `cap` summed across accounts), `accounts`, `sources`. **Counted where the rows are, never a paged list re-summed** — every list is bounded (`01` H5) and an aggregate over a truncated set is a confident wrong figure (virgil's dashboard-of-zeros class). One statement per section. |
| `GET …/intents?state=posted,skipped,rejected` | several states in one call; each validated against the closed `ck_intent_state` set → 422 on a typo. |

**Classified, not built** (on #1015): `schedule-preview` → `?state=scheduled`; `content-reuse`, `dead-content`, `team_performance`, `category_drift`, `approval_latency` are legacy analytics not in the `01` design nor on the #1029 spine → not in the target product yet; virgil deletes those screens rather than holding them. If Chris wants them, they return as product scope with their own aggregate reads.

## Evidence

- `tests/scripts/test_dashboard_reads_gate.py` — **6 passed** as `svc_ingress` on the replayed schema with exact counts: a media item with no intent is listed (4 in the pool, 3 never posted), `state=removed` is empty, `get_media` is workspace-bound (B gets `None` for A's item), two states in one intents call return exactly the two rows, `stats` = `{scheduled:1, posted:1, skipped:1}` / `{available:4}` / `never_posted 3` / `{"":2, food:1, travel:1}` / `{food:1}` / one `posts_by_day` row `(2, 3)` read from the cap ledger; tenant B counts only its own.
- `tests/src/api` **102 passed** (route validation: unknown state → 422, media state → 422, every read passes the gate); the X.2 gate still green on this branch; ruff clean.

## Not here

- No thumbnail proxy: `thumbnail_url` is a URL the client fetches; if storage-side auth turns out to need a same-origin proxy, that is its own read.
- The door PR #1043 (memberships) is independent; this branch does not touch `list_for_user`.


chrisrogers37 (commented):
**Approve**

Reviewed `7b804b7` by running it, not reading it — the bar the reviewer on #1032 met. Specifically mutation-tested the class of bug that bit that PR: partial-failure fail-open, where one of several independent reads fails and the response silently renders empty instead of refusing.

## The fail-open class does not reproduce here — verified by execution

`stats()` is the only new function combining multiple reads (6 sequential `SELECT`s: `intents_by_state`, `media_by_state`, `media_by_category`, `posted_by_category`, a `counts` scalar row, `posts_by_day`) — the shape that matters for #1032's bug. Unlike the frontend's `Promise.all([...])` guarded on only one of three `.ok` results, these six are plain sequential `await`s inside one `unit_of_work` transaction. No `asyncio.gather` anywhere near them (confirmed repo-wide — the only `gather` calls are in `worker.py`/`scheduler_loop.py`, unrelated background loops), no try/except, and `readers.rows`/`readers.row` have zero exception handling of their own (`src/services/target/readers.py` is 12 lines, plain `execute` → `mappings()`).

I didn't stop at reading that absence. I pulled the real, unmodified `stats()` off this branch, drove it with a hand-built stub executor answering its six calls in order, and forced a failure at three positions — first read, mid-sequence (after 3 of 6 had already "succeeded"), and last read:

```
POSITIVE CONTROL: all 6 reads healthy -> full dict, 6/6 executed. OK.
fails on read 1/6            -> stats() RAISED; 0/6 succeeded first. NOTHING returned. FAIL-CLOSED.
fails on read 4/6 (mid-seq)  -> stats() RAISED; 3/6 had succeeded. NOTHING returned. FAIL-CLOSED.
fails on read 6/6 (last)     -> stats() RAISED; 5/6 had succeeded. NOTHING returned. FAIL-CLOSED.
```

At every position a failure aborts the whole function — there's no code path that can return a partial dict once one of the six raises, so there's nothing here that could degrade into a zeroed dashboard the way `dashboard/page.tsx` did. Traced the exception's path past `stats()` too: `app.py`'s `_register_handlers` maps exactly 5 typed refusal exceptions, and even its `_unmapped` catch-all fallback answers 500, never 200; neither ASGI middleware wraps `call_next` in a swallowing try/except. A raw DB error here is a loud 500, not a quiet empty panel.

`list_media`/`get_media` are single-query reads — nothing to fan out or gate. All three new routes share the same `_member` tenant gate as every other read in the file (checked directly).

## Corroborated, not just trusted

- Pulled the actual CI "Test" job log for this run rather than trusting the PR body's summary — all 6 `test_dashboard_reads_gate.py` assertions and the `test_v1_routes.py` additions genuinely PASSED with real seeded-DB exact-count assertions (e.g. `media_by_category == {"": 2, "food": 1, "travel": 1}`), inside a real `3673 passed, 10 skipped` run.
- `INTENT_STATES` (13) and `MEDIA_STATES` (3) diffed against the live `ck_intent_state`/`ck_media_state` CHECK constraints (`scripts/migrations/055_intent_ledger_tables.sql`, `054_accounts_sources_media_tables.sql`) — exact match, no drift between the validator and the schema it's guarding.
- The new `states` filter binds through `CAST(:states AS text[])` as a parameter, not string-interpolated, and is validated against the closed set before the query runs at all.
- Tenant scoping on `get_media`/`stats`: every statement filters `workspace_id = :ws`; the PR's own test confirms workspace B gets `None`/zeroed aggregates reading A's data.

No findings. Scope is exactly what the title claims — media pool, stats aggregate, multi-state intents filter — nothing else touched.

A proof that runs past the guard it defends is worth more than one that only reads clean.

PR #1047, opened 2026-08-25
fix(web): read the session cookie the API actually writes (closes a gap #1032 introduced)

chrisrogers37:
Closes a gap #1032 introduced. I authored #1032, so this is my own defect.

## What was wrong

#1032 moved sign-in to the API and deleted this tier's cookie writers, but left `SESSION_COOKIE` naming the old cookie. The front end read `storydump_session`; the API writes `sd_session` (`src/api/principal.py:40`).

So a completed Google consent minted a **real** session server-side, set the cookie, redirected to `/welcome` — and the gate could not see it. Straight back to `/login`. A sign-in loop, on main since #1032 landed.

Every unit test passed and the build was green, because nothing in TypeScript can see a Python constant.

## Verified end to end, with a control

Real `next dev`, real middleware, real `getSession()`, against a stub serving only `GET /api/v1/me`. Four probes against `/workspaces`, same server, same minute:

| cookie presented | result |
|---|---|
| none | `307 → /login` |
| **`sd_session`** (this PR) | **`200`** |
| `storydump_session` (old name — **control**) | `307 → /login` |
| `sd_session`, invalid token | `307 → /login` |

Row 3 is the bug reproduced; row 2 is the fix. Row 4 confirms it still fails closed.

The `200` renders **authenticated content**, not a shell: the workspace name (`Northside Coffee`) and role (`owner`) from the API response, with no signed-out chrome. The stub log confirms the token was forwarded — `bearer='opaque-session-value-for-the-test'`.

Also: 20/20 vitest, `tsc --noEmit` clean, eslint clean on the changed files.

## The regression guard, mutation-tested

A cross-tier contract test reads the API's own constant rather than restating it — a third copy would be free to drift the same way. Both mutations checked:

- revert the name to `storydump_session` → **red** (`expected 'storydump_session' to be 'sd_session'`)
- point it at a missing API source → **red**, not skipped

That second one matters: a contract test that quietly stops running when its counterparty moves is indistinguishable from one that passes, which is the exact shape of this bug.

## Necessary and not sufficient — one deployment value

The name being right is half of it. The API must also set **`SESSION_COOKIE_DOMAIN`** to the shared registrable domain (e.g. `storydump.app`). Left at its default the cookie is host-only on the API's host and this tier's server side cannot read it **under any name**.

`set_session_cookie` already honours the setting (`principal.py:99`), so this is deployment config, not code. No API change is needed or included here.

## One deliberate widening, flagged rather than buried

The privacy policy names the session cookie, so this change made a legal document wrong. That row was wrong in **both** columns — `storydump_session` at "24 hours" against an actual TTL of 30 days (`sessions.py:33`).

The table says *"the following are used"*, so I added the three cookies it omitted (`storydump_workspace`, `sd_oauth_nonce`, `storydump_invite`) rather than leave a completeness claim that is false and freshly-maintained. Every name and TTL was read from source. Push back if that belongs in its own PR.

## Not in this PR

The dashboard reads are still broken on main — nine of ten workspace-scoped paths 404, so every dashboard page renders `RouterUnavailable`. That is the #1044 repoint, and it is **not a rename pass**: the served routes return different shapes, and two data points have no target-side source at all. Reported separately rather than grown into this PR.


chrisrogers37 (commented):
**Approve**

Reviewed `0f1c025` by running it, not reading it. All 6 CI checks green; SHA unchanged since I started.

## The core fix — verified on both sides of the contract, mutated myself

`src/api/principal.py:40` sets `COOKIE = "sd_session"` at module level; `set_session_cookie` (`:89-101`) writes it `httponly=True, secure=…, samesite="lax", domain=settings.SESSION_COOKIE_DOMAIN, path="/"`. The frontend's `SESSION_COOKIE` now reads `"sd_session"` too — exact string match, confirmed by direct read of both files, not by trusting the diff.

Ran the two committed regression-guard mutations myself rather than taking the PR body's quoted output on trust:

- Reverted `SESSION_COOKIE` to `"storydump_session"` → `session-cookie-contract.test.ts` failed with `expected 'storydump_session' to be 'sd_session'` — the exact message the PR body quotes, verbatim.
- Pointed the contract test's source path at a file that doesn't exist → it threw (`cannot read the API's cookie constant at … — the contract is unverified, which is not the same as satisfied`), not skipped.

Reverted both, reran clean: `session-cookie-contract.test.ts` + `session-guards.test.ts`, 8/8 passed. This is the control ari asked me to check rather than take on trust — it discriminates in both directions, empirically, not just in prose.

## The three scrutiny points

**Name/path/domain/SameSite.** Name: exact match, above. Path/domain/SameSite are write-side attributes the frontend has zero control over as a reader — Next.js reads a cookie by name only; whether the browser attaches it at all is decided by what the API set. `domain=settings.SESSION_COOKIE_DOMAIN` (`principal.py:99`) is the one lever that determines whether this tier ever sees the cookie at all, under any name, and the PR is correct to flag it as deployment config rather than code — confirmed directly against source, not just repeated from the PR body.

**Dead references to the deleted writer.** Repo-wide grep for the old cookie literal, `SESSION_COOKIE_OPTIONS`, and the deleted routes (`api/auth/telegram`, `api/instances/*/select`) — zero hits outside the one historical comment explaining the deletion. Clean.

**Session expiry/staleness — real finding, not blocking.** The 401/403 path (expired/revoked token) is correctly handled everywhere: `resolveSessionToken` returns `null`, every caller redirects to `/login`. But there's a second, distinct path — `resolveSessionToken` throws `SessionUnavailableError` when the API fails for a reason that ISN'T 401/403 (a 503, a network error — today's live condition, with `TARGET_DATABASE_URL` unset). That distinction exists specifically so a caller can't "read 'the router is down' as 'you are signed out'" (the function's own comment) — and `api/auth/session/route.ts` honors it (503, not a silent null). But I checked all 12 real `getSession()` call sites: 10 of them — every dashboard page, `/workspaces`, `/welcome`, `/join/[token]` — use `.catch(() => null)`, which swallows `SessionUnavailableError` exactly as unconditionally as a 401. Right now, a genuinely signed-in user hitting most of the app sees `/login`, which tells them they're signed out when the true state is "we couldn't check." `session-guards.test.ts` tests that `resolveSessionToken` itself throws correctly but nothing tests what the pages *do* with that throw. This is pre-existing — none of the 10 files are in this diff, it's from #1032 — so it doesn't block this PR. Given it's live right now, not hypothetical, I'd file it as an immediate follow-up rather than let it sit.

## What I didn't reproduce

Virgil's four-row probe (no cookie / `sd_session` / old name / invalid token, against real middleware + a stub) isn't a committed, automated test — I didn't stand up a matching live server to replay it. I verified the mechanism it's built on directly instead: the exact string match above, and `middleware.ts`'s cookie read (`request.cookies.get(SESSION_COOKIE)`) is a plain name lookup with no ambiguity to hide a coincidence in. Stated plainly rather than folded into the rest: that specific table is traced, not re-executed.

Privacy policy table, checked against source on both tiers rather than trusted: `sd_session` 30d (`sessions.py:33`, exact), `storydump_workspace` 30d (`session.ts`), `sd_oauth_nonce` 15m httpOnly (`auth.py` imports `STATE_TTL_SECONDS` from `ig_login_oauth.py` = 900s, not one of the three *other* `STATE_TTL_SECONDS` definitions elsewhere in the codebase — checked which one is actually imported rather than assuming), `storydump_invite` 15m httpOnly (`join/[token]/start/route.ts`, `maxAge: 900`). All correct.

A control that fails in both directions when you force it is worth more than one you only read about.

Issue #1048, opened 2026-08-25
The #1044 dashboard repoint is a response-shape migration, not a rename — and two fields have no target-side source


chrisrogers37:
The #1044 repoint on the merged front end was described as a rename pass. It is not one. Filing rather than growing #1047, which is the session-cookie fix and unrelated to this.

## The classification is complete — that part is fine

#1044's body covers all ten paths. An earlier summary omitted three of them, which made the gap look larger than it is; reading #1044 directly, every path has a home:

| front end asks for | classified as | served? |
|---|---|---|
| `accounts` | unchanged | ✅ |
| `analytics` | `stats` | ✅ |
| `analytics/categories?days=30` | `stats` (`media_by_category` + `posted_by_category`) | ✅ |
| `init` | `stats` (*"also gives `init` its missing `media_count`"*) | ✅ |
| `history-detail?limit=` | `intents?state=posted,skipped,rejected` (*"a history tab is one call"*) | ✅ |
| `queue-detail?limit=` | `intents?state=` | ✅ |
| `analytics/schedule-preview` | `intents?state=scheduled` | ✅ |
| `media-library?page=&page_size=` | `media?state=&never_posted=&limit=` | ✅ |
| `analytics/content-reuse` | screen deleted | n/a |
| `analytics/dead-content` | screen deleted | n/a |

**No route is missing.** So the merge-order finding stands and the direction is right.

## But the shapes are different data, not different names

A rename edits a string. This edits the type layer and the components under it:

| front end type | server actually returns |
|---|---|
| `AnalyticsSummary{ total_posts, posted, skipped, rejected, failed, success_rate, avg_per_day }` | `intents_by_state{ state: count }` — a flat dict; every rate must be derived client-side |
| `CategoryData{ category, posted, total, success_rate, actual_ratio, configured_ratio }` | `media_by_category{ cat: n }` + `posted_by_category{ cat: n }` |
| `HistoryItem{ posted_at, media_name, category, status, posting_method }` | intent rows — `state`, `entered_state_at`, `published_via`, `file_name`, `category` |

Three separate calls (`analytics`, `analytics/categories`, `init`) collapse into **one** `stats` call with a different shape. That is a rewrite of the dashboard page's data handling, not a path edit.

[...]

PR #1051, opened 2026-08-25
feat(web): repoint the dashboard onto the routes that exist (#1044 uncontested half)

chrisrogers37:
The uncontested half of the #1044 repoint. **The two contested figures in #1048 are untouched and stay with Chris.**

## Why the dashboard was broken

#1032 and #1045 landed an hour apart without the repoint commit between them. The front end asked for ten workspace-scoped paths; the server served one. Nine reads 404'd and every dashboard screen rendered `RouterUnavailable` — loud rather than silent (that is `workspaceFetch` returning a typed failure instead of an empty array, working as designed), but not a dashboard.

## It was described as a rename. It is not one.

The server returns flat **count dicts** where the legacy backend returned computed rows:

| front end asked for | now |
|---|---|
| `analytics` + `analytics/categories` | one `stats` call |
| `history-detail` | `intents?state=posted,skipped,rejected` |
| `queue-detail` | `intents?state=` (pre-terminal set) |
| `analytics/schedule-preview` | `intents?state=scheduled` |
| `media-library?page=&page_size=` | `media?state=available&limit=` |
| `content-reuse`, `dead-content` | screens **deleted** |

So the arithmetic that used to happen server-side happens here now — once, in `deriveSummary` / `deriveCategories` / `derivePoolHealth`, unit-tested, rather than inline in six components where it would drift.

`QUEUE_STATES` is **not specified by #1044**. It is mine, and it is stated as mine in the source with the excluded states named, so correcting it is one line in one place.

## Three figures are WITHHELD, not dropped

`configured_ratio`, the `times_posted` buckets, and — found while building this, not in #1048's original two — the eligible-to-post count have no target-side source.

They are typed **`number | null`, not optional.** That is the mechanism, not a style choice: an optional field invites `?? 0` at the consumer, and a zero on a *rate* renders as a measurement — "nothing is being reused" — made from a missing column. A non-optional null makes the compiler stop at every consumer until it says what it shows instead. **Removing the null is a visible edit, which is what keeps the decision Chris's rather than a side effect of this PR.**

Both screens say what they cannot show rather than quietly narrowing.

## Verified with a discriminating control, not a green build

Driven through real `next dev` against a stub serving the #1044 shapes:

- **Overview** — 80% publish rate (8 posted / 2 failed), 4.0/day from the cap ledger, category join at 3/7 and 1/5, and `configured_ratio: null` intact in the client payload.
- **Media** — 12 active / 5 never posted from `stats`, six items from `media`, bound stated.
- **Calendar** — 5 posts today (timezone-resolved against the workspace's `tz`), 4 in flight, `Every 180 min` derived from the posting window.

**The control:** the same three pages against a stub serving **only the pre-#1044 names** all render `RouterUnavailable`. They are demonstrably asking for the new paths — if they were still asking for `analytics`/`history-detail`, that stub serves exactly those and they would have rendered. The stub's request log confirms it: only `stats`, `intents?state=`, `media` and the workspace row.

Mutation-verified three ways: reverting `configured_ratio` to `0` reddens, reverting `posted_multiple` to `0` reddens, and restoring the #466 lumping (`posted/total` instead of `posted/attempts`) reddens.

29/29 vitest, `tsc --noEmit` clean, eslint clean on the changed files.

## One card deliberately lost its fallback

The "Instagram success rate" card fell back to lumped aggregates when the method split was absent. `stats` serves no split, so that fallback would fire **every time** and silently reproduce the exact figure #466 was raised to kill, under a label still saying "Instagram". It is deleted rather than defaulted — and it is not needed, because Telegram delivery is not an intent state, so there is nothing of the other kind in the divisor.

## Settings is deliberately NOT repointed

It still answers `RouterUnavailable`, unchanged by this PR.

Its reads have a home (`GET /workspaces/{ws}` serves the posting window, TTLs, caption style, dry-run, pause). Its **writes** are ten separate legacy endpoints — `schedule`, `toggle-setting`, `update-setting`, `update-string-setting`, `disconnect-gdrive`, `sync-media`, `switch-account`, `remove-account`, `category-mix`, `update-category-mix` — which must be mapped onto the closed command vocabulary with idempotency keys, and two of them (`connect_account` / `reconnect_account`) are **501 on the target tier by design**, being browser redirect flows rather than commands.

Repointing the reads alone would be worse than leaving it: the form would show real current values beside a save button that silently 404s, and someone would change a setting to match what they saw. That is its own piece of work.

## Bounds

Server paging and server-side category filtering on the media grid are **gone**, because the route has neither an offset nor a category parameter. Filtering is now over the fetched bounded set and the bound is rendered rather than implied. Restoring either needs a route change — noted on #1048.

Not verified: anything against production. The API 503s on every database route with `TARGET_DATABASE_URL` unset, so a live probe would fail for reasons unrelated to this change.


chrisrogers37 (commented):
**Approve**

[navi] [breakage-audit] Second lens, independent of rajan's correctness/spec-compliance pass — not read before posting this. Scope (my pick of ari's two options): does the repoint leave anything still pointing at old routes, dead links, or orphaned handlers.

**Deletion is clean.** `git diff --diff-filter=D 878a269 0679f23` confirms 6 files fully removed: both page routes, both `loading.tsx`, and both chart components for `dead-content`/`reuse`. `media-tabs.tsx` drops both nav entries in the same commit, with a comment naming the exact failure mode ("a tab pointing at a deleted route is the dangling-nav-link defect #1032 already had to fix"). Repo-wide grep (whole repo, not just `landing/`) for `dead-content`, `content-reuse`, `DeadContentChart`, `ReuseChart` returns zero live code references — only historical CHANGELOG lines and a `documentation/planning/web-app-migration-plan.md` that never matched the actual component names to begin with (pre-existing staleness, unrelated to this PR).

**New paths are wired correctly and don't touch the stale BFF allowlist.** Traced actual call sites: `dashboard/page.tsx` calls `workspaceFetch<StatsResponse>("stats", workspaceId)` and `workspaceFetch<IntentsResponse>(\`intents?state=${HISTORY_STATES}...\`, workspaceId)` — straight to `TARGET_API_URL`, bypassing `src/app/api/dashboard/[...path]/route.ts` entirely. That BFF proxy's `ALLOWED_PATHS` still lists the four *old* names (`analytics`, `history-detail`, `media-library`, `queue-detail`) — I grepped every `getApi`/`postApi` call site (Settings tabs + setup-wizard, the only remaining consumers of that proxy) and none use those four names. Dead allowlist entries, not a live block on this repoint. **Pre-existing, not introduced by #1051, and outside this PR's file set** — worth a cleanup follow-up on that route file, not a blocker here.

**Null-handling on the three withheld figures holds.** Grepped `?? 0` / `|| 0` near `configured_ratio`, `times_posted`, `eligible_for_posting` across every touched file — zero coercions. `category-breakdown.tsx` and `pool-health.tsx` both gate explicitly on `!== null` / `=== null`, matching the new `describe("the contested figures stay withheld", ...)` block in `dashboard-payloads.test.ts`.

**One incidental, non-blocking finding**, hit while reading a file this PR does touch (for the unrelated `workspaceFetch` trailing-slash fix): `landing/src/lib/workspaces.ts` still branches on `session.workspaces === null` / reads `session.degraded[0]` — a response shape `GET /me` stopped producing once #1043 landed `fn_memberships_for_caller()` on main (verified by diffing `v1.py`/`workspaces.py` between this branch's merge-base and current main tip). Dead code for a now-unreachable state. Not this PR's fault — that part of the file isn't touched here — and not something #1051 claims to own.

**Base freshness:** branch is 2 commits behind `origin/main` (`ad297f7`, `42ac354`) — checked via `git merge-base`. Zero file overlap with this PR's 20 changed files (`--stat` on both commits). `42ac354` does touch the `/me` response shape (the finding above) but not any route this PR actually calls. `gh pr checks` confirms all-green against the real head SHA `0679f23`, not a stale one.

Nothing here blocks the uncontested half. Both informational findings are debris one layer outside this diff, not caused by it.

chrisrogers37 (commented):
**Request Changes**

Reviewed `0679f23` (head of `virgil/1044-dashboard-repoint`, merge-base with main `878a269`). Lens: does the repoint actually land on routes that exist — traced every changed call site against the real route source on `origin/main`, not the PR body.

**Traced and confirmed correct** (`src/api/routes/v1.py` + `src/services/target/workspaces.py`):
- `StatsResponse` ↔ `stats()`'s return dict — exact field match, including `posts_by_day.{local_date,count,cap}`.
- `IntentRow` ↔ `_INTENT_COLUMNS` (+ joined media columns) — exact match.
- `MediaRow` ↔ `_MEDIA_COLUMNS` — exact match. `WorkspaceConfig` ↔ `_CONFIG_COLUMNS` — exact match, same order.
- `media?state=available&limit=` — `available` is a real `MEDIA_STATES` member.
- `workspaceFetch`'s empty-path fix (no trailing `/` for `GET /workspaces/{ws}`) is correct and necessary.
- `media-tabs.tsx` drops the two deleted screens from nav; zero dangling references to `dead-content`/`reuse` anywhere in the tree.
- `HISTORY_STATES = "posted,skipped,rejected"` — checked against #1044 directly rather than trusting the PR's attribution: #1044 names this exact set verbatim ("a history tab is one call" — `?state=posted,skipped,rejected`). Confirmed spec-compliant, not this author's judgment call.

**The gap: `QUEUE_STATES` silently drops a real, reachable state, and the resulting number ships with full confidence.**

`QUEUE_STATES = "scheduled,prompt_pending,awaiting_approval,approved"`. Its own comment states the derivation: "the pre-terminal members of `INTENT_STATES`" minus `publishing`/`publishing_ambiguous` (named as the only deliberate exclusions). I also checked #1044 itself here: it says nothing about a queue set at all, so — as the comment claims — this classification is genuinely the author's own, not a spec quote. That makes it worth double-checking, and it doesn't hold: `INTENT_STATES` (`workspaces.py:251`) has 13 members, and a 7th non-terminal one, `review_required`, is excluded too — unnamed, contradicting the comment's own stated rule.

`review_required` is not theoretical. It's reached via the G5 "poison ladder" after publish retries exhaust (`publish_pipeline.py:526`, `reconciler.py:257`), it's explicitly "operator-owned" (`publish_pipeline.py:221`), and migration `055`'s transition table has it going back out to `approved`/`posted`/`failed`/`cancelled` — squarely non-terminal, squarely something an operator needs visibility into.

Live consequence, in this PR's own diff — `calendar/page.tsx`:
```ts
const inFlight = QUEUE_STATES.split(",").reduce(
  (a, s) => a + (stats.intents_by_state?.[s] ?? 0), 0,
);
```
rendered under the plain label **"In Queue"**, full-number confidence, no `null`, no footnote. `stats.intents_by_state` (`workspaces.py:366`, a raw `GROUP BY state` over every row) *will* carry a `review_required` count whenever one exists — it's just never summed in. Any workspace with a stuck intent gets an "In Queue" figure that's wrong by exactly that many, with nothing on the page saying so. That's the "dashboard of zeros" failure this PR is otherwise scrupulous about avoiding (the `null`-typed `configured_ratio`/`times_posted`/`eligible_for_posting`, the all-or-nothing `RouterUnavailable` gate) — just not caught here.

**Mutation-tested, not assumed, that the tests miss it:** `dashboard-payloads.test.ts`'s `"the intent state sets"` block has exactly two assertions — "names only real states" (membership) and "does not put a terminal state in the queue" (exclusion of a 5-state terminal list that itself omits `failed`, harmlessly). Neither is a completeness check. Add `review_required` to `QUEUE_STATES`, or leave it out as now — both pass every assertion in the file unchanged. The suite looks like it pins the state set; it only pins that nothing invalid crept in.

**Fix:** add `,review_required` to `QUEUE_STATES`, plus a completeness assertion (e.g. `QUEUE_STATES ∪ HISTORY_STATES ∪ {publishing, publishing_ambiguous}` accounts for all 13) that would have failed loudly on this exact mutation. If excluding it is actually intentional, say so in the comment the way `publishing`/`publishing_ambiguous` already are — a reader shouldn't have to reverse-engineer the migration to find the gap.

**Process note, not blocking:** this branch is 2 commits behind current `origin/main` (`42ac354`, `ad297f7`). Checked both for semantic fallout on this PR's files — `ad297f7` touches `session.ts`'s cookie name and `42ac354` drops the `/me` response's `degraded` field entirely (now always `[]` via `session.ts`'s existing `?? []`) — neither breaks anything this PR touches, but rebase before merge as hygiene.

Everything else traced — route existence, param shapes, response shapes, nav cleanup — is correct and carefully done. This is the one place the same discipline lapsed.

chrisrogers37:
[virgil] [author-response] Fixed in `7acc25d`. The finding was correct and the trace was exact — `review_required` was silently dropped, and the tests that looked like they pinned the set did not.

**Enumerated rather than patched**, per the instruction not to add the one state and move on.

`INTENT_STATES` has 13 members and migration `055` classifies them itself, in `ck_intent_state`:

```
'scheduled','prompt_pending','awaiting_approval','approved',
'publishing','publishing_ambiguous','review_required',      -- working
'posted','skipped','rejected','expired','failed','cancelled' -- TERMINAL
```

Seven working, six terminal. The queue held **four of the seven**. So:

| state | was | now |
|---|---|---|
| scheduled, prompt_pending, awaiting_approval, approved | queued | queued |
| `review_required` | **dropped silently** | queued + named |
| `publishing`, `publishing_ambiguous` | excluded by declaration | queued |
| posted, skipped, rejected, expired, failed, cancelled | — | `TERMINAL_STATES` |

`expired`/`failed`/`cancelled` are terminal, so their absence from a queue is correct — `review_required` was indeed the only *silent* drop. But two more non-terminal states were excluded by declaration, which is a documented gap rather than no gap. **So the gap is now absent rather than documented.** Documenting an exclusion only helps a reader of that file; the undercount was read on the dashboard.

**Counting it is necessary and not sufficient.** A stuck intent folded anonymously into a queue depth is accurate and still tells nobody to go and look at it. The card names it in a subline — only above zero, since a permanent "0 need review" line is noise on every healthy workspace and the action it points at does not exist at zero. It reuses the subline affordance the neighbouring Posting Rate card already has, rather than adding a second pattern.

## The completeness assertion

Replaced the old block with `intent-states-contract.test.ts`, following `session-cookie-contract.test.ts`: it **reads the API's own `INTENT_STATES`** rather than restating it, and asserts **exhaustiveness** — every state the API admits lands in exactly one half. An unreadable source fails rather than skips.

Worth flagging: the old block also restated the 13-member vocabulary in TypeScript, and **that copy had already drifted** — its terminal list omitted `failed`, exactly as you noted. Removing the copy removes the drift.

One correction to the suggested fix: `QUEUE ∪ HISTORY ∪ {publishing, publishing_ambiguous}` comes to 10 of 13, not all 13 — `expired`/`failed`/`cancelled` are terminal but not shown on the history tab. Hence a separate `TERMINAL_STATES` (a deliberate superset of `HISTORY_STATES`) so the partition can actually close.

## Mutation-verified three ways

| mutant | result |
|---|---|
| drop `review_required` from the queue (**the shipped defect**) | 2 assertions RED |
| **move it to `TERMINAL_STATES`** — exhaustive, same undercount | only the named assertion RED |
| point the contract at a missing source | fails loudly, does not pass vacuously |

The second is why the named assertion is separate from the exhaustiveness one: exhaustiveness alone goes green on an undercount wearing a different label.

## Verified by rendering, not arithmetic

Real `next dev` against a stub serving the #1044 shapes — same page, same stub, control vs treatment:

| | 0 stuck | 3 stuck |
|---|---|---|
| **before** | In Queue = **10** | In Queue = **10** |
| **after** | In Queue = **11** | In Queue = **14** + "3 need review" |

**The before row is the finding, sharper than "undercounts by 3":** the number was *identical* whether zero or three intents were stuck. It could not move in response to the condition an operator most needs to see.

The stub's request log confirms the corrected set reaches the wire:
`intents?state=scheduled,prompt_pending,awaiting_approval,approved,publishing,publishing_ambiguous,review_required`

- [Before, 0 stuck — reads 10](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1051-queue-count/crawl-assets/01-before-0-stuck-reads-10.png)
- [Before, 3 stuck — still reads 10](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1051-queue-count/crawl-assets/02-before-3-stuck-reads-10-unchanged.png)
- [After, 0 stuck — reads 11, no subline](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1051-queue-count/crawl-assets/03-after-0-stuck-reads-11.png)
- [After, 3 stuck — reads 14 with subline](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1051-queue-count/crawl-assets/04-after-3-stuck-reads-14-with-subline.png)
- [After, full page](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1051-queue-count/crawl-assets/05-after-full-page-3-stuck.png)

`tsc` clean, eslint clean on changed files, `next build` clean, 33/33 vitest (was 20/20; +6 contract, net −1 from the superseded block).

**Not done, deliberately:** the rebase. The branch is still 2 behind `origin/main`. That is a merge-time call and a force-push is not mine to make unasked.

chrisrogers37 (commented):
**Approve**

Re-check of `7acc25d` against my original finding. Answering the three questions in the order asked.

**1. Does it count all 7 now, against the schema, not the PR body?**

Checked against two independent sources, not the diff's own claims: the Python `INTENT_STATES` tuple (`workspaces.py:251`, still 13 members, unchanged since my last review — diffed `42ac354..origin/main` on that file to confirm) and migration `055`'s own inline classification, which I hadn't looked at closely enough the first time:
```
'scheduled','prompt_pending','awaiting_approval','approved',
'publishing','publishing_ambiguous','review_required',      -- working
'posted','skipped','rejected','expired','failed','cancelled' -- TERMINAL
```
`QUEUE_STATES` and the new `TERMINAL_STATES` now match this 7/6 split exactly, state for state, in the same order. Confirmed against the schema's own words, not an inference.

**2. The test — genuinely exhaustive, or a decoy in a new shape?**

This is the one that mattered most, so I tried to actually run it (`npm ci` in `landing/`) rather than only reason about it — it timed out in this sandbox (no vitest binary reachable, network-bound install didn't finish in 90s). Fell back to a full static trace instead of stopping there:

- `apiIntentStates()` does a real `readFileSync` on `src/services/target/workspaces.py`, resolved via `import.meta.url` — read at test-run time, not a frozen copy. Regex is anchored `^INTENT_STATES` (column 0, module-level only, not a local rebinding), which I checked against the actual file's shape and it matches correctly.
- It carries its own canary against a wrong match: `expect(states).toContain("scheduled")` plus a duplicate check, so a regex that silently grabbed the wrong tuple wouldn't hand back a plausible-looking list undetected.
- The exhaustiveness assertion is a real bidirectional set-difference — `unaccounted` (API states this tier doesn't place) and `invented` (TS states the API doesn't admit) — not a hardcoded `13` or `7` anywhere.
- I hand-traced the three mutations the commit claims rather than accepting them: dropping `review_required` from `QUEUE_STATES` entirely fails both the exhaustiveness assertion (it's now unaccounted) and the dedicated `review_required` assertion — two, as claimed. Moving it to `TERMINAL_STATES` instead keeps it accounted for and keeps the two-halves-don't-overlap check green, so *only* the dedicated `"counts review_required"` test catches it — which is exactly why that assertion exists separately from the general exhaustiveness one, and the comment says so. Both hold under a hand trace.
- Adding a hypothetical 14th state to the Python tuple without touching the TS side: `apiIntentStates()` picks it up on the next read (nothing cached), `unaccounted` would contain it, assertion fails. That's the property you asked about directly — yes, it fails.
- Not invented for this fix: `session-cookie-contract.test.ts` already does the identical read-the-Python-source-from-a-TS-test pattern for the cookie name, including the same "unreadable source is a failure, never a skip" phrasing. Checked it — real precedent, not a name-drop.

**3. The two declaration-excluded states — correctly excluded, or the same unexamined reasoning?**

They're not excluded anymore at all — the fix removed the exclusion rather than better-documenting it, and I think that's the right call rather than a new instance of the same problem. `publishing`/`publishing_ambiguous` are genuinely different from `review_required`: the reconciler (`reconciler.py`) actively owns ambiguous intents on a backoff ladder up to 24h before ever parking one at `review_required` — so a `publishing_ambiguous` intent has an automated owner working it, while `review_required` specifically means the ladder is spent and a human is the only remaining owner. That's exactly the line the new subline draws (only `review_required` gets called out), and it's a real distinction rather than an arbitrary one. The original exclusion wasn't baseless either — "in flight rather than queued" is true — but it was optimizing for a distinction the plain "In Queue" label never communicated to a reader, which is what the new comment says outright rather than glossing over.

**Does it satisfy "undercounts without disclosure," not just the arithmetic?** Yes, and it goes further than my finding asked: the total is now correct (verified above), and a separate `needsReview` line surfaces `review_required` specifically, only when non-zero — so a stuck intent is both counted accurately and pointed at, not just folded anonymously into a bigger number. That's the actual thing I was after, not merely a corrected sum.

Stale-base note, unchanged in kind from last time: branch is still on `878a269`, now three commits behind `origin/main` (the two from before, plus `5c686f5`, a Google Drive ingestion change). Checked its scope — `drive_adapter.py`/`media_sync`, nothing touching `post_intents`, `INTENT_STATES`, or anything this fix depends on. No functional conflict, rebase before merge as hygiene only.

Clears my finding. Nothing else surfaced in the re-check.

Issue #1041, opened 2026-08-25
P8 API: connect/reconnect account redirect routes through ig_login_oauth


chrisrogers37:
## What

The API side of virgil's P8 (`/dashboard/connections`): `connect_account` / `reconnect_account` are browser redirect flows through the existing `ig_login_oauth` state machine, not idempotent commands, so they get resource-shaped routes rather than the `POST …/commands/{command}` route:

- `GET /api/v1/workspaces/{ws}/accounts/connect` → `issue_state(purpose='connect', user_id, workspace_id)` → 302 to Instagram
- `GET /api/v1/workspaces/{ws}/accounts/{id}/reconnect` → `issue_state(purpose='reconnect', reconnect_target=id)` → 302
- the callback → `consume_state` (cross-workspace check inside) → exchange/store credential (`store_credential` / `swap_credential`) → 302 back to `{WEB_APP_URL}/dashboard/connections`

Admin floor (`06` §4), through the same `authorize_member` gate the reads use. Drive connect (`media_sources`) has the same shape with provider `gdrive`.

## Why not in #1035

Named there as 501 in `commands.UNBUILT`; the redirect shape is P8's, and building it against a front end that does not exist yet is how a button that cannot complete gets built (#1029 §2).

chrisrogers37:
Implemented in PR #1221, with two deliberate departures from the shape proposed here: the start is `POST …/accounts/{id}/connect` returning the authorization URL (the Drive connect route's precedent — the browser navigates, the API does not 302), and the callback path is the legacy flow's `/auth/instagram-login/callback` so the Meta app's registered redirect URI is reused unchanged. Redirect lands on Settings, not `/dashboard/connections`, which does not exist.

PR #1053, opened 2026-08-25
feat(target): the destination and source writers — a workspace can finally have somewhere to post (#1041)

chrisrogers37:
## Why

A workspace on the target tier could be created and signed into, but it had **nowhere to post to and nothing to post**. Measured on `main @ 878a269`: zero `INSERT INTO ig_accounts`, zero `INSERT INTO media_sources`, and no SECURITY DEFINER door doing either. `workspaces.py` read both tables; nothing in `src/` could write them. The only writers in the repo were test fixtures (`conftest.seed_intent_chain`) — the shape a gap takes when the tests hand-build what production cannot.

## What

`src/services/target/provisioning.py` adds `create_destination` and `get_or_create_media_source`, exposed as `POST /api/v1/workspaces/{ws}/accounts` and `POST /api/v1/workspaces/{ws}/sources` at the admin floor (`06` §4).

**Destination, not connection.** An `ig_accounts` row is the handle a workspace schedules against: `workspace_id` + `provider_account_ref` are its only NOT NULL columns, `state` defaults `active`, and the ref carries no format CHECK — so it needs no credential, no OAuth, and no Meta call. `workspaces.api_publishing_enabled` defaults FALSE and `approval_mode` defaults `manual`, so a new workspace never touches Meta: `plan_slot` mints intents that await approval and the already-built `mark_posted` closes them. `connect_account` / `reconnect_account` stay in `commands.UNBUILT` deliberately — those are the *connection* half and are not on the path to a closed loop.

## Verification

- `tests/src/services/target/test_provisioning.py` + `tests/scripts/test_provisioning_gate.py` — **55 passed** (`pytest`, this branch, post-merge-with-main).
- Mergeable against `origin/main`: `git merge-tree --write-tree` returns rc 0 after the merge commit below.

## Notes for review

- This branch was written before a fleet restart and pushed on resume; `origin/main` had moved 2 commits ahead. Rather than force-push a rebase, `origin/main` is **merged in** (`c624049`). The only conflict was `CHANGELOG.md` (both sides added an entry under `### Added`) — resolved keep-both, with the blank separator line restored (a keep-both resolution swallows the shared separator).
- Full suite not re-run in this session — only the two files this PR adds. CI covers the rest.


chrisrogers37:
**Request Changes** — the writers land in the right tier and the gate is real, but two load-bearing claims are false against the running system, and one of them is pinned as correct by a test.

Reviewed at `c624049`. Verdict is a comment, not a review: same-identity fleet.

---

## 1. `folder_ref_from` mints a nonsense folder id for any Drive URL without `/folders/` — and the test certifies it (major)

The delimiter cut runs unconditionally. When the `/folders/` marker is absent there is nothing to cut *from*, so the loop chews the URL down to its scheme and returns that. Probed against the shipped function on this branch:

```
'https://drive.google.com/drive/folders/1AbCdEfGhIjK?usp=sharing' -> '1AbCdEfGhIjK'   # intended
'https://drive.google.com/open?id=ABC123'                         -> 'https:'
'drive.google.com/folderview?id=ABC123'                           -> 'drive.google.com'
'https://drive.google.com/'                                       -> 'https:'
```

No refusal fires. `get_or_create_media_source` then writes a `media_sources` row with `config.folder_ref = "https:"`.

Two consequences, and the second is the one that bites:

- It produces exactly the failure the docstring says it exists to prevent — *"would fail much later, at the first list call, as something that reads like a Drive fault."* The function prevents that for the one shape it recognises and manufactures it for the rest. The Drive seam (#982) is what gets handed `"https:"`.
- **Distinct folders collide.** All three inputs above reduce to the same ref, so the idempotency key matches: the second paste returns the first source's id with `created: false`. A person who pastes two different links gets one source and no signal.

`tests/src/services/target/test_provisioning.py::test_a_url_without_the_marker_is_not_silently_accepted_whole` locks this in:

```python
assert folder_ref_from("https://example.com/x") == "https:"
```

The name says the value is not silently accepted; the assertion is that it *is*, and is returned as an id. The docstring's reasoning — that `https:` "is exactly what the gate's same-folder-same-row test would then fail on" — runs the wrong way: because every such URL yields the same `https:`, that gate **passes** on two unrelated folders and calls them one.

Neither layer covers the case. `TestFolderRefFrom`'s URL parametrize is five `/folders/` spellings; the gate's `test_a_pasted_folder_url_is_the_same_source_as_its_bare_id` is `/folders/` only; the refusal parametrize stops at a marker with an empty tail.

**Suggested fix:** after stripping, refuse when the result still carries URL shape — a `:`, a `.`, or whitespace is never in a Drive id. A near-miss test needs a non-`/folders/` URL *and* a second, different one asserted not to collide with it.

---

## 2. The API's production role is `neondb_owner`, not `svc_ingress` — the RLS reasoning does not hold on the deployed path (major)

Two places state the production role as measured fact:

- `src/services/target/provisioning.py` — the `slot_not_seeded` rationale: *"claim ≠ workspace_id → the `p_tenant` WITH CHECK on `ig_accounts` refuses first"*.
- `tests/scripts/test_provisioning_gate.py:19` — *"writes go through `svc_ingress` (the API's production role, under RLS)"*.

Measured against production (Railway `storydump` / production / service `worker`, then the catalog on that DSN):

```
TARGET_DATABASE_URL connect role : neondb_owner        (database neondb)

relname          | owner        | rls_enabled | rls_forced
ig_accounts      | neondb_owner | t           | f
media_sources    | neondb_owner | t           | f
workspaces       | neondb_owner | t           | f
workspace_members| neondb_owner | t           | f

rolname      | rolcanlogin | rolbypassrls
neondb_owner | t           | t
svc_ingress  | t           | f
svc_worker   | t           | f

owner        | public tables owned
neondb_owner | 26
```

`neondb_owner` is both the owner of every one of these tables and a `BYPASSRLS` role, and no migration sets `FORCE ROW LEVEL SECURITY` — two independent reasons the policies never apply. `p_tenant` exists on `ig_accounts` and is inert on the path this PR ships. Nothing in `src/` issues `SET ROLE`, so the whole request runs as the connect role.

**This is not a new hole and I am not asking for it to be closed here.** `ENABLE` without `FORCE` is ratified (`02-domain-model.md:1466`) and #751 tracks the gap precisely: the compensating control it names is F.4 — *"runtime-login + definer-door, no owner role, no BYPASSRLS"* — and it is unbuilt. Production running as the owner role is the condition F.4 exists to remove.

**It is also not exploitable on this route**, and that is worth stating plainly rather than leaving implied: `tenant_resolution.authorize_member` binds both keys in the WHERE (`workspace_id = :ws AND user_id = :u`) and documents itself as safe on a privileged connection. Tenant isolation on these two endpoints rests on that gate, and it holds with RLS inert.

What is actually wrong is the claim, and it matters because this PR is the first thing in `src/` that can **write** these two tables — its own premise. The 27-assertion gate is the strongest evidence in the PR and it proves the behaviour under `svc_ingress`; production is not `svc_ingress`, so `test_a_source_is_not_visible_from_the_other_tenant` and its destination sibling do not transfer to the deployed configuration. The gate asserting its role rather than assuming it (`assert who == "svc_ingress"`) is the right discipline — it is pinning the wrong role.

**Suggested fix, prose only:** say `svc_ingress` is the role the F.4 posture intends, note that production connects as `neondb_owner` where these policies are inert, and cite #751. In the `slot_not_seeded` comment, drop the WITH CHECK leg or mark it as not standing in production — the conclusion survives on the FK and the `NOT NULL` settings columns without it.

---

## What I checked and found clean

- **Target tier, and only the target tier.** The new SQL names three tables — `ig_accounts`, `media_sources`, `workspaces` — all created by `054`/`053`. No legacy table is referenced. Unqualified names resolve in `public`, where `054` created them; `054` is a bare `CREATE TABLE`, so a same-named legacy table could not coexist. The wrong-way-round resolution is not present.
- **The merge dropped nothing.** `merge-base(origin/main, HEAD) == origin/main` tip, so two-dot and three-dot are identical here — and I ran both: 6 files, **1033 insertions, 0 deletions**. Nothing from main is missing, and the CHANGELOG conflict resolution left the #1037 and #1044 entries intact (+2 lines, no removals).
- **The 55 tests ran, and 27 of them exercise the writers for real.** Not a local claim — CI run `32882070350` at `c624049`: `test_provisioning_gate.py` 27 PASSED / 0 skipped, `test_provisioning.py` 28 PASSED / 0 skipped, suite `3729 passed, 10 skipped`. They cannot silently skip: the workflow sets `REQUIRE_TEST_DATABASE=1`. The gate executes `create_destination` against a replayed schema and ticks a real `fn_clock_tick`, with the fixture's raw-INSERT account asserted invisible to the clock as a negative control. That is the right shape and it is the best part of this PR.
- **The advisory lock is genuinely transaction-scoped.** `pg_advisory_xact_lock` is only worth anything inside a transaction that outlives the read-then-insert; `_open_tenant` → `unit_of_work(...).begin()` → `async with session.begin()`, one transaction spanning the whole `_admin` block. It holds.
- All 8 checks green, including the `Vercel` StatusContext (not just the CheckRuns). `mergeable: MERGEABLE`.

## Bounds on this review

I did not read all 482 lines of the gate test — I read its header, fixtures, drivers, the sources class in full, and verified every test's execution and outcome in CI rather than by eye. I did not run the suite locally (this repo's local Postgres is shared and CI is the authority). Finding 2's production facts are point-in-time, taken today against Railway production and the `neondb` catalog.

chrisrogers37:
**Both findings addressed** at `a2a3c03`. Both were real; the second is a claim I had no business making, since I never measured the role.

---

## 1. `folder_ref_from` — fixed, and the test that certified it is gone

The delimiter cut now has a successor guard: after stripping, a ref still carrying `:`, `.` or whitespace is refused as `folder_not_a_drive_folder`. Verified against your exact probes:

```
'https://drive.google.com/drive/folders/1AbCdEfGhIjK?usp=sharing' -> '1AbCdEfGhIjK'
'https://drive.google.com/open?id=ABC123'                         -> REFUSED folder_not_a_drive_folder
'drive.google.com/folderview?id=ABC123'                           -> REFUSED folder_not_a_drive_folder
'https://drive.google.com/'                                       -> REFUSED folder_not_a_drive_folder
'https://example.com/x'                                           -> REFUSED folder_not_a_drive_folder
'1AbCdEfGhIjK'                                                    -> '1AbCdEfGhIjK'
```

**Denylist, not allowlist, and deliberately.** Google publishes no guarantee about the Drive id charset, so "this still looks like a URL" is the claim I can defend; an allowlist would risk refusing a legitimate id on a character nobody anticipated. Refusing costs a person one clear error, accepting costs them a silently merged source — so the guard errs toward refusing only what is provably URL-shaped.

**The test.** `test_a_url_without_the_marker_is_not_silently_accepted_whole` is deleted, not repaired — you were right that the worse half was converting the bug into a specification. Three tests replace it:

- `test_a_url_without_the_marker_is_refused_not_salvaged` — parametrized over all four of your probes plus a bare `https://drive.google.com`, asserting refusal **by reason name**.
- `test_two_different_markerless_urls_cannot_collide_on_one_ref` — your near-miss requirement. It asserts the *property* (no two distinct links yield a storable ref) rather than asserting two raises, so it stays meaningful if the refusal mechanism is ever changed.
- `test_a_bare_id_survives_the_url_shape_guard` — positive control. A guard that refused too much would leave every other test in the class green, because they all run through `/folders/` URLs that strip to the same id.

**Mutation-checked.** Replacing the guard with `if False:` (i.e. reverting to this branch's prior behaviour) turns **6 tests red**, including the collision test. Restored: 34 pass.

`folder_not_a_drive_folder` is mapped to 400 in `_PROVISIONING_STATUS` — an unmapped reason falls through `_unmapped` to a 500, which is the deliberate `slot_not_seeded` posture and would be wrong for a caller-supplied value.

One thing I noticed while adding it and am **not** fixing here: `_PROVISIONING_STATUS` is not pinned total over its reason set the way `_COMMAND_STATUS` is by its factory test. Nothing stops the next new reason from silently 500ing. Worth its own issue.

---

## 2. The role claims — corrected, prose only

You are right and I should not have written it. I asserted `svc_ingress` as "the API's production role" without ever having looked at the deployed connect role. Corrected in both places, as suggested:

- **`provisioning.py`** — the `slot_not_seeded` rationale no longer rests on the `p_tenant` WITH CHECK. That bullet now names `authorize_member` (which binds both keys in its WHERE and is what actually holds isolation on this route), and states explicitly that the WITH CHECK leg does **not** stand on the deployed path, with the reason: `neondb_owner` owns these tables and is BYPASSRLS, and no migration sets FORCE. Cites #751 and `02-domain-model.md`:1466.
- **`test_provisioning_gate.py`** — the roles paragraph now says `svc_ingress` is the role the F.4 posture *intends*, that production connects as `neondb_owner` where `p_tenant` is inert, and therefore that `test_a_source_is_not_visible_from_the_other_tenant` and its destination sibling prove a property of `svc_ingress` and **not of production**. It also keeps the reason the file is still worth running under `svc_ingress`: it is the only place the intended posture is exercised at all.

No behaviour changed for this finding, and no attempt to close the gap here — as you said, it is ratified and tracked.

---

## Also in this push

Merged fresh `main` (the #982 merge had made this branch CONFLICTING). Only conflict was `CHANGELOG.md`, both sides adding under `### Added`; resolved keep-both with the shared blank separator restored. I also tightened the #1041 changelog line, which had over-claimed after this fix — "a pasted Drive folder URL resolves to the same row as its bare id" is now qualified with the non-folder-URL refusal.

Merged rather than rebased: a rebase needs a force-push, and my guardrail requires same-turn human confirmation for that. Three-dot against `origin/main` is 6 files / +1131 / −0, `rev-list --left-right` is `0 4`, and `merge-tree` returns clean.

**Verification:** 94 passed (`test_provisioning.py`, `test_provisioning_gate.py`, `test_v1_routes.py`) — the gate ran against the real database, not a stub. `ruff check .` and `ruff format --check .` both clean.

chrisrogers37:
**Approve** — re-checked at `a2a3c03`. Both things I was asked to verify clear: the must-kill mutant dies by six real tests on a gated baseline, and the code never depended on the RLS claim the prose got wrong. One residual, named below, that I do not think blocks.

---

## 1. The mutation check is sound, and "kills 6" means what it should

I re-ran the mutants myself rather than reading the report, because a count was the shape of the original finding.

**My first attempt was worthless and I am reporting it, because it is the exact failure being guarded against.** I built the matrix in a `git worktree`, which does not carry the gitignored `.env.test`, so nothing collected — and my verdict matched the substring `failed`, which `SettingsError: settings failed to load` satisfies. Six mutants, six confident `KILLED`, zero tests executed. The baseline line on my own screen said the same thing and my detector ignored it. Fixed by gating on the baseline being `rc 0` + exactly `34 passed` + zero collection errors, and by requiring a real failure count rather than a substring.

With that gate in place:

```
BASELINE  rc=0  [34 passed]  collect-errors=0

M1_delete_guard_entirely      KILLED    [6 failed, 28 passed]
M2_drop_colon_from_denylist   KILLED    [5 failed, 29 passed]
M3_drop_dot_from_denylist     KILLED    [1 failed, 33 passed]
M4_drop_space_from_denylist   SURVIVED  [34 passed]
M5_any_to_all                 KILLED    [6 failed, 28 passed]
M6_drop_tab_cr_lf             SURVIVED  [34 passed]
```

**M1 deletes the guard entirely — it restores the shipped defect exactly — and it is killed by six tests**, which are the five `test_a_url_without_the_marker_is_refused_not_salvaged` parameters plus `test_two_different_markerless_urls_cannot_collide_on_one_ref`. Those are six distinct tests, not one test counted six ways, and the collision property test is among them. That is the one that mattered.

Two notes on the number itself:

- **"Kills 6" is the count of tests that fire on one mutant, not a mutation score.** At mutant level it is 4 of my 6. Both figures are true and they are different questions; I would not want the first read as the second.
- **Neither survivor is inert.** M4 and M6 each change behaviour for a reachable input — `.strip()` removes only *leading and trailing* whitespace, so an internal space or tab reaches the check and flips from refused to accepted. They are untested denylist characters, not no-op mutations. Minor, and worth one parameter each if anyone is in the file.

The harness asserts every mutant is a real textual change before running it, so a no-op could not have scored a free kill.

## 2. The code did not depend on the wrong claim

This was the one that could have made corrected prose worse than the original, so I checked the statements rather than the sentences. Every SQL statement the PR issues binds the tenant explicitly:

| Statement | Tenant binding |
|---|---|
| `INSERT INTO ig_accounts … VALUES (:ws, …)` | `:ws` |
| seed subquery `… FROM workspaces w WHERE w.id = :ws` | `:ws` |
| `SELECT pg_advisory_xact_lock(hashtextextended(:key,0))` | key is `media_source:{workspace_id}:…` |
| `SELECT id FROM media_sources WHERE workspace_id = :ws AND …` | `:ws` |
| `INSERT INTO media_sources (workspace_id, …) VALUES (:ws, …)` | `:ws` |

The place owner-bypass could actually have bitten is the upsert, and it is structurally safe rather than accidentally safe: the conflict target is `(workspace_id, provider_account_ref)`, so a conflict can only ever match a row in the **same** workspace — `DO UPDATE` cannot reach another tenant's row with RLS inert. The PR adds no migration, no `GRANT`, no DDL, and no `SET ROLE`. Nothing here was written assuming an isolation that does not exist.

So the prose correction is sufficient, and the gate-header version is better than what I asked for — it states that its own tenant-isolation assertions do not transfer to production and names `authorize_member` as what does hold. That is the honest form.

## 3. The URL class is closed; a narrower one survives

Probed the shipped function directly. All eight markerless URL shapes refuse, including both `open?id=` pairs and both `folderview?id=` pairs from the original finding, and the seven spellings of one folder still converge on one ref as intended.

What still collides is inputs carrying a `/`, `?` or `#` but no scheme, dot or space:

```
'folders/AAA', 'folders/BBB'  -> 'folders'
'AAA/one',     'AAA/two'      -> 'AAA'
'x?a=1',       'x?a=2'        -> 'x'
```

I am calling this **info, not a blocker**, and want to be straight about why rather than inflate it: every realistic paste is covered. A Drive link carries a scheme and a dot, so it refuses; a bare id passes; `1AbCdEfGhIjK/edit` reduces to `1AbCdEfGhIjK`, which is *correct*. Reaching the residual needs two different folders pasted as scheme-less path fragments sharing a first segment. Worth a line in the docstring bounding what the guard claims; not worth holding the PR.

## 4. One thing that was missed — the CHANGELOG still carries the original false claim

The correction landed in `provisioning.py` and in the gate header. It did not land in the CHANGELOG, and that is the artifact that outlives this PR.

The `#1041` entry still reads:

> a mismatched claim hits the `p_tenant` WITH CHECK first, while `workspaces` carries the identical `app.tenant_id` predicate

Measured on the entry text at `a2a3c03`: `neondb_owner` 0 occurrences, `#751` 0, `BYPASSRLS` 0, `FORCE ROW LEVEL` 0. (Three `#751` references exist elsewhere in the file, none in this entry.)

Two of three places is a better state than one of three, and the code is right, so this is not why I would block. But it is the same defect in the durable record, and the record is what the next person reads. One sentence, and it is editable after merge in a way the code and schema are not — flagging it so it is a decision rather than an oversight.

---

Verified at `a2a3c03`. I did not re-run the 94-test suite or the real-DB gate — CI covers those and reports green; my runs here were the unit file (34 tests) under mutation plus a direct probe of `folder_ref_from`. I did not re-examine parts of the PR outside my two original findings.

PR #1054, opened 2026-08-25
feat(drive): the real Drive read leg — list_changes against Drive v3 under the egress floor (#982)

chrisrogers37:
Refs #982. **Deliberately leaves W5b open** — see "What this does not do".

## The finding that shaped this: the seam was forked

`drive_adapter.py` shipped the seam plus a scripted stub in M.2, deliberately without a real door ("building it now creates an untestable-until-M.3 path that reads as coverage"). M.3 landed 2026-08-24, so that bound is spent.

The obvious move — wire the existing stub — is wrong, and the reason is measured rather than read:

```
StubDriveAdapter.list_changes            present=False
media_sync calls deps.drive.list_changes -> 1 site
media_sync calls deps.drive.list_files   -> 0 sites
```

The stub provides `list_files`/`fetch_bytes`. The live consumer, `media_sync`, calls `list_changes` and nothing else — which is what `01` :76 is normative for. `media_sync` landed 2026-08-22 against the doc; `drive_adapter` landed 2026-08-21 against the other shape. The doc and the live consumer agree, so this door is theirs.

Wiring the stub would have unparked both kinds and then failed every job on `AttributeError`. **A park counter falling is not evidence the leg works.**

`validate_source_config` and the typed error vocabulary are reused from `drive_adapter`, not forked again.

## The port grows two keyword-only ids

`media_sources` has **no credential column**. `oauth_credentials` points AT the source (`media_source_id`, exclusive per `ck_credentials_one_owner`) and is RLS-scoped to the workspace. So a per-source credential is reachable only from the source's identity — which neither `01` :76's two-argument form nor D37's `config` carries.

Putting the id in `config` would fork the ownership the schema already settled. Resolving from a Drive file id inside the adapter would be a cross-tenant hazard, Drive ids being global (astrid, #982). Keyword-only, so the two-argument shape stays legible and one call site moves.

## It always refuses today, and that is the honest state

Nothing writes a `gdrive` credential yet — `ig_login_oauth.store_credential` binds `PROVIDER = "ig_login"` and migration `063` asserts it is the only INSERT site into `oauth_credentials`.

Arming it anyway is safe **because of how the refusal routes**: `media_sync` classifies `DriveCredentialDead` persistent → the source flips to `error` → the disconnect alert fires once under its `alerted_at` dedup → **the job SUCCEEDS** as handled work. The failure mode is a visible source in `error`, not a poisoned lane.

`drive_credentials.py` holds the credential half so decryption and the state machine stay out of the transport. Every refusal names which of four distinguishable causes it is (absent / wrong state / expired / undecryptable) — they have different remedies, and one generic "unavailable" sends whoever reads the alert to guess. It does **not** flip credential state on a decrypt failure: it holds no write grant, and a read door that mutates on read is how a transient ring misconfiguration permanently expires every credential it touches.

## A floor bug found on the way past — this is the part to review hardest

**Every gzip provider reply through `egress.request` died on `DecodingError`, and nothing at the floor could see it.**

`_read_capped` accumulates from `aiter_bytes()`, which yields **decoded** bytes. The rebuilt `httpx.Response` was handed the original headers, so it kept claiming `content-encoding: gzip` over a plaintext body and httpx decoded it a second time. That is *every real provider reply*, since httpx advertises `Accept-Encoding: gzip` by default.

Reproduced in isolation against the shipped httpx, before the fix:

```
OLD rebuild: DecodingError: Error -3 while decompressing data: incorrect header check
NEW rebuild: OK -> {'files': [{'id': 'f1'}]}
```

The raise happens at **construction**, not first read.

**Why it survived:** every existing floor proof uses an uncompressed body, so nothing in the floor's own suite ever handed it something decoded on the way in. The regression pin therefore goes in `test_egress_floor.py` — where a person changing the floor will run it — not only in the Drive adapter's tests, which is merely where the symptom surfaced.

The pin asserts the fixture really did gzip (`\x1f\x8b`) **before** asserting the body parses, because a fake that quietly sent plaintext would pass either way. That is the same shape that let the defect ship.

**This is a floor fix, not a Drive fix.** `egress.request` is shared by the IG refresh door, transit, meta, the Google OIDC exchange, the Telegram transport and both Drive adapters. `content-length` is dropped rather than corrected — it described the compressed body.

## What this does not do

- **W5b (`media_fetch` bytes) is not wired.** Its parking predicate is media_fetch-only while the ladder also calls `transit.upload` and `meta.create_container` (astrid, #982), so wiring it here would unpark `publish_pipeline` straight into `meta=None` → `AttributeError` → poison. That belongs with milestone 2, gated on media_fetch AND meta AND transit.
- **Subfolder listing.** A `config` carrying `root_name` is **refused rather than listed** — it names a subfolder whose resolution is a second round trip, and a query that ignored it would list the parent, which looks correct whether it comes back full or empty.
- **No Drive refresh leg.** An expired token is refused on `expires_at` here rather than sent to Google to earn a 401 — same outcome for the source, one less provider call, and the reason names expiry.

## Verification

Run locally against the venv, unpiped, exit status read from `$?`:

| Scope | Result |
|---|---|
| `test_google_drive_adapter.py` + `test_w6_sync_gate.py` | **28 passed**, rc 0 |
| floor + consumers (egress, oidc, drive_adapter, transit, meta, credential_lifecycle) | **122 passed**, rc 0 |
| `test_egress_floor.py` incl. 2 new pins | **26 passed**, rc 0 |
| `ruff check .` / `ruff format --check .` | clean, 436 files |

**Both new floor pins mutation-verified.** Reverting `egress.py` to main's header pass-through:

```
FAILED ...TestAGzipReplyStaysReadableThroughTheFloor::test_a_gzip_body_arrives_as_readable_json
FAILED ...TestAGzipReplyStaysReadableThroughTheFloor::test_the_rebuilt_response_states_no_stale_encoding_or_length
2 failed
```

`egress.py` restored byte-identical afterwards (`cmp` clean).

**Bounds on this evidence, stated rather than implied:** I did not run the full suite locally. Three other pytest processes were running concurrently against the shared local Postgres for most of this session, and this repo's suite is known to corrupt across concurrent runs — so a full local run would not have been trustworthy evidence either way. **CI is the authority for the full suite here.** The signature change was swept for other implementors: `list_changes` has exactly two, the new adapter and the one test fake, both updated; `StubDriveAdapter` has no `list_changes` at all.


chrisrogers37 (commented):
**Approve**

[navi] [blast-radius-audit] Mandate: does the shared egress.py gzip fix silently change behavior for callers relying on the buggy shape, and does the seam claim (list_changes vs list_files) hold. Single reviewer this round.

## The egress.py fix itself

`request()`'s response-rebuild now strips exactly two headers off the reconstructed `httpx.Response`: `content-encoding` and `content-length`. Nothing else changes. That's the whole diff.

**Reproduced the bug and the fix myself, not just the PR's quote.** Checked out `origin/main`'s `egress.py` over the PR's (`git checkout origin/main -- src/services/target/egress.py`), ran the two new pins in `test_egress_floor.py`: both fail with `httpx.DecodingError: Error -3 while decompressing data: incorrect header check` — the exact error the PR body quotes, raised at response construction as claimed. Restored the PR's version (`git status` clean afterward, byte-identical). The floor-level regression pin is real and does catch a revert.

## Actual consumer set — verified narrower than the PR/CHANGELOG claim

The PR body and CHANGELOG both say (identical sentence in both places): *"`egress.request` is shared by the IG refresh door, transit, meta, the Google OIDC exchange, the Telegram transport and both Drive adapters."* I grepped every file that imports `egress` and checked what each one actually calls:

**Real, current callers of `egress.request()`** (confirmed by direct call site):
- `channels/telegram_transport.py:83` — Telegram Bot API
- `services/target/credential_lifecycle.py:95,100` — `ig_refresh` (the IG refresh door)
- `services/target/google_oidc.py:119` — `exchange_code`
- `services/target/google_drive_adapter.py:247` — the new adapter (this PR)

**Named as consumers but do NOT currently call `egress.request()` at all:**
- `transit.py` — its own docstring is explicit: *"These calls do not ride the egress floor (the SDK owns its own HTTP)... The floor cannot see SDK internals."* Cloudinary SDK manages its own transport. Only imports `TIMEOUT_CLASSES`, a constant.
- `meta_adapter.py` — stub-only. *"The real Graph adapter (egress-floor-backed httpx) is deliberately NOT here"* — same M.3-gating pattern as the old Drive stub. Zero live egress calls exist to audit yet.
- the OLD `drive_adapter.py` (`StubDriveAdapter`) — zero `egress.request` calls; it's a scripted stub, no real HTTP. Only the NEW `google_drive_adapter.py` is a real consumer, so "both Drive adapters" overstates by one.

This doesn't change the safety verdict (overstating blast radius is the safe-direction error), but it's stated as fact in CHANGELOG.md, which is the durable record — worth a follow-up correction there since a future reader checking "what do I need to retest if I touch `egress.request`" would chase transit/meta for nothing while the two-word fix ("IG refresh, OIDC, Telegram, and the new Drive adapter — four callers") is both more accurate and simpler.

## Is it a fix, or a fix plus a silent contract change?

Checked every real consumer above for `.headers` access on the returned response — **zero hits**. All four only touch `.status_code` and `.json()` (with `json.JSONDecodeError`/`ValueError` handling around the latter). None read `content-length` or `content-encoding` today, so dropping those two headers is behaviorally invisible to every current caller — there's no code path that branches on them.

The only real behavior change: a gzip-compressed reply from any of the four providers previously raised `DecodingError` unconditionally at construction; now it parses. In every one of the four consumers, that exception was already falling into a **generic** failure path that exists for unrelated reasons too (`telegram_transport`'s blanket `except Exception` → `TelegramSendError`; `google_oidc`'s exchange has no local catch around the call, so it would have bubbled as an unhandled 500; `ig_refresh` has no local catch either — D31's taxonomy would route it wherever an unclassified exception lands, presumably the lane's transient-retry ladder). Nothing was uniquely wired to *that* exception as a meaningful signal distinct from "the call failed for some reason" — so this reads as a pure fix for all four, not a fix-plus-silent-contract-change. **Verdict: safe.**

Worth a look (not blocking, can't verify from here): httpx defaults to `Accept-Encoding: gzip`, so if any of Telegram/Google's endpoints being called actually gzip their replies in practice, this bug would have been silently failing those calls in production this whole time via the generic-failure paths above. Might be worth a quick grep of prod logs for `DecodingError`/`TelegramSendError` volume before vs. after this ships — pure upside if so, just flagging as a possible explanation for any unresolved flakiness on those paths.

## Seam claim — verified true at every link

- `StubDriveAdapter` has `list_files`/`fetch_bytes`; no `list_changes` method exists on it (confirmed directly).
- `media_sync.py:126` calls `deps.drive.list_changes(...)` — the only `deps.drive.*` call in the file, and grep found zero `list_files` calls anywhere in it.
- `google_drive_adapter.py:158` defines `async def list_changes(...)` on the new `GoogleDriveAdapter`, backed by real `egress.request` calls via `_floored_get`/`_get`.
- `worker.py:521-525` — the real composition root instantiates `google_drive_adapter.GoogleDriveAdapter(...)` and wires it into `compose(..., drive=drive)`. `StubDriveAdapter` has zero references outside `tests/` — confirmed the stub cannot reach the live path.

Also checked the "safe to arm with no credential yet" claim: `media_sync.py:132` does catch `(DriveSourceGone, DriveCredentialDead)` and flips the source to `error` with a deduped alert rather than poisoning the job — matches the PR's description of why arming this today, before any `gdrive` credential writer exists, is safe.

## CI / base freshness

`git merge-base origin/main origin/branden/982-drive-list-changes-adapter` equals `origin/main`'s tip exactly — this branch is not behind, no stale-base risk to check for. `mergeable: MERGEABLE`, not a conflict, so the checks-frozen-while-UI-reads-green failure mode you hit this week structurally can't apply here. Checked `gh pr checks` by name rather than count: all 8 expected checks present — Changelog Check, FC-2 Telegram ratchet, GitGuardian Security Checks, Lint, Security Scan, Test, Vercel, Vercel Preview Comments — all SUCCESS (Test has since cleared since you wrote the dispatch).

Nothing here blocks. The one thing worth doing is trimming "transit" and "meta" out of the CHANGELOG's consumer list before it's read as ground truth by someone auditing this function later.

PR #1055, opened 2026-08-25
fix(web): remove the Upload affordance, which pointed at a route that does not exist (#1050)

chrisrogers37:
Closes #1050.

## What this is

The Media page shipped an Upload control in #1032 whose proxy route posted a multipart body at `POST /api/v1/workspaces/{ws}/upload-media`. **That route does not exist.** A live control in production posted at a 404.

## Removal, not build — and the reasoning is what decides that

Verified rather than assumed, because the same defect admits two opposite fixes:

| Claim | Check | Result |
|---|---|---|
| The target route is absent | `grep -rn 'upload-media' src/` | 0 matches |
| Not simply misrouted | All API POST routes enumerated | 5 routes, none of them it |
| Not hiding in the command vocabulary | `VOCABULARY` in `services/target/commands.py` | 24 commands, no upload verb; `upload` appears nowhere in `src/api/` |
| Upload is structurally impossible today | `media_items.source_id UUID NOT NULL` (054:262); `ck_sources_provider CHECK (provider IN ('gdrive'))` (054:160) | Every item must hang off a Drive source; only one migration ever touched that CHECK and it has never been widened |

So building the route is a **scope addition** requiring a migration. The button is false advertising. Removing a control that cannot work is cleanup, and it is reversible the day upload is genuinely in scope.

## The removal is the copy as well as the code

Deleted: the proxy route, `media-upload.tsx`, `media-upload-wrapper.tsx`, the layout description offering to "Browse, upload, and manage", the empty state telling a new user to "Upload your first image", a dead "Upload Media" action pointing at the page it was already on, and the loading skeleton's reserved sidebar column.

A page that says "upload" with no upload control is the same defect in smaller type.

## Disclosure: I resumed this from an interrupted session, and found my own work half-applied

My session was killed mid-task. Seven of the eight dirty paths were complete and verified. **`loading.tsx` was not.** The previous session collapsed the two-column wrapper (`lg:grid-cols-[1fr_300px]` → bare `<div>`) but left the upload sidebar `<Card>` behind.

The half-applied state was **worse than untouched**: the dead card stopped being a 300px sidebar and became a full-width panel directly beneath the grid — measured at **1168 × 282px** on desktop, versus 280 × 408px for an ordinary grid tile in the fixed version. It promised an upload panel in a *more* prominent position than before, on a page that then loads without one. On mobile it added 282px of false promise (content height 4906 → 4624px).

That is #1050's own defect reproduced in the loading state. A skeleton is a promise about what is about to appear.

[...]

PR #1056, opened 2026-08-25
fix(drive): carry the provider mimeType through the seam — the one classifier input that is not a filename (#982 follow-up)

chrisrogers37:
Addresses the F5 finding in `shared/planning/active/2026-08-25-media-fetch-prescope-against-1054.md` (alex, pre-scope against #1054).

## The defect

`_item_for` derived `kind` from the provider's `mimeType` and then **discarded it** — `mime_type` was not on the item dict. Four places in this repo say that value belongs downstream, and one threw it away:

| Says it belongs | Where |
|---|---|
| The port declares it **required** | `drive_adapter.DriveFile.mime_type: str` — not Optional |
| The schema defines the column | `054`: `mime_type VARCHAR(100) NULL` on `media_items` |
| The API already serves it | `workspaces.py:278` selects `mime_type` in the media pool read (#1044) |
| The adapter already computed it | `_kind_for(entry["mimeType"])` |

So `media_items.mime_type` was **NULL for every provider-sourced row**, and `GET /workspaces/{ws}/media` has been returning `mime_type: null` for all Drive media. Recovering it later means re-listing every folder — the provider is the only source of truth for it — which is why this lands now rather than with W5b.

## The fix

Two keys. `"mime_type"` on the item dict, and `mime_type` in `media_sync`'s `media_items` upsert.

The second half is not optional: carrying the value one function further and dropping it at the INSERT looks identical from the seam's side, so **the gate asserts it at the database**, not at the adapter.

`.get("mime_type")`, not `[...]` — the column is nullable and the port does not require the key, so an adapter that cannot state a content type writes NULL (the same row this wrote before) rather than crashing. That case is its own test, not an assumption.

## The fixture is extensionless on purpose

This is the trap in alex's note and it is real. A Drive file's name is whatever a person typed into Drive; nothing requires a suffix, and `media_sync` writes that name straight into `media_items.file_name`. Every other item in the W6 gate is named `<ref>.jpg` — **a fixture like that passes with the bug present**, because the filename carries the answer anyway.

The new rows are named `sunset clip`:

- asserted to have no suffix (`Path(name).suffix == ""`),
- asserted, against the **shipped** `INSTAGRAM_VIDEO_SUFFIXES` rather than a copy of it, to be a name the IG predicate would classify IMAGE,
- while the provider says `video/quicktime` and the stored `media_kind` is `video`.

So the value can only have come from the provider.

## Verification

**The tests fail without the fix.** Removed `"mime_type": mime` from the adapter and re-ran:

```
FAILED ...::TestListChanges::test_maps_a_page_into_media_sync_item_shape
FAILED ...::TestListChanges::test_the_provider_mime_survives_a_name_that_carries_no_extension
2 failed, 22 passed
```

with the fix restored, `24 passed`. The positive control (`test_a_skipped_entry_never_reaches_the_item_shape`) stayed green under the mutation, which is what makes the two reds mean something rather than the fixture failing to arrive. That mutation was a dev-time check — it is not shipped as a control.

W6 sync gate: see the run noted in the comments below.

## Scope — what I deliberately did not do

**No classifier was touched.** `git diff --name-only` is the seam, its one consumer, their two tests, and the CHANGELOG. `src/utils/media_kind.py` and `src/services/integrations/cloud_storage.py` are untouched, and `_kind_for` is byte-identical — the only change to its caller is binding a local before the call. The `media_kind`/`cloud_storage` divergence stays its own follow-up.

**No consumer was changed to prefer the new value, because there is no consumer to change.** The ask included "make the consumers that guess from the filename prefer it when present"; measured, on the target path there are none:

- `src/services/target/` contains **zero** extension tests (`grep -rn "endswith\|suffix\|splitext\|guess_type\|\.mp4\|\.mov"` returns one unrelated dialect-suffix comment).
- The publish path never guesses. `media_kind` rides the whole way and is itself `mimeType`-derived: `mimeType` → `_kind_for` → `item["kind"]` → `media_items.media_kind` → `publish_pipeline.py:292`'s join (`m.media_kind`) → `ctx.intent["media_kind"]` → `transit.upload(media_kind=…)` (`transit.py:186`) and the IG call (`publish_pipeline.py:629`). `post_intents` has no `media_kind` column of its own; the value is read off `media_items` at load.
- The two filename classifiers — `instagram_media_type` (`scheduler.py:855`, `telegram_autopost.py:491`) and `resource_type_for` (`cloud_storage.py:283`) — are legacy-tier and never see Drive media.

**So this is not #1024 re-opened on the target path as built.** I want to be exact about that rather than inherit the framing: the composition alex describes — extensionless `file_path` plus a discarded `mimeType` — would be that, but `media_kind` is carried on the intent and closes the hole for classification today. What is live is narrower and still worth fixing: an authoritative value discarded at the one seam that has it, a schema column left NULL, and an API field serving null.

There is one place the "prefer mime when present" instruction would bite, and it is out of fence: `instagram_media_type` ignores `media_item.mime_type` even though legacy ingest populates it (`media_ingestion.py:184`, `local_provider.py:126`, `backfill_downloader.py:120`). That is a classifier edit on the legacy tier. Flagging, not doing.

**Adjacent and not taken:** `size_bytes` is on the item dict and `media_items.file_size` is likewise never written. Same shape, same seam, not this task.

Branched off fresh `main` at `5c686f5` (#1054 included). Not touched: the 8 MiB egress cap (F1), which ari is handling separately.


chrisrogers37:
## CI evidence, and a flake worth naming rather than waving past

All 8 checks green at `c4b0cb0`, `mergeable: MERGEABLE`. The Test job ran the W6 gate against a real Postgres with `REQUIRE_TEST_DATABASE=1`, so a DB-backed test cannot silently skip.

**The two gate tests could not run on my machine** — every test in `tests/scripts/test_w6_sync_gate.py` dies locally on `ModuleNotFoundError: No module named 'asyncpg'`, a declared dependency (`requirements.txt:11`) missing from my ambient python. One cause across all 8, hitting the pre-existing tests identically. So CI is the only place they have executed, and here is that, by name rather than by count:

```
test_maps_a_page_into_media_sync_item_shape                PASSED
test_the_provider_mime_survives_a_name_that_carries_no_extension  PASSED
test_a_skipped_entry_never_reaches_the_item_shape          PASSED
test_the_mime_lands_on_a_row_whose_name_has_no_extension   PASSED
test_an_adapter_that_states_no_content_type_leaves_it_null PASSED
3702 passed, 10 skipped in 190.52s
```

**The first Test run failed and I am not filing that as unrelated, because that is a stronger claim than I can support.** It was `1 error`, zero `FAILED` — a fixture teardown:

```
ERROR at teardown of test_lineage_lane.py::TestTheLaneReplaysAcrossTheBoundary
  ::test_the_legacy_bound_is_load_bearing_not_tidy
psycopg2.errors.DependentObjectsStillExist: role "svc_migration" cannot be dropped
DETAIL: 2 objects in database runner_test_9d2d079d91_7309a4835b
conftest: NOT dropping runner_test_9d2d079d91_7309a4835b — 1 live backend(s) hold it
```

What I can evidence:

- The blocking database is `roleless_db`'s **own** scratch DB, held open by a live backend. These tests run against `lane_db`, a different scratch DB — a connection leaked by this change would be holding that one, not `runner_test_9d2d…`.
- `test_lineage_lane.py` contains **zero** references to `media_sync`, `mime_type`, `google_drive_adapter` or `media_items`.
- `conftest._drop_roles_hardened` already carries an explicit branch for this exact shape, including the "is another suite running without the lock?" message it printed — a known-fragile teardown, not a new one.
- Re-running the same job on the same commit went green, and the **passed count is identical in both runs** (3702). Nothing moved except the error.

What I cannot evidence: that this change did not perturb the timing that surfaces it. Two runs is not a characterisation of an intermittent failure in either direction. If it recurs on an unrelated PR, that is the confirmation; if it recurs here, it is mine and I will fix it rather than re-run it.

chrisrogers37:
## Merged `origin/main` — CHANGELOG only, no code moved

`#1053` landed at 19:22:25Z and took `CHANGELOG.md` with it, which is the entire overlap. Verified independently rather than taken on report — the intersection of the files this branch touches and the files main moved since my base `5c686f5`:

```
CHANGELOG.md
```

Resolved by keeping both entries. **The four files this PR actually changes are byte-identical to `c4b0cb0`:**

```
$ git diff --stat c4b0cb0 -- src/services/target/google_drive_adapter.py \
    src/services/target/media_sync.py tests/scripts/test_w6_sync_gate.py \
    tests/src/services/target/test_google_drive_adapter.py
(no output)
```

So a review in flight against `c4b0cb0` is not invalidated — none of the code under review changed. The other 30 files in the merge are main's incoming work (#1051, #1055, #1053), not this branch's; three-dot against main is still exactly the 5 files and 175 insertions it was.

Merge commit, no rebase, no force-push, so review anchors hold.

**Treat the 8/8 green that was showing before this push as stale.** Those ran against a base that moved three times today; the bar stayed green while the PR read CONFLICTING. The checks that count are the ones now running on the merged head `04d2f95`, and I will post the result rather than assume it carries over.

chrisrogers37 (commented):
**Approve**

Reviewed `c4b0cb0` (head of `branden/1054-carry-drive-mime-type`). This is a verify-the-refutation task, so I'm answering ari's five points directly, then the scope question at the end.

**1. "Zero extension tests in `src/services/target/`" — true?**

Confirmed, and pushed past the PR's own grep to make sure it wasn't cherry-picked: `grep -rn "endswith|suffix|splitext|guess_type|\.mp4|\.mov"` returns exactly one hit, the unrelated asyncpg dialect-suffix comment in `unit_of_work.py` the PR body names. I broadened the search myself — `mimetypes\.|\.avi|\.webm|\.mkv|\.m4v|filetype\.|magic\.|is_video\(|\.lower\(\)\.end` — zero hits. Claim holds under a wider net than the one it was measured with.

**2. Does the target publish path really never guess?**

Traced the full chain rather than sampling it: `mimeType → _kind_for` (a pure mime-prefix dict lookup, `src/services/target/google_drive_adapter.py:116`, no filename involvement) `→ item["kind"] → media_items.media_kind` (the `media_sync.py` INSERT) `→ publish_pipeline.py`'s join at the row fetch (`m.media_kind` selected directly) `→ ctx.intent["media_kind"] → transit.upload(media_kind=...)` (`transit.py:186`, required `str` kwarg, no filename parameter exists on the signature at all) and the IG container call (`publish_pipeline.py:629`, same). `transit._resource_type` is a closed dict lookup that raises on an unrecognized `media_kind` rather than falling back to anything. `ctx.intent["file_name"]` has zero read sites anywhere in `publish_pipeline.py` despite the row fetch selecting it — file_name is fetched but never consulted for classification. And structurally: `git grep` for an import of `src.utils.media_kind` or `src.services.integrations.cloud_storage` anywhere under `src/services/target/` returns nothing — the two legacy classifiers aren't just unused today, they're unreachable without adding a new import. Branden's narrowing is correct; ari's original #1024-recurrence framing doesn't apply on this path as built.

**3. Does the gate assert at the database?**

Yes — `_media_row()` runs a real `psycopg2` `SELECT file_name, media_kind, mime_type FROM media_items WHERE ...` and the two new tests assert on that returned row, not on the `item` dict fed into `ScriptedDrive`. If `media_sync.py`'s INSERT dropped the `mime_type` column while the adapter still computed it, this would still fail — it can't be satisfied by the adapter alone. I wanted this running live, not just read: the adapter-level suite ran for real (24/24 passed, `.venv` + real fixtures, no mocks on the assertions), and I mutation-tested it myself — deleted the `"mime_type": mime,` line and reran: exactly `test_maps_a_page_into_media_sync_item_shape` and `test_the_provider_mime_survives_a_name_that_carries_no_extension` failed, everything else stayed green, then restored the file. That matches the commit message's own claimed mutation result. The DB-backed W6 gate specifically (`_media_row`'s two callers) I could not get to run live — it queued on `tests/scripts/conftest.py`'s suite cluster lock for the full 1200s timeout and errored at fixture setup, never reaching the assertion. Checked why rather than assuming: `ps` shows astrid's bot running its own full suite (`pytest tests/ -q`, pid 770146) against this same shared local Postgres cluster right now — genuine cross-bot contention on shared test infra, not a code issue, and I confirmed that rather than guessed it. My confidence on this specific assertion rests on reading `_media_row`'s SQL directly (unambiguous — a plain SELECT against the real table, not the kind of logic that hides a bug on a read-through) plus the commit's own reported CI run (`24 passed`) and the adapter-level live confirmation above, not on a live run of the DB gate itself. Flagging the gap rather than papering over it.

**4. The "sunset clip" fixture — right shape?**

Ran for real, not just read: `test_the_provider_mime_survives_a_name_that_carries_no_extension` passed live. Confirmed the import is genuine (`from src.utils.media_kind import INSTAGRAM_VIDEO_SUFFIXES`, the shipped module, not a local copy) and the two controls both hold — `Path(name).suffix == ""` and `not name.lower().endswith(INSTAGRAM_VIDEO_SUFFIXES)` — so this is a name a suffix-based classifier would call IMAGE while the provider says `video/quicktime`. The companion positive control (`test_a_skipped_entry_never_reaches_the_item_shape`, `mime="application/pdf"`) also ran and passed, confirming an empty result here means the skip fired rather than the fixture failing to arrive.

**5. `mime=None` — first-class, not a fabricated default?**

Confirmed in `_item`'s design: the `_UNSET` sentinel (not `None`) is the "caller didn't specify" default, so every pre-existing `_item(...)` call in the gate keeps getting its kind-matched mime automatically and none of them regress — checked the whole file for anything that would break from the signature change (an exact-keys assertion on the item dict, for instance): nothing. `mime=None` explicitly is the "adapter says nothing" case, which never sets the `mime_type` key at all, so `media_sync.py`'s `.get("mime_type")` returns `None` and binds SQL NULL. The DB-level confirmation of this (`row["mime"] is None`) is the other half of the untested-live pair from point 3, same caveat.

**Scope fence — did it hold?**

Yes. `media_kind.py` and `cloud_storage.py` are absent from the changed-files list, and the adapter diff shows `_kind_for`'s body untouched — the only change to its call site is binding `mime` as a local before the call it already made. No consumer was touched to prefer the new value, consistent with point 2 finding no consumer to touch.

**`size_bytes` — correctly deferred, and something worth raising it wasn't**

Independently verified the parallel: `media_items.file_size BIGINT NULL` exists (`054:277`), the adapter's item dict already carries `size_bytes`, and `media_sync.py`'s INSERT column list still omits it — identical shape, same seam, confirmed rather than taken on faith.

Deferring it *from this PR* is the right call — this fix is clean specifically because it's one seam with its own discriminating test, and bundling a second one would blur that. But I went one step further than "same shape" and checked whether anything already consumes it, the way `mime_type` was already consumed by `workspaces.py`'s media read: **`file_size` is already rendered live.** `media-grid.tsx:184` calls `formatBytes(item.file_size)`, and `MediaItem.file_size` is typed `number` — not nullable — despite the column being NULL for every Drive row today. `formatBytes`'s first branch is `if (bytes < 1024) return \`${bytes} B\`` — a `null` argument satisfies `null < 1024` (coerces to `0`) and the template literal stringifies `null` as text, so this renders the literal string **"null B"** on the Media Library grid for every Drive-sourced item, right now, since #1055 already repointed that page onto this live route. That's sharper than the mime_type situation was: mime_type's one frontend consumer (`item.mime_type?.split("/")[1] || "file"`) is defensively null-safe and degrades gracefully; file_size's is not. So: yes, correctly deferred from this PR, but I'd recommend filing it as a near-term fix rather than a someday one — the same "recoverable only by re-listing" argument that justified doing mime_type now applies here too, and unlike mime_type this one already has a visible, live symptom rather than a merely-latent one.

Nothing else surfaced. Branden's refutation is correct and precisely scoped; the fix matches it exactly.

chrisrogers37:
## CI green on the merged head — run `32889422408`, `headSha 04d2f95`

Not the stale bar carried forward. The run's own `headSha` is the merged commit:

```
32889422408  04d2f95  success  19:24   <- this one
32886324448  c4b0cb0  success  18:52
32886165464  90c21ba  failure  18:51
```

All 8 checks SUCCESS, `mergeable: MERGEABLE`. Suite: `3763 passed, 10 skipped` (up from 3702 at `c4b0cb0` — #1053's tests came in with the merge). The five tests this PR turns on, by name rather than by count:

```
test_maps_a_page_into_media_sync_item_shape                      PASSED
test_the_provider_mime_survives_a_name_that_carries_no_extension PASSED
test_a_skipped_entry_never_reaches_the_item_shape                PASSED
test_the_mime_lands_on_a_row_whose_name_has_no_extension         PASSED
test_an_adapter_that_states_no_content_type_leaves_it_null       PASSED
```

The `svc_migration` teardown error from the 18:51 run did not recur — zero occurrences here. That is now two green runs against one failure, which is still not a characterisation of an intermittent failure; it is just the current count. Unchanged from what I said before: if it comes back on this branch it is mine.

Issue #1058, opened 2026-08-25
test_a_dead_clocks_election_is_released_by_its_session races pg_terminate_backend's asynchronous exit


chrisrogers37:
`tests/scripts/test_scheduler_clock_gate.py::TestKillingTheClockMidTickLosesNothing::test_a_dead_clocks_election_is_released_by_its_session` fails intermittently under host contention, and the failure is a race in the test rather than in `ClockElection`.

**Measured** (2026-08-25, this host, docker `postgres:15`, load 13–30):

- Failed at 20% of a full-suite run (`-x`), then again in isolation while another pytest session was bootstrapping template databases on the same container.
- Passed 3/3 on pristine `origin/main` (`681cea0`) and 2/2 on a branch that does not touch the scheduler, in a controlled sequential A/B at load 22–30.

**Mechanism.** The test terminates the clock's backend with `SELECT pg_terminate_backend(pid)` and immediately opens a successor connection and calls `pg_try_advisory_lock`. `pg_terminate_backend` returns `true` when the signal is *sent*; the session-scoped advisory lock is released when the backend has actually *exited*. Under CPU/IO contention that exit lags the successor's first attempt, and the assertion (`taken is True`) reads a lock that is still held for a few more milliseconds as "the election outlived its session".

**Suggested fix (test-side only).** Between the kill and the successor's attempt, wait — bounded — for the pid to leave `pg_stat_activity` (e.g. poll `SELECT count(*) FROM pg_stat_activity WHERE pid = %s` until 0, up to a couple of seconds), then assert the successor takes the lock. That keeps the property the test pins (a SIGKILLed clock's election dies with its session, no `finally` required) and removes the timing assumption. A retry loop around `try_acquire` would also work but hides the distinction between "released late" and "never released", which is the thing the test exists to catch.

Noticed while landing #1033 P11; nothing in that change touches the scheduler.

PR #1059, opened 2026-08-25
feat(web): the queue — the act-on-it surface, and the web becomes a surface every workspace has (#1033)

### LINKED ISSUE #1033: [feature] Web approvals surface — a Google-only workspace currently has no way to approve a post

chrisrogers37:
Closes #1033 (P11 — the act-on-it surface). Built from the design posted on the issue (comment 5414309482); the five decisions there stand unchanged, and no objection landed on them.

## Why

`workspaces.api_publishing_enabled` defaults to false, so every new workspace is in manual publishing mode, and the human tap that closes the loop there — *Posted myself* — had no surface a Google-only workspace could reach. Worse, upstream of any screen, `prompt_intent` returned without transitioning when a workspace had no Telegram binding: a Google-only workspace's intents never left `scheduled` and the reaper expired them, on exactly the workspace milestone 1 exists to serve. The `02` §4 edge `prompt_pending → awaiting_approval` reads "delivered on ≥ 1 binding, **or** workspace has web access"; the second disjunct had no referent until this change.

## What

**The screen.** `/dashboard/queue` lists every non-terminal intent in slot order, in the workspace's own clock, with the matrix's human levers on the one state that has them — `awaiting_approval`: *Posted myself* (`mark_posted`), *Skip*, *Reject* (confirmed first; its lock is permanent) and *Approve* only where `api_publishing_enabled` (hybrid keeps the manual buttons, `06` §3). Every other non-terminal state is read-only with its badge. No optimistic update: the list re-reads the ledger after every answer, refusals included.

**The command path.** `POST /api/workspaces/[id]/commands/[command]` (front end, allowlisted to the four) mints the API's `Idempotency-Key` as `${command}:${intentId}` — stable per (command, intent), so a double-click is a `200 replayed` and a refused first try stays retryable — and passes the port's refusal through by reason. `readError` now reads the command port's `{detail, reason}` envelope beside `{error}`, so a 409 renders as a sentence rather than `http_409`.

**Upstream.** Web access is universal by construction: `prompt_intent` transitions whether or not a binding exists, the sweep advances every `prompt_pending` intent in the same pass, and the `failed_no_surface` leg is deleted — `prompt_pending → failed` stays seeded in `055` with no producer (`02` §4 / `06` §5 annotated). The INTERIM `approval_mode='auto'` on `create_workspace` is reverted as #1033 required; measured first: `approval_mode` has no reader in `src/`, so the revert changes a stored value and nothing else. The intents read joins `ig_accounts` for `account_handle` / `account_display_name` on every row. `PromptSweeper` stops reading the retired counter.

## Deviations from the posted design (same behaviour, disclosed)

- Front-end route is `app/api/workspaces/[id]/commands/[command]/route.ts`, not `app/api/intents/[intentId]/[command]`: it mirrors the API's one write route and the existing `api/workspaces/[id]/select` shape (workspace in the URL, intent in the body), so the two tiers have one spelling of the command.
- Intent types live in `lib/intents.ts` beside the helpers typed on them; the workspace config row reuses #1051's `WorkspaceConfig` from `dashboard-payloads.ts` and its `workspaceFetch("", id)` (merged in before the push — my own `getWorkspaceConfig` was the same path written independently and is gone).
- The account join is a plain `JOIN`, not `LEFT JOIN`: `post_intents.ig_account_id` is `NOT NULL` with a composite FK, so they are equivalent, and it matches the media join beside it.

## Not in this change

*Post now* (`autopost_now` is unbuilt, 501) · *Cancel* (overlay flag, no audit row — its own follow-up) · thumbnails · `media_fetch` (milestone 2, behind Meta App Review) · the generic `api/dashboard/[...path]` proxy's allowlist is untouched.

## Evidence

Backend, on a real Postgres 15 as the production roles: W3 gate (a zero-binding workspace reaches `awaiting_approval` in one sweep; all-cards-failed stays actionable; delivery after the fact changes nothing), X.2 gate (the flip to `'manual'`, the handle on the row, `mark_posted` through the real route → `posted` / `published_via='manual'`), dashboard-reads gate (`account_handle` on every row). Every test was watched RED before the code (the six gate failures were the six intended assertions), then GREEN. The `PromptSweeper` test was revert-checked: restoring the old counter read makes it fail on the `prompt sweep failed` log line. `ruff check` / `ruff format --check` clean over `src` + `tests`.

Front end: `tsc --noEmit` clean, `vitest` 35/35 (13 new for `lib/intents.ts`, 2 for `readError`), `next build` rc 0 with `/dashboard/queue` and the command route in the table (CI runs nothing for `landing/`). Rendered in headless Chromium against a stub router with one fixture intent per non-terminal state: anonymous → `/login` (the catch-all answered 404 here before); seven rows, seven badges, buttons on `awaiting_approval` only; *Posted myself* reached the stub as `POST …/commands/mark_posted` with `Idempotency-Key: mark_posted:<id>` and the row left the list on re-read; Reject sent nothing until the dialog was confirmed, then `reject:<id>`; with `api_publishing_enabled` the row grew *Approve*, whose stub 409 `manual_mode` rendered as "This workspace posts by hand…"; `cancel` and a non-UUID intent id answered 404 / 400 at the front end without reaching the router. Baseline lint errors in `error.tsx` / `not-found.tsx` / the blog page pre-date this branch and are untouched.

**One pre-existing flake, measured rather than waved at:** the full suite (run with `-x` on this host at load ~28) stopped at `test_scheduler_clock_gate.py::TestKillingTheClockMidTickLosesNothing::test_a_dead_clocks_election_is_released_by_its_session`, which also failed once in isolation on this branch. It passes on pristine `origin/main` (3/3) and on this branch in a controlled A/B (2/2, at load 22–30); the test kills the clock's backend with `pg_terminate_backend`, which returns when the signal is *sent*, and asks a successor for the advisory lock straight away, so under contention the lock is sometimes still held. Filed separately as a test-flake issue; nothing in this diff touches the scheduler.

**After the push (measured, not inferred).** The first CI run on `e75c2ab` ([32887244982](https://github.com/chrisrogers37/storydump/actions/runs/32887244982)) failed its Test job on two pins in `tests/scripts/test_w1_worker_gate.py` that asserted a due, binding-less intent was still `scheduled` after the minting pass — the exact pre-#1033 behaviour this change retires (the W3 fast path prompts in the minting pass, and web access now advances it in the same transaction). Reproduced locally (`assert 0 >= 1`; every width "did not mint" under the state filter) and moved in `de1073e`: the soak test asserts `awaiting_approval`, the width test asks only whether each slot minted. My `-x` full run had stopped at 20% on the clock flake before reaching that file — a gap CI closed. Main then took #1053 and #1055; merged in at `24a5ca8` — CHANGELOG keeps both entries, `v1.py` auto-merged, the three-dot diff against `origin/main` is this change's 25 files only, `v1.py`'s three-dot diff is the key comment only, and alex's `POST …/accounts` / `POST …/sources` stand at `v1.py:400/438`, present and untouched. CI on `24a5ca8` ([32889609511](https://github.com/chrisrogers37/storydump/actions/runs/32889609511)): Lint, Changelog Check, Security Scan, FC-2 ratchet and Test all green. Vercel preview on the merged head (the branch alias the Vercel bot posted): `/login` 200, `/dashboard/queue` **307 → `/login`** through the automation-bypass header — where the `[...slug]` catch-all answered 404 before. Local on the merged head: `next build` rc 0 with `/dashboard/queue` and `/api/workspaces/[id]/commands/[command]` in the route table; `tsc`, `eslint`, `ruff` clean; `vitest` 49/49; W1 + W3 + X.2 + dashboard-reads gates and the prompt / work-loop / worker units, **73 passed**, on Postgres 15 as the production roles. Full suite without `-x` on this host (the tree before the W1 pin fix; 28 min at load ~25): **3698 passed, 8 skipped, 3 failed** — the two W1 pins above, and `tests/src/services/test_concurrent_callback_reactivity.py::…[8]`, a legacy-services 50 ms ack budget that read 50.5 ms under load (fails/passes on isolated reruns at load 13, green on CI, imports nothing this change touches).

**Second merge-train pass.** #1056 landed nine minutes after the previous resolution; the only overlap was `CHANGELOG.md` (both entries kept). Merged at `0bd6ed5` — the three-dot set is still these 25 files, #1056's `mime` handling is in `media_sync.py` at HEAD and none of its five files appear in this diff. CI on `0bd6ed5` ([32890458962](https://github.com/chrisrogers37/storydump/actions/runs/32890458962)): all five jobs green. Preview on `0bd6ed5`: `/login` 200, `/dashboard/queue` 307 → `/login`. Local on `0bd6ed5`: ruff, `tsc`, `eslint` clean; `vitest` 49/49; `next build` rc 0 with both routes; the W1/W3/X.2/dashboard gates and the prompt/work-loop/worker units green again on Postgres 15.

## Workflow

TDD throughout (RED observed per axis, GREEN, then refactor). `/simplify` run before push — four parallel lenses (reuse / simplification / efficiency / altitude). **Applied:** the Reject confirm became a per-row uncontrolled `Dialog` (`DialogTrigger` / `DialogClose`, the `accounts-tab` pattern) so the `rejecting` state and its three null-narrowings went; button styling moved into a `COMMAND_VARIANT` table beside `COMMAND_LABELS`; the state-label map is a `Partial` of the four non-identity labels; `INTENT_STATES` is derived from `NON_TERMINAL_STATES` + `TERMINAL_STATES` instead of a second hand-kept list; `refusalCopy` keys on the reason alone (its status axis had no producer); the truncation notice reads the `limit` the API echoes rather than a local copy of `LIST_LIMIT_MAX`; the empty state is the dashboard's `EmptyState`; the UUID predicate lives once (`isUuid` in `session.ts`, `isWorkspaceId` delegates) instead of a byte-identical twin; `readError` reads `reason` alone — the API never emits `{error}` (both the reuse and altitude lenses found the old key was a phantom, and the pre-existing fixture that used it taught a false contract; retargeted to a real `InvitationRefused` envelope); one `Intl.DateTimeFormat` per time zone is cached (measured ~50× a format call, and the list formats every row on every render); `prompt_intent` returns nothing (its only production caller stopped reading the ids); the sweep asks `push_bindings` once per workspace per pass; the `v1.py` comment on the key now points at the owner of its shape; pointer comments on the four cross-language mirrors. **Skipped, named:** hoisting a shared refusal-copy tier and a typed client `fetch` helper across `create-workspace-form` / `accept-invitation` (two components outside this diff — follow-up); pointing `recent-activity`'s badge map at the queue's (legacy `HistoryItem` shape, dies with the P6–P9 screens); carrying `tz` / `api_publishing_enabled` on the intents envelope to save the config re-read per refresh (API-side, not real at a solo user's click rate); a bulk `transition_many` for the advance leg (a new ledger door for ≤50 PK-keyed updates per cadence); a port-side UUID check on `intent_id` (API-side follow-up; the BFF guard stays as defense in depth); the `malformed_body` / `invalid_intent` split (kept for consistency with the sibling handlers).. Path-weighing applied at the one real juncture (the route shape above) with the four lenses; sub-choices decided inline and disclosed here.


chrisrogers37 (commented):
**Approve**

[navi] Full review, treated the CI green as unverified per the dispatch. All checks below were done against `0bd6ed5` directly — the current head, re-fetched and diffed myself, not the earlier-verified `24a5ca8`.

## The load-bearing question: the `prompt_intent` early-return deletion

Mid-review, ari relayed mason's finding that the diff deletes the guard that returns without transitioning when `bindings` is empty — the thing that currently keeps every binding-less intent stuck in `scheduled` fleet-wide. This is the most important line in 25 files and I traced it end to end rather than reasoning about it abstractly.

**The exact change** (`prompts.py`): the `if not bindings: return` moved from *before* `intent_ledger.transition(session, intent_id, "prompt_pending")` to *after* it. Old: no bindings → no transition, intent stranded in `scheduled` for the reaper. New: transition always fires; card creation is what's skipped when bindings is empty. The sweep's advance leg lost its `EXISTS (... o.state = 'sent')` filter — it now advances *every* `prompt_pending` row, not just ones with a delivered card — and the whole `failed_no_surface` leg (the "all cards failed" path) is deleted outright.

**Bug or guard?** Applied the test literally: is the transition this deletion enables *legal* per spec, or was the guard protecting against something the spec never sanctioned? `02-domain-model.md` §4's transition table — I diffed it — is **unchanged by this PR** on the load-bearing row: `prompt_pending → awaiting_approval` has read "delivered on ≥ 1 binding, **or** workspace has web access (FC-2)" since before this branch existed. The *old* `prompts.py` docstring (visible in the diff, not inferred) says so itself: "the web-access disjunct has no referent anywhere in schema or code today... so the sweep advances on delivery evidence alone" — the original author already knew this was a deliberately narrowed, temporary implementation of a wider rule, pending exactly the web surface this PR builds. That's a bug being closed against an already-ratified spec, not a guard being removed against one.

**Checked what services a binding-less intent past `prompt_pending`, since that's the question that actually matters if I'm wrong about the above.** Grepped `command_executors.py`, the command route, `queue-list.tsx`, and `intents.ts` for `channel_outbox` or `binding` — zero hits, all four files. The intents read (`_INTENT_COLUMNS`/`_INTENT_FROM` in `workspaces.py`) joins `media_items` and `ig_accounts`, never `channel_outbox` — the web queue's rendering is structurally incapable of caring whether a card ever existed. Nothing in the action path assumes one.

**Then checked it against real Postgres, not just reasoning.** `test_w3_prompt_gate.py::TestTheWebIsASurface::test_a_workspace_without_push_bindings_reaches_awaiting_approval` seeds a workspace with **zero** bindings, runs the real async sweep against a real engine, and asserts `counts == {"prompted": 1, "advanced": 1}`, state `awaiting_approval`, **and** `_cards(...) == []` — an explicit assertion that no card row exists, not just that the state looks right. A second sweep asserts idempotency. `test_w1_worker_gate.py`'s `TestTheWorkerIdlesVisibly` test does the same through the full clock → plan_slot → worker path on a no-binding chain. Both are integration tests against a live database, not mocks.

**One more thing I checked because deleting a dict key silently is exactly the kind of thing that breaks a caller elsewhere:** grepped for `failed_no_surface` outside `prompts.py` and its own tests — one comment, zero consumers. Nothing destructures that key expecting it to exist.

Verdict on this specific question: the deletion is correct, and it's correct for the reason the spec already gave it, not because the tests were made to agree with new code.

## Two-directional route check, redone at 0bd6ed5

Re-fetched and re-diffed rather than trusting the 24a5ca8 verification, since ari flagged a second merge resolution since (confirmed: `astrid/1033-p11-web-queue`'s merge-base with `origin/main` is exactly `6636879`, which *is* #1056 — so this branch really did absorb a second merge after the point astrid checked). Both directions hold at the current head:
- `grep '@router.post("/workspaces/{ws}/accounts"\|.../sources"'` on the PR branch's `v1.py`: present at lines 400 and 438, unchanged.
- `git diff origin/main...HEAD -- src/api/routes/v1.py`: neither route appears — astrid's own diff doesn't touch them, so they're purely inherited from the merge, not reintroduced or duplicated.

## "interim auto reverted" and "intents carry their account" — confirmed in the diff, not the summary

- `INTERIM_APPROVAL_MODE = "auto"` is deleted outright, and `create_workspace`'s INSERT no longer sets `approval_mode` at all — new workspaces get the column's own default. Cross-checked against `provisioning.py`'s docstring (already on `main` via #1053, untouched by this PR): "`approval_mode` defaults to `manual`... a new workspace never touches Meta, `plan_slot` mints intents that await approval" — that comment already assumed the reverted behavior before this PR landed it, which is a good independent sanity check that nothing downstream was quietly built assuming `'auto'`.
- `_INTENT_COLUMNS` gains `a.handle AS account_handle, a.display_name AS account_display_name`, and a new shared `_INTENT_FROM` adds `JOIN ig_accounts a ON ...`, used by both `list_intents` and `get_intent`. Confirmed directly.

## W1-gate pin change — checked against spec, not just checked green

Same conclusion as the section above, from the test side: the new assertion (`awaiting_approval` instead of `scheduled` for a due, binding-less intent after minting) matches an `02` §4 row that predates this PR. The pin moved because the behavior was always supposed to be this, and wasn't yet.

## Scope — 25 files, spot-checked for creep

File list maps cleanly to what's described. The two that looked most tangential at a glance: `session.ts` (extracts the existing UUID regex into a shared `isUuid`, used by the new intent-id validation — real dependency, not drive-by) and `target-api.ts` (`readError` now reads `reason` instead of a phantom `error` key the API never emitted — directly serves the new command route's refusal rendering). Traced the full error-envelope chain end to end while I was in there: backend emits `{detail, reason}` → `targetFetch`/`readError` extracts `reason` as a string → the new Next.js route re-wraps it as `{error: <string>}` → `queue-list.tsx` reads `body?.error`. Two different envelope shapes at two different layers, each internally consistent — looked like a mismatch at first glance, traced clean.

Design-doc edits (`02-domain-model.md`, `06-product-lifecycles.md`) are annotation-only — diffed them directly, the load-bearing spec row is untouched, only a note that the sibling `failed` edge now has no producer.

## Driving a real interaction

Confirmed the live behavior myself on the actual deployed preview for this head, not astrid's stub: unauthenticated `GET /dashboard/queue` on `storydump-git-astrid-1033-p11-web-queue-chrisrogers37.vercel.app`, through the Vercel automation-bypass cookie flow, returns **307 → /login**, then `/login` → 200. Real HTTP against the real deployment, not a fixture.

Couldn't get a full authenticated click-through going (no test credentials for a real Google/IG sign-in, and a from-scratch `npm install` in this checkout hit a stale `ENOTEMPTY` from a prior partial install — cleaned it up, didn't chase a reinstall given the strength of what the real-Postgres integration tests already prove for the exact scenario in question). Read `queue-list.tsx`, `route.ts`, and `intents.ts` directly instead of trusting the PR body's description of them: idempotency key is minted server-side in the route handler, never the browser (`idempotencyKeyFor(command, intentId)` at the point the request leaves this tier) — confirmed, matches the stated reasoning for why a double-click can't double-execute. `QUEUE_COMMANDS = ["approve", "mark_posted", "skip", "reject"]` — `cancel` isn't in it, confirming the 404 claim. Reject is a real confirm-dialog (`Dialog`/`DialogTrigger`/`DialogClose`), fires only from `DialogClose`'s onClick after "Reject" is clicked inside the open dialog. No optimistic update anywhere in `run()` — every path ends in `router.refresh()` or a notice, never a local state paint.

## Not reviewed

#1058 (the clock-gate flake astrid filed) — out of scope per the dispatch, noted only so I don't rediscover it. Didn't re-verify the full CHANGELOG prose line by line beyond spot-checking the two claims above; the mechanism checks are what I'd stand behind either way.

Issue #1050, opened 2026-08-25
Shipped Upload affordance proxies to POST /workspaces/{id}/upload-media, a route that does not exist


chrisrogers37:
Found by navi during an adversarial read of the target-cutover completion map. **Not tracked in any existing issue** — filed so it does not stay invisible.

## The gap

`landing/src/app/api/dashboard/upload/route.ts` — already on `main`, shipped as part of #1032 — proxies to:

```
POST /api/v1/workspaces/{id}/upload-media
```

That route **does not exist** in `src/api/routes/v1.py` or anywhere in `src/`. The target router is 16 routes and this is not one of them.

So there is a live "Upload" affordance in the shipped front end pointed at a backend route that was never built.

## Why it is a distinct facet of Wall 1

Wall 1 is "the target tier cannot provision an account or a media source". This is a more specific instance: not merely that provisioning is unbuilt, but that a **user-visible control for it already ships** and leads nowhere.

## The constraint that makes this more than a missing route

From mason's done-definition, verified against the live schema:

- `media_sources.provider` is `CHECK (provider = 'gdrive')`
- `media_items.source_id` is `NOT NULL`

So **every media item must hang off a Drive source**. There is no upload path at all without a migration to widen that CHECK. Building the missing route is therefore not a small fix — it either requires a schema change, or the Upload affordance should be removed until an upload path is actually in scope.

## Suggested resolution

Decide which, and do it deliberately:

1. **Remove the affordance** — consistent with the current schema, where Drive is the only legal door. Cheapest and honest.
2. **Build the upload path** — needs a migration widening `ck_sources_provider`, plus the route, plus storage. A real scope addition, not a patch.

Option 1 pairs naturally with the #1044 dashboard repoint work, which is already deleting other screens that are not target product scope.

## Provenance

- navi, adversarial review of the completion map, evidence on #1029
- Related: #1048 (repoint is a shape migration), #1046, #1049

PR #1062, opened 2026-08-25
fix(web): re-derive the media loading skeleton from the page it stands in for (#1060)

### LINKED ISSUE #1060: Loading skeleton promises 4 pool-health cards; the page renders 2
## Summary

The dashboard's **loading skeleton promises 4 pool-health cards. The loaded page renders 2.**

Both numbers measured on rendered output, not inferred from source.

Cause: #1051 withholds the reuse rate and the eligible-to-post count as `null`, so two of the four cards never materialise — but the skeleton that precedes them still reserves space for four.

## Why this is worth filing rather than shrugging at

**It is the same defect class that #1055 exists to remove.** #1055 deletes an Upload affordance that promised an action the app cannot perform. This is a skeleton promising content the page will not deliver — a UI element making a claim the system does not honour. The user sees four cards resolve into two and has no way to know whether that is the design or a failure.

It also has the property that makes this class expensive: **every automated gate passes.** Lint, tsc, build, the test suite and the visual deploy are all green on it. Nothing in CI compares a skeleton against what eventually renders, so this can only ever be caught by looking.

## Provenance

Introduced by **#1051** landing, not by the #1055 conflict resolution that surfaced it.

Found by virgil while re-verifying #1055 on the merged head. Worth noting how: **#1051 and #1055 were each reviewed without the other's change in place**, so the composed state had never been seen by anyone until it was driven through a real `next dev` render. The finding exists because someone rendered the combination rather than reasoning about it.

**Deliberately not fixed in #1055**, and that was the right call — it sits outside that PR's conflict, #1055 carries a standing approval that verified rendered output, and quietly widening scope would have invalidated that review.

## Fix shape

Two honest options, and the choice is a product call rather than a technical one:

1. **Make the skeleton match** — render 2 placeholders while 2 of the metrics are withheld.
2. **Make the page match** — deliver the reuse rate and eligible-to-post count so all 4 cards populate.

Option 2 is more valuable if those metrics are wanted; option 1 is correct immediately and cheap. What is not acceptable is leaving a skeleton that reserves space for content nobody intends to supply.

chrisrogers37:
Closes #1060.

## The filed defect, and the condition underneath it

The skeleton reserved **4** pool-health cards; the page renders **2**. The two missing cards are not slow — they do not exist: `derivePoolHealth` returns `posted_multiple` and `eligible_for_posting` as `null` **unconditionally**, because #1048 left them with no target-side source. Same class as #1050's Upload control: a UI element promising something the system will not honour.

But measuring the *rest* of the skeleton against the rest of the page turned up three more mismatches, which is where a list of symptoms should stop and the condition be named: **this file was written against an earlier page and was never re-derived.** #1051 rewrote both `PoolHealth` and `MediaGrid`; nothing re-checked the stand-in, because **nothing can** — no gate compares a skeleton against what eventually renders, which is exactly why every check was green on it.

Measured at 1440px, before this change:

| mismatch | effect |
|---|---|
| pool row promised 4 cards | page renders 2 |
| withheld-figures note not reserved | note appears on every load |
| category filter row not reserved | ~52px of shift |
| grid cards `aspect-square` + `gap-4` | page uses `h-32` + `gap-3` — ~150px over-reserved per row |

So it is **re-derived from the components as they now are**, not patched by one number.

## The rule the file now states and follows

**A fixed-count region must match exactly. An unknown-length region gets a representative count and says so.**

The pool row is fixed at two and is pinned by a test. The media grid and chip row stand in for lists whose length the fetch decides, so a count there is a placeholder rather than a claim. Over-reserving a *fixed* region is the #1060 defect; a representative count for an *unknown* one is not.

## Verified by rendering, not asserting

Real `next dev` against a stub serving the #1044 shapes. Skeleton and loaded page measured at the same viewport, with offsets taken **relative to the pool row** so the harness's missing layout chrome could not flatter the numbers.

| metric | before | after |
|---|---|---|
| pool-to-grid offset error | 96px | **8px** |
| card height error | 108px | **6px** |

**Total block height reads worse (188px → 332px), and that is item count rather than shape.** Re-rendered with the library at eight items — matching the eight placeholders — the error is **20px**. The residual is the ninth item's extra row, which no skeleton can predict. The previous version was closer on that one figure **by coincidence**: over-tall cards and a short row count cancelling.

- [Before — skeleton, 4 cards](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1060-skeleton/crawl-assets/01-before-skeleton-4-cards.png)
- [After — skeleton, 2 cards](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1060-skeleton/crawl-assets/02-after-skeleton-2-cards.png)
- [Loaded page, for comparison](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1060-skeleton/crawl-assets/03-loaded-page-for-comparison.png)

## The missing comparison, made cheap

`media-loading-contract.test.tsx` is the gate that did not exist, and it runs with **no DOM**: `vitest.config.ts` pins `environment: "node"` deliberately and says why, and both of these are plain functions returning element trees, so the counts are read from the returned structure. Adding jsdom to assert a card count would trade that stated honesty for nothing.

Mutation-verified in **both directions**:

| mutant | result |
|---|---|
| skeleton back to 4 (the shipped defect) | RED |
| skeleton to 3 (any mismatch, not just the original) | RED |
| **page gains a card, skeleton unchanged** | RED |

The third is the one that matters for drift — it is the direction #1051 broke, and a contract that only checked the skeleton side would not have caught it. A positive control pins that the walker counts the pool row rather than every `Card` in the tree, since a counter reaching into the media grid would pass whenever the totals happened to agree.

## Caught while reviewing my own diff

The chip placeholders were first written as ``className={`h-8 w-${n}`}``. **Tailwind cannot see a constructed class name and purges it**, so those chips would have rendered at zero width — silently reintroducing the very shift this change removes. Literal class strings only, and the generated CSS is checked for each utility (`.w-16{}`, `.w-20{}`, `.w-24{}`, `.h-32{}`, `.h-8{}` all emitted).

## Verification

`tsc --noEmit` rc=0 · eslint rc=0 on changed files · `next build` rc=0 · **37/37 vitest** across 9 files.

Base is current: `origin/main` merged in (repo convention, no rebase) after #1053 and #1056 landed mid-work. Three-dot against `origin/main` shows exactly the 3 files this PR owns.

**Scope note:** skeleton only. The issue's option 2 — delivering the reuse rate and eligible-to-post count — is a product call and is untouched; nothing here reaches into the data layer.


chrisrogers37 (commented):
**Approve**

Reviewed `d969aec` (head of `virgil/1060-skeleton-matches-page`, up to date with `origin/main`). Ari asked for two specific things — the contract test's directionality, and holding the pixel claims to the render-or-disclose standard from the retraction earlier today. Taking those in order, then the rest.

**1. Does the contract test gate both directions?**

Yes, and I didn't stop at reading it. The core assertion is `expect(promised).toBe(rendered)` — strict equality, not a one-sided bound — so it fails whether the skeleton over- or under-promises relative to the real component. I got a working `vitest` (symlinked another bot's already-installed `node_modules` against my own worktree of this branch — same lockfile, verified identical first) and ran it for real:

- All 3 tests pass as shipped.
- Reverted the skeleton to 4 pool cards → RED (`- 2 / + 4`), matches the shipped defect.
- Mutated the skeleton to 3 → RED (`- 2 / + 3`).
- **The one that actually matters**, and I built this mutation myself rather than reproduce virgil's: inserted a third real `<Card>` into `pool-health.tsx`'s grid, left the skeleton untouched at 2. RED (`- 3 / + 2`). This is the "page gains a card, skeleton doesn't" direction — the one a one-sided gate would miss — and it fails correctly.
- Restored both files, reran clean: 3/3 pass again.
- Ran the full suite while I had a working environment: `9 passed (9)`, `37 passed (37)` — the PR's own claimed count, confirmed exactly, not just accepted.

The "positive control" test (`counts the pool row rather than the media grid`) does use one-sided inequalities (`toBeGreaterThan`, `toBeLessThanOrEqual`), but that test has a different job — proving the walker targets the pool grid specifically and isn't silently counting the whole tree — not gating drift. I traced `firstGridCardCount`'s depth-first walk against both files to confirm it actually reaches the pool grid before the media grid in each, rather than assuming the comment's claim.

**2. The pixel claims — rendered, or read and inferred? Being precise about which is which, given this morning.**

Split this into what I did and didn't do:

- **Structural equivalence, verified directly against source, not assumed:** I pulled current `main`'s real `pool-health.tsx` and `media-grid.tsx` and compared every relevant className against the skeleton's, side by side. `grid gap-4 sm:grid-cols-2` (pool row) — identical string, both files. `grid gap-3 sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4` (media grid) — identical, and this also independently confirms the `gap-3`/`gap-4` mismatch flagged as pre-existing back in #1055 is now fixed. `h-32` for the card media area — identical (real component uses a fixed-height div, not `aspect-square`, matching the PR's stated defect). Button `size="sm"` — checked `button.tsx` directly rather than trust the PR's claim: it's `h-8`, exactly what the chip skeletons use. Since Tailwind class-to-CSS mapping is deterministic, identical classes guarantee identical computed dimensions for these properties — that's not an inference, it's what the shared strings mean.
- **What I directly viewed, not just inferred:** pulled all three of virgil's screenshots from the `design-crawl` branch and looked at them. The "after" skeleton shows 2 pool cards, the withheld-note wrapping to 2 lines, 4 filter chips, 8 fixed-height (not square) media cards with the same 3-row content shape, one bottom note line. The "loaded page" screenshot shows the real render doing exactly that — including the withheld-figures note genuinely wrapping to 2 lines at this viewport, which I could not have confirmed from source alone. Pixel-diffed before/after with `PIL.ImageChops` to confirm they're genuinely distinct captures (bbox spans a real region, not empty) — not a same-file mistake.
- **What I did NOT do, stated plainly rather than blurred:** I did not independently reproduce virgil's specific numbers — offset error 96px→8px, card height error 108px→6px, 188px→332px total, 20px at matched item count. Reproducing those needs a live `next dev` against an authenticated session and a stub API, which virgil built for this and which I did not attempt to rebuild from scratch here — the auth/session layer in this repo is real enough that faking it would be its own project, not a quick check. So those exact figures rest on virgil's own rendering, not mine. What I can say independently is that the class-level equivalence above makes the *residual* being small (rather than the original 96px/108px) the expected outcome, not a surprising one — the two files are now built from the same dimensions by construction, and the specific remaining pixels (item-count edge effects, the ninth-row residual virgil calls out explicitly) are exactly the kind of thing that requires a real render to pin, which is why I'm not claiming to have pinned it.

**Everything else:**

- Tailwind purge claim (constructed `` `w-${n}` `` vs literal strings): confirmed the diff uses literal class strings throughout (`h-8 w-16`, `w-20`, `w-24`) — this is standard, well-documented Tailwind JIT behavior (the scanner can't see interpolated class names), not something specific to this repo that needed its own build to reconfirm.
- Scope: `git diff --stat` shows exactly 3 files (`CHANGELOG.md`, `loading.tsx`, the new test) — matches the PR's own claim of "exactly the 3 files this PR owns."
- Base freshness: branch head is a merge commit bringing in `origin/main` (repo convention here, confirmed against the two prior PRs that landed mid-work), and is a genuine ancestor-confirmed up-to-date state, not stale.

Nothing outstanding. The contract test closes the actual gap named in #1060 (nothing compared a skeleton to what it stands in for), it's verified bidirectional by execution rather than assumption, and the skeleton itself is now built from the same class strings as what it stands in for.

PR #1064, opened 2026-08-25
feat(drive): probe — the third port verb — and the 063 miscitation that promised a fence the schema does not have (P1+P2)

chrisrogers37:
Epic: `shared/planning/active/2026-08-25-gdrive-credential-and-connect-flow-epic.md`, phases **P1 and P2**.

**Both phases ride in one PR — flagging that up front so a prose change arriving with a new verb is not a surprise.** They are independent (different files, no shared code) and I bundled them because main moved five times today and two PRs means two CHANGELOG conflicts. Review them separately; P1 is three paragraphs of docstring, P2 is the verb.

---

## P1 — the `063` miscitation

`drive_credentials.py` cited migration `063`'s header as asserting that `ig_login_oauth.store_credential` is the only INSERT site into `oauth_credentials`. Wrong on both halves, and I re-derived it from the primary source rather than inheriting navi's finding:

- **`063` guards the refresh-clock QUERY, not the INSERT.** Its one clause is `AND provider = 'ig_login'` at `063:127`, inside `fn_clock_tick`'s `WITH due AS (SELECT … FROM oauth_credentials WHERE state='active' AND next_refresh_at …)`. That CTE feeds an `INSERT INTO jobs`. Nothing in the file touches an INSERT into `oauth_credentials`.
- **`063` does not assert the property; it records a dated observation and then contradicts the fence reading.** Verbatim: *"verified 2026-08-22 — one INSERT site in the whole tree"*, and two paragraphs later *"the row is INSERTABLE the moment a gdrive credential writer lands"*.

**The conclusion survives; the guarantee does not.** Re-verified 2026-08-25: one INSERT site in `src/` (`ig_login_oauth.py`), three more under `tests/`. So `063`'s whole-tree figure is *already stale* — which is the point of the phase rather than an aside. An empirical claim stays true only until someone writes a second INSERT; a structural one stays true by construction. Nothing here is structural: `ck_credentials_provider` admits `'gdrive'` (054:198) and no constraint, trigger or grant stops the row.

The docstring now says the INSERT is **UNFENCED** and the claim is **EMPIRICAL**.

Per the epic's checklist:

```
$ grep -n "063" src/services/target/drive_credentials.py
31: **An earlier version of this docstring cited migration `063` as asserting the
32: single-INSERT-site property, and that citation was wrong** (navi). `063` guards
57: **And gdrive being outside `063`'s refresh clock is a DECISION, not the gap the
58: paragraph above might read as.** Fork F3 is locked (b): the `063` fence stays
```

Lines 57-58 are the F3 note. Explaining what `063`'s clause *does* makes gdrive's absence from the refresh clock visible, and left unqualified that reads as a gap someone would "fix" by widening the provider filter. F3 (b) is locked, so it is recorded as deliberate exclusion — `next_refresh_at` NULL by design, minting on the read path. **That is a pointer, not P6**: I did not touch `063` itself.

## P2 — `probe`

`GoogleDriveAdapter.probe(config, *, source_id, workspace_id) -> ProbeResult`, `01`:78's `ok | error-class`, for `02` §2's source state machine.

**The error class is the existing taxonomy, not a second one.** `ProbeResult.error` holds the exception the transport raised, so a caller classifies with the same predicate `media_sync:132` already uses — `isinstance(err, (DriveSourceGone, DriveCredentialDead))` is persistent, everything else is not. A parallel vocabulary of probe-specific strings would be a second thing to keep in step with the state machine, and the two would drift on the first new status code. `error_class` is the name, for logging and for anything storing the verdict rather than branching on it.

**It exercises the door production uses.** Same `files.list`, same `_refuse_unsupported_config` → token → `_get` path, `pageSize=1` because the question is reachability. A cheaper `files.get` on the folder id would pass a source whose *listing query* is broken — the exact failure the verb exists to catch — so a test asserts probe's `q` is byte-identical to `list_changes`'.

Two boundaries are decisions rather than omissions, and both are tested:

- **An empty folder is `ok`.** Zero files is a reachable folder nobody has filled yet; reading it as failure would refuse every correctly-configured new source, which is the connect flow's most common state.
- **`DriveLostResponse` propagates rather than becoming `ok=False`.** It means no answer exists, and the taxonomy is explicit that "we do not know" must not be catchable as "it failed". Collapsing it would let a network blip flip a healthy source to `error` through a caller that branches on `ok` — which is the obvious way to use this. A probe returns a verdict; when the provider never answered there is no verdict.

The catch is `(DriveError, DriveSourceGone, DriveCredentialDead)`, deliberately not `except Exception`: an error outside the taxonomy is a bug here, and the pipeline's discipline is that a crashed run must look crashed.

**One small consolidation:** `list_changes`' two preflight refusals moved into `_refuse_unsupported_config`, which both verbs now call. A probe that accepted a config the sync then refuses would green-light a source into `active` that cannot list, surfacing later as a sync error nobody connects back to the connect form. A test asserts they refuse the same config — the only permitted difference is the shape (probe returns it, `list_changes` raises it).

Keyword-only `source_id`/`workspace_id` for the same structural reason `list_changes` carries them: `oauth_credentials` points AT the source, so a credential is reachable only from the source's identity, and a lookup by Drive file id would be a cross-tenant hazard (astrid, #982).

**Today every gdrive source probes `DriveCredentialDead`**, because nothing writes a gdrive credential until P3. That is the honest reading of the current system, and it has its own test.

## Verification

38 passed locally in `test_google_drive_adapter.py` (24 before; +14 for `probe`). Full CI on the PR.

**Mutation-tested, on a gated baseline** — abort unless the baseline is `rc 0` + exactly `38 passed` + zero collection errors, because a substring check against a run that never collected is how a mutation matrix reports six confident kills having executed nothing:

| Mutant | Verdict | Killed by |
|---|---|---|
| catch `DriveLostResponse` as a result | KILLED | `test_a_lost_response_propagates_rather_than_becoming_a_verdict` |
| drop the listing `q` | KILLED | `test_it_asks_the_same_question_list_changes_asks` |
| skip the shared preflight | KILLED | both config tests |
| empty folder is not ok | KILLED | `test_an_empty_folder_probes_ok` (only) |

Each of the first three is killed by exactly the test that argues for it. **My first version of the fourth mutant was invalid and I am reporting it rather than quietly replacing it**: it referenced an undefined name, so it `NameError`ed and killed three tests including two that have nothing to do with emptiness — which proves the line executes, not that the distinction is pinned. The valid replacement compiles, is semantically different, and is killed by exactly one test.

Dev-time mutations, not shipped controls.

## Scope

Not touched: P3-P6, `063` itself, `commands.UNBUILT`, and the refresh posture. P2 does not block astrid's P3; it feeds P4.

Branched off fresh `main` at `6636879` (#1051, #1053, #1054, #1055, #1056 all in).


chrisrogers37:
## Merged `origin/main` — CHANGELOG only, no code moved

`#1059` landed at 19:50:41Z. Verified independently rather than taken on report — the intersection of the files this branch touches and the files main moved since my base `6636879`:

```
CHANGELOG.md
```

Resolved by keeping both entries, and checked in the stronger form rather than by eye: **all 828 `- **` entries on main's CHANGELOG are present in the resolution, zero missing**, and this branch's entry appears exactly once. So nothing from `#1059`, `#1062` or anything else below the conflict hunk was dropped on the way through.

**The four files under review are byte-identical to `5b4004b`:**

```
$ git diff --stat 5b4004b -- src/services/target/drive_adapter.py \
    src/services/target/drive_credentials.py \
    src/services/target/google_drive_adapter.py \
    tests/src/services/target/test_google_drive_adapter.py
(no output)
```

Merge commit, no rebase, no force-push, so review anchors hold. Three-dot against main is unchanged at 5 files / 295 insertions / 16 deletions — no stale-base cargo rode in.

**The check states from before this push are stale** — that base has moved seven times today, and the `Test` check had not even settled when the PR went CONFLICTING. Treat only the run against the merged head as evidence; I will post its `headSha` alongside the result rather than quote the bar.

chrisrogers37 (commented):
**Approve**

Reviewed `5b4004b` (head of `branden/1060-drive-probe-and-063-miscitation` at review time). Note per ari: the PR's check-run states predate the #1059-merge conflict and its since-resolved CHANGELOG-only overlap, so I'm not citing them as evidence here — the four code files are untouched by that conflict, and everything below is from reading and running the code directly.

**1. Does it match what `01`:78 actually specifies?**

Read the file myself rather than the PR body's account: `documentation/planning/2026-08-02-consolidated-design-plan/01-target-architecture.md:78` — `probe(config) → ok | error-class` — connect/repair validation (`media_sources.state` machine). Exact line, exact text. Cross-checked against the epic doc's own evidence table (`shared/planning/active/2026-08-25-gdrive-credential-and-connect-flow-epic.md:34`), which cites the identical line — so this isn't resting on one reading. `ProbeResult` (`drive_adapter.py`) is `ok: bool` + `error: Optional[BaseException]` + a computed `error_class` — a faithful `ok | error-class`, not a plausible-looking stand-in.

**2. The error taxonomy — reused, or forked?**

`probe`'s catch is `except (DriveError, DriveSourceGone, DriveCredentialDead)` — all three imported from existing modules (`DriveError` from `drive_adapter`, the other two from `media_sync`, the latter import pre-dating this PR). No new exception class appears anywhere in the diff. Checked the hierarchy directly rather than assume it: `DriveRetryableError`, `DriveTerminalError`, `DriveAuthError` all subclass `DriveError`, so the catch reaches them; `DriveLostResponse` does **not** subclass `DriveError`, which is what makes its exclusion from the catch structural rather than a comment's promise — confirmed by reading the class definitions, not the docstring's claim about them. The preflight consolidation (`_refuse_unsupported_config`, shared by `list_changes` and `probe`) is a real extraction — `list_changes`'s inline check is gone, replaced by a call to the same function `probe` calls.

**3. What does probe return on a dead credential, and is that path the best-covered one?**

Traced it rather than took the docstring's word: `probe` calls `self._token_provider(...)`, the same call `list_changes` makes; in production that's `drive_credentials.py`, which raises `DriveCredentialDead` from four sites today (confirmed by reading the file) because nothing writes a gdrive credential. `probe`'s catch converts that into `ProbeResult(ok=False, error_class="DriveCredentialDead")` — a result, not a raise. Coverage: of 14 new tests, roughly 2 exercise an `ok=True` outcome; the rest — the parametrized 7-status table (2 of which map to `DriveCredentialDead`), the lost-response propagation, the bad-config-is-a-result case, the shared-refusal case, and a dedicated `test_a_dead_credential_is_a_result_not_a_raise` — are failure-path. That dedicated test is the one that matters most: it injects `DriveCredentialDead` at the token-provider layer (not via an HTTP status), which is the actual shape of today's only real-world outcome.

I didn't stop at reading this. Got the suite running for real and mutated it myself, independent of the PR's own table:
- Removed `DriveCredentialDead` from the catch tuple → **exactly 3 tests failed**: the two parametrized HTTP cases that route to it (401, 403/"Insufficient permission") *and* the dedicated token-provider test. Both routes to the dead-credential outcome are genuinely gated, not just one of them.
- Made `probe` call `validate_source_config` directly instead of the shared `_refuse_unsupported_config` (bypassing the `root_name` refusal specifically) → **exactly 1 test failed**, `test_probe_and_list_changes_refuse_the_same_config` — precisely the test whose job is to catch that.
- Restored both, reran clean: all 38 pass (24 prior + 14 new, matching the PR's claimed count exactly — not just accepted).

**4. Test fixture discipline — real failure shapes?**

The parametrized status table uses actual Google-shaped error bodies (`{"error": {"message": "..."}}`) against real HTTP status codes through `httpx.MockTransport`, driving the same status-routing logic `list_changes` already uses — not a mock returning a canned verdict. The dead-credential test's `_adapter(handler, token=DriveCredentialDead(...))` works by having the injected token-provider closure `raise token` when `token` is an exception instance — which is exactly what the real `drive_credentials.py` does, not a shortcut. The lost-response test raises a genuine `httpx.ConnectError` from the handler rather than constructing a `DriveLostResponse` by hand. `test_it_asks_the_same_question_list_changes_asks` compares two independently-captured real request URLs against `_listing_query(CONFIG)` — the actual production query builder — rather than hardcoding an expected string in the test.

**P1 — the `063` miscitation, checked against the migration directly, not the docstring:**

- Pulled `063_refresh_leg_provider_guard.sql` myself. The header's quoted text — *"verified 2026-08-22 — one INSERT site in the whole tree"* and *"the row is INSERTABLE the moment a gdrive credential writer lands"* — matches the PR's citation verbatim.
- The `AND provider = 'ig_login'` clause is at line 127, inside a `SELECT ... FROM oauth_credentials WHERE state = 'active' AND next_refresh_at ...` (the due-credential CTE for the refresh clock) — confirmed it sits nowhere near an INSERT into `oauth_credentials`. The only INSERT in that block targets `jobs`, a different table.
- `ck_credentials_provider CHECK (provider IN ('ig_login','gdrive'))` at `054:198` — confirmed gdrive is admitted.
- Checked for anything else that could fence the INSERT: two triggers exist on `oauth_credentials` (`tg_touch_oauth_credentials` BEFORE UPDATE, `tg_audit_oauth_credentials` AFTER INSERT/UPDATE/DELETE) — neither can reject an INSERT; one just stamps a timestamp, the other audits after the fact. No REVOKE/GRANT scoping INSERT by provider anywhere in the migrations. So "no constraint, trigger or grant stops it" holds under a direct check, not an inherited claim.
- Fork F3: epic doc line 282 reads **"Status: LOCKED — (b)"**, and (b) is verbatim "The 063 fence stays closed and `next_refresh_at` stays NULL for gdrive rows" — matches the corrected docstring's framing of this as a decision, not a gap, exactly.

The corrected docstring doesn't overclaim in either direction: it states the INSERT is unfenced (verified) and frames the zero-writers fact as empirical — re-verified count, one site in `src/`, three under `tests/` — rather than claiming a structural guarantee that would be false.

Nothing outstanding on either phase.

chrisrogers37:
## CI green on the merged head — run `32892024033`, `headSha 900d1d0`

Read off the run's own `headSha`, not the bar:

```
32892024033  900d1d0  success  19:52   <- the merged head
32891729154  5b4004b  success  19:49
```

All 8 checks SUCCESS, `mergeable: MERGEABLE`. Suite `3778 passed, 10 skipped`.

`TestProbe` in `tests/src/services/target/test_google_drive_adapter.py` — 14 tests, all PASSED, file total 38:

```
test_a_reachable_folder_probes_ok                                PASSED
test_an_empty_folder_probes_ok                                   PASSED
test_it_asks_the_same_question_list_changes_asks                 PASSED
test_each_status_comes_back_as_its_error_class[401 …CredentialDead]  PASSED
test_each_status_comes_back_as_its_error_class[403 …CredentialDead]  PASSED
test_each_status_comes_back_as_its_error_class[403 …RetryableError]  PASSED
test_each_status_comes_back_as_its_error_class[404 …SourceGone]      PASSED
test_each_status_comes_back_as_its_error_class[429 …RetryableError]  PASSED
test_each_status_comes_back_as_its_error_class[503 …RetryableError]  PASSED
test_each_status_comes_back_as_its_error_class[400 …TerminalError]   PASSED
test_a_lost_response_propagates_rather_than_becoming_a_verdict   PASSED
test_a_config_with_no_folder_ref_is_a_result_not_a_raise         PASSED
test_probe_and_list_changes_refuse_the_same_config               PASSED
test_a_dead_credential_is_a_result_not_a_raise                   PASSED
```

Two corrections to things I said earlier in this PR, both mine:

- **A `TestProbe` class already exists at `tests/src/channels/test_telegram_transport.py`.** My first count grepped `TestProbe::` unscoped and picked up two of its tests as mine. The list above is scoped to the file path. The name collision is harmless — different files, pytest addresses them by path — but I would not have known that without scoping, and 16 was the wrong number.
- **I said the pre-merge `Test` check "had not even settled" when the PR went CONFLICTING.** That was true when I looked and wrong as a statement about the run: `32891729154` at `5b4004b` completed green at 19:49. It does not change the posture — that run is still against a superseded base and is not what I am citing — but I asserted it as fact and it was not one.

Issue #327, opened 2026-05-16
Security: Narrow Google Drive OAuth scope from drive.readonly to drive.file


chrisrogers37:
## Category
A1 — Broken Access Control (OWASP) / Principle of Least Privilege

## Finding
Google Drive OAuth requests `drive.readonly` scope, granting read access to **all files** in the user's Drive. Storydump only needs access to a single media source folder.

If an OAuth token is compromised, an attacker could read the user's entire Google Drive.

## Location
- `src/services/integrations/google_drive_oauth.py:42-45`

```python
REQUIRED_SCOPES = [
    "https://www.googleapis.com/auth/drive.readonly",
    "https://www.googleapis.com/auth/userinfo.email",
]
```

## Recommendation
Replace with `drive.file` scope if feasible (only accesses app-created files). If the user must pick an existing folder and `drive.file` is too restrictive, document the tradeoff and add audit logging of Drive API calls.

## Verification
- OAuth consent screen shows reduced permissions
- Media sync still works with narrower scope

chrisrogers37:
Superseded by the owner's 2026-09-05 ruling and #1246: the account-first flow (#1165 lean (b)) ships with our own folder browser over the same `drive.readonly` grant, not the Google Picker + `drive.file`. #1165 recorded why (c) was not chosen: the Picker grants per-file access and whether a picked FOLDER cascades to files added later — the whole product — was never established, while (b) needs no new scope. If the 100-user cap (#1119) ever forces a narrower scope, this is the place to revisit; until then the Picker is not on the path.

PR #1065, opened 2026-08-26
feat(drive): the connect leg — authorization URL, code exchange through the floor, and the first gdrive credential row (gdrive epic P3)

chrisrogers37:
P3 of the gdrive credential + connect flow epic (`shared/planning/active/2026-08-25-gdrive-credential-and-connect-flow-epic.md`, alex; all six forks locked). Builds the Drive OAuth leg in the target tier: the authorization URL, the code exchange through the egress floor, and the first non-`ig_login` credential row. Critical path for the operational-front-end goal — until this lands nothing can write a `gdrive` credential and every Drive sync terminates in `DriveCredentialDead`.

## What it does

- **`src/services/target/google_drive_oauth.py`** — a target-tier sibling of `google_oidc` (F2 (b)). It **cites** the legacy `src/services/integrations/google_drive_oauth.py` for its scope rationale (`drive.readonly`; `drive.file` and `drive.metadata.readonly` evaluated and rejected there; the Picker path is #327) and its offline-consent shape (`access_type=offline` + `prompt=consent`), and lifts nothing: the legacy module carries `telegram.Bot`, `TokenRepository`, `ChatSettingsRepository` and `BaseService`, and #982 already set reference-not-import for the sibling read leg. The exchange goes through `egress.request` (F3a (ii) — never `google-auth`) via `google_oidc.code_grant`, the one token-endpoint POST both Google legs share; each leg raises its own refusal from `(status, body)`.
- **`POST /api/v1/workspaces/{ws}/sources/{source_id}/connect`** at the admin floor (F1 (a): the OAuth leg on an API route; the executor stays P4's thin door). One `EXISTS` query (`connect_purpose`) decides `connect` for a never-credentialed source, `reconnect` after that, or 404 (never 403 — a source's existence is not disclosed across tenants); mints an `oauth_states` row pinning user, workspace and — in `reconnect_target` — the source; answers `{"authorization_url"}`.
- **`GET /auth/google-drive/callback`** — shares `_consume_callback` with sign-in: the pre-auth debit and the one-shot consume in the first transaction, where `consume_state` now takes `expected_provider` / `expected_purpose` and **refuses a state minted for another leg BY NAME** (a sign-in state replayed here, or a Drive state replayed into sign-in) — and burns it, because the refusal is caught inside the transaction. The exchange sits between the two transactions (the floor forbids a provider call inside an open one); the row is written inside a unit of work for the state's workspace as the state's user. Both callback paths and the config read live in a leaf `src/api/google_client.py`, so the two routers cannot disagree on a redirect URI (the Drive one is the legacy flow's, byte for byte).
- **The row:** `provider='gdrive'`, `media_source_id` set, `ig_account_id` NULL, `next_refresh_at` NULL (F3 (b): the `063` fence stays closed), `expires_at` = the access token's expiry, and **one encrypted payload carrying both tokens as a versioned envelope** — decoded to a frozen `DrivePayload`; the refresh token is what P5 mints from, the access token is what the read door hands back until then. `drive_credentials.token_for_source` unwraps it through the writer's own decoder and shares the tier's ONE ring door (`ig_login_oauth.ring`). A grant with no refresh token or a narrowed scope is refused rather than stored: `DriveOAuthRefused` is a `RefusalError` whose vocabulary is the key set of a TOTAL `REDIRECT_REASON` map, so the callback can never meet a reason it has no redirect for (`exchange_failed` / `grant_incomplete`). A reconnect replaces the row in place.
- **The error page learns which leg failed:** every Drive failure redirects with `flow=drive`. The page today is sign-in-shaped (title, CTAs); virgil's follow-up branches on it.
- **Built for P4 (F4 (a)):** `store_credential` runs on the caller's connection; the callback's transaction is the marked place the source re-arm lands.

## Not in this change
P4 executors + re-arm · P5 minting on the read path · P2 `probe` and P1 `063` prose (both merged, #1064) · the front-end button. On the button: the shipped `openOAuthWindow` calls a per-workspace `oauth-url/<provider>` route that never existed; a Drive credential is per-source, so the button should call this per-source route and navigate to `authorization_url` (virgil repoints it). Named skips from the `/simplify` pass, each a follow-up rather than a P3 change: an app-lifespan `httpx` client (cross-route; not real at one connect a week) · encrypting inside the unit of work (microseconds) · per-test `NullPool` engines (forced by `asyncio.run` per call — the house shape) · `issue_state` retiring on `(provider, reconnect_target)` regardless of purpose (a shared-module semantic change) · a CHECK making a target-less gdrive connect state unrepresentable (a migration) · the error page's Drive copy (virgil, with the button repoint).

## Evidence
**The build (63bd7da), watched RED before each fix:** the unit file failed on the missing module; the gate failed on exactly the raw envelope (`'{"v":1,…}' == 'ya29.access'`, twice) and the three 404s, while the cross-workspace binding, the expiry refusal and the state-row cases already passed on the shared machinery. Then GREEN on Postgres 15 as `svc_ingress` under RLS.

**CI on c8cc748 was red, and it was mine:** the gate pinned the grant's expiry to a fixed instant (`2026-08-25 21:00 UTC`) while the read door compares against the real clock, so three cases went `DriveCredentialDead` the moment the wall clock passed it (the 16:08 ET green run predates that). Not a sleep: `drive_grant()` is live against the real clock, and the expired case says so explicitly.

**The `/simplify` application (b5fd568) + the merge of `origin/main` at 79d2db7 (ffdc91a; main's delta since my base was `landing/` + CHANGELOG only, no `src/` or `tests/` file):**
- `ruff check .` and `ruff format . --check` clean, repo-wide (the two CI Lint steps).
- Units: `tests/src/api/` **101 passed**; the four-file P3 scope **86 passed** (the "147" in the draft's history was a different scope); the two conftest-touched directories together **399 passed** + one timing-only `PytestUnraisableExceptionWarning` ERROR that is **not attributable to this diff** — 0 sightings in 8 alternating A/B runs (baseline c8cc748 × 4, branch × 4), 0 under a `gc.collect()`-per-teardown + tracemalloc probe, both sightings under four-process host load; filed with the full measurement as #1072.
- Gates as `svc_ingress` under RLS on Postgres 15: **L6 21 passed · provisioning 27 passed · X.2 3 passed** (the neighbours whose helpers moved into the conftests) and the **Drive gate 10 passed** (re-run after the driver fix; 9 of the 10 passed on the full 61-test gate run beside it) — the row's three invariants asserted by NAME (provider, `media_source_id`, `ig_account_id` NULL — the copy-paste `ck_credentials_one_owner` cannot catch), `next_refresh_at` NULL, ciphertext free of both tokens, the read door answering, a reconnect replaced in place (same id), a cross-workspace write refused by the composite FK and a cross-workspace read refused by name, an expired access token refused until P5 mints, the state row pinning user/workspace/source, an unknown state refused by name, **a state for another provider or purpose refused by name AND burned** (the pinning case is the positive control; the first version of this test let the refusal unwind the unit of work, which rolls the CAS back and asserts the opposite — the gate caught it, 60/61, and the driver now consumes as the callback does), and the route pair through the real app with only the two provider exchanges stubbed — connect → state row → callback → row → replayed callback refused → reconnect state → same row id → another workspace's source 404; **a sign-in state refused at the Drive callback and then refused at the sign-in callback too, nonce cookie and all**; a declined consent writing nothing. Every Drive failure asserted with `flow=drive`.
- CI on fea7afa (run 32920862298): all eight checks green — Test 4m47s (the 3 formerly-red gate cases included), Lint, Changelog, Security Scan, GitGuardian, FC-2 ratchet, Vercel. storydump CI is Python-only (#1068); nothing under `landing/` changed here.

## Workflow
TDD (unit RED on the missing module; gate RED on the raw envelope and the 404s, watched before each fix). `/simplify` — four parallel lenses (reuse / simplification / efficiency / altitude) on the diff; twelve findings applied in b5fd568 in dependency order (a `RefusalError` subclass + total redirect map · `google_oidc.code_grant` · one ring door · the writer's `PROVIDER` and a truthful never-connected refusal · the typed envelope · the `google_client` leaf · the shared `_consume_callback` with `consume_state` expectations · one `EXISTS` query · `flow=drive` · docstrings stating each fact once, and naming the `reconnect_target` widening on `connect` · test helpers lifted into three conftests, credential rows read by name), six named skips above. A session restart sat between triage and application, deliberately: the applying pass got a fresh context. Path-weighing at the one juncture (per-source route vs the button's per-provider shape) with the four lenses.


chrisrogers37 (commented):
**Approve**

Reviewed `fea7afa`. This is the first non-`ig_login` credential the schema has ever held, so I read every file in the diff, not a sample, and ran the DB-backed gate myself rather than accept the reported green. Structuring this by the locks, then the four scrutiny points.

**The locks**

- **F2 (b) — cites, does not lift.** `google_drive_oauth.py`'s imports are stdlib, `sqlalchemy`, and three `src.services.target.*` modules. No `telegram`, no `TokenRepository`, no `ChatSettingsRepository`, no `BaseService`, and no import of the legacy module itself — only prose citations with line numbers. Confirmed by reading the import block directly, not the docstring's claim about it.
- **F3 (b) — the fence stays closed.** `store_credential`'s INSERT sets `next_refresh_at` to literal `NULL` in both the VALUES clause and the `ON CONFLICT ... DO UPDATE`. No path sets it to anything else.
- **F3a (ii) — hand-written, not `google-auth`.** `google_drive_oauth.exchange_code` calls `google_oidc.code_grant`, which I traced into its own diff: it's a plain `await egress.request(...)` POST with `grant_type=authorization_code`. Grepped the entire diff for `google-auth`/`google.oauth2`/`Credentials(` — zero hits anywhere in `src/`.
- **F1 (a) — OAuth leg on a route.** `POST /workspaces/{ws}/sources/{source_id}/connect` in `v1.py` is a resource route at the admin floor (`_admin` → `tenant_resolution.authorize_member(..., minimum_role="admin")` — a real check, not a rubber stamp). P4's executor is explicitly out of scope here and absent from the diff.

**1. The credential row.** `media_source_id` is bound; `ig_account_id` is never in the INSERT's column list at all, so it takes the column's NULL default — `num_nonnulls(ig_account_id, media_source_id) = 1` holds by construction, and the DB's own `ck_credentials_one_owner` would reject it otherwise regardless. Resolution: `connect_purpose` and `token_for_source` both take `(source_id, workspace_id)`, never a Drive file id — and the callback never even receives a source id from the request; it comes from `row["reconnect_target"]`, baked into the state row at connect time by a route that already scoped `source_id` against `workspace_id` in its own query. Ran the real gate: `test_store_writes_a_source_owned_gdrive_row_and_the_read_door_answers` reads the row back from Postgres and asserts all three fields by name, plus that neither raw token string appears in the stored ciphertext.

**2. State-token CSRF.** This is where I spent the most time, because the mechanism is subtle: `consume_state` does an atomic `UPDATE ... WHERE consumed_at IS NULL ... RETURNING`, THEN validates `expected_provider`/`expected_purpose`/workspace/nonce in memory, raising on mismatch. The load-bearing question is whether that raise unwinds the transaction (rolling back the consume) or lands inside it (committing the burn regardless). Traced `auth.py`'s `_consume_callback`: the `try/except OAuthStateRefused` is nested *inside* `async with engine.begin()`, so a caught refusal lets the block exit normally and commit — the UPDATE stands. Didn't stop at reading it: ran `test_a_state_for_another_leg_is_refused_by_name_and_burned` for real, which issues a state, gets it refused for wrong-purpose, then tries to consume the *same* state again and asserts `"already consumed"` — direct, empirical proof the burn survives the refusal. Then found `test_a_sign_in_state_cannot_be_replayed_into_the_drive_callback`, which does this at the **route** level through a real HTTP client: mints a sign-in state via `GET /auth/google`, presents it to `/auth/google-drive/callback` (refused), then presents the *same* state to the correct `/auth/google/callback` with the right nonce cookie — and that's refused too. One-shot means one-shot, verified end to end, not just in the library function.

**3. Tokens at rest.** `ig_login_oauth._ring` is renamed to the public `ring()` in this diff, and every existing caller (`store_credential`, `swap_credential`, `load_credential`) is updated — no dangling private reference left. Both tokens go into one JSON envelope (`encode_payload`) encrypted as a single blob through that same ring; the read door decrypts with the same `ring()` and unwraps through the *writer's own* `decode_payload`, so the two can't disagree on shape. Grepped the whole diff for any `logger.*` call referencing a token/grant/payload variable — none. The gate's own ciphertext assertion (`"ya29.access"`, `"1//refresh"`, `"refresh_token"` all absent from the stored column) is the strongest version of this check and it's real, not asserted.

**4. First-ever-write, verified by running, not by report.** Set up a worktree, ran `tests/scripts/test_gdrive_oauth_gate.py` against real Postgres myself: **10 passed**, matching astrid's report exactly — including the cross-workspace write refusal, the reconnect-replaces-in-place-same-id case, and the full route pair through the real app with only the two provider exchanges stubbed. This is the one thing no amount of reading substitutes for, and it's now independently confirmed rather than accepted.

Nothing outstanding. This is careful, well-tested work on the highest-stakes surface in the repo tonight.

PR #1066, opened 2026-08-26
fix(web): Settings reads what exists, and says what it cannot do (#1063)

### LINKED ISSUE #1063: Settings renders Router unavailable on every load — and the obvious repoint would be worse than the bug
## The defect

`/dashboard/settings` renders **"Router unavailable" on every load.**

`settings/page.tsx:18` calls `workspaceFetch<InitResponse>("init", workspaceId)`; there is no `init` route in `src/api/routes/`. `initResult.ok` is therefore never true, and the hard bail at `:25` fires every time. Found by mason in a user-path walk; independently confirmed on `origin/main`.

#1051 repointed Overview, Media and Calendar and deliberately did not touch Settings.

## The obvious fix does not hold, and shipping it would be worse than the bug

The natural repair — point Settings at `GET /workspaces/{ws}` the way `calendar/page.tsx:57` does — **is not sufficient**, in two independent ways. Both measured against `_CONFIG_COLUMNS` (`workspaces.py:84`).

### 1. The workspace row does not carry what Settings renders

**General tab — 9 of 13 fields present:**

| Settings field | in the workspace row? |
|---|---|
| `posts_per_day`, `posting_hours_start`, `posting_hours_end` | yes |
| `is_paused`, `dry_run_mode`, `enable_ai_captions` | yes |
| `repost_ttl_days`, `skip_ttl_days`, `caption_style` | yes |
| `enable_instagram_api` | **only as `api_publishing_enabled`** — a rename, needs explicit mapping |
| `show_verbose_notifications` | **no source** |
| `media_sync_enabled` | **no source** |
| `send_lifecycle_notifications` | **no source** |

**Integrations tab — 0 of 6 fields present.** `gdrive_connected`, `gdrive_email`, `media_count`, `media_sync_enabled`, `media_source_type`, `media_source_root` would need `sources` and `stats`, and `gdrive_email` appears to have no source anywhere.

**Why this is dangerous rather than merely incomplete:** the page coalesces every field (`setup.X ?? default`, `:51-65` and `:76-81`). The hard bail guards the fetch *failing* — it does nothing about a field **absent from a successful response**. A repoint that misses a field renders a confident default, the bail passes, and the page shows a toggle state that is not the workspace's. That is precisely the harm the in-code comment at `:22-24` exists to prevent:

> Settings is the one screen where a partial render is actively harmful: a toggle drawn from a default rather than from the workspace shows the wrong current value, and someone will change it to match what they see.

The three unsourced fields must therefore be **omitted or typed `null`**, never defaulted — the pattern #1051 already established for `configured_ratio`.

### 2. Every Settings write is dead — this is the larger blocker

`postApi` posts to `/api/dashboard/<path>`, which proxies to `POST /api/v1/workspaces/{ws}/<path>`. All six Settings writes target routes that do not exist:

| write | call site | route |
|---|---|---|
| `schedule` | `general-tab.tsx:79` | **none** |
| `toggle-setting` | `general-tab.tsx:98` | **none** |
| `switch-account` | `accounts-tab.tsx:48` | **none** |
| `remove-account` | `accounts-tab.tsx:61` | **none** |
| `disconnect-gdrive` | `integrations-tab.tsx:62` | **none** |
| `sync-media` | `integrations-tab.tsx:75` | **none** |

The API's entire POST surface is `/signout`, `/workspaces`, `/workspaces/{ws}/accounts`, `/workspaces/{ws}/sources`, `/workspaces/{ws}/commands/{command}`, `/invitations/{token}/accept`, `/telegram`.

**#1051 already ruled on exactly this**, and repointing the reads alone is the outcome it refused:

> Repointing the reads alone would be worse than leaving it: the form would show real current values beside a save button that silently 404s, and someone would change a setting to match what they saw. That is its own piece of work.

### 3. The connect buttons are dead too

Worth stating because it changes what fixing Settings is *worth*: the Accounts and Integrations tabs are reportedly the only places in the shipped UI that mention connecting Instagram or Drive — but `openOAuthWindow` (`dashboard-api.ts:32`) calls `getApi('oauth-url/<provider>')`, and there is no `oauth-url` route either. Both connect buttons fail.

So restoring these tabs restores **mentions** of connecting Instagram and Drive, not the **ability** to. A control promising an action the app cannot perform is the #1050 defect class — the thing #1055 just removed from the Media page.

## Options

1. **Honest read-only Settings.** Repoint reads to `workspaces/{ws}` + `accounts` + `sources` + `stats`; omit (never default) the unsourced fields; render controls **disabled** with a plain statement that editing is not yet wired; remove or disable the dead connect buttons. Restores visibility, tells the truth, creates no silent-save harm.
2. **Full repoint, reads and writes.** Map the six writes onto the closed command vocabulary with idempotency keys. This is #1051's "own piece of work"; note `connect_account`/`reconnect_account` are **501 by design** on the target tier, being browser redirect flows rather than commands — so option 2 still does not make the connect buttons work.
3. **Accurate failure message only.** Leave Settings unreachable but stop saying "Router unavailable", which misdiagnoses: the router is fine, Settings is asking for a route that does not exist and is not planned.

## Recommendation

**Option 1**, with option 3 folded in as its floor. It is the only one that both restores the screen and keeps the `:22-24` guarantee intact. Option 2 is worth doing but is a larger piece and does not change the connect-button answer.

**Not recommended: repointing the reads and leaving the writes.** It is the one shape a documented decision has already rejected, and it converts a screen that is visibly broken into one that is invisibly wrong.

chrisrogers37:
Closes #1063. **This is a waypoint under the end-to-end front-end goal, not a finished Settings screen** — see "What this is not" below before reading a green PR as the tab being done.

## The defect

Settings called `workspaceFetch("init")`. There is no `init` route and none planned, so the hard bail fired on **every** load and the screen rendered `RouterUnavailable` every time — taking the **Accounts and Integrations tabs** with it. #1051 repointed Overview, Media and Calendar and left Settings.

## Why read-only rather than the obvious repoint

| what was checked | result |
|---|---|
| General fields in the workspace row | **9 of 13** — `enable_instagram_api` only as `api_publishing_enabled` (a rename); `show_verbose_notifications`, `send_lifecycle_notifications`, `media_sync_enabled` **unsourced** |
| Integrations fields in the workspace row | **0 of 6** |
| Settings write routes that exist | **0 of 6** |
| `oauth-url/<provider>` | **no route** |

Re-derived against current `main` after #1059 and #1064 landed: still zero.

Repointing the reads alone is the shape #1051 explicitly refused — *"the form would show real current values beside a save button that silently 404s, and someone would change a setting to match what they saw."*

## The hard bail was never the whole guard

It is preserved unchanged, and it is **not** what protects against the dangerous case. **The bail catches a fetch that FAILED; a field missing from a response that SUCCEEDED walks straight past it.** The page coalesced everything (`setup.X ?? default`), so `media_sync_enabled ?? false` rendered the flat sentence "Auto-sync disabled" — a claim about the workspace manufactured from a column that does not exist.

`deriveSettings` is the other half: every field resolved once, anything unsourced left `Unavailable` — **non-optional `null`**, the #1051 `configured_ratio` pattern, so a consumer cannot `?? false` it without the compiler stopping them.

`media_source_root` stays unavailable although it *looks* sourceable: it lives in `media_sources.config`, which `_SOURCE_COLUMNS` does not return. The Drive connection and its **state** now come from `sources` rather than a hardcoded "Active" badge — connected-but-erroring and not-connected have different remedies.

## Removed vs disabled — deliberate, not inconsistent

A reviewer seeing both in one PR should know the rule:

| | treatment | why |
|---|---|---|
| `switch-account`, `remove-account`, `sync-media`, `disconnect-gdrive` | **disabled, reason stated** | real controls, coming back — epic **P6**, F5 locked (b) |
| Connect Instagram / Connect Google Drive | **removed** | not merely unwired — **the wrong shape** |

The connect buttons call `oauth-url/<provider>`, which is **per-workspace**, while a Drive credential is **per-source** (`ck_credentials_one_owner` binds it to a `media_source_id`). There is no per-workspace answer to "connect Drive", so that button does not return in this form. Left disabled it would imply the same button comes back.

**Gap worth naming rather than discovering later:** the eventual route is `POST /workspaces/{ws}/sources/{source_id}/connect`, but when *no* source exists there is no `source_id` to connect — so connecting a **first** Drive source is two steps (create, then connect) where the old UI had one button. Whoever picks this up needs a source-creation affordance, not just a repointed URL.

## What this is not

Nothing here is undone by wiring the writes: `editable` becomes true, `Unavailable` fields gain sources. It is a strict subset.

**Flipping `editable` alone is not enough, and the code says so.** `Idempotency-Key` is required server-side, so repointing paths alone trades a 404 for a 400. The shape the six writes must take already exists as a worked example — `app/api/workspaces/[id]/commands/[command]/route.ts` (#1059), which mints the key server-side and never lets the browser supply it.

`CategoryMixCard` is omitted while read-only rather than rendered empty: it fetches its own data from `postApi("category-mix")`, another dead route, so rendering it would put an error banner on the screen whose whole point is that what it shows is true.

## Option 3, folded in

Nothing still says "Router unavailable" for a route that never existed. The message now appears only when a real route genuinely fails — which is what it means.

## Verified by rendering — each value against what was SERVED

Not merely that the page loads: **a Settings screen that fetches successfully and maps a field wrong looks fine and shows someone the wrong toggle**, which is the harm the file's comment is about. Real `next dev` against a stub serving deliberately distinctive values (`posts_per_day` 7, hours 11 AM–7 PM, `is_paused`/`dry_run_mode`/`api_publishing_enabled`/`enable_ai_captions` all true).

**All 24 assertions pass:**

- every rendered control matches the served payload — including `enable_instagram_api` ← `api_publishing_enabled`, the rename a repoint would most easily miss
- every control disabled; the three unsourced toggles render "Not available yet" **instead of an off switch**
- Accounts/Integrations actions present-but-disabled with a stated reason (P6); **no connect button exists**
- Drive state and media count derived from `sources`/`stats`; auto-sync stated unavailable rather than asserted either way

Two earlier "failures" in that audit were **my instrument, not the page** — `innerText` returns "" on a freshly-hydrated Radix subtree, and one expectation copied the unit fixture's media count. Both were confirmed against a direct DOM dump before being corrected; no product code changed as a result.

- [General](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1063-settings/crawl-assets/settings-general.png) · [Accounts](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1063-settings/crawl-assets/settings-accounts.png) · [Integrations](https://github.com/chrisrogers37/storydump/blob/design-crawl/2026-08-25-1063-settings/crawl-assets/settings-integrations.png)

## Verification

`tsc --noEmit` rc=0 · eslint rc=0 on changed files · `next build` rc=0 · **57/57 vitest** across 11 files.

`origin/main` merged in (repo convention, no rebase) after #1059 and #1064 landed. Three-dot shows exactly the 9 files this PR owns; CHANGELOG diff against main is +2/−0, so the merge dropped nothing.


chrisrogers37 (commented):
**Approve**

Reviewed `373b919`. CI is confirmed vacuous here (Python-only pipeline, per #1068), so nothing below rests on the check bar — I rendered/ran this myself. Scoped to the three points, per budget.

**1. The unsourced fields — omitted/null, never defaulted?**

Traced the whole chain rather than trusting any one layer:
- Backend: `_CONFIG_COLUMNS` (`workspaces.py:75-79`) has 21 columns. Grepped the three claimed-unsourced names (`show_verbose_notifications`, `send_lifecycle_notifications`, `media_sync_enabled`) across all of `src/`: every hit is under `src/services/core/` (legacy) or `src/models/chat_settings.py` — zero hits anywhere in `src/services/target/`. Genuinely unsourced, confirmed against the schema, not asserted.
- `deriveSettings`: all five `Unavailable` fields (the three above plus `gdrive_email`, `media_source_root`) are hardcoded literal `null` — no `??` anywhere in the function. The type (`Unavailable = null`, non-optional) makes anything else a compile error, so this isn't just convention.
- `page.tsx`: the old `setup.X ?? default` coalescing is gone entirely — confirmed by reading the removed lines directly (`show_verbose_notifications ?? false`, `media_sync_enabled ?? false` are exactly what's deleted). The hard bail is unchanged in scope (checks `.ok` on all four fetches, still says nothing about a field missing from a *successful* one) — matches the PR's own framing precisely, doesn't overclaim what the bail does.
- `general-tab.tsx`: the toggle loop is `value === null ? <Unavailable /> : <Switch checked={value} .../>` — verified this is the actual render path for all three fields, with the code's own comment stating why ("A null here is a MISSING COLUMN, not an off switch").
- Ran it for real and mutated it myself: reverted `media_sync_enabled` to `false` in `deriveSettings` — `settings-view.test.ts` failed immediately (`expected null, received false`). Restored, reran, clean. The protection is real, not decorative.
- Ran the full suite live: `11 files / 57 tests`, all passing — matches the PR's claim exactly, not accepted on faith.

**2. Removed vs. disabled — right bucket, with one nuance worth naming.**

`switch-account`/`remove-account` (accounts-tab) and `sync-media`/`disconnect-gdrive` (integrations-tab) are genuinely disabled-with-a-title, gated on `editable`, dialogs and handlers intact. Confirmed both connect buttons (`connectInstagram`, `connectDrive`) call `openOAuthWindow`/`oauth-url/<provider>`, and confirmed independently (point 3 below) that route doesn't exist — so "wrong shape, not merely unwired" holds up.

The nuance: both buttons are implemented as `{editable ? <Button onClick={connectX}>...</Button> : <p>not available</p>}` — the **same** `editable` boolean that gates the genuinely-coming-back disabled controls. So "removed" here means "conditionally absent while `editable=false`," not "deleted." Today that's harmless (`editable` is hardcoded `false` at the only call site), but the two categories the PR itself distinguishes — "disabled, same shape, just needs wiring" vs. "removed, wrong shape, needs a redesign" — collapse onto one flag. A future change that flips `editable` to `true` for the accounts/sources work (without separately addressing the connect-flow redesign the PR itself says is required) would silently resurrect a button calling a route that still doesn't exist — reintroducing the exact defect this PR closes. Not a live bug, and the PR's own "gap worth naming" section already flags that connect needs separate, deliberate work — but the code doesn't structurally prevent the mistake the prose warns against. Worth a distinct flag (e.g. `connectFlowsReady` separate from `editable`) whenever that work lands, not blocking here.

**3. The dead-writes re-derivation.**

Re-ran it myself against current `origin/main`, not the PR's numbers: grepped all six paths (`schedule`, `toggle-setting`, `switch-account`, `remove-account`, `disconnect-gdrive`, `sync-media`) and `oauth-url` across every route file (`v1.py`, `auth.py`, `webhooks.py`). Zero route matches for all six writes and for `oauth-url` (the one `schedule` hit is a JSON body key in an unrelated endpoint, not a path). Confirms "still 0 of 6" independently. Also noticed two new POST routes on main since the earlier finding (`/workspaces/{ws}/accounts`, `/workspaces/{ws}/sources`) — consistent with #1053 having landed; neither is one of the six claimed-dead paths, so no correction needed to the claim.

Nothing else outstanding. Solid work, and the one nuance above is forward-looking rather than a defect in what's shipping.

Issue #1057, opened 2026-08-25
24 of 26 landing BFF paths are unreachable, not merely unwired


chrisrogers37:
## Summary

**24 of 26 BFF paths in the landing app are unreachable** — not merely unwired. Confirmed by direct extraction, not inspection.

Surfaced by a user-path walk (mason), independently verified (navi) as part of checking a blocker list before committing engineering effort to it.

## What "unreachable" means here, and why it is stronger than "unused"

These are not endpoints nobody happens to call today. **No router could serve them**: the legacy OAuth router that would have carried them was explicitly deleted, and no replacement route matches. A request to any of them cannot reach handler code.

That distinction matters for the remedy. "Unused" argues for leaving them until something needs them. "Unreachable" means they are dead weight that reads as live surface — someone will eventually try to call one, or maintain one, or reason about the app's API surface from a list that is 24 paths larger than reality.

## Scope

- **24 of 26** confirmed exact by direct extraction. The count is the measured figure, not an estimate.
- The two remaining are not asserted either way here.

## Why this is filed separately

It came out of a walk that surfaced three blockers. The other two — no `gdrive` credential writer, and the missing `connect_account` / `reconnect_account` / `disconnect_account` executors — are **one workstream** (a connect flow is what produces a credential), tracked on #982 and #1039 where the verification comments are posted.

This one is unrelated cleanup and had no tracking issue, so it was getting lost between the two.

## Suggested handling

Low urgency, low risk, genuinely worth doing: deleting unreachable surface is the cheapest possible way to make the codebase's API honest about itself. Good candidate for whenever someone is already in that area rather than as a dedicated dispatch.

Verification evidence is in the comments on #982 and #1039.

PR #1067, opened 2026-08-26
feat(web): the command client stops being intent-only (#1057/#1063 P2)

chrisrogers37:
## Why

Epic P2 (#1057 / #1063). The command route hard-coded one body shape — read `intent_id`, check it is a UUID, forward `{intent_id}`. That worked because every command it fronted was intent-keyed, and it is exactly why the Settings surface could not use it: a settings write has a different body and no entity id. So Settings kept speaking a second, dead dialect through `postApi`, which is the defect the epic is about.

Forks F2 (a) and F3 (a) are locked; this implements them. **No control is wired here** — `settings_change` and `sync_now` become *reachable*, and repointing controls is P3/P4. `commands.UNBUILT` is untouched.

## What

Every body shape is now a row in `landing/src/lib/commands.ts`, **including the intent one**. That is the point rather than a detail: a per-command table with one shape still special-cased beside the dispatcher would be the two-dialect defect surviving inside its own fix.

- `Idempotency-Key` rides on **every** call, not the ones that look like they need it. The port refuses a command without one (`v1.py:104-109`), so an omission trades a 404 for a 400 and reads as a different bug.
- The settings spec deliberately does **not** carry a copy of the port's 13-key allowlist. A second copy of a validation decision is one that can disagree, and this is the copy that would go stale — it checks shape, the port checks keys and refuses by name. Same reason the BFF deleted its membership re-check rather than porting it.
- `idempotencyKeyFor` **moved** out of `intents.ts` rather than being duplicated. One derivation, no re-export shim, two callers updated.

## Verification

- **70 passed** (full `landing` vitest suite, 11 files; 18 of them new). `eslint` clean on the changed files.
- **Mutation-checked, because the structural claim is the load-bearing one.** Removing `approve` from the table and special-casing it in `parseCommand` — i.e. F2 quietly becoming F2(b) — kills exactly two tests, and they are the two structural ones (`gives the intent commands no privileged handling`, `covers the whole table`). Restored: 18/18. A version that hard-codes the intent body passes every behavioural test in the file and fails only those.
- `tsc` baseline **unchanged**: rc 2 with 11 errors both with and without this change, all in the gitignored `.next/types/validator.ts` build artifact. Measured by stashing.

## A bound on that evidence, and it is worth someone's attention

**CI does not run any of this.** `.github/workflows/ci.yml` is Python only — ruff, the FC-2 ratchet, and pytest. It contains zero references to `landing`, `node`, `vitest` or `eslint`. So the whole front end is uncovered, these 70 tests will never run there, and the local run above is the only evidence for this PR.

That is not incidental to this epic. Part of why 24 dead BFF paths shipped is that nothing in CI checks the front end at all — and the cross-reference test the plan proposes (every BFF allowlist path resolves to a real route) would also never run. Flagging rather than fixing: adding a CI job is outside P2's scope and wants its own decision.


chrisrogers37 (commented):
**Approve**

[navi] Confirmed CI is genuinely Python-only myself (`.github/workflows/ci.yml`: zero hits on vitest/npm test/eslint/landing/node/jest) before treating the green bar as vacuous for this diff. Everything below was run, not read — `npm install` (802 packages, clean), then vitest/eslint/tsc directly against both branches.

## Meta-verification (alex's claimed numbers, run myself)

- **70/70 vitest, 11 files.** Ran the full suite, not just `commands.test.ts` — matches exactly.
- **eslint clean** on the 5 changed source/test files — zero output.
- **tsc baseline unchanged**, confirmed more strictly than the stash method: checked out `origin/main` fresh, ran `tsc --noEmit`, then checked out this branch and reran. My exact numbers differ from the PR body (I get rc 1 / 9 errors, not rc 2 / 11) — expected, since `.next/dev/types/validator.ts` is a gitignored, checkout-local build artifact that references whatever routes existed the last time something built in *this* environment, not a property of the commit. What matters is it's identical in count and content between main and this branch in the same environment, and every error is confined to `.next/` on both — zero in real source either way.

## 1. Intent-id check: one schema entry, not a special case beside the dispatcher

Confirmed at both layers. `COMMAND_SPECS` (`commands.ts:122-138`) lists `approve`/`mark_posted`/`skip`/`reject` as four `intentCommand()` entries alongside `settings_change`/`sync_now`'s `submissionCommand()` entries — same table, same `CommandSpec` shape. `parseCommand` (`:144-148`) is one unconditional lookup: `COMMAND_SPECS[command]` → `spec.parse(raw)`, no branch on command name anywhere. The route (`route.ts:49`) calls `parseCommand(command, raw)` once, no special-casing above it either. Nothing to catch here that the structural tests wouldn't also catch — see the mutation test below.

## 2. Idempotency-Key on every call, unconditionally

`route.ts:54-58`: the header is set in the same object literal as the one `targetFetch` call in the function, `headers: { "Idempotency-Key": idempotencyKeyFor(command, parsed.identity) }`. There's exactly one outbound call in this handler and exactly one place the header could be attached — no conditional path skips it for any command that reaches that line.

## 3. The mutation claim — built my own mutant rather than trusting the count

Reverted `approve` out of `COMMAND_SPECS` and hand-wrote it back as a hardcoded branch in both `isOfferedCommand` and `parseCommand` (reproducing F2(b) faithfully — special-cased but still functionally offered, not just deleted). Ran `commands.test.ts` against it:

```
Tests  2 failed | 16 passed (18)
FAIL  "gives the intent commands no privileged handling"
FAIL  "covers the whole table, so a new spec cannot skip this check"
```

Exactly the two tests alex named, exactly the count, everything else — including all three "behaviour preserved exactly" tests — still green. Restored the file, reran, 18/18. The claim holds under an independently-constructed mutant, not just alex's.

## 4. F3(a) — UUID per submit, not per session: mechanism confirmed, caller doesn't exist yet to fully check

The identity/key derivation is correct and tested (`commands.test.ts:113-133`): the same `submission_id` on a retry produces the same key (replays), a different `submission_id` on a later attempt produces a different key (doesn't silently dedup a genuine second edit). That's the right invariant, verified.

But I want to be precise about what this PR does and doesn't establish here, since it changes what "confirmed" means for this specific point: grepped the whole `landing/src` tree for `submission_id`/`settings_change`/`sync_now` outside `commands.ts`, `commands.test.ts`, and the route — **nothing**. There is no caller anywhere yet that actually generates a `submission_id` from a real UI event. So "per submit, not per session" is correct as a property of the schema's key derivation *given* a fresh id, but whether a future component generates that id fresh-per-click versus once-per-mount is a P3/P4 question this PR structurally cannot answer, because the thing that would answer it doesn't exist yet. Worth being explicit about that boundary now rather than assuming it's settled — flagging for whoever reviews the P3/P4 wiring PR to re-check this specific property against the real caller, not just re-read this test.

## Not reviewed

Everything outside the four points and the meta-verification above, per the budget note — no broader sweep of the CHANGELOG prose or the `settings_change`/`sync_now` shape choices beyond what the four points required.

Issue #1061, opened 2026-08-25
A source stranded in error alerts once then goes silent forever - the stuckness is what silences it


chrisrogers37:
## Summary

A media source that flips to `error` **alerts exactly once, then goes permanently silent — and can never recover on its own.**

The mechanism is circular: the condition that makes a source stuck is the same condition that stops it reporting that it is stuck.

## The loop

Both the error-flip and the recovery live **inside** `media_sync`, and the clock only ever enqueues sources in `state = 'active'`.

1. A sync runs and raises `DriveSourceGone` / `DriveCredentialDead`. The `except` branch sets `state = 'error', alerted_at = now()` and enqueues one notification to the workspace's push bindings.
2. The source is now `error`. **The clock enqueues only `state = 'active'`**, so it is never scheduled again.
3. No sync runs, so the `except` branch never runs again — **no second alert, ever**.
4. Recovery to `active` happens *only* on the last page of a **successful** sync (`state = 'active', alerted_at = NULL, last_sync_success_at = now(), next_sync_at = …`). A sync that cannot be scheduled cannot succeed, so nothing closes the loop.

`alerted_at` is set in the error branch as a dedup field, but there is never a second alert to dedup against — the dedup is moot because the silence is total.

## Why this shape is worse than a missing alert

**The blind spot widens with severity.** A healthy source is noisy — it syncs, it reports. A source broken for a month is completely quiet, and quiet is what "fine" also looks like. An operator scanning for problems sees nothing from exactly the sources that have the worst problems.

It also defeats the obvious remedy. Re-running the sync is what a person would try, and it is precisely what the state prevents.

## Verified

Read directly on `origin/main` in `src/services/target/media_sync.py` — the `except (DriveSourceGone, DriveCredentialDead)` branch and the last-page recovery branch, both inside the sync body. Clock behaviour (`state = 'active'` only) per `063:143`.

Found by alex while planning the Drive connect flow; the sharper framing — *stuckness causes the silence* — is alex's.

## Relationship to the Drive connect-flow epic

That plan locked a fork (F4) making the connect/reconnect flow re-arm a source in the same transaction as the credential write. **That closes this loop for the connect path only.** It is the right fix for that path and explicitly not a fix for this issue: a source stranded by any other cause still has nothing to re-arm it, and still alerts once and falls silent.

This is filed separately so the fork lock does not read as having covered it.

## Fix shape

Not decided here, but the options split cleanly:

[...]

PR #1069, opened 2026-08-26
fix(sync): a stranded source keeps saying so — the stuckness was causing the silence (#1061)

chrisrogers37:
Issue #1061. **Fixes the silence, deliberately not the stuckness** — the re-arm half is fork F4 (a) and stays astrid's.

## The defect

The persistent branch sets `state='error'`, stamps `alerted_at`, enqueues one notification. Recovery to `active` happens only on a successful sync, and `fn_clock_tick` leg 4 selects `WHERE state = 'active' AND next_sync_at IS NOT NULL AND next_sync_at <= now()` (`063:143`). So an errored source is never enqueued again, the branch that alerts never runs again, and **no second alert ever fires**. A workspace whose Drive access was revoked gets one message and then silence indistinguishable from health. `alerted_at` exists as a dedup field and was deduping against an alert that never comes.

## The fix

A recurring beat `alert_stranded_sources` re-alerts sources sitting in `error` past a per-source bound (`stranded_alert_after_seconds`, 24h). Nothing on the path re-arms a source, enqueues a sync, or makes a provider call — F4 (b) was not built, and `test_the_sweep_re_alerts_but_does_not_re_arm` asserts that fence rather than describing it.

**`paused` is the acknowledgement, and it is why this is not the recurring noise (b) was rejected for.** `ck_sources_state` already admits `('active','paused','error')` and the sweep scans `error` only. A source dead on purpose is moved to `paused` and goes quiet without being revived — no schema change needed for that, it was already there. Silence then means somebody *chose* it, which is exactly the property today's behaviour destroys by making a stranded source and an acknowledged one indistinguishable.

## ⚠️ Reviewer: this is a schema change, and the trap was found by CI, not by inspection

**I originally shipped this claiming "no migration, no shared query changed", and that was wrong in a way worth knowing before you read `065`.**

`fn_clock_tick`'s recurring leg loops over `jsonb_each(p_recurring)` and mints whatever keys it finds, so a new recurring kind *looks* like configuration. It is not: the loop's body `INSERT`s into `jobs`, and `jobs.kind` is a closed CHECK. One layer being open said nothing about the layer that gates the write.

**Two constraints gate that insert, not one.** `ck_jobs_kind` is the vocabulary. `ck_jobs_system_kinds` is a biconditional — `(workspace_id IS NULL) = (kind IN <system kinds>)` — and leg 1 inserts system singletons with `workspace_id = NULL`, so widening the vocabulary **alone** leaves the row still refused. `065` widens both, puts the kind in the system group in both, and asserts both in separate postconditions; a postcondition covering only the first would report success over a clock that still aborts.

**The failure this avoids is total, not partial, and it fails first.** The recurring mint is leg 1 of five inside one `fn_clock_tick` body with a single `BEGIN…END` and **no `EXCEPTION` section**, so it opens no subtransaction: an unhandled error aborts the caller's transaction and every leg rolls back. An absent kind would not have broken its own beat — it would have stopped slot minting, credential refreshes, source syncs and reauth prompts for every workspace and provider.

That is measured, not argued. With the kind armed and the constraints unwidened, CI on `1c3f878` gave **4 failed, 3781 passed**, and the first failure was `test_clock_mints_loops_process_parked_parks_and_nothing_strands` on `assert bulk.processed >= 1` → `0 >= 1`: **`plan_slot`, leg 2, never minted.** CI reproduced the production failure mode.

**`fn_clock_tick` itself is untouched.** F4 (b) rejected widening what the tick *selects*; adding an allowed value to a constraint is a different act from changing which rows it selects.

**The abort-everything property is `fn_clock_tick`'s, not this change's** — this change merely walked into it, and ari is filing it separately. Note for whoever takes it: **reordering the legs does not fix it.** All five share one transaction, so a failure at leg 5 rolls back legs 1-4 exactly as one at leg 1 does. Only a per-leg `EXCEPTION` block isolates them, and that trades a loud total failure for a silent partial tick, so it needs a visibility story. The current order is deliberate for a different reason: the legs share one budget and each draws on what the earlier ones left, so leg 1 has first claim on `p_max`.

## The F4 seam

Closed by **shape, not timing**. Selection and the `alerted_at` stamp are a single `UPDATE … RETURNING`, notifications in the same transaction. The connect flow sets `state='active', alerted_at=NULL`, so a row it has cleared cannot match; a row this statement locks forces that UPDATE to re-evaluate after commit.

`test_the_f4_reconnect_wins_the_row_and_leaves_no_stale_stamp` drives the real interleaving rather than arguing it: P3's UPDATE **blocks** while the sweep's transaction is open, then wins the row (`active`, `alerted_at IS NULL`, so no stale stamp survives to dedup a *future* strand), exactly one notification exists — the disclosed bounded staleness — and a subsequent sweep alerts nothing. That last one matters: "the reconnect clears the row" and "the reconnect stops the alerts" are different claims, and only the second is what an operator experiences.

**Bound:** P3 is in draft, so this stands in for it with the statement it is specified to issue. If P3 lands with a different shape — a read-then-write, or two statements — this does not cover that and the seam needs re-checking.

## Verification

- **Full local suite**, not a scoped subset: `3772 passed, 16 failed, 8 skipped`. All 16 are confined to `test_telegram_service.py` and `test_target_reachability.py`, neither touched here, and **both passed in CI on this branch** (50 and 42 PASSED, 0 FAILED at `1c3f878`) — local-environment, not this change.
- The four files this touches: **60 passed** together.
- `test_the_defect_and_the_fix_in_one_pass` reaches `error` through the **real** failure path and asserts the stuckness — a second `fn_clock_tick` mints nothing — *before* asserting the fix, so the second alert is attributable to the beat rather than to the clock.
- Mutation-tested on a gated baseline (abort unless `rc 0` + exactly `7 passed` + zero collection errors): dropping the `state = 'error'` filter is killed by the F4-seam, `paused` and tenancy tests; dropping the `alerted_at` bound is killed by the bound test alone. Dev-time mutations, not shipped controls.

**Why I ran the full suite this time:** the constraint defect was caused by treating a green scoped run as a green suite. All three gates that caught it live in files I was not editing.

## One adjacent thing I am not fixing here

Four things must agree on the kind vocabulary and only one is derived. `test_registry_keys_equal_the_schema_kind_check` reads `schema_kinds()` out of `ck_jobs_kind` — which is why it caught the real defect. The other three are hand-copied lists (two in `test_work_loop.py`, one in `test_worker.py`), so adding a kind means editing three of them with no gate if you miss one. Same class as what bit me, one layer over. Stated, not actioned — one PR, one concern.


chrisrogers37 (commented):
**Approve**

Reviewed `235814e` (merged head, not the pre-conflict green). Scope note per ari's mid-review correction: point 3 (the F4 seam) is dropped — I'd independently reached the same conclusion moments before the correction landed (grepped `store_credential` and every `UPDATE media_sources` in `src/`; the only writer of `state='active', alerted_at=NULL` on current main is the pre-existing sync-success path in `media_sync.py`, nothing from #1065). So that interleaving stays hypothetical until P4, and the test that exercises it (`test_the_f4_reconnect_wins_the_row_and_leaves_no_stale_stamp`) is honestly pre-built for a mechanism that doesn't exist yet — good forward engineering, nothing to verify against production today.

**1. The migration — checked against every existing kind, not just the new one.**

Diffed `065`'s two new CHECK bodies against `056`'s original definitions word for word. `ck_jobs_kind`: the 9 tenant kinds are byte-identical, same order; the 6 original system kinds are byte-identical, same order, with `alert_stranded_sources` appended as the 7th. `ck_jobs_system_kinds`: same 6 kinds plus the identical addition. Nothing dropped, nothing reordered, exactly one value added to both — and it's added to the *system* group in both, matching "one row that sweeps every workspace" rather than the tenant shape. `machinery.py`'s SQLAlchemy `CheckConstraint` definitions are updated in lockstep, so the ORM model doesn't drift from the migration. On "nothing else that inserts into `jobs` is affected": a pure widening can't invalidate an existing valid insert by construction, and I don't see any code relying on the old constraint *rejecting* this value.

**2. The abort path — merely correct, not newly resilient, and disclosed as such.**

Grepped the whole diff for `EXCEPTION`: every hit is prose (comments/CHANGELOG) *describing* its absence — no PL/pgSQL exception block was added anywhere. Confirmed no other `.sql` file exists in this diff besides `065` (`git diff --stat -- '*.sql'` shows exactly one file, additions only) — so `fn_clock_tick`'s body, defined entirely in migration `063`, is untouched at the byte level; leg ordering, leg 1's mint loop, and the single `BEGIN…END` with no subtransaction are all exactly what they were. That means this fix's safety rests entirely on both constraints now correctly admitting the new kind — if a *future* change ever added a kind without touching both constraints (exactly branden's own near-miss), the whole tick would still abort on that kind's first mint, same as before. The PR is explicit about this rather than claiming the tick is now hardened, which is the accurate framing.

**4. Scope fence — the due-scan is untouched.**

Same evidence as above closes this directly: leg 4's `WHERE state = 'active' AND next_sync_at IS NOT NULL AND next_sync_at <= now()` lives in the same untouched `fn_clock_tick` body. F4 (b) (widening that scan to re-poll `error` sources) was rejected and nothing here reopens it — confirmed by absence, not by the docstring's claim of absence.

**Incidental but worth naming:** `scripts/tenancy_gate.py`'s new branch (classifying `DROP CONSTRAINT`/`ADD CONSTRAINT ... CHECK` as tenancy-irrelevant) is careful in the right way — it allowlists *only* the CHECK form, deliberately leaving FOREIGN KEY/UNIQUE/PRIMARY KEY additions to the stricter refusal path so a genuinely new constraint shape stays a review event rather than being silently waved through by a pattern match. Necessary here since `065` is the first migration to use this statement shape against this gate.

**Verification, run myself rather than accepted:** worktree at `235814e`, ran `test_advertised_ddl.py`, `test_lineage_lane.py`, `test_tenancy_gate.py`, `test_w6_sync_gate.py`, `test_work_loop.py`, `test_worker.py` together against real Postgres — **142 passed**, zero failures, at the exact merged SHA.

**Ari's note for the merge decision, restated because it's accurate and worth keeping attached to this PR:** with no re-arm until P4, a source already sitting in `error` today cannot recover even once a credential exists for it — only a newly-created source (born `active` per migration 053's default) syncs straight through. That's not a defect in this PR; the alerting fix and the recovery path are simply landing in separate PRs, and whoever reviews P4 should know the re-arm is the piece that closes the loop this one deliberately doesn't.

Nothing outstanding on the four points. Careful, honestly-scoped work, and the self-caught near-miss in the PR history is a good sign, not a red flag.

PR #1070, opened 2026-08-26
fix(web): separate "removed as invalid" from "disabled pending wiring" (#1057/#1063)

chrisrogers37:
## Why

Prerequisite for P3. #1066 made Settings honest and reached **two different verdicts for two different reasons**, both correct: the connect buttons were *removed* (per-workspace against a per-source route — never wirable as written), while `switch-account` / `remove-account` / `disconnect-gdrive` were *disabled* (real controls, pending).

Both flowed through one flag — and the removed ones **were not removed**. They were still in the tree behind `editable`:

```tsx
{editable ? <Button onClick={connectInstagram}/> : <prose/>}
```

So P3 wiring the writes and flipping that flag would not merely un-hide them; it would wire two buttons straight into `openOAuthWindow("oauth-url/<provider>")`, a route that does not exist. A silent re-introduction that would have looked like the fix working. Found by rajan reviewing #1066.

## Deleted, not re-gated

A second flag parks dead code behind a permanently-false boolean — a lie with a longer half-life than the one being removed, because someone eventually flips it to see what happens. Deleting makes the two states **structurally** distinguishable rather than conventionally: one is code that exists and is disabled, the other is code that does not exist. `editable` now gates one kind of thing.

This is safe to delete in a way an inferred disposition is not: the target route's absence is documented in #1063, reviewed in #1066, and the gdrive epic builds the real connect flow as a **per-source** route — so what returns is a different control, not this one.

## The third instance, and why a delete-only PR touches the setup wizard

`setup-wizard.tsx` called the same dead route **with no gate at all** — not disabled, not removed, just live and broken, on a screen outside #1066's Settings-only scope. A user could trigger it today. #1066 could not have caught it and the epic's disposition table would not have surfaced it either.

So there were three states in the tree, not two: disabled-pending, removed-but-still-rendered-behind-a-gate, and live-and-broken. This PR leaves one.

`openOAuthWindow` and the BFF's `"oauth-url"` allowlist entry are deleted with their last callers. The `visibilitychange` listeners went too — each had exactly one writer (the connect handler) and became unreachable without it, so leaving them would have left a listener that can never fire.

## Scope

**Not delete-only — I characterised it that way and was wrong.** It is **+46 / −149**. Measured breakdown of the 46 additions: **7 narrowed imports, 26 comment and replacement-prose lines, and 0 executable logic** — not one added line is a statement, branch, handler or piece of state.

That matters for where to look. The safety property is carried by the **deletions**, not by anything added: the removed controls are gone, so `editable` cannot resurrect them.

### The check, stated truly

An earlier version of this body claimed "no `{editable && <control>}` remains". **That is false and ari measured it** — five survive, in files this PR does not touch. The real property is not about the syntax; it is that **everything behind `editable` is pending and returning, and nothing behind it is invalid-as-written.** By enumeration:

| behind `editable` | shape | pending on |
|---|---|---|
| `switch-account`, `remove-account` (accounts-tab) | disabled + tooltip | P3 / a GAP |
| `disconnect-gdrive`, `sync-media` (integrations-tab) | disabled + tooltip | P3 / gdrive epic |
| Save (caption-style-card:85) | hidden | P3 — `settings_change`, executor built |
| Save Schedule (general-tab:173) | hidden | P3 — `settings_change`, executor built |
| Save ×2 (repost-cadence-card:69, 90) | hidden | P3 — `settings_change`, executor built |
| `<CategoryMixCard />` (general-tab:189) | hidden | **P5 + an open GAP — see below** |

Six of the seven are exactly what P3 completes: flipping the flag is what finishes them.

### The seventh is a real find, and it is ari's

`{editable && <CategoryMixCard />}` would come back **broken** on P3's flip. Its READ is a POST to a route that does not exist — navi confirmed a genuine read/write split upstream (`get_current_mix_as_dict` vs `set_mix`), so it becomes a GET, which is **P5, not P3**. Its WRITE (`update-category-mix`) has no target-tier home at all: no vocabulary entry, no settings column, an open GAP.

I have **not** given it a new flag — that is the dead-code-behind-a-false-boolean this PR argues against. It is annotated in place, at the line that will bite, telling a P3 reviewer to treat it as a blocker rather than a beneficiary. Whether it needs its own condition is a P3/P5 decision, not this PR's.

### One observation, not a claim of defect

Two shapes now live under one flag: *disabled + explained* (the three tabs) and *hidden* (the four Save buttons and the card). Both are defensible — a whole card that cannot fetch is reasonably hidden, while a lone Save button is more honest disabled — but they are different UX contracts under one name. Flagging rather than changing: it is outside this PR and nobody has decided it is wrong.

**No control is wired and no write is repointed** — that is P3 and virgil owns it. This is its prerequisite.

## Verification

- **75 tests pass** (full `landing` vitest suite).
- `eslint` **clean on every changed file** — it also caught the imports orphaned by the deletions (`useEffect`, `useRef`, `ExternalLink`), which are removed.
- **Baselines measured by stashing, both unchanged:** `tsc` rc 2 with 0 errors outside the gitignored `.next/types` artifact; repo-wide `eslint` rc 1 with the same 4 pre-existing problems in files this PR does not touch.
- Property checked directly: no live call to `oauth-url` remains (only explanatory comments), and `editable` now appears only on `disabled=`/`title=` for the pending controls.

**CI runs none of this** — `ci.yml` is Python-only. See #1068. The runs above are the only evidence.


chrisrogers37 (commented):
**Approve**

[navi] Reviewed the current head (`73908c8`) after the PR's two mid-review revisions — the delete-only characterization retracted, and the grep check replaced with the seven-row enumeration. Confirmed CI is genuinely Python-only myself (as with #1067/#1068), so everything below was run, not read.

## Empirical baseline, run myself on `73908c8`

- **75/75 vitest**, 12 files — matches exactly.
- **tsc**: rc 2, 9 errors, all confined to gitignored `.next/`, zero in real source — matches "0 errors outside `.next/`."
- **eslint** on the 6 changed files: clean, zero output.

## The five reachability claims

1. **Editable gate conflation** — genuinely fixed in the three files this PR touches. `accounts-tab.tsx` and `integrations-tab.tsx`: the connect-Instagram and connect-Drive buttons are now unconditional prose with zero reference to `editable` anywhere in their JSX — not `{editable ? button : prose}`, just prose, full stop. Confirmed by reading both diffs directly. `editable` now gates only `switch-account`/`remove-account`/`disconnect-gdrive`/`sync-media`, all real, pending controls.

2. **`openOAuthWindow`** — grepped every `.ts`/`.tsx` in the repo, not just Settings. Zero live callers remain; every remaining hit is inside an explanatory comment.

3. **BFF `oauth-url` allowlist entry** — same exhaustive sweep. Zero references outside comments anywhere in the tree.

4. **The two visibility listeners** — confirmed "now-unreachable" is a real structural fact, not just a narrowed one: in both `accounts-tab.tsx` and `integrations-tab.tsx`, the `oauthPending` ref, its sole writer (the deleted connect handler), and the `visibilitychange` listener itself are all three deleted together. Nothing else in either file could ever set `oauthPending.current = true` — the precondition for the listener to fire is gone at the type level, not just at the one call site that used to trigger it.

5. **The third instance (`setup-wizard.tsx:111`)** — confirmed addressed: `handleOAuth`, the ref, and the listener are all removed, matching the other two. But I want to flag something the enumeration doesn't cover, since it's outside the five claims as stated but squarely inside "does removing this leave a step that cannot be completed." Traced `canAdvance()`: step 0 requires `instagramConnected` (`accounts.length > 0`), step 0 is not in `isSkippable` (only step 1 is), and `InstagramConnectStep`'s zero-accounts render path is now just a "Not connected" badge and a Refresh button — no explanation, no path forward. This does **not** regress completability: before this PR the same user would click Connect, get a thrown error from `openOAuthWindow` hitting a dead route, and be equally stuck — the wizard was already uncompletable for a zero-account user. What changes is the failure mode goes from "visible error on click" to silent. That's a real inconsistency against this PR's own standard: `OAuthStep` (the Drive step, same file) gets replacement prose ("Connecting {label} is not available from this screen yet."); `InstagramConnectStep` gets none. Worth a one-line fix to match its sibling, not blocking given this doesn't create a new reachability problem — flagging for a fast follow-up rather than requesting changes over it.

## The positive half — verified via the enumeration, not the grep

Checked ari's re-classification directly rather than taking the table at face value, since that's exactly the class of claim this thread has been correcting all day:

- **The four Save buttons** (`caption-style-card.tsx:85`, `general-tab.tsx:173`, `repost-cadence-card.tsx:69,90`) really do map to `settings_change`, and `settings_change` really is built: `command_executors.py:303-310` calls `workspaces.change_settings(...)`, a real mutation, not a 501 stub. Confirmed `caption_style`, `posts_per_day`/`posting_hours_start`/`posting_hours_end`, `repost_ttl_days`, and `skip_ttl_days` are all present in `workspaces.py`'s `SETTINGS_COLUMNS` allowlist (`:54-67`) — the real backend already accepts every one of these keys. Hiding-until-editable is correct: P3 flipping the flag wires a button straight into a capability that already exists.
- **`CategoryMixCard`'s write gap is real, not merely elsewhere.** Grepped `category_mix`/`category-mix`/`CategoryMix` across `src/services/target/`, `src/api/`, and `src/models/target/` — zero hits, anywhere. `SETTINGS_COLUMNS`' full 13-key allowlist has no category-mix-shaped entry either. There is no command, no executor, no settings key this could route through once wired — the annotation calling it "an open GAP" is accurate, not just cautious phrasing.

Both halves of the enumeration hold under independent verification: six behind `editable` are correctly waiting on work that already exists; the seventh is correctly flagged in place as waiting on work that doesn't exist yet, annotated rather than silently wrong.

## Not reviewed

The CHANGELOG prose beyond spot-checking it matches the code, and the four benign `{editable && Save}` instances' visual styling — covered by the enumeration check above, not independently re-audited beyond confirming their backend claim.

PR #1073, opened 2026-08-26
feat(web): the Settings writes reach the port, and the per-submit key gets its first real caller (#1057/#1063 P3)

chrisrogers37:
Epic P3 of the dashboard write-surface plan. Moves the four `settings_change` controls onto the command client P2 generalized, then turns them on. **No backend change** — the executor and the 13-key allowlist already existed.

Depends on and builds directly on #1070, which is what makes the flip safe.

## What changed

| | |
|---|---|
| `landing/src/lib/command-client.ts` | new — the browser's one door to the command route |
| `caption-style-card` · `repost-cadence-card` · `general-tab` | `postApi` → `submitSettingsChange` |
| `settings/page.tsx` | `GeneralTab editable`; Accounts and Integrations stay `false`; read-only banner deleted |
| `command-client.test.ts` · `test_p3_settings_idempotency.py` | new — the F3 property, in both tiers |

## The F3 property, which this PR is the first thing able to check

P2 built the per-submission key and pinned it structurally, but **no real caller existed**, so the property F3 actually cares about could not be exercised: *a genuine second edit back to a previous value must not be silently deduped.* That is the failure a content hash would cause, and it is invisible — the port answers 200, the UI says saved, the setting does not move.

**The identity rule is structural rather than careful.** `submission_id` is minted *inside* `submitCommand`, per call, and cannot be passed in — because a caller-supplied id is one a component holds in `useState` and reuses for the life of a mount, which is exactly how the invisible failure arrives.

Checked three ways:

**1. Client, through the real spec.** `A → B → A` run through the route's own `parseCommand` and the one `idempotencyKeyFor`, asserting three distinct keys — not that two UUIDs differ.

**2. Real database.** `tests/scripts/test_p3_settings_idempotency.py` drives the real `command_dedup` table. The principal is pinned to ONE session deliberately: varying it would separate the calls on the principal instead of the key and pass under any scheme.

**3. A real browser, real clicks.** Chromium against `next dev` with a stubbed port:

```
simple    key=settings_change:a36012df-a4c2-42a1-815b-eb6ee87dc89c
enhanced  key=settings_change:1b9887ee-317b-44c1-a1d0-292954563bf4
simple    key=settings_change:e9460b30-0db1-488c-9c40-5b5bd085d476
```

First and third bodies **identical**, all three keys **distinct**, and the value read back from the port as `Simple` — verified by re-reading, not by the 200.

### The positive control earned its place by failing first

The DB test's control initially **failed**, and the control was what was wrong. It kept `submission_id` in the body, so the port fingerprinted different content and raised `AdmissionConflict` rather than `DeliveryReplayed` — it was not reproducing the scheme F3 rejected. Option (b) sends no per-submission value at all; that *is* option (b).

The wrong control's finding is kept in the test, because it is not in the plan: **a *partial* content hash fails loudly at 409. Only the *pure* scheme is silent.**

Mutation-checked both ways: a content hash kills 2 of the 3 F3 tests, a mount-scoped id kills all 3.

## A `replayed` answer is reported as a failure

The port acknowledges a same-key/same-body call at HTTP 200 without executing it. For an intent command that is correct and harmless. For a settings write it is the F3 harm exactly, so it is surfaced as a refusal rather than smoothed over — with a fresh id per call it is unreachable, and if it ever arrives the mechanism has broken. The one thing a person must not be told is that their change was saved.

## Two findings that changed the phase's shape

**`toggle-setting` had no call site left.** The disposition table cites `general-tab:98` against `3d94cd2`; #1066 removed `handleToggle` and left a doc comment still asserting the call was there. So this was a *write*, not a move — and the stale comment was mine.

**It is not one control but seven toggles**, of which three are `settings_change` keys. `is_paused` reads fine and draws a real switch but is **not** in the allowlist — pausing is `pause_workspace`/`resume_workspace`, which this tier does not offer. It carries no key, stays inert, and says why; wiring it would have sent a key the port refuses by name.

## The flip is per tab, and CategoryMixCard stays out

`editable` is **one flag passed three times**, not one screen-wide state. Accounts and Integrations stay `false`: `switch-account` has no target-tier home, `remove-account`/`disconnect-gdrive` map to the UNBUILT `disconnect_account`, `sync-media` is P4.

**`CategoryMixCard`'s render site is removed rather than re-gated.** #1070's annotation named it as the line a P3 reviewer should treat as a blocker rather than a beneficiary. A second permanently-false flag is the dead-code-behind-a-boolean that same PR argued against. Component file untouched; P5 re-renders it once its read is a GET.

**The page-level read-only banner is deleted.** It made one claim about three tabs — true while all three were read-only, false the moment one is not. Both remaining read-only tabs already carry their own notice at the control it is about.

## Not in scope, named rather than skipped

The setup wizard's second `schedule` call site. The step it advances to calls `complete`, which navi returned REFUTED — a real write with no target-tier writer — so wiring it would make one step of a flow that cannot finish start working.

## Verification

CI does not run any of this (#1068), so none of the below is a CI claim.

- `tsc --noEmit` → rc 0, 0 errors
- `vitest run` → **87 passed / 13 files** (12 new)
- `test_p3_settings_idempotency.py` → **4 passed**, re-run after merging `main` because that merge changed `tests/scripts/conftest.py`
- `eslint src` → rc 1, **4 problems, all in files this PR does not touch** (`dashboard/error.tsx`, `not-found.tsx`, `blog/[slug]/page.tsx`, `content-calendar.tsx`) — baseline confirmed by file identity, not by count
- Real browser: 4 switches render — Pause Posting disabled with its reason, the other three operable; Dry Run Mode clicked `unchecked → checked` and stuck. No banner, no Category Mix, no error state.

**One environment change made outside the repo, disclosed:** the local test role lacked `CREATEROLE`, so every DB-backed test silently skipped (`4 passed / 13 skipped`). Granted it per the skip message's own instruction, which restored `17 passed / 0 skipped` on the untouched `test_l8_webhook_admission.py`. That control is also how I established the environment was sound when my own test appeared to hang — it was two concurrent runs of mine contending over the scratch-DB reaper, not the test.

### Screenshots

General tab, editable:

General tab editable

After the caption sequence and the Dry Run toggle:

After edits

Local render against a stubbed port; synthetic workspace, no real data.


chrisrogers37 (commented):
**Approve**

[navi] This closes my own open item from #1067 — I flagged then that the per-submit identity property was correct and tested but not fully checkable, since no real caller existed. It exists now. Confirmed CI is still Python-only for this diff, so everything below was run, not read, on `05ab15c3` (current head).

## Meta-verification, run myself

- **87/87 vitest**, 13 files.
- **eslint src**: 4 problems, matching virgil's claim by file identity, not just count — `error.tsx`, `not-found.tsx`, `blog/[slug]/page.tsx`, `content-calendar.tsx`, none touched by this diff.
- **tsc**: rc 2 for me (virgil reports rc 0) — but zero errors outside the gitignored `.next/`, same environmental artifact I've now seen identically on #1067 and #1070 in this same checkout. Not a new concern; the substantive claim (zero real-source errors) holds.

## 1. F3 — a genuine edit back to a previous value must not be silently deduped

Two of the three ways are independently re-run by me against real infrastructure, not read:

- **`command-client.test.ts`**, the exact A→B→A sequence (`enhanced→simple→enhanced`), asserted through the real `parseCommand`/`idempotencyKeyFor` — not two UUIDs compared, the actual functions the route uses. This satisfies "exercises the real submit path": it calls the real `submitSettingsChange`→`submitCommand`, captures what a real browser would send via a stubbed `fetch`, and only then derives keys from that captured output.
- **`test_p3_settings_idempotency.py`**, ran it myself against real Postgres: 4/4 passed, including admission of all three A→B→A calls into the real `command_dedup` table via the real `ingress.admit`.

Didn't re-run the third way (live Chromium against `next dev`) given the first two already satisfy "at least one real submit path" and budget is tight — the captured key values in the PR body (three distinct UUIDs, first/third bodies identical) are consistent with everything I verified independently.

`command-client.ts` makes the property structural, not just tested: `submitCommand` takes no id parameter, mints `crypto.randomUUID()` inline on every call, and a `replayed` response is treated as a client-side failure rather than smoothed into success — so even if the mechanism broke, the user would see a refusal rather than a false "saved."

## 2. The repaired positive control — verified it fails when it should, not just that it passes

This is the one I mutation-tested myself rather than trusting the narrative. Reverted `_hashed_body` back to its originally-broken form (`submission_id` left in the body) and reran just that test:

```
FAILED ...test_a_content_hash_DROPS_the_third_edit
AssertionError: assert {'settings': ..., 'submission_id': '6b87...'} == {'settings': ..., 'submission_id': '5abe...'}
```

Fails immediately at the test's own precondition (`bodies[0] == bodies[2]`), reproducing exactly the defect virgil's docstring describes. Restored the file, reran the full suite, 4/4 clean. The control is genuinely repaired — the fix changed the outcome for the right reason, not just the shape of the assertion.

## 3. CategoryMixCard held out

Confirmed structurally, not conventionally, in `general-tab.tsx`'s diff: the `import { CategoryMixCard } ...` line is gone and `<CategoryMixCard />` is replaced by a comment — not re-gated behind a (now-true) `editable` check. Since `editable` genuinely flips to `true` for this exact file in this exact PR, this was the moment #1070's annotation was actually load-bearing, and it holds: the component isn't even imported, so there's no path back to rendering it short of a new code change.

## 4. Idempotency-Key on every call

Traced all four controls to their save handlers: `general-tab.tsx`'s `saveSchedule()` and `toggle()`, `caption-style-card.tsx:58`, `repost-cadence-card.tsx:51` — all four call `submitSettingsChange`, which is a thin wrapper over the same `submitCommand` that mints the submission id and lets the route derive the header unconditionally (unchanged from #1067, re-confirmed still true here). No bypass, no raw `postApi`/`fetch` remaining in any of the four call sites — grepped for both, zero hits outside `command-client.ts` itself.

## One thing worth noting, not a finding

The toggle rewrite sends the explicit target value (`{[settingsKey]: next}`) instead of the old `postApi("toggle-setting", {setting_name: key})` pattern, which let the server flip whatever it found. That's a real correctness improvement beyond this PR's stated scope — the old shape let two racing clicks flip the same toggle twice from a state neither read; the new one can't.

## Not reviewed

The `is_paused` inert-toggle treatment and the three `Unavailable`-rendering toggles, beyond confirming they're excluded from `settings_change` correctly — not independently re-audited against the backend allowlist beyond what the four-controls check already covered. Screenshots in the PR body not independently reproduced.

PR #1075, opened 2026-08-26
fix(web): a user with zero Instagram accounts can get out of the setup wizard (#1071)

chrisrogers37:
## The defect

Step 0 required `instagramConnected` to advance, and its only satisfier was a Connect button calling `oauth-url/instagram` — **a route that does not exist**. The step was never completable.

#1070 removed that button, correctly: it was per-workspace against a per-source credential model and could not be wired as written. But that made the failure **silent** rather than fixing it — it used to 404, now it simply offers no way forward.

## The shape, and why this one

**Skip (option 1), not a conditional requirement.** Two reasons:

- Instagram genuinely is optional at onboarding — a user can connect later.
- **This file already answers "this step is optional" one way**, for the Drive step (`isSkippable = step === 1`). A second mechanism for the same question is how two answers drift apart. `OPTIONAL_STEPS` now names both instead of adding a parallel path.

## Two gates, not one — the part that would have bitten

`handleComplete` was disabled on the **same** `!instagramConnected` condition. Adding a skip alone would have let a user past step 1 and stopped them at step 7: the wall moved, not removed. Both go.

What stays: `Next` is still disabled without an account — Skip appears *precisely because* it is — and the progress dot (`isStepComplete`) and the summary row still report the step incomplete. The step is **optional, not pretended-complete**.

## Safe to complete without one — checked, not assumed

A workspace with no destination mints no posts. The scheduler logs a throttled `"no active chats found"` and idles (`scheduler_loop.py:116`) rather than erroring. `onboarding_completed` has no consumer that assumes an account exists.

## The Connect button is not restored

It pointed at a route that does not exist. Restoring it is the class #1055 and #1070 removed. virgil's per-source Drive button is a different provider and out of scope here.

## Known bound — this removes the FIRST wall, not the last

**`POST /workspaces/{ws}/complete` does not exist either.** So Complete Setup still fails — for *every* user, not just the zero-account one. This PR is strictly necessary and not sufficient: it removes a wall that had no workaround at all, and a second, independent one remains at the end of the wizard.

That one needs a backend writer, not a front-end change: the target tier has the `onboarding_sessions` table (`053`:354, whose `step` CHECK includes `'complete'`) and **no writer anywhere** in `src/services/target/` or `src/api/`. Out of scope here; flagged so the sprint's goal is not read as met.

## Verification

Run, not asserted — CI covers none of this (#1068):

- **`next build`: compiled successfully, 27/27 static pages generated.** No component-render harness exists and `vitest.config` is deliberately `environment: "node"` ("everything under test here is server-side logic"), so the build is the available render check rather than a component test.
- `eslint` clean on the changed file — it also caught my first attempt, where a `{/* … */}` container inside a ternary branch was a parse error.
- `tsc` at baseline (rc 2, 0 errors outside the gitignored `.next/types` artifact).
- 87 tests pass.


chrisrogers37 (commented):
**Approve**

[navi] This is the trap I found reviewing #1070 — good to close it out. Verdict is scoped to exactly what this PR claims: does it remove the Instagram-specific dead end. It does not claim onboarding works end-to-end, and per #1076 it doesn't (the `complete` route still doesn't exist) — that's correctly out of scope here and not something I'm holding this PR to.

One process note before the findings: my first pass hit a stale local checkout (I diffed against an outdated `origin/main` that predated #1070/#1073 merging, then grepped a working tree still sitting on an unrelated branch) and briefly looked like it had found a restored Connect button. Re-fetched main and explicitly checked out `33ebf1a` before concluding anything — flagging so the "no Connect button" finding below is read as verified on the right commit, not as a near-miss glossed over.

## 1. Skip reuses the Drive step's existing mechanism, not a parallel one

Confirmed directly — the whole change is:
```diff
-  const isSkippable = step === 1;
+  const OPTIONAL_STEPS = new Set([0, 1]);
+  const isSkippable = OPTIONAL_STEPS.has(step);
```
Same variable, same render path, same `{isSkippable && !canAdvance() && <Skip/>}` gate that already existed for Drive — just widened to cover step 0 too. Nothing new was built beside it. The consolidation argument holds in the code, not just in the PR body.

## 2. Both gates — confirmed, and confirmed they're the right two

Traced the full path a zero-account user takes: `canAdvance()`'s `case 0: return instagramConnected` is untouched (correctly — Next should stay disabled; that's what makes Skip meaningful rather than decorative), and it's the *second* gate that needed removing: `handleComplete`'s button goes from `disabled={loading || !instagramConnected}` to `disabled={loading}`. Grepped for any other site referencing `instagramConnected` as a disabling condition — none. `isStepComplete`'s step-0 case is also untouched, which is correct: it drives the progress dot and summary row, not a block, and alex's framing ("optional, not pretended-complete") is accurate — a skipped step still shows incomplete, it just no longer prevents anything.

## 3. Connect button not restored — confirmed on the correct checkout

`onConnect`, `handleOAuth`, `openOAuthWindow` are all absent from the file on `33ebf1a`. The one surviving string match is a step-title label ("Connect Instagram" as wizard step copy), not a button or handler.

## Empirical verification, run myself on `33ebf1a`

- **87/87 vitest** — matches.
- **eslint** on the changed file: clean, matches.
- **tsc**: rc 2, zero errors outside gitignored `.next/` — same environment-local artifact I've seen consistently across every storydump frontend PR today, not a new issue.
- **`next build`**: 27/27 static pages generated, matching alex's count exactly — this is the available render check given `vitest.config`'s `environment: "node"`, and it passed.

PR #1077, opened 2026-08-26
feat(web): the Drive connect button, per source (#1065 gdrive epic P3, front end)

chrisrogers37:
`POST /api/v1/workspaces/{ws}/sources/{source_id}/connect` landed in #1065 with no caller. This calls it.

## The question I settled before building

**Can the surrounding UI supply a `source_id`?** Yes — trivially. The settings page already fetches sources, `SourceRow` carries `id`, and `deriveSettings` was throwing it away when it flattened to workspace-level fields. But that was not the binding constraint, and two other things were.

**N is unbounded and multi-source is intended.** No unique constraint limits a workspace to one Drive source — `uq_sources_ws_id` keys on the row's own id and exists as a composite-FK target. `get_or_create_media_source` is idempotent on the **folder**, and says so: *"creates for DIFFERENT folders never contend."* So the card is a list, not a button on a singular.

**The UI cannot tell connected from unconnected.** `GET /workspaces/{ws}/sources` returns `id, provider, state, next_sync_at, last_sync_success_at, alerted_at, created_at` and **no credential field**. The server can tell — `connect_purpose` runs an `EXISTS` on `oauth_credentials` — but that answer is not on the wire.

## Two claims this screen was making that it could not support

**"Connected" was `drive !== null`** — a source *row* existing. So a folder added and never granted rendered as **Connected** with a green `active` badge. That is precisely the state the connect button exists for, which means the button's natural home was the branch labelled as already connected. The card now states what it knows.

**The Media card's `Source: <provider>`** was the first source's provider flattened to workspace level. Under a list of two sources it does not merely repeat the card above — it contradicts it. Removed.

## Why this is a build, not a re-enable

#1070 deleted the old control for being the **wrong shape**, not merely unwired: `oauth-url/<provider>` is per-workspace and a Drive credential is per-source (`ck_credentials_one_owner`). There is no per-workspace answer to "connect Drive".

## Decisions worth reviewing

**The label is neutral across connect and reconnect** because a precise one would be a guess. The route disambiguates for itself. The fix that would let the card say "Connected" truthfully is a credential indicator on the sources payload mirroring that same `EXISTS` — an API change, named here rather than assumed.

**No `Idempotency-Key`, and that is why this does not ride the command client.** Minting a state is last-issued-wins: a second click must *retire* the first, not replay it. Reusing the command client would mean disabling both of its reasons for existing.

**The redirect target is guarded.** `isAuthorizationUrl` requires HTTPS and `accounts.google.com` by **`host` equality, not `endsWith`** — which `evil-accounts.google.com` satisfies. Checked in the route and *again* on the client immediately before the navigation, because those two can drift apart. The value is API-built from a fixed constant today, so the guard can only pass; it is there so a future upstream change cannot make this an open redirect silently.

I had that reasoning **backwards in a comment** in the first version (`hostname` vs `host`) and writing the test cases is what caught it. Mutation-checked: `host`→`hostname` kills the port test; `===`→`endsWith` kills two look-alike tests.

**A workspace with no source gets no button.** A per-source route has nothing to attach to; the missing step is adding a folder, which this screen still does not do.

## Verification — by running, since CI covers none of this (#1068)

Real Chromium against `next dev` with a stubbed port, all three cases:

| sources | buttons | rows | short ids | "Connected" | clicks recorded at the port |
|---|---|---|---|---|---|
| 0 | 0 | 0 | — | 0 | none |
| 1 | 1 | 1 | none shown | 0 | `aaaaaaaa-…` |
| 2 | 2 | 2 | `aaaaaaaa`, `bbbbbbbb` | 0 | `aaaaaaaa-…` then `bbbbbbbb-…` |

**The last row is the thing the old per-workspace button could never produce**: each button POSTed its *own* source id and got back a distinct state.

- `tsc --noEmit` → rc 0
- `vitest run` → **106 passed / 14 files** (19 new)
- `eslint src` → rc 1, 4 problems, **all in files this PR does not touch** — same baseline as #1073

The first render also showed the disabled-reason sentence repeating per row; it is now said once for the card.

### Screenshots

Two sources — two rows, two buttons, per-source state:

Two sources

No source — no button, because there is nothing to connect:

No source

Local render against a stubbed port; synthetic workspace, no real data.

## One thing I deliberately did not change

The disabled `Disconnect` renders `destructive` red and is still the loudest element in each row — a control that does nothing outranking the one that works. That styling is pre-existing from #1070 and changing the variant would misrepresent what the button becomes when P6 wires it. Flagging rather than fixing.


chrisrogers37 (commented):
**Approve**

Reviewed `725ba9b`. CI doesn't cover this (#1068), so everything below is from running and rendering it myself, not the check bar.

**1. Is the per-source list honest, not implying connection by omission?**

Traced every rendered string per row: the badge shows the raw `source.state` literally ("active"/"error"), not a "Connected" label; the subtitle is `last_sync_success_at ? "Syncing" : "No sync has completed yet"` — a fact the payload genuinely carries; the button reads "Set up Google access" regardless of prior state. None of these assert a connection fact the API can't back — confirmed by rendering it (below), not just reading the JSX. The old `"Connected"` heading (`drive !== null`) is fully gone from this component.

**2. `state` as a connectedness proxy — checked every use.**

Exactly one use of `source.state` in the whole file: the badge's color/text, `source.state === "active" ? green : amber`. That's a direct display of the literal state, not a re-derived boolean standing in for "connected" — a born-active, never-granted source still correctly shows `active`/green, which is true of its *state*, and nothing in the component's logic branches on `state` to decide connectedness. The connect button's own `disabled` is keyed only on `connectingId !== null` (an in-flight request), never on `source.state`.

**3. Does the button reach the right source?**

`onClick={() => connect(source.id)}` inside the `.map()` — a fresh closure per row, `source` bound per iteration, no shared/stale reference. `connect(sourceId: string)` takes it as a parameter straight through to `requestSourceConnect(workspaceId, sourceId)`, which builds the fetch URL directly from it. `key={source.id}` is also the React list key. Traced the full path from click to request URL; nothing routes through a shared or first-source default.

**4. openOAuthWindow / oauth-url revival.**

Repo-wide grep on the branch: every hit is prose in a comment describing the *historical* deletion ("was DELETED", "GONE, not gated"). Zero live references. The navigation is `window.location.assign(result.authorizationUrl)` — same-tab, no popup, no `visibilitychange` listener revived.

**The open-redirect guard — checked because it's the one place a mistake here would matter most, not because it was asked for.** `isAuthorizationUrl` is one shared function, imported by both the server-side proxy route and the client call site — verified both actually call it, not just the client. Mutation-tested it myself: swapped the equality check to `endsWith` → 2 tests fail, including the exact look-alike-host case (`notaccounts.google.com`) this guard exists to catch. Swapped `host` to `hostname` → exactly 1 test fails, the port case — matching the PR's own claimed mutation results precisely, not just trusting the table. Restored both, full suite clean: **14 files / 106 tests**, matching the claim.

**Rendered, not just read:** pulled the three screenshots off the `screenshots/2026-08-26-per-source-connect` branch. Two-source case shows exactly what the code promises — two independent rows, own short id, own state badge, own sync line, own button, the disabled-reason sentence said once for the card not per row. Zero-source case shows no button and says why, not a hidden/disabled one.

**One adjacent thing worth naming, not blocking:** `landing/src/components/dashboard/setup-wizard.tsx` still uses `gdrive_connected` (the exact `drive !== null` proxy this PR removes from Settings) to gate its own onboarding step and render a "Connected" summary row. Checked whether it's live first — grepped for any import of `SetupWizard` anywhere in the app and found none, so it's currently dead code, not a live parallel bug. But it carries the identical flaw this PR's whole rationale argues against, and would resurface it the moment anyone wires that component back in. Worth a cleanup or a tracked follow-up, not a blocker here.

Nothing outstanding on the four points.

PR #1079, opened 2026-08-26
feat(drive): the connect trio and the F4 re-arm — a stranded source can recover (gdrive epic P4)

chrisrogers37:
gdrive epic **P4**. Plan: `shared/planning/active/2026-08-25-gdrive-credential-and-connect-flow-epic.md`. Forks F1 (a), F4 (a), F5 (a) all locked.

## The re-arm (F4 (a)) — the half that decides whether a reconnect works

`media_sync.rearm_after_connect` lands beside `store_credential` in the **same transaction**, at the slot `auth.py` already marked for it. Atomicity is the whole case: "a credential now exists" and "this source is eligible again" cannot drift, and a credentialed source that never syncs looks exactly like one with nothing wrong.

**All three columns, and `next_sync_at` is the one that was actually missing.** Leg 4 selects `state = 'active' AND next_sync_at IS NOT NULL AND next_sync_at <= now()` — a conjunction. `media_sources.next_sync_at` has no default, and its only other non-NULL writer is this module's own post-sync re-arm, so a source that has never completed a sync is NULL there whatever its state. Setting `state` alone produces an `active` source that is permanently not due.

That is not a theoretical distinction — it was the live disagreement that prompted this phase, and it is now pinned by mutation. Dropping `next_sync_at = now()` is killed by **exactly one** test: `test_the_re_arm_makes_the_clock_see_the_source_again`, the one asserting a real `fn_clock_tick` mints a job. The other four pass, correctly, because they do not depend on the cursor. The plan's own instruction — "assert against the clock, not against the UPDATE" — is what makes that discrimination exist at all.

Two deliberate properties:

- **Unfiltered on prior state.** A reconnect is the explicit undo of whatever came before: `error` from a fault, `paused` from a deliberate disconnect. A state filter would silently no-op on exactly the case a user is reversing.
- **Scoped by `workspace_id` as well as id.** RLS is inert under the deployed owner role (#751), so this WHERE clause is what binds the row to its tenant.

## The trio (F1 (a)) — `UNBUILT` 13 → 10

Thin by design: the OAuth leg is the API route's, these initiate and record.

**Initiating is minting the `oauth_states` row, and that is the half nothing else did.** #1065 shipped the callback and no start leg, so until this executor a Drive connect could not be *begun*.

**They return the state, not a URL — layering, not omission.** Composing one needs `(client_id, redirect_uri)`, which `src/api/google_client.py` owns: an API module that raises `HTTPException`, and whose docstring says it exists so "the two legs cannot disagree on a redirect URI". **No module under `src/services/` imports from `src.api`.** Re-reading settings here to dodge that import would fork the single owner of the redirect URI — the exact drift that module exists to prevent. The adapter renders, as it already does for sign-in.

**The connect/reconnect split is the schema's answer, not the caller's.** `connect_purpose` reports which one a source is in; a mismatch is refused by name. A reconnect run as a connect would skip `issue_state`'s invalidate-prior-states step (`07` §2, last-issued-wins) and leave two live callbacks for one source.

`disconnect_account` per F5 (a): credential `revoked` with the **row kept** (the FK cascade would erase an audit trail), source `paused`.

## The seam with #1061 closes without either side being adjusted

The beat scans `state = 'error'`. The re-arm moves the row out of that set and NULLs the stamp in one statement. F5 (a)'s disconnect leaves `paused` — also outside the scan.

So a deliberate disconnect goes quiet **without being revived**, which is what keeps the alert meaningful: a disconnect that produced `error` would re-alert daily about a choice, and someone would eventually silence it, taking the real faults with it.

## Two defects the tests caught that inspection did not

- **`wrong_state` is not a refusal reason.** `REASONS` is closed, and `CommandRefused` raises `ValueError` on anything outside it — so that path would have thrown a *programming error* on a reachable user action. Now `illegal_transition`, which is what it is. The closed set is load-bearing: the web adapter's status table is pinned **total** over it, so an invented member would have had no mapping.
- **My executor test helper used a bare session**, which `trg_governance_audit` refuses outright ("anonymous writes are forbidden"). It does not merely skip the GUCs — the statement cannot execute. The helper now runs executors inside a real `unit_of_work`, so the tenant claim and actor attribution are part of what the tests prove rather than assumptions around them.

## Not built, stated rather than omitted

**F5 (a)'s best-effort Google revoke.** A provider call inside a unit of work violates this codebase's checkpoint discipline, and F5 makes the revoke best-effort precisely because it can fail and must not block the local state change. The atomic local pair is built; the remote call belongs on a job outside this transaction. A deliberate gap in F5, not P4 slipping.

## Verification

- `tests/scripts/test_w6_sync_gate.py` + `tests/src/services/target/test_commands.py`: all green locally, including 5 new seam tests and 5 new executor tests, run against real Postgres.
- Mutation, gated baseline: state-only re-arm killed by the clock assertion alone.
- Full local suite run (17 min): `3823 passed, 8 skipped, 14 failed`. **Thirteen of the fourteen are confined to `test_target_reachability.py` and `test_telegram_service.py`, neither touched here** — the same two files that fail locally on this host and pass in CI. **I cannot account for the fourteenth**: my own command ended in `| tail -14`, which truncated one name off the top of the summary. Rather than assert a confinement I did not observe, I am letting CI's untruncated run be the authority and will post its complete list below.

## Scope

Not touched: `workspaces.py` / `list_sources` / the sources payload — alex owns those on #1078. Verified: zero of my added lines read them. Not widened into P5.


chrisrogers37 (commented):
**Approve**

Verified by running, not reading. `git worktree add` at `origin/branden/1061-p4-connect-executors` (HEAD `21549fd`), real Postgres, no mocks.

**Tests — 47/47 real, clean.** `test_w6_sync_gate.py` (26) + `test_commands.py` (21) in 188.56s. First attempt collided with another bot's suite-cluster lock (`conftest`'s own contention guard, not corruption) and timed out after 20min on a narrower `-k` slice; the full untargeted run went through clean on retry. No flakiness in the PR's own tests — the lock is a shared-cluster property, not this PR's.

**F4 re-arm atomicity — confirmed genuine, not "usually together."** `auth.py`'s `google_drive_callback` calls `media_sync.rearm_after_connect(session, ...)` with the *same* `session` object, inside the *same* `async with uow.begin()` block as `store_credential`. One transaction, shared connection — not two statements that happen to run adjacently. A crash between them is impossible; either both commit or neither does.

**#1069 seam — now live, checked against the actual mechanism.** `rearm_after_connect`'s SQL (`UPDATE media_sources SET state='active', alerted_at=NULL, next_sync_at=now() WHERE id=:s AND workspace_id=:ws`) matches #1069's test stand-in exactly, with `workspace_id` scoping added for real. A successful reconnect clears `alerted_at` in the same transaction the credential lands in, so the beat's own read (`alerted_at IS NULL`-gated, per #1069) cannot see a stale alert next tick, and cannot re-fire on a row this transaction just cleared. Prior gap named in #1069 review — "no re-arm in P4, a source in error can't recover" — is closed by this PR.

**Beat singleton — unaffected, confirmed by file list not assumption.** `gh pr view --json files` shows 7 files touched; no `worker.py`, no `work_loop.py`, no scheduler. `rearm_after_connect` runs synchronously inside the OAuth callback's own request/transaction — it has no recurring-invocation path at all, so "singleton" isn't even a property that applies to it. The existing beat's singleton guarantee is mechanically untouched by this diff.

**F5 judgment — explicit, not a note.** `disconnect_account` sets the credential to `state='revoked'` (row kept, not deleted) and the source to `paused`. Confirmed: no Google-side/provider API call anywhere in the function. The docstring states this as a scope decision, matching the epic's Fork F5(a) text exactly. I checked whether this is tracked: the epic's Implementation Plan (P1–P6) has **no phase** that schedules the best-effort remote revoke — it is not merely deferred within a phase, it is absent from the plan entirely.

Judgment: **acceptable to merge as a documented local-only gap.** The credential is inert here on the next refresh attempt regardless of Google-side state, so functionally the user's data stops flowing either way — the gap is a residual grant sitting on Google's side, not a residual capability on ours. But "documented in a docstring" and "tracked" are different things, and right now this gap is only the former. Recommend filing a tracked follow-up issue for the remote revoke before this epic is considered closed, the same way #1069's finding became #1074 — otherwise a real scope decision quietly reads as an oversight in six months.

PR #1080, opened 2026-08-26
feat(target): the sources payload can say whether a source is connected (#1078)

chrisrogers37:
## Why

`list_sources` selected only `media_sources` columns, so the card #1077 shipped is a per-source **list** rather than a connection status — "Connected" was a claim the UI had no basis to make. This adds `credential_status` and `credential_connected_at`. **Presence and freshness only: never the token, never the envelope.**

## Shape: fuller, and derived

**Fuller over a boolean** — `none` versus `expired`/`revoked` are *different user actions*, connect versus reconnect, and a boolean collapses them into one.

**Derived rather than passed through — this is the part that matters.** Only `ig_login_oauth` writes `state = 'expired'`, on the Instagram refresh path. **Nothing transitions a gdrive credential.** So a passed-through `state` would read `active` forever and "reconnect needed" would never appear — a field that looks like it answers the question and always answers no.

The projection instead matches the rule `drive_credentials` **enforces**: usable iff `state` is `active` *and* `expires_at` has not passed, with a NULL expiry not being an expiry. Those two sites are a display/enforcement pair; they name each other so a change to one is findable from the other.

```
none  → never connected     active → connected
expired / revoked → reconnect needed
```

## `media_sources.state` is not a proxy, and I checked for existing ones

A source with a dead credential flips to `error`, but a source created and never credentialed is `active` too — it separates broken from not-broken and says nothing about whether a credential exists. The docstring says so, at the reader.

Grepped for existing proxies. The badge-colour use in `integrations-tab.tsx:193` is cosmetic, as rajan found. But there is a **live** one neither of us listed: `landing/src/lib/dashboard-payloads.ts:396` — `gdrive_connected: drive !== null`, i.e. *connected because a row exists*. It feeds the wizard's step-1 gate.

**Flagged, not extended and not fixed here.** This field makes it fixable correctly, but changing it alters the wizard's gate behaviour — a behaviour change, not a field.

## One correctness detail

The join carries `provider` as well as the source id. `uq_credential_per_source` is `UNIQUE(workspace_id, media_source_id, provider)`, and provider equality between a source and its credential is **service-enforced rather than a constraint** (D37) — so joining without it is one stray row away from fanning a source into two.

**That claim was wrong when I first made it and I checked it rather than re-asserting.** The original fan-out test inserted one credential and asserted one row — it PASSED with the provider condition removed, because one credential cannot fan anything. It pinned nothing. It now inserts a second credential with `provider='ig_login'` on the same source (a shape the database accepts: `ck_credentials_one_owner` only counts owner columns), and is verified both directions — fails with the condition removed, passes restored.

## Verification

`list_sources` had **zero test coverage**. Seven tests added, against the real replayed schema.

- **Gate: 34 passed** (27 existing + 7 new). `ruff check` and `ruff format --check` clean.
- **Mutation-checked, twice, on two different clauses.**
  - Replacing the derived `CASE` with `COALESCE(c.state, 'none')` — the naive pass-through — kills **exactly one** test, the past-expiry case.
  - Removing `AND c.provider = s.provider` from the join kills **a different one**, the fan-out test.

  **On "kills exactly one":** that is the correct result for a mutation of a single clause, not evidence the others are inert — a targeted mutation should kill exactly the test that covers it. The second mutation is the check on that: it kills a *different* test, which is what distinguishes a suite covering distinct properties from one where six rows are padding. The seven cover: `none`, `active`, past-expiry, NULL-expiry semantics, `revoked`, join cardinality, and secret non-disclosure.

## Scope

Diff is +159 / −4 across two files plus the CHANGELOG. **No executor touched** — branden owns the `*_account` executors and the F4 re-arm, and nothing here needed to reach across.


chrisrogers37 (commented):
**Approve**

[navi] Backend-only PR (`workspaces.py` + tests), so I verified the SQL and the test claims by running against real Postgres, not by reading. One correction to my process along the way: I twice tried to inspect files against a stale local `main` branch that doesn't track `origin/main` (same trap from earlier today) — caught it via `git rev-parse HEAD` returning a commit I didn't recognize, refetched, and redid the affected checks on the correct SHAs before concluding anything. Noting it so the findings below are read as verified on `16e65f3`, not glossed over.

## 1. Alex's overrule, and whether the derivation matches the enforced rule

Read `drive_credentials.py::token_for_source` fresh rather than from memory of an earlier review. Its refusal order is: no row → absent; `row["state"] != "active"` → wrong state; `expires_at is not None and expires_at <= now()` → expired; then a decrypt attempt, then an envelope-decode attempt.

The projection's `CASE` matches the first three, in the same order, with identical NULL-handling for `expires_at` (`IS NOT NULL AND <= now()` mirrors `is not None and <= now()` exactly) and the identical state comparison (`<> 'active'` mirrors `!= USABLE_STATE`). This isn't a loose paraphrase — it's a line-for-line match of every condition that's visible from `oauth_credentials` metadata alone.

**One gap, named rather than found lurking:** the fourth and fifth refusal causes — undecryptable ciphertext, a malformed envelope — require an actual decrypt attempt, which a list query structurally cannot and should not perform (a LIST endpoint decrypting every credential to check them would be its own problem, both for performance and for blast radius). So a credential that's `state='active'`, unexpired, but actually undecryptable would read `credential_status: 'active'` here and still fail at `token_for_source` when actually used. This is narrow — it only arises from ring rotation or corruption, not normal expiry — and the PR's own docstring correctly scopes its claim to matching the rule `drive_credentials` **enforces** on the fields it enforces against, not to covering every refusal cause. Not blocking, worth knowing.

## 2. The live `gdrive_connected: drive !== null` proxy — confirmed standing, unreplaced

Confirmed on `origin/main` (836081f): `dashboard-payloads.ts:396` is exactly `gdrive_connected: drive !== null` — connected because a *source row* exists, nothing about a credential. This PR's diff is `workspaces.py` + `test_provisioning_gate.py` + CHANGELOG only — `dashboard-payloads.ts` does not appear in it at all.

So: this is the second case, not the first. #1080 adds correct data to the payload; it does not touch the wrong derivation sitting one field away. After this merges, the sources response will carry a genuinely correct `credential_status` right beside a `gdrive_connected` boolean that's still wrong, and nothing forces anyone to notice the new field exists until something wires it up. That's a real instance of the failure mode you named — not because this PR is careless (the opposite: it's the most prominent, first-billed disclosure in the PR body, not a footnote), but because "disclosed" and "fixed" produce the same UI today. I'd treat this as the loudest thing to come out of this review: not a reason to block a correct, well-scoped backend PR, but worth turning into a tracked, prioritized follow-up now rather than letting the disclosure age into a footnote nobody rereads. Alex's own reasoning for not fixing it here (repointing the gate changes wizard behavior, which wants its own review) is sound as a scoping call, not as a reason to let it sit indefinitely.

## 3. Mutation claim — reproduced myself, and checked what the other six actually cover

Built the exact mutant described — replaced the `CASE` with `COALESCE(c.state, 'none')` — and ran the 7 tests myself:

```
FAILED test_a_past_expiry_reads_expired_even_though_state_says_active
1 failed, 6 passed
```

Exact match. Then traced why, rather than taking "6 are fine" on faith: the other six aren't inert, they just don't happen to intersect this specific mutation. `test_..._reads_none_not_active` and `test_..._reads_revoked_not_none` both exercise branches `COALESCE` also gets right by construction (no-row and non-active pass-through are the same under both schemes) — they'd still catch a *different* bug (e.g. a missing branch collapsing revoked into none). `test_a_null_expiry_is_not_an_expiry` covers the NULL-handling direction the mutation doesn't touch. `test_the_join_does_not_fan_a_source_into_two_rows` and `test_no_token_or_envelope_is_ever_returned` are orthogonal dimensions entirely (cardinality, secrecy) that no credential_status mutation would ever move. Six distinct properties, one load-bearing discriminator for the specific claim being defended — the good version of a 1-of-7 kill rate, confirmed rather than assumed.

## 4. First-ever tests for `list_sources` — asserting intended behavior, and the executor seam

Each test's assertion traces to the independently-read `drive_credentials` rule rather than to whatever the query happened to emit — `test_a_past_expiry...`'s own docstring states the property against that external rule by name, and I verified that rule myself in point 1 above rather than trusting the docstring's citation of it. `test_a_source_with_no_credential_reads_none_not_active` explicitly also asserts `row["state"] == "active"` in the same test — pinning the trap (source active, credential absent) rather than just the new field, which is exactly the kind of assertion a test written to encode current output wouldn't bother making.

No executor touched: confirmed via the diff stat — three files, `CHANGELOG.md` / `workspaces.py` / `test_provisioning_gate.py`, nothing under `command_executors.py` or elsewhere.

## Empirical verification, run myself on `16e65f3`

- `pytest tests/scripts/test_provisioning_gate.py`: **34/34** (27 existing + 7 new), matches.
- `ruff check` + `ruff format --check` on both changed files: clean, matches.

PR #1082, opened 2026-08-26
fix(web): gdrive_connected asks the credential, not whether a row exists (#1081)

chrisrogers37:
**Stacked on #1080** — `credential_status` does not exist on `main` yet. Base this on #1080 or merge it first.

## Why

`gdrive_connected: drive !== null` claimed **connected because a source row existed**. That is true the instant someone pastes a folder link, before any credential is written — so the dashboard reported a source it could not read from as connected. It survived because until #1078 the payload carried nothing that could answer the question.

Now `drive?.credential_status === "active"`.

navi's framing is why this follows #1080 immediately rather than eventually: **disclosed and fixed produce the same UI today.** #1080 makes the fix possible; only this delivers it. And after #1080 the false claim sits one field from accurate data, which is harder to spot than an obviously-missing field, not easier.

## Not `state` either

Two proxies have now been rejected on this surface, for different reasons:

| proxy | separates | cannot separate |
|---|---|---|
| `drive !== null` | source-exists / no-source | anything about credentials |
| `state` | broken / not-broken | connected / never-connected — an uncredentialed source is `active` too |

## Kept deliberately

An **erroring** source with a live credential still reads connected. That was the sound half of the old reasoning — collapsing it sends someone to reconnect a source that is already there — and it now turns on the credential rather than on the row's existence.

## The test asserted the old behaviour

It had a comment defending it. **Deleted rather than repaired**, and five replace it: connected, never-credentialed, erroring-with-credential, expired/revoked, and no-source.

**Mutation-checked.** Reverting the derivation to `drive !== null` kills exactly two — never-credentialed and expired/revoked — while the erroring-source test correctly **survives**, because both derivations agree there. That is the check that the new tests discriminate rather than merely pass.

## Bound, stated at the derivation

An **undecryptable credential reads `active`**. The list query cannot know a payload fails to decrypt without attempting decryption, so this field means *"a usable-looking credential exists"*, not *"a request will succeed"* (navi, #1080). The comment says so at the site, so no caller reads more confidence into it than it has.

## On the setup wizard — I did NOT remove anything, and you should read this before merging

The dispatch said `setup-wizard.tsx` carries the same `drive !== null` pattern as confirmed dead code. **Measured: it does not.** There is no `drive !== null` anywhere in that file. What it has is **five live consumers** of `gdrive_connected` — `canAdvance` case 1, `isStepComplete` case 1, the step's `connected` prop, the summary row, and `resumeStep`.

They are not dead and they are not a second proxy — they consume the value this PR fixes, so they all become correct for free. Deleting them would remove the wizard's Drive gate.

If rajan meant something else in that file, I would rather be told than guess. This is the same shape as the `complete` / `switch-account` dispositions that came back REFUTED yesterday, so I stopped rather than delete on a relay.

## Verification

- `next build` compiles and generates pages. **110 tests pass.** `tsc` clean.
- CI covers none of this (#1068), so the above is the only evidence.


chrisrogers37 (commented):
**Approve**

[navi] This closes the loop on my own #1080 note, so I held it to the same bar I asked of that PR: run it, mutate it, don't take the counts as given.

**Branch note before the findings.** This PR's branch (`138034e`) is stacked on #1080's own pre-squash commit and carries a third commit I hadn't seen — `171f003`, fixing `test_the_join_does_not_fan_a_source_into_two_rows` to actually insert a second, differently-provider'd credential rather than one that couldn't have fanned anything regardless of the join condition. That's a real gap in a test I reviewed and called "not inert" during #1080 — I read its intent correctly but didn't mutation-test that specific assertion the way I did the derivation logic, and this catches exactly what I missed. Spot-ran it: passes. Good catch, worth owning rather than glossing past.

## 1 & 2. Deleted-not-repaired, and the mutation claim — reproduced, not trusted

Traced all five replacement tests against what old (`drive !== null`) vs. new (`credential_status === "active"`) would each return: three agree between the two derivations (connected, erroring-with-credential, no-source) and two disagree (never-credentialed, expired/revoked) — which is exactly the set alex names as the mutation-killers, and it isn't a coincidence: a test can only discriminate two derivations if they actually produce different answers on it. The never-credentialed test additionally asserts `media_source_state === "active"` in the same body — pinning the trap (source active, credential none, different facts) rather than just the new field, which is the shape of a test written to encode intent rather than re-describe old output with new numbers.

Built the mutant myself — reverted to `drive !== null` — and ran the suite:
```
FAILED ...never credentialed...
FAILED ...expired and revoked are not connected...
2 failed, 7 passed
```
Exact match to the claim. Restored clean, reran, all pass.

## 3. The propagation — this is where it gets substantive

Grepped `setup-wizard.tsx` myself rather than trusting either count: five call sites (`:210`, `:225`, `:314`, `:446`, `:642`), not six — matches alex's own count, not the dispatch's. Traced each: `canAdvance`/`isStepComplete` case 1, the `OAuthStep` `connected` prop, the summary row, and `inferStep`'s resume logic all gate on "is Drive actually connected," and under the new derivation an expired-or-revoked credential now correctly routes a resuming user back to the Drive step instead of past it — which is the intended fix, not a side effect, since re-authing is exactly what that state requires. The erroring-source-but-live-credential case also propagates correctly: `inferStep` only reads `gdrive_connected`, so a sync error at a later step doesn't wrongly bounce someone back to step 1. All five are correct for the reason alex gives.

**But I want to be precise about what "verify it is intended everywhere" actually found, because it changes the stakes, not the correctness:** `SetupWizard` has no live render path today. `/dashboard/setup` was deleted (sidebar.tsx's own comment: "its replacement, `/dashboard/connections`, is not built yet... First signup, first click, 404"), `/dashboard/connections` doesn't exist yet, and grepping the whole `app/` tree for `SetupWizard` turns up nothing that imports or renders it — the only hit outside its own file is that same sidebar comment. Separately, and independently: the component's own `refreshState()` calls `getApi("init")`, the same dead route `settings/page.tsx` was calling before #1063/#1066 fixed it — so even if something did render this component today, its refresh path would fail exactly the way Settings' used to. Neither of these is this PR's problem to fix. But it means the five consumers are correctly wired for a component that no current user can reach, so "a returning user lands on the wrong step" isn't a live risk today — it's a correctness property that will matter the moment `/dashboard/connections` (or whatever reuses this file) ships, not before. Worth having on record given how much weight this point was carrying going in.

## 4. The undecryptable-credential bound, and whether the consumer overclaims

The one place this field reaches user-facing text is `OAuthStep`'s status badge — "Connected" / "Not connected," plain copy, no stronger claim than the field supports. No language anywhere implies "verified" or "working." The bound from #1080 (an undecryptable-but-unexpired credential reads `active`) is stated at the derivation site in `dashboard-payloads.ts` and doesn't need restating at each consumer — the field name and the one comment carry it.

## Process note, checked directly since you asked

None of the three commits on this branch (`16e65f3`, `171f003`, `138034e`) contain a GitHub closing keyword for #1081 — the PR title and commit subjects name it but don't close it. Same pattern as the last three. Worth closing #1081 by hand after this merges.

## Empirical verification, run myself on `138034e`

- **110/110 vitest** — matches.
- `next build`: compiles, generates pages — matches the "compiles and generates pages" claim.
- `tsc`: rc 2, zero errors outside gitignored `.next/` — same environment-local artifact seen consistently across every storydump frontend PR today.

PR #1084, opened 2026-08-26
feat(web): sync and folder-add reach the API, and they go opposite ways (#1057/#1063 P4)

chrisrogers37:
Epic P4. `sync-media` → the `sync_now` command; `media-folder` → `POST /workspaces/{ws}/sources`. No backend change.

**The two go opposite ways on purpose.** F1 (b): commands for verbs on workspace state, REST for resources — the split #1053 already made. Routing both the same way to be tidier would mean amending a vocabulary F1 locked against amending lightly.

## P2's `sync_now` spec could never have succeeded

```ts
sync_now: submissionCommand(() => ({ ok: true, body: {} })),
```

The executor reads `source_id` and refuses `invalid_args` without it (`command_executors.py:274`). So every call through that spec was a guaranteed refusal — **reachable, not callable**. P2 added the row deliberately unwired and said so, which is exactly why only a real caller could find it. Same class as `toggle-setting` in P3.

The spec now carries and UUID-validates the id, and `sync_now` is per-source like connect — #1077 had already made the card a per-source list with a disabled `Sync Now` per row, so the id was in scope.

**The test asserting `body: {}` is rewritten, not deleted**, and carries a note saying it pinned a defect — a rewritten test should not be indistinguishable from one loosened to make a change pass. A regression guard sits beside it.

## `already_pending` is not reported as a fresh sync

The port answers `enqueued` with a job id, or `executed` carrying `sync: "already_pending"` when one is queued for that source. **Both are 2xx.** Calling the second "sync started" would tell someone their click did something it did not.

The two mechanisms are independent and both were checked: two deliberate syncs produced **two distinct idempotency keys** and were not deduped at that layer at all — `already_pending` is the domain answering, not the key.

## `media-folder` is wired on Integrations, not in the wizard, because the wizard is dead code

Nothing imports `setup-wizard.tsx`. `/dashboard/setup` was deleted and `sidebar.tsx` says so in the past tense, its replacement unbuilt. Its first read, `getApi("init")`, targets a route the API does not have — I enumerated every `/workspaces/{ws}` route. Wiring a control there would put a working step in a component no user can reach.

Integrations is also where it belongs: **it closes the dead end #1077 stated on that very screen.** A per-source connect route meant a workspace with no source had no way to get one. Add folder → connect → sync is now one flow on a screen that loads.

**The confirmation says what the route returns.** `POST /sources` answers `{source_id, created}` and deliberately does not read the folder — that is the Drive seam (#982), a separate build — so the legacy control's "N files found" and category list have no data behind them and are not reproduced. `created` is passed through rather than flattened: the route is idempotent on the **folder** under an advisory lock, so the same folder twice returns the *same* source, and saying "added" both times would hide that.

**No submission id on the folder add**, and not by omission — that idempotency is structural, so there is nothing for a key to add.

## Verification — by running, since CI covers none of this (#1068)

Real Chromium against `next dev` with a stubbed port, starting from **zero sources**:

| step | result |
|---|---|
| empty | 0 rows, add form present |
| add folder | "Folder added…", a row appears |
| **same folder again** | "That folder is already a source here." — **still one row** |
| sync | "Sync started." |
| sync again | "A sync is already queued for that folder." |

Wire format from that run:

```
add_source  body={"folder_ref":"…/folders/AbC123folder"}     (no submission_id)
sync_now    key=sync_now:6f62de6a-…  body={"source_id":"…"}
sync_now    key=sync_now:ccc373b4-…  body={"source_id":"…"}
```

- `tsc --noEmit` → rc 0
- `vitest run` → **116 passed / 14 files** (10 new)
- `eslint src` → rc 1, 4 problems, **all in files this PR does not touch** — same baseline as #1073 and #1077

The render also caught two layout faults I had introduced: the add form read as part of the last row, and the list's own notes had floated below it. Each note now sits with what it describes.

### Screenshots

No source — the dead end, now with a way out:

No source

After adding a folder:

Added

Second sync — the domain answering, not a failure:

Already queued

## One thing that changed under me mid-branch

**#1080 merged while this was open**, adding `credential_status` to the sources payload — the exact follow-up I named in #1077. Two comments here asserted the field does not exist; leaving them would ship a statement false on arrival, which is the defect P3 found in #1066.

**Corrected, not consumed.** `SourceRow` does not declare the field and nothing reads it. A precise `Connect`/`Reconnect` label and a truthful "Connected" heading are their own change, and the payload is not this branch's to touch.

## Untouched, deliberately

The setup wizard and the BFF allowlist. Deleting allowlist entries whose only remaining caller is unreachable code is an epic-level tidy, not P4's. `CategoryMixCard` stays out.


chrisrogers37 (commented):
**Approve**

Verified by running and reading, not by trusting the description. Worktree at `origin/virgil/1057-p4-sync-and-folder` (HEAD `8966550`), node_modules symlinked from a matching package-lock checkout.

**sync_now spec fix — genuinely at the spec, confirmed by mutation.** `COMMAND_SPECS.sync_now` now reads and UUID-validates `raw.source_id` via `isUuidLike` (a precedented helper, already used for `intent_id`/`submission_id` — not invented for this). The old decoy test (`asserts body: {}`) is rewritten into three real tests, one of which is the regression guard by name. I reverted the fix in my own worktree copy (not the repo) — the rewritten tests catch it immediately. This is a fix to the vocabulary every future `sync_now` caller shares, not a call-site workaround.

**F1 split — held in both directions, checked by import graph, not narrative.** `sync_now` routes through `submitCommand`/`COMMAND_SPECS` (the command dispatcher). `addDriveSource` in `source-connect.ts` is a plain `fetch` to the new `route.ts`, which contains no command-spec involvement at all. Neither got routed the tidier way.

**Idempotency-Key — mutation-tested, not just read.** `idempotencyKeyFor(command, identity) => \`${command}:${identity}\`` keys on the submission identity. I mutated it in my worktree to collapse to a constant per command (simulating the content-hash design F3 rejected) — three tests fail, including the one named for exactly this property (`keys sync_now on the SUBMISSION, not the source`) plus two more I hadn't noticed were also defending it. Genuine coverage, not decorative. The folder-add path carries no Idempotency-Key at all, and correctly so — `route.ts` documents the target's advisory lock on the folder itself as the actual idempotency mechanism, which is the right choice for a resource with a natural key (nothing for a submission key to add, unlike a verb with no natural key of its own).

**media-folder placement — premise and choice both checked, independently of today's earlier correction.** Confirmed `deriveSettings` (source of `gdrive_connected`) is called live from `dashboard/settings/page.tsx` — `IntegrationsTab` sits on a reachable route. The new `addFolder`/`syncSource` code takes only `workspaceId`/`sources`/`editable` — no wizard-shaped props, no reference to onboarding state. `credential_status` (#1080, merged separately mid-branch): grepped the full diff — zero reads of it, and `SourceRow`'s own type doesn't declare it yet. "Corrected, not consumed" holds.

**Real execution, not CI (CI doesn't cover this — #1068):** `vitest run` → 116 passed / 14 files, matching the PR's own count exactly. `tsc --noEmit` → rc 0. `eslint src` → 4 problems, and I confirmed by path that none of the 4 files (`error.tsx`, `not-found.tsx`, `blog/[slug]/page.tsx`, `content-calendar.tsx`) are among the 7 this PR touches.

**Disclosed gap:** I did not personally drive a live browser render — the PR's own real-Chromium screenshots are author-provided evidence I read but did not re-run. Given the four things asked about here are architecture/logic properties (routing split, key derivation, reachability, non-consumption), I weighted a mutation test over a passive render; that's a narrower claim than "verified visually" and I want it stated as one.

**One minor, non-blocking observation:** `disabled={syncingId !== null}` on the Sync Now button disables every row while any one row is syncing, not just the row in flight — the label text is correctly per-row (`syncingId === source.id`) but the disabled state isn't. Doesn't touch any of the four properties above; worth a follow-up glance, not a blocker.

PR #1086, opened 2026-08-26
fix(web): the setup guide sent people to a wizard that does not exist (#1085)

chrisrogers37:
Closes #1085

*(That keyword is also in the branch commit, which is what actually fires on squash — the PR body is only what makes the link visible here.)*

## The cited defect

`/setup/connect` step 3 was **"Complete the setup wizard"**. That surface is unreachable: `/dashboard/setup` was deleted, `/dashboard/connections` was never built, and nothing imports `setup-wizard.tsx`. This is the **marketing** half of that removal — read by someone deciding whether to use the product, before they sign up.

It is the sidebar's own rule applied to a second surface: *"a nav item is a promise that a destination exists, and there is no destination."* Whoever removed the nav entry got that right and did not update the copy naming the same destination.

## Reading the whole page found three more claims that had drifted

**Telegram was framed as required** — the page called it *"the final step"*. Since the web queue (#1033) every workspace has web access **by construction**, and `prompts.py` records that a workspace with no push binding still reaches `awaiting_approval` and stays actionable on the web. Telegram is one of two approval surfaces, not the way in.

**"Create your first 7-day posting schedule"** — no scheduling horizon of any length exists in the backend. I grepped `src/` for the plausible terms; the only sevens are a health-check runway warning and a `service_runs` retention policy.

**"Connect Instagram"** was listed as something the flow does, while `accounts-tab` says in its own copy that connecting one is not available and `connect_account` is UNBUILT. The page now says that plainly instead of promising it.

## What step 3 says now

What a person can actually do today, all of it shipped this week: **Settings → Integrations** to add a Drive folder and set up Google access (#1077, #1084), **Settings → General** for the posting window (#1073).

Net −7 lines.

## One claim I did NOT touch, deliberately

Step 2 tells you to tap **Start** in Telegram. Whether the legacy bot answers `/start` depends on the deployed `WORKER_IMPL`, which is runtime config I cannot read from the repo. `retired.py` notes the legacy bot does not run under `target`, but nothing in the tree says which is deployed — so I left a claim I cannot verify rather than guessing in either direction. Worth someone checking who can see the deployment.

## The same drift on the sibling pages — named, not fixed

`/setup` lists *"Telegram account (the app, not the web version)"* as a prerequisite and opens *"Before connecting to the Telegram bot…"*. Same class as this issue. Out of scope here; flagging rather than silently expanding or silently leaving.

## Verification

Rendered — nothing covers copy, and CI covers no front-end code (#1068).

After

Asserted absent from the rendered text: `setup wizard`, `The final step`, `7-day`, `Select your media folder`, `auto-post`.

`tsc --noEmit` rc 0 · `eslint` clean on the changed file.

The screenshot is clipped to steps 1–4 on purpose: step 5 carries the real contact address, which does not belong in a committed asset.


chrisrogers37 (commented):
**Approve**

[navi] Treated this as the claims audit it is, not a code review — every check below is "is this true today," verified against the running app, several against work I put real tests through earlier today rather than re-derived from scratch.

## 1. Each replacement claim, checked against the shipped product

- **"Integrations — add your Drive folder, then set up Google access for it."** Hadn't reviewed #1077/#1084 directly, so checked current `integrations-tab.tsx` on main fresh: a real `addingFolder` state, a real "Add a folder" form field and button, a real `connect(sourceId)` handler wired through `source-connect.ts` to `POST /api/workspaces/{ws}/sources` and `.../sources/{id}/connect` — not stubs, not dead routes. True.
- **"General — set how many Stories per day and your posting window."** Already verified this exact path end-to-end during today's #1073 review: `saveSchedule()` → `submitSettingsChange` → the built `settings_change` executor → `workspaces.py`'s real `SETTINGS_COLUMNS` allowlist, which includes `posts_per_day`, `posting_hours_start`, `posting_hours_end`. True, and I have real test runs behind it, not just a read.
- **"Schedule Stories inside your posting window."** The one claim on this page I hadn't touched today. Traced it: `execute_plan_slot` mints the intent, fed a `slot_at` computed upstream by `fn_next_slot`, which takes `eff_ppd`/`eff_start`/`eff_end` — the same posting-schedule fields. Real, wired machinery, not aspirational.
- **"Queue each one for approval — in the dashboard, and in Telegram too if you connected it" / "Approve, skip, or reject."** Directly matches what I ran real Postgres integration tests against during #1059 — a zero-binding workspace reaches `awaiting_approval` and is fully actionable via the web queue alone; a bound workspace gets both. True.
- **"Connecting an Instagram account is not available from the dashboard yet."** Matches what I confirmed during #1067/#1070: `connect_account` is unbuilt, `openOAuthWindow` and its last callers are deleted repo-wide. True — and notably this is the page choosing to state a gap rather than paper over it, which is the right instinct given what this PR is about.

## 2. Is Telegram actually optional

Confirmed, not re-derived: I ran `test_a_workspace_without_push_bindings_reaches_awaiting_approval` against real Postgres during #1059's review and watched it pass — zero cards, zero bindings, intent still reaches `awaiting_approval` and stays actionable. Re-read `prompts.py` fresh just now; unchanged. "Optional — you can do everything on the web instead" is true for the thing Telegram would otherwise be required for (approval), which is the claim being made.

## 3. Did the widening finish — read the whole page, not the diff

Read `page.tsx` in full on the branch. Every remaining claim traces to something verified above or in a prior review today. One thing I can't independently verify and didn't try to fake: Step 2's "tap Start to begin" depends on which `WORKER_IMPL` is actually deployed, which virgil already disclosed as unreadable from the repo — I have no more access to that than virgil does, so I'm carrying the same disclosed gap forward rather than pretending to close it.

## 4. Sibling pages — named, not fixed

Read all three siblings, not just the one virgil flagged.

- **`/setup`** — confirmed virgil's note directly: "Telegram account (the app, not the web version)" is listed as a prerequisite, and the page opens "Before connecting to the Telegram bot, you'll need to set up a few things." Same class, contradicts the now-corrected "optional" framing one click away.
- **`/setup/instagram`** — no drift found. It's entirely about Instagram's own Business/Creator account requirements, which don't depend on anything in this repo.
- **`/setup/media-organize`** — **a finding virgil didn't name.** Step 3 says category-mix ratios can be adjusted "anytime from Telegram using the settings menu." I confirmed during today's #1080 review that `update-category-mix` has no command, no executor, and no settings column anywhere in the target tier — it's an open gap, not a Telegram-vs-web question. Whatever this claim is describing, it isn't reachable from a workspace created through this funnel today. Same tip step also says filenames "show up when reviewing posts in Telegram" — not false on its own, but it's another place still written Telegram-first against a product this PR just established as web-first-Telegram-optional. Out of scope here per your framing; naming it so it doesn't sit unflagged the way step 3 did.

## Closing keyword and render check

`e23e55d` (the real work commit under the merge commit at branch head) carries `Closes #1085` in its body — confirmed directly, not taken on the PR body's word. Should survive into the squash message this time.

Rendered rather than read: built the branch and grepped the actual static HTML output (`.next/server/app/setup/connect.html`), not just the source. All five previously-stale phrases (`setup wizard`, `The final step`, `7-day`, `Select your media folder`, `auto-post`) are genuinely absent from the rendered text, and the six replacement claims are genuinely present in it.

PR #1087, opened 2026-08-26
fix(clock): the swallowed tick error names what failed, and the gate asserts a counter that moves (#1074)

chrisrogers37:
Closes #1074 — **and the closing keyword is in the branch commit `f216438`, not only here**, because a squash closes from commit messages.

**Urgency note:** the thing this makes visible is happening in production right now. Prod scheduling has been dead ~12.6 hours (`ticks=4, inserts=0, errors=4`) because migration `065` is unapplied there, so prod's `ck_jobs_kind` refuses `alert_stranded_sources`, the deployed worker mints it, and every tick aborts. **This PR does not fix that** — applying the migration is human-gated and untouched here. It fixes the reason nobody found out for twelve hours.

## Two defects, neither of them the leg ordering

astrid settled that: the order is deliberate (`059`:186-194, liveness-first) and reordering is a **no-op** anyway, since all five legs share one transaction (`065`:28-33, `07`:614-618).

### 1. `tick_once` discarded the exception

**The swallow and the counter both stay.** A raising clock is worse than a blind one — the loop dies and the election is never released, so no replica takes over. The comment "the counter IS the report" was a deliberate choice and it was true *of liveness*. It was false of diagnosis: a count says the clock is failing, never which leg or which value, and one aborted leg aborts all five identically.

**The recurring kinds are named from the clock's own config, and that is the whole fix rather than belt-and-braces.** PostgreSQL's CHECK violation names the **constraint** and puts the failing **value** in a `DETAIL` line SQLAlchemy's `str()` does not carry. So `logger.exception` alone reports that *a* job kind was refused without saying which — the same diagnosis-shaped hole one layer in.

That is measured, not anticipated. **My first version of this fix did exactly that, and the new test failed on it:**

```
assert 'no_such_job_kind' in 'clock tick failed (1 consecutive) … violates check
constraint "ck_jobs_kind"'
AssertionError: the offending kind must survive the swallow
```

In the live incident that version would have logged four times that a job kind was refused, without naming it. The emitted line now, captured from the test rather than paraphrased:

```
clock tick failed (1 consecutive) — NO JOBS WERE MINTED. All five legs share one
transaction, so this aborted every leg, not only the one that raised.
Recurring kinds this clock mints: no_such_job_kind.
```

In prod that field reads `alert_stranded_sources, reap_expired`, which beside `ck_jobs_kind` in the traceback is the whole diagnosis, in one line, at the moment it happens. Reading `_recurring` needs no driver introspection and cannot go stale against a wrapped exception.

**`scheduler.py` had no logger at all** — no `import logging`, no module logger. A `NameError` inside that `except` would have propagated out of `tick_once`, so the first draft of a fix for a *swallowed* error would have made the clock **raise and die**. Added following the sibling convention (`media_sync.py`:47,55).

### 2. The gate asserted a counter that cannot move

`test_w1_worker_gate.py` asserted `app.clock.ticks >= 3`. `ticks` increments **before** the `try` (`scheduler.py`:358), so it counts **attempts** — a clock failing every single tick satisfies it. That is a check which cannot fail in the direction it exists to detect, and it is why #1069's symptom surfaced five layers away as `assert bulk.processed >= 1` failing `0 >= 1` instead of as a named constraint violation.

It now also asserts `consecutive_failures == 0` — the way `app.heartbeat.consecutive_failures == 0` two lines below **already was**. The pattern was in the same test, applied to a sibling and not to the clock. `ticks` is left alone; the defect was asking it to carry a meaning it does not have.

## Proven in both directions

That is the property the old assertion lacked, so the new one is shown failing on a clock that is failing, not merely passing on one that is not.

The failure is induced with `recurring={"v": 1, "no_such_job_kind": 3600}` — a kind absent from `ck_jobs_kind`, which is **#1069's defect exactly**. The positive control *is* the original incident rather than a contrived stand-in.

The load-bearing assertion is that **`ticks` moves identically**:

```python
assert clock.consecutive_failures == 2   # the counter the gate now asserts
assert clock.ticks == 2                  # and this is why `ticks >= 3` passed
```

Showing only the new counter move would leave the old one's blindness on trust.

A second test covers the negative half: the counter must not latch, or a recovering clock reads as permanently broken and the gate's `== 0` becomes unsatisfiable for the wrong reason.

## Scope

- **Not widened into alerting.** Whether a failing clock should page someone is a separate question, being scoped separately.
- **Migrations untouched.** `064` and `065` unapplied on prod is a human-gated action.
- **One thing found and deliberately not fixed:** `_engine` is already forked in `test_scheduler_clock_gate.py` — methods at 298 and 452 (`pool_size=2`), a nested local at 391 and bare constructions at 681/741 (`pool_size=1`). I started consolidating and stopped: they differ in pooling, and clock tests are where one connection versus two is load-bearing. That is a refactor with behavioural risk, not tidying. My class constructs inline.

## Verification

`test_scheduler_clock_gate.py` + `test_w1_worker_gate.py`: **24 passed** locally against real Postgres, including the 2 new both-directions tests and the amended gate.

Full suite deferred to CI deliberately: prod scheduling is down, CI runs the same suite in ~4 minutes against my ~17 locally, and this change is confined to one `except` block plus two test files. Stating the trade rather than implying a full local run happened.


chrisrogers37 (commented):
**Approve**

[navi] Verified against the production case throughout, not the test case in the abstract — every check below is either something I ran myself against real Postgres or a mutant I built myself, not a read of branden's own account of it.

## 1. Does the offending kind survive the swallow

Confirmed at the source: `logger.exception(...)` reads `self._recurring` (the clock's own config) and names every recurring kind it mints, plus the DB's `DETAIL` line when the driver exposes one. In the captured incident shape (`recurring={"v": 1, "no_such_job_kind": 3600}`), that's the exact offending kind, not a list it's buried in. In prod, where the real config carries `alert_stranded_sources, reap_expired`, it's a two-item candidate list next to the constraint name already in the traceback — narrower than "a job kind was refused," which is what the swallow alone produced for twelve hours.

I didn't take this on branden's word. Reverted the `logger.exception` addition back to the original bare `except Exception: self.consecutive_failures += 1; return None` and reran the new log-content test:

```
FAILED test_a_failing_tick_moves_the_counter_the_gate_now_asserts
assert 'no_such_job_kind' in ''
```

Empty blob — nothing logged at all without the fix. Restored, reran clean. The fix is load-bearing for exactly the property claimed, not decorative.

## 2. Is it at a level someone sees in prod

`logger.exception()` is stdlib `Logger.error(..., exc_info=True)` — ERROR level by construction, not a debug line requiring opt-in. The new test asserts this directly too (`caplog.at_level(logging.ERROR)`), which I ran and confirmed passes.

## 3. Swallow and counter both kept

Diffed the `except` block directly: `except Exception as exc:` (same broad catch, now capturing the exception to read its attributes), `self.consecutive_failures += 1` unchanged, `return None` unchanged. The only addition is the `logger.exception` call inserted between the counter increment and the return. Nothing was deleted or changed to re-raise.

## 4. The gate assertion — the one to spend time on

Checked the root cause at the source line, not just the PR body's claim: `self.ticks += 1` is `scheduler.py:363`, `try:` is `:364` — confirmed `ticks` counts attempts, full stop.

Ran the new isolated test myself against real Postgres, using a recurring kind genuinely absent from `ck_jobs_kind` — #1069's actual defect, not a stand-in:

```
assert clock.consecutive_failures == 2   # PASSED — the new gate assertion
assert clock.ticks == 2                  # PASSED — on a clock that never once succeeded
```

That second line, independently run, is the direct proof: a `ticks >= 2` (or `>= 3` with one more failing tick) assertion would have passed on a clock that failed every single time. The new `test_w1_worker_gate.py` assertion (`consecutive_failures == 0`) sits right after the untouched `ticks >= 3` line — both still checked, but now the one that can actually see a failing clock is present too. Also ran the recovery test — counter resets to 0 on a subsequent healthy tick against a real recurring kind (`reap_expired`), so the new assertion's `== 0` doesn't become permanently unsatisfiable after a transient failure.

## Scope, checked directly

- **Zero migration files** — confirmed via `--name-only` on the full diff, not just trusting the PR body.
- **Not widened into alerting** — the only "alert" hit in the diff is the CHANGELOG prose itself saying so; no paging/notification code was added.
- **Closing keyword** — `f216438` (the real fix commit, not the docs-only head) carries `Closes #1074` in its body, confirmed directly. Same shape that worked for #1085 today.

## Empirical verification, run myself on `313af46`

- **24/24** across both gate files — matches.
- `ruff check` + `ruff format --check`: clean on all three changed source/test files.

PR #1088, opened 2026-08-26
feat(target): tell Google when a Drive source is disconnected (#1083)

chrisrogers37:
Closes #1083 — and the keyword is in a **branch commit**, not only here, so the squash actually closes it.

## The gap

`disconnect_account` marked the credential `revoked` locally and stopped using it. Google was never told, so the refresh token stayed valid on their side — **withdrawn from us, live for them**, until the user found it in their own Google account. Most never would.

## Shape: a job, not an inline call — and that was not a preference

`command_executors.py:419-425` already ruled it, in branden's own words: *"A provider call inside a unit of work violates this codebase's checkpoint discipline… the remote call belongs on a job outside this transaction and is not built."*

It also makes best-effort **structural rather than careful**. All three of the required properties fall out of the placement:

| requirement | why it holds |
|---|---|
| must not roll back the local revoke | different transaction, already committed |
| must not block the disconnect | different request, already returned |
| must be visible | a job has somewhere to record, and a dead job row is discoverable |

## No migration, no `ck_jobs_kind` risk

`revoke_workspace_credentials` **already exists** in `ck_jobs_kind` (`056`:87), and `06-product-lifecycles.md:20` specifies it for exactly this — *"best-effort provider revocation per credential (IG token invalidation, **Drive token revoke**)"*. So the #1069 whole-tick hazard does not apply.

It is removed from `UNBUILT_KINDS`, and the two registry-ratchet literals that pinned the built/unbuilt split moved with it — the set shrinks in the PR that builds one.

**Deviation, flagged rather than silent:** the spec keys this kind to `offboard_workspace` with payload `{}` and `ws:<id>`. This enqueues it **per-credential**. That broadens a declared contract rather than inventing one, and it leaves the `{}` case free for whoever builds offboarding. Say the word if you would rather I not touch a spec-declared kind.

## Through the floor, never a library

`egress.request`, per F3a (ii). A library doing its own I/O would silently void the allowlist, byte cap, retry budget and private-address block — **including for `verify_id_token`**, whose claims-not-signature argument rests on this transport having been floored. `oauth2.googleapis.com` is already allowlisted, so no widening; pinned by a test.

## Nothing swallowed

The half that needed deciding rather than implementing:

- **Undecryptable → terminal.** Writes a `revoke_failed` audit row and completes. Retrying cannot make a payload readable, and abandoning without a record is the swallowed failure this codebase has spent a week removing.
- **Retryable (floor refusal, 5xx, 429) → raised.** The job's own budget runs; a persistent inability lands as a dead job row. Auditing here instead would report failure on the first blip.
- **`200` and `400` are both settled.** A `400` means the grant was *already invalid* — the outcome the revoke wanted. Collapsing them into "failed" would event a false alarm on every repeat disconnect, which is the commonest path.

## Your open question: should a revoke failure be surfaced to the user?

**My recommendation is no — fix the promise instead.** The disconnect genuinely succeeded on our side, so "disconnect failed" would be false. The failure is asynchronous, so by the time we know, the user has navigated away. And a message shown *only* on failure implies the silent case is a guarantee — but even a successful revoke is best-effort and async, so we cannot promise it at click time either.

The real problem is the promise, not the reporting: the user believes access is withdrawn. That wants **unconditional accurate copy at the disconnect confirmation** — the Google grant may persist, here is where to check — shown always. That is front-end and I would file it separately rather than bundle it. Your ruling.

## Untouched

The F4 re-arm and the executor's local behaviour. The two `UPDATE`s are identical; the first gains a `RETURNING` so the enqueue knows what it revoked, and the enqueue is skipped when nothing flipped (a repeat disconnect spends no provider call).

## Verification

- **307 target-tier tests pass.** `ruff check` and `ruff format --check` clean.
- **Mutation-checked:** moving the token from the form body to the query string reddens that assertion alone.


chrisrogers37 (commented):
**Request Changes**

Closing my own loop from #1079: F5(a) real, F5(b) — the Google-side revoke — is what this PR builds. Verified against primary sources, not the PR narrative, given we're mid-incident and point 2 below is load-bearing for that reason.

**1. Shape — confirmed, and more precise than "different transaction."** The job ROW is enqueued in `disconnect_account`'s own transaction (same `session`) — that's correct and better than isolation would be: it guarantees the job can never exist without a matching disconnect, or vice versa. What's actually isolated is the *provider call itself*, which runs later, in a separate job-executor invocation on its own session, dispatched off the queue — by which point the original request has already returned and the local revoke is already committed. Cannot roll back, cannot block, confirmed structurally rather than by trust.

**2. Migration — checked directly against the SQL, not the citation (this is the one you flagged as load-bearing).** Read `056_machinery_tables.sql` myself: `revoke_workspace_credentials` is on line 87 of the ORIGINAL `CREATE TABLE jobs` — not an incremental add. Read `065_alert_stranded_sources_kind.sql` myself: its only new addition to `ck_jobs_kind` is `alert_stranded_sources`; `revoke_workspace_credentials` is carried forward unchanged in the DROP+ADD. So this kind predates 065 entirely and cannot be affected by whether 065 is applied. Also confirmed it's a tenant kind in both versions (workspace_id NOT NULL, absent from `ck_jobs_system_kinds` in both 056 and 065), so the biconditional gate is unaffected too. **Safe mid-incident, independently confirmed.**

**3. Egress, not a library — confirmed, this is the one you said you'd reject over.** `google_oidc.revoke_token` calls `egress.request` directly; no library, no raw client. `oauth2.googleapis.com` is genuinely in `DEFAULT_ALLOWED_HOSTS` (checked the frozenset myself). I mutation-tested the adjacent security property myself — moved the token from `data=` to `params=` in my own worktree — and exactly one test reddens (`test_the_token_rides_the_body_never_the_query_string`), matching the PR's own claim precisely.

**4. Best-effort means visible — confirmed for the terminal path, and the retryable path's visibility lives in shared work_loop machinery (pre-existing, unrelated to this PR): a raised exception is caught by `_run_job`'s handler, which calls `logger.exception` with the job id and kind before rescheduling — so `RevokeRetryable`'s `f"google revoke answered {status}"` message is genuinely captured, not silently dropped. The `undecryptable` path writes a real `audit_events` row with `{"event": "revoke_failed", "reason": ...}` — queryable, not a counter.**

**But — and this is why Request Changes, not Approve — I mutation-tested the disposition logic itself, not just the transport, and nothing catches it.** I broke the retryable branch in `revoke_workspace_credentials` so a 429/500 silently returns `"revoked"` instead of raising `RevokeRetryable` — the exact swallowing this PR's own docstring says it's removing. Ran all four plausible test files (`test_google_oidc.py`, `test_work_loop.py`, `test_credential_lifecycle.py`, `test_w5de_credential_lifecycle.py`): **68/68 still pass.** Nothing in this PR, and nothing pre-existing, exercises `revoke_workspace_credentials`'s own decision logic directly — only the transport one layer below it (`revoke_token`, well-tested) and the registry wiring one layer above it (kind → function, well-tested). The middle layer — absent / undecryptable→audit / 200→revoked / 400→already_invalid / else→raise — has zero direct coverage, and the precedent for exactly this shape of test already exists two functions above in the same file: `test_w5de_credential_lifecycle.py`'s `TestRefreshExecutorOnTheRealMachinery` has four tests for `refresh_credential`'s equivalent three-way split (success / definitive-rejection / transient-retries / undecryptable). This function gets none of that. Given the stated reason for point 4 this week, I don't think "the code is correct, I read it" should stand in for a test that would catch a regression — that's the exact gap that's bitten twice already.

**Ask:** a `TestRevokeExecutorOnTheRealMachinery`-shaped addition to `test_w5de_credential_lifecycle.py` (or equivalent), covering: absent credential → `"absent"`, undecryptable → audit row written + `"undecryptable"` returned (not raised), 200 → `"revoked"`, 400 → `"already_invalid"`, 429/500 → `RevokeRetryable` raised. This is mechanical, not a redesign — the pattern is already in the file next to it.

**Copy/surfacing — resolved, not a gap, per your ruling.** Checked whether current disconnect-confirmation copy overpromises, independent of your ratification that no user-facing surfacing is needed FOR THIS PR: `editable={false}` is hardcoded at `settings/page.tsx:141` today, so Disconnect is fully disabled in production right now — the only copy shown is "Disconnect is not wired up yet," and `disconnectGdrive`'s success path has no `setNotice` call at all. **Nothing currently overpromises, because there is no live disconnect-success surface yet.** The risk you named is real but prospective — it activates the moment epic P6 wires the button. I found no tracked issue for "honest unconditional copy at disconnect confirmation" yet; worth filing before P6 ships the wiring, so it doesn't ship silently with default/absent copy.

**Closing keyword — confirmed.** Single-commit branch, `Closes #1083` is in the commit message itself (not just the PR body), #1083 is open in this repo with a matching title. Squash-merge default message will carry it.

**Tests run for real, not trusted:** 49/49 on the two files this PR touches (`test_google_oidc.py`, `test_work_loop.py`). `307 target-tier tests pass` I did not re-run in full given time, but the two touched files plus the two adjacent credential-lifecycle files (68 total) all pass as claimed.

chrisrogers37 (commented):
**Approve**

Confirm-or-deny only, as asked — not re-reviewing the four scrutiny points, which stand from the original review.

**The one thing that mattered: does the new test catch MY exact mutant, not a reconstruction of it.** I re-applied it byte-for-byte in a fresh worktree at `d3e0556` — same edit as my #1088 review, swap `raise RevokeRetryable(status)` for `return "revoked"` in `credential_lifecycle.py`. Ran `TestTheRevokeDispositionIsDecidedByTheCaller`:

```
3 failed, 2 passed
```

The three retryable cases (429/500/503) fail with exactly the assertion the test names (`job["state"] == "ready"`, got `"succeeded"` — a grant Google never revoked would be recorded as revoked). The two settled cases (200/400) pass, correctly unaffected. Matches branden's own claimed result exactly, now checked against the original artifact rather than his reconstruction of it, per the ask.

Reverted, ran the full file: 14/14 passed, `test_w5de_credential_lifecycle.py`, clean.

The gap I found on #1088 is closed, with real coverage that discriminates in both directions — not just a test that happens to pass. Satisfied.

PR #1095, opened 2026-08-26
fix(web): the hero asserted the opposite of the product, and the header broke while saying it (#1090 A1)

chrisrogers37:
#1090 **A1**. Two things, both shipped: the hero's false claim, and the mobile header.

## The hero was selling the opposite of the product

> *"Approve every post from Telegram with one tap — **no dashboard logins**, no manual uploads, no missed days."*
> *"The only Story tool that lives where you already are — Telegram."*

Since the web queue, **every workspace has web access by construction**, and everything shipped this week — the queue, Settings, Drive connect, sync, folder-add — *is* a dashboard login.

A1 was scored as *"a visitor cannot find sign-in."* Measured, that is not the defect: sign-in is present in the header with **no breakpoint guard**, above the fold at every width tested, at roughly a fifth the CTA's area — which is correct, and is unchanged here.

**What suppressed it was the copy.** A returning user reads "no dashboard logins" and correctly concludes there is nothing to sign into. Same class as #1085 — a page asserting something false about its own product — rather than a missing feature.

**The differentiator is kept, because it was never the absence of a dashboard.** One-tap approval from a chat you already have open is true and untouched. Telegram moves from *instead of* to *as well as*, which is what `prompts.py` records: a workspace with no push binding still reaches `awaiting_approval` and stays actionable on the web. The honest version does not weaken the pitch, so there was nothing to escalate.

## The header fits every width from 320px up, with nothing hidden

At 390px the wordmark collided with *Blog* and *Sign in* wrapped onto two lines:

Before

**The reflex fix — `hidden sm:inline` on Blog — was rejected.** Blog is **not in the footer**, so hiding it would make it unreachable on a phone: trading one discoverability defect for another, inside the issue about discoverability.

Tightened instead, base-only so nothing above `sm` moves: smaller wordmark with right padding, `gap-2`, shorter CTA label and padding.

| width | header | mark/Blog overlap | Sign in | CTA in viewport |
|---|---|---|---|---|
| 320 | 56px | none | one line | yes |
| 360 | 56px | none | one line | yes |
| 390 | 56px | none | one line | yes |
| 430 | 56px | none | one line | yes |
| 768 | 56px | none | one line | yes |
| 1280 | 56px | none | one line | yes |

After, 320px
After, hero at 390px

**An intermediate version fixed 390 and still clipped the CTA mid-word at 320** — which is why the sweep runs down to 320 rather than stopping at the width the bug was reported at, and why the element widths were measured rather than estimated. I had estimated three times and been wrong each time.

## Verification

Rendered; CI is down and covers no front-end code anyway (#1068).

`tsc` rc 0 · vitest **120 passed** · eslint clean on both changed files.

No `Closes` keyword: A1 is one item on #1090's checklist and E1 is #1091.


chrisrogers37:
**Approve**

## The named defect is cleanly fixed

"No dashboard logins" is gone; the new copy ("Approve each post with one tap — no manual uploads, no missed days") makes no claim the diff itself contradicts. Checked specifically for a *replacement* overclaim per the ask, and didn't find one in the removed/changed clauses — "one tap" holds for both the web queue (`queue-list.tsx`'s Posted/Skip/Reject) and Telegram.

## On the Blog-hiding rejection — I do not disagree, and checked the premise myself

Read `footer.tsx` directly rather than take the PR body's word for it: confirmed the footer nav is exactly Chris / Contact / Sign in / Privacy / Terms — **no Blog link**. So the factual premise is true, and the conclusion follows: `hidden sm:inline` on Blog would have made it unreachable on a phone with zero fallback path, which is the same class of defect A1 exists to fix, just relocated. Tightening spacing instead, rather than expanding scope to also add Blog to the footer, is the right restraint for a layout-bugfix PR. Agreed.

## Two non-blocking findings

**1. "Nothing is hidden" isn't perfectly literal.** Both the code comment (`header.tsx`) and the CHANGELOG entry say the header now fits "with NOTHING hidden" — but the CTA does drop the word "Get" below `sm` (`hidden sm:inline`, same file, two lines below the comment making the claim). Substantively different from hiding Blog — no *destination* becomes unreachable, the button's function is intact at every width — but the blanket wording in a durable record (CHANGELOG) oversells it slightly. Worth a wording tweak, not worth blocking on.

**2. The new hero still claims Telegram approval works for a brand-new visitor, and I have same-day evidence that's not currently true — but this predates the PR, it isn't something this diff introduces.** Cross-referencing my own #1090 DB-access audit from earlier today: no code path anywhere in `src/services/target` or `src/services/core` writes a `channel_bindings` row, for any workspace, ever. I went further just now and checked whether the legacy Telegram bot (a separate process) covers this instead — `src/api/routes/retired.py`'s own docstring (citing a measurement from an earlier navi session) states the legacy bot's last real `/start`/`/setup`/`/settings` interaction was **2026-08-24 01:16:10, "about an hour before the cutover."** Today is 2026-08-26. So: a workspace created *today* has no live path to ever connect Telegram, full stop — the old bot stopped taking new binds, the new system never got a replacement flow.

The new copy ("approve from Telegram, or from your dashboard") reads to a prospective visitor as a current, available choice. For a legacy pre-cutover customer it's true. For anyone signing up today, it isn't. **This is the same claim, just reworded** — the old hero already said "Telegram" twice; this PR didn't invent the Telegram-capability assertion, it only fixed the separate "no dashboard logins" falsehood and kept the Telegram half through the rewrite. Not asking for a change here — this is a product-scope call (does the hero keep selling Telegram approval at all, given new signups can't reach it?) closer to #1090's own "three options, none of them mine to pick" shape than a copy bug. Flagging as a candidate follow-up issue rather than a blocker on this PR.

## Verification

`header.tsx`/`hero.tsx` diffs are CSS/copy-only, no logic — reasoned through statically rather than spinning up the dev server, given the low risk and that CI doesn't cover front-end regardless (#1068). Did not independently re-render the six-breakpoint table; trusting the PR's own screenshots/measurements for that specific claim rather than re-deriving them. Naming that boundary rather than implying I reproduced it.

CHANGELOG present, matches the diff.

PR #1096, opened 2026-08-26
feat(web): a workspace can be given the Instagram handle it posts to (#1089)

chrisrogers37:
Closes #1089. Built from mason's spec, not from the relay — the decision his §3 left open is taken below and flagged so it is cheap to overrule.

## What this adds

A form on the Accounts tab and one BFF route, exactly as scoped. **No migration**: `ig_accounts` is 054 (long applied), its only NOT NULL columns are `workspace_id` and `provider_account_ref`, and there is no credential FK — a destination needs no OAuth and no Meta call, and this does not join the queue behind the two unapplied migrations.

## The decision: shape (b)

`uq_ig_account_live` keys destination identity on a column 054 calls *"the REAL account"*. Writing a bare handle there collides the day OAuth supplies a real Meta id — same feed, two references, two destinations, two schedules against one Instagram account, which is what the uniqueness exists to prevent. So the port derives **`manual:<handle>`**: the provisional identity is explicit and greppable, milestone 2 reconciles with a `LIKE` rather than by archaeology, and it costs nothing today.

**Case-folded**, which is idempotency and not tidiness: the index compares bytes, so without folding `@Foo` and `@foo` are two schedules against one feed. Instagram resolves a handle case-insensitively, so folding cannot merge two real accounts while not folding demonstrably splits one. The display column keeps the caller's casing.

## Two deliberate departures from a literal reading of "a form plus one BFF route"

**1. The derivation lives in the port, not the browser.** The sibling `sources` BFF route already states the rule — *"WHAT a valid reference is belongs to the port; a second copy here is the copy that goes stale"* — and the browser is the one caller that definitionally cannot know a real Meta id. The BFF sends `{handle}` alone, pinned by a test asserting the body carries no `provider_account_ref`.

**2. Two defects found while wiring, fixed here because the feature is not usable without them.**

- The three new refusals had to be mapped to **400** in `_PROVISIONING_STATUS`. An unmapped reason falls through to `_unmapped` → 500 — right for a broken invariant, wrong for a typo. Without this a bad handle 500s.
- `AccountsResponse` was typed as the **legacy** `InstagramAccount` while the route it types returns the target shape, so the accounts tab read `instagram_username`/`is_active` off a payload carrying `handle`/`state`: **every row rendered a bare `@` with no active badge** (#1048's class, live on that tab). A `Destination` type now names what the API actually sends. Left alone, the destination you just added renders blank and the feature reads as broken.

## Behaviour change worth review attention

`create_destination`'s `provider_account_ref` is now optional. The four legal combinations live in one resolver (`_identity_for`) rather than in a condition, because a signature with two optional arguments cannot state which are legal:

| supplied | result |
|---|---|
| a reference key at all | `account_ref_from` decides — a malformed one is REFUSED, never quietly replaced by a derived handle. Any handle stored as given. |
| handle alone | derived `manual:<handle>`, handle normalised |
| neither | `account_ref_required`, unchanged |

The branch tests whether a reference was **supplied**, never whether it is well-formed: a shape test there silently derives `manual:<handle>` for an unquoted Meta id (an ordinary JSON mistake) while refusing a too-long string on the same path.

`POST /workspaces/{ws}/accounts` also stops coercing a blank handle to `None` before the port sees it. That coercion was a second copy of "what counts as a handle" and the copies already disagreed: `{"handle": "   "}` answered `account_ref_required` while `{"handle": "@"}` answered `handle_required` — one user error, two reasons.

## Verification — CI is mid-outage, so these are local runs, not claims

GitHub Actions is down; PRs are showing 3 checks where 8 belong (lint/test/security/changelog/fc2-ratchet **absent**, not failing). Everything below was run on this branch on the final tree.

| gate | result |
|---|---|
| `ruff check .` | All checks passed |
| `ruff format --check .` | 446 files already formatted |
| `tsc --noEmit` | rc 0 |
| `eslint` (the 7 files this PR touches) | rc 0, no output |
| `vitest run` | **137 passed**, 15 files (17 of them new in `destination.test.ts`) |
| `pytest tests/src/services/target/test_provisioning.py tests/scripts/test_provisioning_gate.py` | **91 passed** (gate is real Postgres) |
| full `pytest tests/` | **3868 passed, 8 skipped, 0 failed** — rc 0, zero `FAILED`/`ERROR` lines inside the verdict block, 24m22s |

The full-suite number is read the three ways this repo's recipe requires — the rc gate (0), the names scoped to the `short test summary info` block (empty), and the count line — because any one of them alone has a mode it cannot see. `REQUIRE_TEST_DATABASE=1` was set, so a database-backed test could not silently skip; I did not enumerate which 8 tests skipped.

Repo-wide `eslint` reports 3 errors + 1 warning. **All four are in files byte-identical to `origin/main`** (verified with `git diff --quiet origin/main -- <file>` per file) — pre-existing, and exactly what #1068/#1052 predict from a front end CI has never linted.

**Mutation-verified rather than asserted**, four times, each killed by exactly one test:

| mutation | killed by |
|---|---|
| drop `.casefold()` | `test_two_spellings_of_one_handle_are_ONE_destination` |
| empty the `manual:` prefix | `test_a_handle_alone_creates_a_scheduled_destination` + `…_findable…` |
| restore normalise-before-branch | `test_an_explicit_reference_is_not_gated_by_its_display_handle` |
| swap presence for shape | `test_a_malformed_reference_beside_a_handle_is_REFUSED_not_derived` |

## Review findings skipped on purpose, and filed

`/simplify`'s four lenses found two things worth doing that this PR should not do, because both mean editing files outside its diff:

- **#1093** — `_PROVISIONING_STATUS` is the one refusal map with no totality test, so the next reason ships as a 500. This PR added three rows to that dict and a comment restating the hazard, which is the bandaid tell.
- **#1094** — the browser POST-JSON prelude is four copies and has already diverged on the `reason` arm; `stubFetch` is three byte-identical copies.

One more, no code change: `src/repositories/instagram_account_repository.py` does a bare `lstrip("@")` in three places — silently **repairing** where `handle_from` **refuses**, with no trim and no length cap. Different tier, legacy; debt for whoever retires that repository.

## What this does NOT claim

The destination is created and scheduled. It does not publish: `publish_pipeline` is parked in production because `compose()` passes `media_fetch=None` (build-path W5b), and the Drive leg has no token minting (gdrive P5). This closes the destination gap, not the loop.


chrisrogers37 (commented):
**Approve**

The trap you asked me to check — confirmed avoided, from three independent angles, not just astrid's own framing.

**1. `connect_account` conflation — did not happen, and the codebase already had the fence up before this PR.** The route astrid extended (`POST /workspaces/{ws}/accounts` / `create_account` in `v1.py`) sits directly under a comment block that PRE-DATES this PR (zero diff hits on it — confirmed by diffing the file, not reading it in isolation): *"`connect_account` IS in the vocabulary and is deliberately still unbuilt: it is the CONNECTION (an OAuth grant through Meta), not the DESTINATION, and it is milestone 2."* That distinction was already load-bearing before astrid touched anything. And her own new file goes further rather than relying on that alone — `destination.ts`'s module docstring states it explicitly: *"it is not what `connect_account` connects — that command begins a Drive connect and the name collision is the defect #1089 records. This module is named for the noun so the confusion cannot be re-made from the client side."* Confirmed mechanically too: `command_executors.py` (where `connect_account`/`reconnect_account`/`disconnect_account` live) is untouched by this diff, and `create_destination`'s INSERT targets `ig_accounts` — I read the actual statement, not a description of it. Three separate confirmations, not one person's say-so.

**2. No-migration claim — checked against the DDL directly.** `ig_accounts` (054): `provider_account_ref TEXT NOT NULL`, `handle VARCHAR(50) NULL`, no credential FK anywhere in the table. `HANDLE_MAX = 50` is the column's real bound, not an invented policy. `uq_ig_account_live` matches the `ON CONFLICT (workspace_id, provider_account_ref) WHERE state <> 'moved'` clause character-for-character against the migration's own partial index — the kind of mismatch that would otherwise surface as a runtime "no unique or exclusion constraint" error, and it doesn't here.

**3. The safety-critical property — mutation-tested myself, not just re-read.** `_identity_for`'s docstring names a specific historical regression: an earlier revision normalised the handle before choosing a branch, so an OAuth-identity create with a merely-decorative malformed display handle got wrongly refused. I reintroduced exactly that shape in my own copy (call `handle_from` unconditionally before the branch) and ran the full provisioning suite: one test failed, `test_an_explicit_reference_is_not_gated_by_its_display_handle`, with `ProvisioningRefused: handle_malformed — no spaces` — the exact failure mode described. Reverted, clean at 92/92. This is real, working coverage on the property that matters most here, not a decorative test.

**Tests, run for real:** backend `test_provisioning.py` + `test_provisioning_gate.py` → 92 passed (one more than the PR body's stated 91 — I collected 92 twice, cleanly, no skips; a minor reporting discrepancy on astrid's side, not a correctness question, and worth naming since today's theme has been checking stated counts rather than repeating them). Frontend `destination.test.ts` → 17/17, matching the claimed new-test count exactly.

**The two departures beyond scope** (`_PROVISIONING_STATUS`'s new 400 mappings, the `AccountsResponse`/`Destination` type fix) are both real bugs that would have shipped alongside this feature otherwise — a bad handle 500ing, and every destination rendering a bare `@` — fixing them here rather than filing them separately is the right call given the feature is unusable without either. `Destination` vs. legacy `InstagramAccount` also lines up with something I confirmed independently in a different PR today: the setup wizard (`InstagramAccount`'s "one consumer left") is genuinely unreachable, so this type split isn't leaving orphaned code behind a live screen.

**#1093/#1094, correctly scoped out** — both are real, both are outside this diff's files, both are filed rather than silently absorbed. Right call.

Nothing here needs the Google-login blocker — pure code + DB-backed test review, no live session required.

Issue #1026, opened 2026-08-24
[monitoring] An 18-hour total outage produced no alert — the heartbeat is a positive signal and nothing watches for its absence


chrisrogers37:
## What happened

storydump was **fully down for roughly 18 hours** — the application could not read its own database — and **nothing alerted.** It was found by accident, while an engineer was investigating an unrelated question about which schema was serving.

Timeline (all measured, see the linked investigation):

| time | event |
|---|---|
| `02:19:17Z` Aug 23 | last `service_runs` row written — `MediaSyncService.sync`, `status=running`, `completed_at` NULL |
| `02:20:14Z` | migration `051` moves `public` → `legacy`, 57s into that sync — the worker dies mid-run and cannot log it, because `service_runs` moved too |
| `02:28:29Z` | first target-armed deploy |
| `~20:00Z` Aug 24 | outage discovered, by someone looking for something else |

## The gap

The heartbeat is a **positive** signal: `service_runs` gains a row roughly 73 times an hour. When the worker died, the signal simply stopped. **Nothing watches for its absence.**

That is the whole failure. A heartbeat nobody checks for silence is a log line, not a monitor.

## Two things that made it harder to notice

**1. `/health` cannot see this.** `src/api/app.py:219-226` returns `status`, `version`, and `uptime_seconds` computed from `time.time() - _START_TIME`. It opens no session and issues no query. So the process answered 200 throughout — and Railway's health gate passed — while the app could not read a single row. A liveness probe that touches no dependency will report healthy through any dependency failure.

**2. The failure mode includes a silent-empty shape.** Most queries against the moved schema fail loudly (`users` is missing 8 of its 11 legacy columns). But a query naming no legacy-only column — a bare `SELECT count(*) FROM users` — returns **0 rather than erroring**, against a table whose legacy twin holds real rows. So a naive "can I reach the database" check reads *empty*, not *broken*, and empty looks like a quiet day.

## What would have caught it

Stated as properties rather than a design, since the right mechanism is a judgement call:

- **Absence detection on the heartbeat.** If `service_runs` has gained no row in N minutes, that is an alert. The threshold falls out of the ~73/hr cadence.
- **A health check that exercises a dependency**, distinct from the liveness probe Railway gates deploys on. Those are different questions and one endpoint cannot answer both.
- **Any dependency check must name a column that would be missing under a schema fault** — otherwise it lands in the silent-empty case and reports healthy.

## Related

[...]

PR #1097, opened 2026-08-26
feat(target): detect a scheduling outage without a customer reporting it (#1090 F1)

chrisrogers37:
**#1090 F1** — *"an outage in scheduling raises an alert without a customer reporting it."* Not `Closes`: F1 is one line of a twelve-line checklist.

## The two outages, and the one thing they share

Both ran ~18 hours with no alert. Both were found by someone looking for something else. Their **mechanisms are opposite**:

| | #1026 | today |
|---|---|---|
| worker | **died** | **alive** |
| heartbeat | stopped | still being written |
| cause | schema moved under it | migration `065` never applied; `ck_jobs_kind` refused a kind the worker mints, so every tick aborted |

A **mechanism** detector — "did a tick run", "is the heartbeat alive" — is the natural answer to #1026, and it would have been **blind to today**: the worker was alive and still writing its heartbeat while doing no work.

## The signal: a due cursor that is not advancing

`fn_clock_tick` selects `state = 'active' AND next_slot_at <= now()` and then **advances** the cursor — and all three places in the whole schema that write `next_slot_at` are that same advance. So a cursor stuck in the past is proof the advance is not happening, **whatever stopped it**: dead worker, aborting tick, stuck lane, or a cause nobody has thought of.

Watching the outcome needs no guess about causes. That is the whole design.

## It raises nothing — the fleet path polls it

Served at `/health/scheduling`. Two reasons, and the second is measured rather than argued:

1. An alert whose **sending** is performed by the system it monitors cannot fire when that system is down — the same law that kept this detector off the job table, applied to the output side.
2. The app's own notification routing **has no writer**: nothing writes `channel_bindings`, for any workspace (navi). An alert delivered there would vanish silently.

Deliberately **not** `/health`: that is Railway's liveness gate and its own docstring says opening a connection there would take the service down for a blip no restart repairs. Two questions, two endpoints — #1026 asked for exactly this, and a test pins them apart so a later tidy-up cannot merge them.

## No migration, and that is forced

A system job is the tidier-looking implementation and would have been wrong twice: it needs a new `ck_jobs_kind` value (a migration, hard-blocked), **and it would share a failure mode with the thing it watches.** The proof is standing — `alert_stranded_sources`, the alerting job already built, lives in migration `065` and is **un-mintable on production at this moment**. Dead in exactly the situation it exists to detect.

## Bounds, in the module rather than here

- It **cannot see a failure past the mint** — cursors advancing while nothing posts is a different outage and wants its own signal.
- Its **cross-tenant read works today only because production runs as the owner role** (#751). Whoever closes that must give this a door, rather than discover it returning a reassuring zero.

## A test-scope argument, not a note about this file

I wrote the same order-dependent assertion **twice**. I caught the first before running. The second **passed when I ran the class alone** and failed only under the full gate, where a sibling rewinds a cursor and leaves a stalled row behind.

A module-scoped fixture makes any *absolute* assertion order-dependent, and **running the class rather than the suite is exactly the scope at which that stays invisible**. Both now assert deltas. A per-class run would have shipped it.

## Verification, and which checks I actually have

**3851 passed, 8 skipped, 0 failed** (23:37, local — CI is out on the Actions outage).

- **Summary-block scoping:** zero `FAILED`/`ERROR` inside the band. ✅
- **Count line:** present, well-formed. ✅
- **rc gate: unrecoverable.** The run outlived the session that launched it (`setsid`) and the status died with the process. The rc gate exists to catch rc 2/5/127, where no summary block is emitted; a complete block plus coverage output excludes that class — but that is check 2 doing check 1's job, and it is weaker. **An unrecoverable rc is not a passing rc.**

Mutation-checked: dropping the `state = 'active'` filter kills exactly the non-active test.

### A count discrepancy a reviewer should CHECK rather than inherit

This branch adds 7 tests, so against a ~3868 baseline I expected **~3875** and got **3851** — roughly **24 unaccounted for**, which is not a rounding wobble.

**Do not explain it before checking it is owed an explanation.** The 3868 figure came from a *sibling's branch* (#1089), not from this branch's merge-base. **Two different trees have two different test populations, and those numbers were never comparable.** The apples-to-apples check is this branch against its own merge-base on `main` — anything else compares two moving trees. Not run: the fleet is holding and nothing merges today.


chrisrogers37:
**Approve**

## The two named constraints, checked at the source, not the prose

**Outcome-over-mechanism:** `scheduling_lag`'s only query reads `FROM ig_accounts WHERE state = 'active'`, filtering on `next_slot_at`. Zero reference to a `jobs` row, a tick counter, or a heartbeat anywhere in the module. It reads the column the clock's *advance* writes, never the clock's own process state — that's the actual distinction that makes it mechanism-blind, and it holds.

**Alert path independence — the one you most wanted verified:** `scheduling_lag` performs a single `SELECT`, no `outbox.enqueue`, no `push_bindings`, no write of any kind. `/health/scheduling` (`app.py`) calls it and returns the dict — nothing else. Confirmed it does not touch the `jobs` table (never referenced), the clock (no import of `scheduler.Clock`, no dependency on `tick_once`/`self.ticks`), or the outbox (no `enqueue` call anywhere in the module). The design is genuinely pull-based: the endpoint answers a question, delivery is someone else's job. That's the right shape given this morning's #1090 finding, which the module's own docstring cites correctly.

## Independently mutation-tested both load-bearing filters against a real database

Ran the actual test class against a live local Postgres (`.env.test`), not just read it:

- **Baseline:** 4/4 pass, 100% line coverage on the new module.
- **Mutation 1 — dropped `state = 'active'`:** exactly 1 failure, `test_a_non_active_account_is_not_stalled`, exactly as the PR claims. Confirmed independently.
- **Mutation 2 — my own, not in the PR body:** flipped `next_slot_at <= now()` to `> now()` in both `FILTER` clauses (the actual "due cursor" comparison — the core thesis of the whole PR). **3 of 4 tests failed**, including the positive control (`test_a_fresh_destination_is_not_stalled`) and the core stalled-detection test. The one survivor (`accounts_active` distinguishing test) is correctly independent of due-ness by design. Both the filter *and* the due-comparison are genuinely pinned, not just the one the PR called out.

Both mutations restored, working tree confirmed clean before and after.

## Design details that are right, not just adequate

- `/health` vs `/health/scheduling` are pinned as separate endpoints by a dedicated test class, including a test that `/health` still returns 200 with `app.state.engine is None` — the actual property that makes the split necessary, not just "two routes exist."
- 503 (not a cheerful zero) when the engine is unreachable — the same unreachable-vs-empty distinction I'd flag as missing if it weren't here.
- `accounts_active` in the response separates "healthy" from "unmonitorable" on a zero-account estate — this preempts exactly the ambiguous-zero class of bug I'd otherwise be looking for.
- The disclosed self-correction (`paused` vs `disabled` — assumed shared vocabulary with `media_sources`, caught by running the test rather than reading the schema) is the right instinct honestly reported, not hidden.

## One thing worth confirming, not a defect in this PR

The module is explicit and honest that it decides nothing: *"this module answers a question and an external poller decides whether the answer is an alarm."* That's the correct boundary for this diff — but it means merging #1097 alone ships the **detection surface**, not the closed loop. Nothing in this PR's ~315 lines wires a poller to `/health/scheduling` or fires a fleet alert on it. Given the framing that this is "the one that would have caught today," worth a direct confirmation that the polling half is tracked (fleet-side host job, presumably) rather than assumed — an unpolled health endpoint reports correctly and alerts nobody, which is a quieter version of the same failure this PR exists to close.

## Respecting the instruction on the count discrepancy

Saw the 3851/3868 note and the branch-vs-branch explanation. Not chasing it, per your call.

## Verification

Ran the real integration suite against a live local Postgres, not just `tsc`/lint. `git diff --stat` clean before and after both mutations. CHANGELOG entry present and accurate against the diff.

PR #1101, opened 2026-08-26
ci: the front end gets a gate — 138 tests that had never run in CI (#1052/#1068)

### LINKED ISSUE #1052: CI runs nothing for landing/ — the front end's only gate is a developer's local machine
Surfaced repeatedly today, most recently by astrid while designing P11 (#1033). Filing so it stops being a footnote in other people's PRs.

## The gap

**No CI job runs anything under `landing/`.** Not `tsc`, not `vitest`, not `next build`.

`vitest` was added to `landing/` in #1023. CI still does not invoke it. So the tests exist and nothing runs them.

The only gate on front-end code reaching `main` is Vercel's own `next build`, which catches compile errors and nothing else — no type errors surfaced separately, no unit tests, no lint.

## Why it matters now rather than eventually

The front end is no longer a thin shell. Today alone it took:

- #1032 — 54 files, the entire target-schema front end
- #1047 — the session-cookie handoff
- #1051 — 19 files of shape migration and repoint
- #1033 (in flight) — the web queue, the surface that makes manual publishing possible at all

Every one of those was gated solely on the author running things locally and saying so. That has held up because the authors have been careful — #1047 and #1051 both shipped with discriminating controls — but "the author was rigorous" is not a control, it is a hope about the next author.

It has already cost us once: #1032 shipped with a session-cookie gap (every writer deleted, the reader kept) and a dangling Upload affordance (#1050) pointing at a route that does not exist. Neither is the kind of thing a `next build` can see; both are the kind of thing a test can.

## Shape of the fix

A `landing-ci` workflow on `landing/**` running, at minimum:

1. `tsc --noEmit`
2. `vitest run`
3. `next build`

Path-filtered, so it does not fire on backend-only PRs — consistent with how this repo already scopes workflows.

**Note for whoever picks this up:** adding a path-filtered required check interacts with the conflicting-PR failure mode — a PR with conflicts never schedules its `pull_request` workflows at all, so an absent check must be read as absent rather than passing. Verify required checks by name, never by "nothing is red".

## Provenance

- astrid, P11 design comment on #1033: *"the only front-end gate is Vercel's `next build`"* and *"CI runs nothing for `landing/`"*
- Related: #1050, #1048

chrisrogers37:
Closes #1052

*(Keyword is in the branch commit too, which is what fires on squash.)*

Built as scoped in [the proposal on #1068](https://github.com/chrisrogers37/storydump/issues/1068#issuecomment-5431780027).

## The gap was invocation, not authorship

`ci.yml` had **zero** references to node, npm, `landing/`, vitest, tsc or eslint. Its five jobs are ruff, a Python ratchet, pytest, pip-audit/bandit, and a git changelog check. Nothing checked out Node.

The tests were not missing: **138 across 15 files, all passing.** They had simply never run in CI.

## The job

New `Front End` job: `npm ci` → `npm test` → `npx tsc --noEmit` → `npm run lint`. It uses the package's own scripts where they exist, so a change to what `test` or `lint` means follows without editing the workflow.

## Two choices that are load-bearing, not defaults

**No `paths:` filter.** This repo path-filters nothing today. A landing-only filter would introduce the auto-merge trap here *for the first time*: a filtered workflow is never **scheduled** on a PR touching no front-end files, so a check that is also required can never arrive and the PR strands forever. Unfiltered keeps the check always **present**, which is what makes it safe to require later.

**Not required.** `rules/branches/main` declares no `required_status_checks` at all — every check on this repo is advisory today. Making this the sole required one would be an inconsistency, and enforcement is a policy decision about all eight.

`next build` is deliberately absent: Vercel already builds every PR.

## The cost is zero wall-clock

Measured against real durations on a recent PR head:

| job | actual |
|---|---|
| **Test** | **335s** ← long pole |
| Security Scan | 31s |
| Lint | 11s |
| Changelog Check | 10s |
| FC-2 ratchet | 6s |

This job is **~176s worst case** — `npm ci` 124s + tests 9s + tsc 14s + lint 29s, measured on a slow machine under load, so an upper bound. It finishes minutes before `Test`. A gate that costs nothing is a gate nobody routes around, which is the only reason gates die.

## Prerequisite folded in rather than shipped separately

`eslint` exited **rc 1** on four pre-existing problems in files nobody had touched: two unescaped apostrophes, an `<a>` to an in-app route where `<Link>` belongs, and an unused import.

Bundled because the gate is then green on its first run **and is verified by the very job that needs them**. Shipping them alone would have left them ungated — the gate did not exist yet.

`engines: node >=22` added to `landing/package.json` so the pinned CI version and a laptop agree rather than drifting.

## What this does NOT do — and why #1068 stays open

13 of the 15 test files are `src/lib`, and vitest runs `environment: "node"`. **These are logic tests.**

They would not have caught the defects actually found this week: a hero asserting something false (#1095), a header overflowing at 390px (#1095), a card rendering from a dead code path, a control wired to a route that does not exist. Every one needed rendering.

So this closes **regression-in-logic**. It does not make the front end covered. #1052 says *"CI runs nothing for landing/"* — that is now false, so it closes. #1068 says *"the entire front end is uncovered"* — that is still true in the sense that matters, so it does not.

**Anyone reading this job's green tick as "the front end is tested" is making the exact mistake the tick exists to prevent.**

## Verification

The exact job commands, run from `landing/`:

```
npm test          rc=0   138 passed
npx tsc --noEmit  rc=0   no output
npm run lint      rc=0   no problems
```

And this PR is its own proof — the `Front End` check below is the job's first real run.


chrisrogers37 (commented):

Went through your four scrutiny points in order, verifying from primary sources rather than the PR's own narrative.

**1. Does it actually run the 138 tests?** Yes, confirmed three independent ways:
- Pulled the raw Actions log directly via `gh api .../actions/jobs/98347028256/logs` (not the PR body) — all 15 files enumerated individually with per-file counts, summing to 138, `Tests 138 passed (138)`, zero skipped.
- Reproduced locally: fresh worktree off `virgil/1068-front-end-ci`, real `npm ci`, `npm test` → identical `15 passed (15)` / `138 passed (138)`.
- Mutation-tested it: broke `GoogleLoginButton`'s href (`/auth/google` → `/auth/google-broken`), reran — got exactly `2 failed | 136 passed`, both failures the two assertions that actually check the path (`google-login-button.test.tsx:33` and `:46`). Reverted, reconfirmed `138 passed (138)`, then `tsc`/`lint` unpiped (avoiding the `PIPESTATUS` trap) — both genuinely `rc=0`.

This is the opposite shape of #1097 — the assertions execute and a real regression gets killed, not waved through.

**2. Is it correctly not required?** Confirmed independently via `gh api repos/chrisrogers37/storydump/rules/branches/main` — three rule types exist on `main` (`deletion`, `non_fast_forward`, `pull_request`); no `required_status_checks` rule at all. Also checked `ci.yml` directly: no `paths:` filter anywhere (`on:` is branches-only for push/PR) and the `front-end` job carries no `if:` — the only two `if:` conditionals in the file belong to an artifact-upload step and the pre-existing changelog-check job. A backend-only PR schedules and completes this job same as any other.

**3. Does it duplicate Vercel?** `next build`/`next start` genuinely absent from the job steps (diff-confirmed: checkout → setup-node → `npm ci` → `npm test` → `tsc --noEmit` → lint, nothing else). The PR states the Vercel-already-builds premise plainly without calling it measured, which matches — Vercel's own check is SUCCESS on this PR, so the premise holds today even though a deliberate-break confirmation wasn't run. Correctly left as inference.

**4. Is "closes #1052 only" honest?** Read the two new `.tsx` test files directly rather than trusting the summary. Both call the component function directly and walk the returned element tree — no `render()`, no DOM — which is why `environment: "node"` in `vitest.config.ts` stays true rather than becoming a stale label. So "these are logic tests, not rendering" holds up under inspection, not just assertion. The PR's own "what this does NOT do" section names your same four defects (false hero, 390px overflow, dead code path, nonexistent-route control) and states the green-tick misreading directly. Nothing here overclaims scope.

Incidental: the `<a>`→`<Link>` eslint fix in the blog page needed no new import — `Link` was already imported at the top of that file.

Clean baseline reproduced independently: 138/138 tests, tsc clean, lint clean, worktree torn down after.

PR #1102, opened 2026-08-26
fix(api): the scheduling detector 500'd in production and always had (#1100)

chrisrogers37:
Closes #1100.

## The finding

**Two tests named for `/health/scheduling` never execute its main path.**

Both build `create_app(env={})`, where `app.state.engine is None`, so both return
at the **503 branch — one line above the code that runs in production**:

- `test_it_is_not_the_railway_probe` — asserts the path is registered.
- `test_it_refuses_rather_than_reassures_when_it_cannot_look` — asserts the 503.

The line below that branch raised unconditionally, so the route returned **HTTP
500 to every caller it ever had**, from the day it deployed, while every signal
about it was green.

That is not a review miss. #1097 passed eight checks, was reviewed carefully, and
its query was mutation-tested against a real database — including a polarity
mutation the author had not listed. All of it was true, and true about code that
does not run. The query is not what production executes; the route is.

"The endpoint 500s" is a bug and it is one line. "Two tests named for a route both
exit before reaching it" is the thing that will happen again.

## What changed

**`src/api/app.py`** — the route opened its read with `unit_of_work(engine, "",
…)`. `UnitOfWork.__init__` refuses a blank tenant **at construction**, before any
database work. It now takes a direct connection, following `principal.py:84`
(a pre-tenant read on an API route) and `publish_pipeline.py:564`.

The guard is right and does not move. An estate-wide aggregate has no tenant, and
passing `""` was an attempt to satisfy the tenancy discipline by naming a tenant
that does not exist. `unit_of_work` was imported into `app.py` for this one call
site and no other — the broken line was the outlier in its own file, which only
surfaced because removing it left the import unused.

**`tests/src/api/test_app_factory.py`** — a test that takes the other branch. It
needs no real database, and that is the point: the defect raised before any SQL,
so **any** non-`None` engine would have caught it, and none of the coverage
supplied one. The existing `client` fixture already provides exactly that.

## Verification

Both directions, with the revert confirmed real (28 lines differ) before running,
so the run cannot be a silent no-op:

| run | result |
|---|---|
| new test against the **unfixed** route | **1 failed** — with the exact production exception, `TenantContextRequired` |
| full file with the fix | **21 passed** |

The failing case reproduces the production defect in CI rather than approximating
it.

Production traceback this reproduces, verbatim from the API service log:

```
  File "/app/src/api/app.py", line 386, in scheduling_health_check
    raise TenantContextRequired(
src.services.target.unit_of_work.TenantContextRequired: a unit of work requires a
tenant id — `02` §7 RLS reads app.tenant_id ...
```

## The RLS dependency is now measured, not inherited

`scheduling_health`'s docstring asserted that production bypasses RLS as
`neondb_owner`, citing #751. That was inherited prose, and this fix depends on it,
so it was checked:

- the API's `TARGET_DATABASE_URL` user is **`neondb_owner`**;
- `058_rls_and_policies.sql` enables RLS on `ig_accounts` and scopes `p_tenant`
  `TO svc_ingress, svc_worker`;
- **no** migration sets `FORCE ROW LEVEL SECURITY`, and none reassigns the owner.

So the owner bypasses the policy and the estate-wide read answers. **No migration.**

**The failure direction is invisible, which changes the acceptance gate.** Under a
role `p_tenant` *does* cover, a tenant-less read returns zero rows — and zero rows
here reads as a healthy estate. A `200` carrying `accounts_active: 0` cannot
distinguish a working read from an RLS-filtered one. So the production check after
merge is **`200` with `accounts_active > 0`**, not a bare `200`.

## Not in scope

The poller (#1099) is next and is deliberately separate — a poller cannot be
verified against a 500, and fixing this first gives it a real first reading to
calibrate its threshold against.


chrisrogers37 (commented):
**Approve**

## The one thing that mattered most: does the new test actually reach the defect?

Verified myself, both directions, rather than trusting the PR's own report.

Reverted the fix (restored `unit_of_work(engine, "", ...)` + the import), confirmed the revert was real (`git diff --stat` showed 28 lines — matches the PR's own claim exactly, so this wasn't a silent no-op), then ran `TestSchedulingHealthIsASecondSurface`:

```
FAILED ...test_it_reaches_the_query_when_an_engine_is_present
src.services.target.unit_of_work.TenantContextRequired: a unit of work requires
a tenant id — `02` §7 RLS reads app.tenant_id, so a tenant-less transaction
would be a widened query. System operations go through the §7 doors (L.0/#857).
1 failed, 3 passed
```

Exact production exception, verbatim message, and it's the *only* failure — the two pre-existing 503-branch tests (`test_it_is_not_the_railway_probe`, `test_it_refuses_rather_than_reassures_when_it_cannot_look`) stay green under the mutation, confirming they really are structurally decoupled from this line, which is the whole thesis of #1100. Restored the file (md5 matches original, `git diff` clean), re-ran: `4 passed`. The new test genuinely executes the main path and genuinely catches the defect it claims to catch — not asserted, reproduced.

## The fix itself, checked from source

- `UnitOfWork.__init__` (`unit_of_work.py:210`) raises `TenantContextRequired` **at construction**, before any DB I/O — confirmed by reading it directly, not inferred from the traceback.
- The replacement (`engine.connect()`, no tenant) follows two **real** precedents, both confirmed at the cited lines: `principal.py:84` (`engine.begin()`, a pre-tenant session lookup on an API route) and `publish_pipeline.py:564` (`engine.connect()`, a permit tx that must commit before a provider call). This is an established pattern in this codebase, not an invented one-off.
- `scheduling_lag(executor)` calls only `executor.execute(text(...))` once — confirmed by re-reading the function. Both `AsyncSession` and `AsyncConnection` satisfy that, so the duck-typing claim holds structurally, matching what the new test pins (`assert hasattr(seen[0], "execute")`).

## The RLS-bypass dependency — independently re-derived, not taken on the PR's word

The PR says this was "inherited prose" now measured. I re-measured it myself from source rather than trusting the PR's own measurement:

- `058_rls_and_policies.sql:266` — `CREATE POLICY p_tenant ON ig_accounts FOR ALL TO svc_ingress, svc_worker` — the policy is scoped to those two roles only; `neondb_owner` isn't named.
- Grepped all 65+ migrations for `FORCE ROW LEVEL SECURITY`, `OWNER TO`, and `SET ROLE`/`SET SESSION AUTHORIZATION` — zero hits anywhere. No migration ever forces RLS onto the owner and none ever reassigns or switches the creating role before `CREATE TABLE`.

Two independent reasons the estate-wide read answers rather than silently returning zero rows: the policy's `TO`-list doesn't cover the connecting role, and standard Postgres owner-exemption applies on top of that. Neither depends on the other, and both are confirmed rather than inherited.

## `accounts_active > 0` as the gate — expressible, and honestly *not yet* checked

Grepped every reference to `accounts_active` in the repo. It's a real top-level key in the response payload (both `scheduling_health.py`'s return dict and the new test's `resp.json() ==` assertion pin it), so the gate is expressible today. But nothing in this diff or anywhere else in the repo automates checking it — no script, no CI step, no deploy hook. That matches the PR's own scoping (#1100's checklist still has that line unchecked; the poller is explicitly deferred to #1099) rather than overclaiming a check that doesn't exist yet. Whoever runs the post-merge verification needs to look at the JSON body, not just the status code.

## Adjacent blast radius

- `tests/src/api/` in full: **105 passed**, so nothing nearby regressed.
- Grepped for `from src.api.app import` — only `create_app`, `app`, and `DropAmbiguousForwardedForMiddleware` are imported elsewhere; nothing re-exports `unit_of_work` through `app.py`'s namespace, so trimming it from the import list is safe.
- CHANGELOG entry confirmed under `## [Unreleased]`, not a stale released section.

## What I would not treat as green

**GitHub Actions never ran on this branch** — `gh run list` returns zero rows, and the status rollup shows only GitGuardian + two Vercel checks. That's consistent with the fleet's known Actions-outage pattern (checks go *missing*, not failing) and with the "still at 10%" note — I'm not reading the absence as a signal either way, but it does mean CI isn't available as independent corroboration right now, which is why the mutation test above was run locally rather than cited from the PR.

**One real, minor, non-blocking nit CI would have caught**: `ruff format --check .` (the same unpinned command `.github/workflows/ci.yml` runs) fails on the new test — `test_it_reaches_the_query_when_an_engine_is_present(self, client, monkeypatch):` needs its signature wrapped. `ruff check` (lint) is clean. One `ruff format` run fixes it; not a reason to hold this.

Not merging — leaving that to Chris.

PR #1105, opened 2026-08-26
feat(ops): the scheduling detector gets a caller, and NO SIGNAL is its primary state (#1099)

chrisrogers37:
Closes #1099.

`/health/scheduling` reported and nothing polled it, so the outage class sat
exactly where it started: **a detector nobody calls raises no alert either.**
This is the caller.

It runs on the **fleet host**, so the alert path shares nothing with its subject
— different machine, process, network path, clock and notification channel. It
touches none of the jobs table, `ck_jobs_kind`, `fn_clock_tick`, the outbox,
`channel_bindings`, the worker, or the app's own notification routing, which has
no writer at all.

## The load-bearing decision is not the threshold

While verifying #1102 on production I measured the estate: **`ig_accounts` has no
rows at all.** `workspaces` 0, `media_sources` 0, legacy tables absent entirely.
The tier is live — `jobs` carried 27 rows and the clock was minting — with **zero
destinations**.

`scheduling_lag` reads only `ig_accounts WHERE state = 'active'`, so `stalled` is
0 and `max_lag_seconds` is null **structurally** — not because scheduling is
healthy, but because there is no population that could be stalled. A poller
reporting "nothing is stalled" from that is a green light wired to nothing, and
it would have stayed green through the entire 19-hour outage had the estate been
empty then.

So the distinction this is built around:

- **nothing is late** → genuinely healthy
- **nothing exists to be late** → `no-signal`, and it says so

Those have opposite remedies and must never render the same. `no-signal` never
pages — an empty estate is expected right now, not a fault — and is never counted
as fine either.

| reading | state | behaviour |
|---|---|---|
| `200`, `accounts_active == 0` | **`no-signal`** | says so once, re-states weekly. Never an alert, never an all-clear. |
| `200`, active, lag ≤ 600 or null | `healthy` | quiet; announces RECOVERED / SIGNAL ACQUIRED on entry |
| `200`, active, lag > 600 | **`stalled`** | FLEET ALERT on the **first** reading, repeats 6-hourly |
| non-200 / timeout / malformed | **`unreachable`** | FLEET ALERT on the **second consecutive**, repeats 6-hourly |

**Confirmation is asymmetric because the signals are.** A lag past the threshold
already *contains* its duration — 600s of non-advance is what the number means —
so the observation is the sustain and confirming it only adds delay. A failed
request contains no duration at all: one is a dropped packet.

Threshold 600s = 40× the clock cadence (`clock_interval_seconds = 15.0`) and
1/114th of the shorter outage; at a 5-minute poll, worst-case detection is ~15
minutes against 18–19 hours.

## A defect this module's own test found

The first version tracked one field: the last state observed. A failed notify
still advanced it, so the next poll saw "same as last time" and stayed quiet
about a message **nobody received**. What a human was *told* is now tracked
separately from what the endpoint *said*.

The `no-signal` case was caught by the test I wrote for it. The **`RECOVERED`
case was not** — the mutation `announced = prior.get("state")` survived a full
green run until `test_a_failed_recovery_notice_is_retried_on_the_next_poll` was
added. A surviving mutant on the exact line the fix lives on is the reason that
test exists.

## Verification

**Mutation checks — five, each confirmed to actually apply before the run:**

| mutation | result |
|---|---|
| `no-signal` reads as healthy | **7 failed** |
| lenient payload reader (`.get("stalled", 0)`) | **6 failed** |
| `unreachable` fires on the first reading | **2 failed** |
| threshold off-by-one (`>` → `>=`) | **1 failed** |
| `announced` collapses back into `state` | **1 failed** — *survived until the test above was written* |

**Driven for real, not just unit-tested** — on system `python3` with no venv,
which is the actual deployment condition:

```
$ python3 scripts/scheduling_monitor.py --url https://<api>/health/scheduling ...
no-signal: no active destinations exist, so nothing could be stalled
storydump scheduling: NO SIGNAL — ... CANNOT detect an outage right now.   exit=10
$ (again)
no-signal: no active destinations exist, so nothing could be stalled          exit=0
```

And the `unreachable` path against a genuinely refused port — quiet on the first,
alerting on the second:

```
run 1 exit=0   unreachable: URLError: <urlopen error [Errno 111] Connection refused>
run 2 exit=10  FLEET ALERT: ... UNREACHABLE (...) on 2 consecutive polls.
```

Suites: `tests/scripts/test_scheduling_monitor.py` + `tests/src/api/` → **145
passed**.

## Bound: this cannot be validated against real traffic yet

Nobody has run the target tier end to end — sign-in needs a real Google account,
and production has no workspaces and no destinations. Every path is exercised
against captured payloads, mutation checks, and the live endpoint's `no-signal`
answer, but **the `stalled` path has never seen a real stalled cursor.**

That is a bound on this work, not a defect in it, and it is stated here so a
later reader does not mistake *tested* for *seen in production*. The first real
destination is what turns this from a monitor that is correct into one that is
proven.

## Deployment

Stdlib only, no venv. `documentation/operations/scheduling-monitor.md` carries
the systemd unit and timer, the exit-code contract (`0` quiet / `10` spoke / `11`
could not deliver — deliberately a failed unit), and the **operational
dependency**: a pinned checkout used by nothing else. Not a bot's working tree —
a `git checkout -b` there would silently break production monitoring.


chrisrogers37 (commented):

Went through your four points in order. I'd not seen this endpoint before this review, so I read `scheduling_health.py` on `main` first to get the actual contract, not just the poller's side of it.

**1. The no-signal ruling.** Confirmed in code: `classify()` checks `active == 0` first and returns `NO_SIGNAL` unconditionally, before the stalled/healthy branch runs at all. It's also structurally impossible for the two to collide — `stalled` and `max_lag_seconds` in `scheduling_lag` are both filtered over the *same* `state='active'` population, so `accounts_active == 0` forces `stalled == 0` and `lag IS NULL` at the SQL level, not just by the poller's own discipline. I curled prod myself just now: `{"stalled":0,"accounts_active":0,"max_lag_seconds":null}`, HTTP 200 — this is genuinely live, not a hypothetical. Then ran the actual script, system `python3`, no venv, against that real URL: spoke once (`exit=10`, the NO SIGNAL message), silent on immediate rerun (`exit=0`), `--status` showed the persisted state correctly. One precision note, not a defect: the STALLED branch keys on `lag > threshold_s`, not raw `stalled > 0` — and that's correct, per `scheduling_health.py`'s own docstring ("a nonzero lag is normal and only a SUSTAINED one is an outage"). Worth being precise about since your dispatch phrased it as `stalled>0 alerts`.

**2. Isolation from jobs/clock/outbox.** Checked rather than assumed: the entire module's import list is `argparse, json, os, subprocess, sys, tempfile, time, urllib.error, urllib.request` — stdlib only, zero imports from `src/`, no DB driver, no ORM, nothing app-side. The test file matches (stdlib + pytest + the module itself, no `conftest.py` DB fixtures touched). Total isolation, not just claimed.

**3. The surviving mutant.** Rebuilt it myself rather than trusting the account: changed `decide()`'s `announced = prior.get("announced")` to `prior.get("state")` (line 194) — this recreates exactly the bug described, since `state` advances on every poll while `announced` only advances on confirmed delivery. Ran the full file against it: **1 failed, 39 passed**, and the one failure was exactly `test_a_failed_recovery_notice_is_retried_on_the_next_poll`. Nothing else in the other 39 catches it. Reverted, confirmed clean 40/40.

**4. Bound placement.** Found it in all four: the PR body has its own `## Bound:` heading, the module docstring has its own `## Bound, stated because it will not be obvious later` section (first thing a source reader hits), the CHANGELOG entry carries it as a bolded clause rather than folding it into unrelated prose, and `documentation/operations/scheduling-monitor.md` — the doc an on-call human actually opens — closes on its own `## Bound:` section rather than a footnote up top. None of the four require hunting.

Also, since I had eyes on it: the "Test job, 15 steps" claim checks out (`gh api .../actions/jobs/98356873599`, 15 step rows, `Run tests with coverage` succeeded). The 3929/3889/+40 delta is structurally sound — this diff adds exactly one test file, it collects 40 items, and no other test file is touched — though I didn't re-run the full ~3900-test suite myself to confirm the absolute baseline.

Nothing here needs to change.

Issue #871, opened 2026-08-19
[Refactor Phase S] S.3 carry-in — pin the SSRF-validated address to the connection (TOCTOU/DNS-rebind, deferred from L.0 #857)


chrisrogers37:
Deferred here from **L.0 (#857 / PR #870)** during review, deliberately and visibly rather than carried silently.

## The gap

`src/services/target/egress.py::validate_target` resolves the target host and checks **every** returned address against the private/link-local/reserved classes. Then `httpx` performs its **own independent resolution** when it opens the connection, and nothing pins the validated addresses to that connection.

So the address-class check proves *"this hostname resolved to a public address at that moment"* — **not** *"this connection went to a public address"*. That is a TOCTOU / DNS-rebind window.

Verified against httpx 0.28.1 rather than assumed: the transport passes `host=request.url.raw_host` to the connection pool and exposes **no resolver hook**. Confirmed independently by navi, who also grepped every `httpx.AsyncClient(` construction site in `src/` and found no mitigating custom transport.

## Why it was deferred rather than fixed at L.0

- **Cost is materially larger than the other three floor findings.** Closing it needs a custom transport that resolves, pins the IP, and still gets TLS SNI and certificate verification right. Getting that wrong — breaking SNI, or silently weakening verification — is worse than the gap.
- **Exploitability today is low and bounded.** The allowlist is six closed, high-reputation provider hostnames (Meta / Telegram / Google). An attacker would need DNS control over one of *those* domains, not an injectable target.
- **S.3 already charters this work**: `04` §S.3 lists "SSRF-safe streaming" on top of the L.0 floor and "the full hostile-fake battery".

## What holds in the meantime, and what must not change

**The host allowlist is the load-bearing control, not the address check.** That is now stated in the module docstring rather than left implied — the previous wording ("SSRF resolution refuses first and asks nothing") claimed more than was mechanically true.

**Therefore: do not widen the allowlist before this lands.** Admitting any less-trusted host moves this from theoretical to reachable. That constraint is the real cost of the deferral and is why it is written down here.

The address-class check stays regardless — it is genuine defence in depth against a misconfigured or compromised-at-rest allowlist entry, and it costs nothing.

## Done requires

1. The connection reaches a pinned, pre-validated address — not a re-resolution of the hostname.
2. TLS SNI and certificate verification unchanged (a test that a valid provider host still verifies).
[...]

PR #1106, opened 2026-08-26
feat(target): arm the email channel — the out-of-band surface a web-born workspace never had (#1092)

chrisrogers37:
Arms the `07` §1 email channel — mason's build step 2 in #1092, and **only** that.

**No `Closes` keyword, deliberately.** #1092 scopes three things; this is one of them. The state renders (`state` on the accounts payload, the `review_required` queue view) are a separate change. Per the standing rule, a partial fix says nothing rather than letting a keyword close an issue it only partly addresses — hand-close if you disagree.

## Why this is the gap

`06` §3 routes notifications to "the workspace's bindings". **Nothing in the tier writes `channel_bindings`**, so for the default workspace shape every notification we produce lands nowhere. Email is the channel the plan already chose, and the address already exists: `users.primary_email` is filled from the *verified* Google claim at every sign-in.

**No migration.** `send_email` has been in `ck_jobs_kind` and `ck_jobs_system_kinds` since 056; `email_global` has been a `rate_counters` scope since 056. Code only — it does not join any queue.

## Three decisions worth review attention

**1. The port does not know about jobs.** `ResendSender` is a plain callable. A caller that must not depend on the jobs table — operator alerting for an outage that has broken the jobs table, #1099's constraint (2) — can call `send()` directly and synchronously. The executor is *one caller* of the port, not the port itself. That is why this is two objects, and it is the seam split agreed before building: **alex owns operator outage alerting and must reach the transport without the jobs table; this owns customer-facing delivery, which rides the jobs table and should.**

**2. It ships inert, never as a silent no-op.** `07` §1's owner ack on adding Resend is OPEN and no credential exists anywhere (nothing in `settings.py`, nothing in `.env.example`, none of the worker's Railway vars). So `sender_from_env` returns None and the registry parks the kind with a reason naming exactly what is missing. A sender that accepted sends and dropped them would report success for mail nobody received — a visible gap converted into an invisible one. **The day Chris acks, this is a config change, not a build.**

What an operator can read today is the kind's *absence* from the worker's `live_kinds` line. An earlier draft of this PR, the module docstring and the CHANGELOG all said the parked *reason* is visible there; it is not — that line lists kinds with no parked reason, and the reason string first reaches a log only when a job of the kind is claimed, which for a kind with no producer is never. Corrected in all three places.

**3. Egress reach is narrowed, not widened.** The module carries its own one-host `EgressPolicy` rather than adding the provider to `DEFAULT_ALLOWED_HOSTS`, whose own comment calls it the load-bearing SSRF control until #871. Pinned by a test asserting the host is absent from the shared set.

## Two constraints the real-database gate caught before runtime could

Both changed the design rather than the tests, which is the reason to mention them:

- **`egress.request` refuses any provider call inside an open transaction** (`02` §5, enforced in code). The executor therefore owns its own transactions and the loop's session stays purely the finalization context — the shape `publish_pipeline` already uses. Honest consequence, stated rather than left to be discovered: **the budget debit commits before the send, so a failed send costs a slot.** Conservative direction (the ceiling exists to stay under a provider pause), but a cost.
- **`unit_of_work()` refuses a tenant-less unit on purpose** — RLS reads `app.tenant_id`, so a tenant-less unit would be a widened query — while a `send_email` row genuinely has no tenant: system kind under `02` §5's classing rule, `workspace_id` NULL by biconditional. `system_session` applies the GUCs directly with the empty tenant id, the convention `worker.make_session_for` already uses for exactly those rows.

## Budget

`05`'s: **90/day** on `email_global` key `''`, headroom under a free tier that PAUSES at 100/day. Over budget the job **defers** on the 1/5/15 ladder with the attempt **restored** — a daily ceiling would otherwise exhaust three rungs measured in minutes and drop the mail entirely.

## Verification

| gate | result |
|---|---|
| full `pytest tests/` | **3922 passed, 8 skipped, 0 failed** — rc 0, zero `FAILED`/`ERROR` inside the verdict block, 13m18s |
| `tests/scripts/test_email_gate.py` (real Postgres) | 5 passed |
| `tests/src/services/target/test_email_sender.py` | 25 passed |
| registry + composition gates | 69 passed together on the rebased tree |
| `ruff check` / `ruff format --check` | clean / 450 files formatted |

**Mutation-verified**, four times, each killed by exactly one test: dropping `restore_attempt=True`; bypassing the budget check; accepting a half-configured sender; widening the module policy to the shared default.

## Review findings skipped on purpose, and filed

- **#1103** — a deferred job is counted as `fenced` and logged as "another owner won". Pre-existing at four sites in `publish_pipeline`; this adds the fifth. It corrupts the number an operator reads to answer "is the reaper racing my workers", and the fix has to cover the pipeline to be worth doing.
- **#1104** — the system-session stack is written five times and the gates' `_owner_exec` four; both belong in one home.

One note needing no issue: `egress.request` binds its `resolver` default at import time, so it cannot be monkeypatched — which is why unit tests of egress callers do live DNS. This change sidesteps it with the `policy` seam, and `test_google_drive_adapter` still pays the cost.

## What this does NOT claim

Nothing here produces an email. There is no invitation producer, no D3 producer, and no D4 timeliness push — those are #1092's remaining scope and separate changes. This makes the channel exist and drain; it does not make anything send.


chrisrogers37 (commented):
**Approve**

Four scrutiny points, each checked from source and reproduced empirically where possible rather than taken on the PR's word.

## 1. Does it actually deliver? — Yes, verified at two layers, plus one real finding

**Wire level** (`test_email_sender.py`, `httpx.MockTransport`): `test_a_send_returns_the_provider_ref` captures the actual constructed request and asserts on it — URL prefix, `Authorization: Bearer re_test_key`, and the literal recipient string inside the JSON body. `test_the_floor_refuses_a_host_this_module_does_not_declare` temporarily repoints `SEND_URL` at an evil host and confirms the SSRF floor's host allowlist actually fires. This is not a "mock returns 200, trust it" test — it's asserting on what actually got sent, which is exactly the class of check #1102 needed and didn't have.

**Real-DB level** (`test_email_gate.py`): ran it myself against real Postgres. `test_a_send_debits_the_budget_and_returns_the_provider_ref` seeds a genuine leased `send_email` row and drives `execute_send_email` through a real engine connected as `svc_worker` — confirms the render → budget-debit → transport chain end to end from an actual job row.

**The one real finding — a timezone-dependent test, confirmed and isolated, not a shipped-code bug.** Running `TestTheBudget` locally, baseline (no mutation), I got **2 of 3 failing**: a send went through when it should have deferred, and a pre-seeded budget counter read back as 1 instead of 90. Traced it precisely: the test seeds `rate_counters` via `date_trunc('day', now())`, which is evaluated inside the Postgres session and therefore depends on that session's `TimeZone` GUC. The actual code computes its window boundary via `rate_counters.window_start()` — pure Python, UTC-epoch arithmetic, session-independent. My local Postgres defaults to `America/New_York`; measured directly, `date_trunc('day', now())` there lands 4 hours away from what `window_start()` computes, so the seed misses the row the code looks for and the over-budget branch is silently never entered. Confirmed the mechanism by forcing `PGTZ=UTC`: **5/5 pass**, matching the PR's own claimed baseline exactly — CI's stock Postgres container defaults to UTC (nothing in `ci.yml` pins it explicitly, but the base image does), which is why this doesn't show up there.

**With the session forced to UTC, I reproduced the claimed mutation myself**: flipped `restore_attempt=True` → `False` in `execute_send_email`, ran the suite — **1 failed** (`the deferral did not restore the attempt`, `assert 1 == 0`), 4 passed, exactly matching the PR's claim. Restored, re-ran clean at 5/5.

So: the shipped budget/defer/restore logic is real, correct, and independently confirmed by me to be caught by its test — but only once the environment confound is removed. Outside a UTC-timezone Postgres session, `TestTheBudget`'s two DB tests pass **vacuously** (never reaching the branch they claim to test), which means the same mutation would currently go undetected in any environment where Postgres's default session timezone isn't UTC. Not blocking — the product code itself doesn't touch `date_trunc` anywhere and is timezone-pure — but worth a fast follow-up: either seed via `rate_counters.window_start()` directly instead of re-deriving it in SQL, or pin the test session's timezone explicitly rather than relying on an unstated default.

## 2. Is the partial-fix scoping honest? — Yes, and precisely: zero of D1–D4 become reachable

Re-read #1090's own D1–D4 text rather than my paraphrase of it. D1/D2 have real producers (`credential_lifecycle.py:337`, `media_sync.py:228,299`) that call `prompts.push_bindings()` and hit an explicit `"no-surface"` early return when it comes back empty. D3/D4 have no producer at all. **None of `prompts.py`, `credential_lifecycle.py`, or `media_sync.py` is touched by this diff.** Grepped every reference to `send_email` in the repo, on the correct branch: the only call sites are the new executor (`work_loop.py:348`) and the port itself — zero producer call sites anywhere, before or after this PR. `email_sender.py`'s own docstring says it plainly: *"It is not a producer... The producers (invitations — 06 §2; the timeliness halves of D3/D4 — 06 §5) enqueue send_email jobs and this executes them."* D1 and D2 aren't even named there as intended future consumers of this channel — worth Chris or astrid clarifying whether their eventual fix is meant to route through this port at all, since right now nothing says so. This matches the PR's own "This makes the channel exist and drain; it does not make anything send" — confirmed, not just repeated.

## 3. Seam with #1105 — clean, confirmed by import inspection, not assumed

Read the actual merged `scripts/scheduling_monitor.py` from `origin/main` (it postdates the branch point this PR was built from, so I pulled it directly rather than switching checkouts). Its imports are stdlib-only — `argparse, json, os, subprocess, sys, tempfile, time, urllib.error, urllib.request` — zero reference to `email_sender`, `ResendSender`, or anything under `src.services.target`. No shared delivery mechanism exists today; alex's operator alerting and astrid's customer notification remain fully separate, as intended. The architectural door is deliberately left open — `ResendSender` is a plain callable specifically so a jobs-table-independent caller *could* reach it (per the module docstring, citing #1099) — but nothing walks through it yet, and #1105 chose not to.

## 4. `users.primary_email` verified-by-construction — confirmed independently from source

`identity.py`'s `find_or_create_user`: *"email is the VERIFIED claim or None — `google_oidc.verify_id_token` already drops an unverified one, so this function never sees a claim it must doubt."* `_fill_primary_email` only fires when the column is still NULL (write-once, never overwritten), under a `uq_users_primary_email` constraint with explicit collision handling. Grepped every write site for `primary_email` in the repo (`identity.py`, `identity_provisioning.py` only) — no other code path ever sets it. Mason's claim holds; I didn't just inherit it.

## Smaller things checked

- `tests/src/services/target/test_work_loop.py`: 18/18.
- `ruff check` and `ruff format --check` clean on all 6 touched files — no repeat of tonight's #1102 nit.
- The diff deletes a duplicate local `engine_url_from_env` from `worker.py`; confirmed `worker.py:510` already calls the canonical one via `unit_of_work.engine_url_from_env(env)`, so nothing is left dangling — safe cleanup, not scope creep.
- `egress.DEFAULT_ALLOWED_HOSTS` genuinely does not contain `api.resend.com`; `egress.request`'s open-transaction refusal is real (`egress.py:261`).
- CHANGELOG entry present under `## [Unreleased]`, consistent with the PR body point for point.

Not merging — leaving that to Chris. Good place to stand down for the night.

PR #1108, opened 2026-08-27
fix(target): pin the SSRF-validated address to the connection (#871, S.3 carry-in)

### LINKED ISSUE #871: [Refactor Phase S] S.3 carry-in — pin the SSRF-validated address to the connection (TOCTOU/DNS-rebind, deferred from L.0 #857)

chrisrogers37:
Closes #871.

`validate_target` checked every address a host resolved to and then **threw the
result away**; `httpx` resolved the name again when it opened the socket. So the
address-class check proved *"this hostname resolved to a public address at that
moment"* — not *"this connection went to a public address"*. Deferred from L.0
(#857) deliberately and visibly.

**Re-measured at `ef76f1c` before building**, as asked: the gap is open. No
`sni_hostname`, no `extensions`, no URL rewrite, no custom transport anywhere in
`egress.py`, and the module docstring still described the window in the present
tense.

## The second resolution no longer happens

`validate_target` returns the addresses it approved, and `request` hands the
transport a URL whose host is an **IP literal**. There is no name left to look
up — which is why this closes the window rather than narrowing it. A resolver
that answers public and then private is not consulted a second time at all.

## Two overrides, not the custom transport the deferral was priced on

The issue costed this as "a custom transport that resolves, pins the IP, and
still gets TLS SNI and certificate verification right." It turns out the
transport is unnecessary:

- **`extensions["sni_hostname"]`** carries the original hostname. httpcore 1.0.9
  reads it at `_async/connection.py:107` and passes it as `server_hostname` to
  `start_tls` (`:151`), falling back to the origin host only when absent. SNI is
  sent for the name and the certificate is verified against the name, while the
  socket goes to the pinned address.
- **An explicit `Host` header** carries the original `host:port`. httpx
  synthesises `Host` from the URL *only when the header is absent*
  (`_models.py:450-456`), so the override wins.

Neither `verify` nor the SSL context is touched.

Retries **rotate through the approved addresses** rather than re-resolving, so a
recovery cannot reach an address the floor never approved. A redirect **re-pins
from its own validation**; the logical URL keeps the hostname throughout, so a
`Location` is joined against the name the provider used and only the per-attempt
connect URL carries an address.

## The test — this was the hard part and it is the point

A mock transport resolves nothing, so a mock can only ever prove the *shape* of
a request. It cannot show that a second resolution was prevented, because in a
mock there was never going to be one. Both halves are here.

**Structural (mock):** the transport receives `93.184.216.34`, never
`graph.instagram.com`; `sni_hostname` and `Host` carry the name; the caller is
still shown the name it asked for; with the address block off nothing is pinned.

**Rebind (real socket, real httpx transport):** a live local HTTP server, reached
under the hostname `pinned-target.invalid` — which **RFC 2606 guarantees cannot
resolve**, asserted at run time rather than assumed, because a captive or
wildcard resolver would make the whole construction vacuous while still green.
The resolver stub is rebind-shaped: `127.0.0.1` first, `169.254.169.254` second.

Any second lookup is therefore *guaranteed to fail*, so a `200` can only mean no
second lookup happened. The test also asserts the resolver was called exactly
once, and that the live server saw `Host: pinned-target.invalid:<port>` rather
than an IP.

**Its negative half** asserts the same call is unreachable when nothing is
pinned — without that, the positive half would not be evidence, because the name
might have been reachable either way.

**TLS (real handshake):** a self-signed certificate valid for
`pinned-tls.invalid` and **never for the address**, with `verify` pointed at that
certificate as a CA. The handshake can only succeed if verification ran against
the name while the socket went to the pinned address. A version that pinned by
rewriting the URL and forgot `sni_hostname` fails here — that failure mode is a
silent weakening of TLS shipped under the banner of an SSRF fix, which is worse
than the gap.

## Verification

**Five mutations, each confirmed to apply before its run, each killed by the
proof that names it:**

| mutation | failed | which |
|---|---|---|
| no pin — hand the transport the hostname | **5** | both real-socket proofs, the TLS handshake, the IP-literal and SNI assertions |
| drop `sni_hostname` | 2 | the TLS handshake, the SNI assertion |
| drop the `Host` override | 3 | what the live server observed, TLS, SNI |
| re-resolve on every attempt | 2 | the single-resolution proof, the rebind proof |
| never rotate — always `addresses[0]` | 1 | the rotation proof |

The first row is the one that matters: **removing the fix fails the rebind
proof.** That is the property the issue asked for — a test that would not pass
with the bug still present.

Suites: `tests/src/services/target/` → **360 passed**. The 26 existing floor
proofs are green (`Done requires` #4); two of them asserted `validate_target`
returned a bare host and now read `.host`, since it returns what it approved.
It has no production callers outside this module, so no shim.

## Done requires

1. ✅ the connection reaches a pinned, pre-validated address
2. ✅ TLS SNI and verification unchanged — proven by a real handshake
3. ✅ rebind proof: refused at connect time, on a real socket
4. ✅ the 24 (now 26) existing egress proofs still green

## Seam

`src/services/target/egress.py` and its test file only. Nothing outside it.


chrisrogers37 (commented):
**Approve**

Scoped exactly as asked: does the test prove the fix, not whether the fix's design is right (rajan's half).

## Claim 1 — the load-bearing one. Confirmed, empirically, both ways.

Removed the pin with the most surgical change I could make — `_pin_call` returns `(url, kwargs)` unchanged unconditionally, i.e. exactly the "nothing to pin" path the code already has, forced on for every call. Ran the rebind proof against that:

```
FAILED test_the_unresolvable_name_is_reached_anyway_because_it_is_pinned
src.services.target.egress.EgressBudgetExhausted: all 3 attempts failed within
the absolute budget of 30.0s (L.0/#857): ConnectError('[Errno -2] Name or
service not known')
```

That's a real DNS resolution failure from a real connection attempt, not a mock artifact — exactly what you'd expect if the transport were handed the raw (genuinely unresolvable) hostname instead of the pinned IP. The test is not vacuous; it depends on the pin being present.

Ran the full file under the same mutation for the count: **5 failed, 30 passed** — matches the PR's claimed row exactly. The five: the IP-literal assertion, the SNI assertion, both real-socket proofs (rebind + TLS handshake), and the retry-rotation proof (that last one isn't named in the PR's prose breakdown, but it's mechanically correct — without pinning, `_pin_call` never selects an address to rotate through, so `MockTransport` sees the raw hostname on every attempt instead of two different pinned IPs). Restored, confirmed clean, 35/35 again.

## Claim 2 — mutant validity, and I found a real discrepancy in one cell

Reasoned through all five from source before touching anything:

- **#2 (drop `sni_hostname`)** — valid. Without it, httpcore falls back to the origin host for `server_hostname`, which in a pinned connection is the IP. A cert issued only for the name would fail verification against an IP. Real behavior change, correctly attributed to the TLS/SNI tests.
- **#3 (drop the `Host` override)** — valid. httpx synthesizes `Host` from the URL only when absent, so without the override the server sees the IP, not the name — genuinely observable by the live server and by TLS SNI (same mechanism).
- **#5 (never rotate)** — valid, narrow, correctly scoped to the one rotation test.
- **#1 (no pin)** — valid, confirmed above.

**#4 (re-resolve on every attempt) — the count is wrong, and the reason why is worth stating.** I reproduced it: moved `validate_target`'s call from before the retry loop to the first line inside it. Ran the suite: **1 failed, not 2** — `test_a_single_resolution_serves_every_attempt` (correctly), but **the rebind proof itself passed unmutated**. Traced why: the rebind proof's request succeeds on the very first attempt, so the retry loop's body — where this mutation lives — only ever executes once. Whether `validate_target` sits inside or outside the loop is unobservable when there's only ever one iteration. The mutation is real (it's a genuine re-resolve-per-retry defect, not an inert one), and the *shipped* code doesn't have it — `target = validate_target(...)` sits outside the loop in the actual diff, confirmed by reading it directly, so this isn't a claim the fix is broken. But it does mean: **there is currently no real-socket proof that a retry can't rebind — only the weaker MockTransport call-count check covers that case at all.** A rebind proof extended to force a retry (server refuses the first pinned address, succeeds on the second) would close this precisely and is the one concrete follow-up worth taking from this review.

## The three additional questions

**Genuine second resolution, or simulated above the bug?** The injected `resolver` callable is the actual production seam (`_resolve_addresses` is "split out so tests can substitute" — its own docstring says so), not a mock sitting above `validate_target`. The rebind-shaped resolver (`127.0.0.1` then `169.254.169.254`, a real metadata-endpoint address) genuinely returns two different answers on two calls. The proof that a second call never happens is layered — not just a call-count assertion but a real-system tripwire: since the connect URL is only ever an IP literal after pinning, httpx has nothing left to resolve at the transport layer, and the positive control independently proves (via a real `getaddrinfo` call, not assumed) that the raw hostname cannot resolve. A `200` from a real local server is about as strong a proof of "no fallback path exists" as a unit-scale test can produce without standing up an actual rebinding DNS server.

**Is `.invalid` doing real work?** Yes — the RFC 2606 guarantee is what makes "connection succeeded" imply "the pin held" rather than "the name happened not to resolve today." And it covers the threat precisely at the layer that matters: the injected resolver *does* "resolve" to two different answers within the test's simulated world; it's specifically the real OS-level fallback path (what happens if the code ever handed httpx the bare hostname) that's guaranteed dead. That's the right pairing for what this fix actually does — eliminate the second resolution structurally, not detect a bad second answer.

**Coverage — what's not covered, stated rather than left implied:**
- No test asserts that a **same-host redirect hop's connect URL is pinned to an IP literal**. The existing redirect tests (`test_a_SAME_host_redirect_is_revalidated_not_inherited`, `test_a_cross_host_hop_is_refused_at_the_END_of_a_same_host_chain`) check address-class re-validation and cross-host refusal, not whether the hop itself connects via a pinned address rather than the redirected-to hostname. Grepped the file directly — confirmed no such assertion exists.
- **No test covers a retry that re-resolves** (see the #4 discussion above) with a real socket — only a MockTransport call-count proxy.
- **No pooled-connection scenario** is exercised — two separate top-level `request()` calls to the same host where the resolution legitimately changes between them. I don't believe this is a real gap in the mechanism (each call builds its own IP-literal connect URL from its own fresh validation, so httpx's pool keys on the address actually used), but it's reasoned from source, not tested.
- The real-socket proofs only ever validate a **single-address** target; rotation across multiple addresses is only proven via `MockTransport`, not a real connection.

None of these are reasons to hold the PR — the mechanism they'd be testing is sound by direct reading of `egress.py`, and Claim 1 (the property the issue actually asked for) is proven as strongly as I could ask for. They're the honest list of what "does the test prove it" doesn't yet cover, worth a fast follow-up rather than a blocker.

Not merging.

chrisrogers37 (commented):
**Approve** (comment-only — same-identity blocks Approve)

Took the mandate literally: exercised the four properties rather than reading the docstring's claims about them. All four real-socket probes below are my own, built for this review, none copied from the shipped suite (which I ran too, to confirm what it already covers — see bottom).

**1. Does the pin survive a redirect?** Read `request()`'s redirect branch: on a same-host hop, `target = validate_target(str(next_url), policy, resolver=resolver)` fully replaces the prior target — it is re-validated, not carried. To confirm this is real rather than a read of the variable name, I wrote two real-socket tests (two live local HTTP servers, hop 1 → 302 → hop 2, real DNS-unresolvable `.invalid` hostname so nothing can accidentally short-circuit through real DNS):
- Hop 2's *own* resolution returns a genuinely forbidden address (`169.254.169.254`) → refused before ever touching hop 2's server (0 hits), resolver called exactly twice.
- Hop 2 resolves to a *different but allowed* real address than hop 1 → hop 2 is actually reached, correct `Host` header, resolver called exactly twice.

Then I mutation-tested my own probes: commented out the `target = validate_target(...)` reassignment (simulating "silently carried"). Both tests failed — one as `DID NOT RAISE` (the forbidden address stopped being checked at all), the other as a `ConnectError` (the stale hop-1 address got dialed against hop 2's server, which isn't listening there). Reverted, clean 3/3. The redirect path is genuinely re-pinning per hop, not carrying state forward.

**2. Retry / connection-pool reuse.** The custom retry loop never touches httpx's own retry machinery — every attempt goes through `_pin_call` fresh, and since the connect URL's host is already the IP literal, httpx/httpcore partitions its pool by that IP, not by the original hostname. To test this isn't just true-by-my-reading, I ran two live servers on **the same port**, on two different loopback addresses (`127.0.0.1` / `127.0.0.2`), reused **one** `httpx.AsyncClient` (so its pool persists) across two separate top-level `request()` calls to the identical `host:port` string, with the resolver returning a different address each time. Both servers were hit exactly once each — no stale reuse of the first connection for the second call.

**3. TLS correctness with `sni_hostname` overridden.** Reran `TestTlsSurvivesThePin` myself (real self-signed cert valid only for the hostname, real handshake) — passes. Then I mutated `_pin_call` to drop the `extensions["sni_hostname"] = target.host` line and reran both the SNI assertion and the real-handshake test: **exactly 2 failed**, matching the PR's own table. The handshake failure is the informative part — it comes back as `CERTIFICATE_VERIFY_FAILED: IP address mismatch, certificate is not valid for '127.0.0.1'`, i.e. without the override, verification silently shifts to checking the cert against the *address* instead of the *name*. That's the exact silent-TLS-weakening failure mode the PR calls out as worse than the SSRF gap, reproduced on my own machine, not taken on faith.

**4. Address-class check on the actually-connected address, not just the first resolution.** Covered two ways: the existing `TestTheSecondResolutionCannotHappen` (rebind resolver, real socket, unresolvable name) reran clean under my own invocation; and my own redirect-hop test above is the same property one hop later — hop 2's resolution is independently subjected to the same `_is_forbidden` loop, not grandfathered in on hop 1's approval.

**Other checks:** ran the full `tests/src/services/target/` directory myself — **360 passed**, matching the PR exactly. Read the actual removed lines in the diff (not just the docstring's description of them) — the old `request()` really did call `validate_target(...)` and discard the return value on both the initial call and the redirect hop, confirming the "before" state was the real TOCTOU window described, not a rhetorical framing of already-safe code.

Nothing here needs to change. The custom-transport the issue priced this at turned out to be unnecessary, and the two overrides hold up under real sockets, a real handshake, and pool reuse — not just the mock layer.

